Issue #170, opened 2014-11-25
Add StoreEventListener to Store SPI tests


vfunshteyn:
This is a new SPI introduced as part of #41

Issue #26, opened 2014-09-04
Add Bulk CRUD API to Cache Interface


alexsnaps:
Add the `putAll`, `getAll` et al methods to the `Cache` interface.
This is an API issue, it doesn't require implementing any further than to please javac. (i.e. throw `UnsupportedOperationException` is just fine)

PR #207, opened 2014-12-05
#170, added test cases for StoreEventListener on expiration and eviction

gmangalick1:
Hi Please review this pull request, few things to notice:

1. I moved TimeSource from impl to core module as i needed access to TimeSource for writing expiration tests.
2. There is a bug in testPutIfAbsentOnEviction test, as putIfAbsent call puts key/value in store even thou store capacity is reached.

Thanks

Gaurav

### DIFF
--- a/xml/config/checkstyle-suppressions.xml
+++ b/xml/config/checkstyle-suppressions.xml
@@ -5,5 +5,5 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-    <suppress files="org[\\/]ehcache[\\/]config[\\/]xml[\\/]model[\\/]" checks="Header"/>
+    <suppress files="org[\\/]ehcache[\\/]config[\\/]xml[\\/]model[\\/]" checks=".*"/>
 </suppressions>


core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -0,0 +1,333 @@
...
+    kvStore.enableStoreEventNotifications(listener);
+    kvStore.put(k, v);
+    timeSource.advanceTime(1);
+    kvStore.compute(k, new BiFunction<K, V, V>() {
+      @Override
+      public V apply(K mappedKey, V mappedValue) {
+        return v2;
+      }
+    });
+    assertThat(kvStore.get(k).value(), is(v2));
```
chrisdennis: This shouldn't be here right?  The compute itself should trigger the expiration and listener fire.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -0,0 +1,333 @@
...
+    assertThat(listener.expired, hasItem(k));
```
chrisdennis: This should come before the get... the listener should already have been fired immediately after the compute returns.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -0,0 +1,333 @@
...
+    timeSource.advanceTime(1);
+
+    kvStore.computeIfAbsent(k, new Function<K, V>() {
+      @Override
+      public V apply(K mappedKey) {
+        return v2;
+      }
+    });
+    assertThat(kvStore.get(k).value(), is(v2));
+    assertThat(listener.expired, hasItem(k));
```
chrisdennis: These assertions should be in the opposite order.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -0,0 +1,333 @@
...
+    timeSource.advanceTime(1);
+
+    kvStore.computeIfPresent(k, new BiFunction<K, V, V>() {
+      @Override
+      public V apply(K mappedKey, V mappedValue) {
+        throw new AssertionError();
+      }
+    });
+    assertThat(kvStore.get(k), is(nullValue()));
+    assertThat(listener.evicted, hasItem(k));
```
chrisdennis: Invert the order here.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -0,0 +1,333 @@
...
+  @SPITest
+  public void testPutOnEviction() throws Exception {
+    Store<K, V> kvStore = factory.newStore(factory.newConfiguration(factory.getKeyType(), factory.getValueType(),
+        1L, null, null));
+    TestStoreEventListener listener = new TestStoreEventListener();
+    kvStore.enableStoreEventNotifications(listener);
+    kvStore.put(k, v);
+    kvStore.put(k2, v2);
+    Map<K, V> map = convertToMap(kvStore.iterator());
+    assertThat(map.size(), is(1));
```
chrisdennis: Should this strictly be lessThan(2)?  And should we assert evicted contents based on how small the size is?

chrisdennis: This applies to probably all of the eviction tests (if we think this is the right thing to do).

chrisdennis:
The failing tests should be @Ignore'd and issues filed for fixing them.

ljacomet:
Two things that make me not go further than this screen:
* Please squash your commits - initial is about 32 files modified, total result of the PR is 8 files. In this context, keeping the whole history of the change before inclusion makes no sense. Request help if that demand is not clear.
* Fix the failing tests or ignore them and file issue(s) as @chrisdennis said.

This PR has been open a long time, would be good to finish it.

gmangalick1:
@ljacomet 
1. I rolled back the changes and rebased my work. Now this is fixed.
2. I already put those tests to ignore, but due to merge failures it was causing some other failures. This is fixed too. 

There is a new pull request with all these changes: https://github.com/ehcache/ehcache3/pull/26

I am hoping to have it merged and close as soon as tomorrow. Thanks

Issue #239, opened 2014-12-23
Ehcache.putAll leaves Store entries after CacheWriter.writeAll failure


cljohnso:
An `Ehcache.putAll` call during which one or more `CacheWriter.writeAll` calls fail old entries in `Store` corresponding to the failed keys remains in `Store`.  Should these entries have been dropped?

The following tests in `org.ehcache.EhcacheBasicPutAllTest` are impacted by this question:

* `testPutAllStoreSomeOverlapWriterNoOverlapAllFail()`

cljohnso:
Discussion during [Ehcache weekly dev meeting - 8 January 2015] (https://plus.google.com/events/c7lgves3acscvgr9l075240rrmg) led to the conclusion that (at least) all declared failed keys should be invalidated in `Store` so, whether or not the write via `CacheWriter` was successful, the next `Ehcache.get` access to any value declared failed will yield the value from the store-of-record.

Converting this issue to a BUG.

Issue #238, opened 2014-12-22
Ehcache.putAll can report successful CacheWriter write as failure via BulkCacheWriterException


cljohnso:
`Ehcache.putAll` can report a successful write through `CacheWriter.writeAll` as a failure if multiple entries are presented to `CacheWriter.writeAll` and `writeAll` throws an exception *other* than `BulkCacheWriterException` after the `CacheWriter` successfully records one or more entries.  Should a different exception indicating uncertain results be thrown instead of the somewhat inaccurate `BulkCacheWriterException` that is now thrown?

cljohnso:
Discussion during [Ehcache weekly dev meeting - 8 January 2015] (https://plus.google.com/events/c7lgves3acscvgr9l075240rrmg) led to the conclusion that (at least) all declared failed keys should be invalidated in `Store` so, whether or not the write via `CacheWriter` was successful, the next `Ehcache.get` access to any value declared failed will yield the value from the store-of-record.

Converting this issue to a BUG.

ljacomet:
Working on this, I realised a number of tests have an invalid state from the start. That is the CacheStore knows mappings that are unknown of the LoaderWriter.
I have updated all these cases to start with a valid state, that is all Cache Store mappings are known to the LoaderWriter - which may in turn contain more mappings.

Issue #234, opened 2014-12-16
Add `Ehcache.putAll` unit tests


cljohnso:
Add unit tests for `Ehcache.putAll`.  (Spun off from Issue #192.)

PR #240, opened 2014-12-30
Issue #234 - Add Ehcache.putAll unit tests

cljohnso:
Add unit tests for `Ehcache.putAll`.

This update does not address issues #238 or #239.

This update changes `EhcacheBasicCrudBase.FakeStore` to update the implementations of the `compute`, `computeIfPresent`, and `bulkCompute` methods and changes the `EhcacheBasicCrudBase.FakeCacheWriter` to implement the `writeAll` method.

The `EhcacheBasicRemoveValueTest` and `EhcacheBasicReplaceValueTest` tests were updated to adapt to the method signature changes introduced in PR #223.

### DIFF
--- a/core/src/test/java/org/ehcache/EhcacheBasicPutAllTest.java
+++ b/core/src/test/java/org/ehcache/EhcacheBasicPutAllTest.java
@@ -24,18 +24,26 @@ import org.ehcache.spi.writer.CacheWriter;
 import org.ehcache.statistics.CacheOperationOutcomes;
 import org.hamcrest.Matchers;
 import org.junit.Ignore;
+import org.junit.Rule;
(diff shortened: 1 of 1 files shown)


ljacomet:
From `org.ehcache.EhcacheBasicPutAllTest#testPutAllEmptyRequestCacheAccessExceptionBeforeNoWriter` I see that we call into the store when passed in an empty map. 
I am in favour of adding a test in the code that short circuits method execution in this case.
Others?

cljohnso:
Agreed -- strikes me as a good optimization.  There's no JSR 107 gotcha for this, is there?  (I don't recall one but there are others better versed in 107 than I.)

PR #243, opened 2015-01-05
add suppress warnings, unsued imports, etc

timeck:
Took a pass at adding warning suppression where I thought it was appropriate, removed unsued imports, set serial UID on our exception types.


timeck:
test this please

ljacomet:
So after discussions, we want to adapt the `CacheLoader.loadAll` contract. Created #245 for that. This also means that the commit about `CacheLoader` generic change should be dropped from this PR.

timeck:
I recreated this PR with the second commit removed

Issue #163, opened 2014-11-20
Cache.putIfAbsent contract in JSR-107 is weird, ours is now weirder - how do we reconcile?


chrisdennis:
This was the topic of the Ehcache 3.0 public meeting today.  The discussion there stands, but the current is actually markedly worse in ehcache/master than I made it out to be in the meeting.

The current code does this:
```java
public V putIfAbsent(K key, V value) {
  V cached = cache.get(key);
  if (cached == null) {
    sor.putIfAbsent(key, value);
    cache.put(key, value);
  }
  return cached;
}
```
This is pretty much useless - any time you putIfAbsent on an uncached value you'll corrupt the cache.

JSR-107 requires this (expanding it's return out to an Object for clarity):
```java
public V putIfAbsent(K key, V value) {
  V cached = cache.get(key);
  if (cached == null) {
    sor.put(key, value);
    cache.put(key, value);
  }
  return cached;
}
```
This works fine as long as the cache and SoR are in sync.  What's weird here is that the behavior is explicitly mandated to be counter to that of a transparent proxy, and does so in a way that allows invalid writes to be made, and returns false information to the user.

After discussion in todays meeting we think we can adopt this:
```java
public V putIfAbsent(K key, V value) {
  V cached = cache.get(key);
  if (cached == null) {
    V stored = sor.putIfAbsent(key, value);
    if (stored == null) {
      cache.put(key, value);
    } else {
      cache.put(key, stored);
    }
    return stored;
  }
  return cached;
}
```
This acts like a transparent proxy unless the cache is out of sync, but unlike the JSR-107 version will not make invalid writes to the SoR (instead the cache might veto some valid writes) and won't return incorrect information to the caller.

Thoughts? Opinions?

PR #244, opened 2015-01-06
Issue #163

lorban:
All the atomic methods (and not just putIfAbsent) now make use of the cache loader when one is configured to get the value from the SoR if there's no mapping in the cache.

There is a new EhcacheServiceConfiguration that can be used to disable the use of the loader in the atomic methods, which is what the 107 caches do by default to meet the spec.

In case a writer but no loader is configured, one get the 107 behavior in both cases.


core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -781,7 +822,7 @@ public V apply(final K k, final V inCache) {
       }
     });
     try {
-      store.computeIfPresent(key, remappingFunction, REPLACE_FALSE);
+      store.compute(key, remappingFunction, REPLACE_FALSE);
```
timeck: Thinking ahead a little bit, how do we think these compute methods will work on clustered stores? If compute() might necessarily end up being slower than computeIfPresent() then it might be worth complicating things here to decide which one to call on the store depending on whether "useLoaderInAtomics" is true and whether there is a cacheLoader.

core/src/main/java/org/ehcache/config/service/EhcacheServiceConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class EhcacheServiceConfiguration implements ServiceConfiguration<EhcacheService> {
+  private boolean noLoadInAtomics;
+
+  @Override
+  public Class<EhcacheService> getServiceType() {
+    return EhcacheService.class;
+  }
+
+  public boolean noLoadInAtomics() {
```
timeck: at first read, using a negative name for this property seems a little strange. maybe consider flipping it to "loadInAtomics"

ljacomet: +1 and also because `Ehcache` uses a `useLoaderInAtomics` flag already.

core/src/main/java/org/ehcache/Ehcache.java
ljacomet: I would apply YAGNI here. Maybe have a low priority issue around to make sure we ask the question again once we know for sure.

core/src/main/java/org/ehcache/config/service/EhcacheServiceConfiguration.java
lorban: This is my convention for stating that the default is true. I can change it if it looks too confusing.

ljacomet:
test this please

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -64,10 +65,12 @@
   private final Eh107CacheWriterFactory cacheWriterFactory;
   private final Jsr107Service jsr107Service;
   private final org.ehcache.config.Configuration ehConfig;
+  private final Eh107CacheService eh107CacheService;
```
ljacomet: Why introduce a new service here? Why not have `Jsr107Service implements EhcacheService`?

107/src/main/java/org/ehcache/jsr107/Eh107CacheWriter.java
```diff
@@ -148,5 +148,42 @@ public V getValue() {
     public <T> T unwrap(Class<T> clazz) {
       throw new IllegalArgumentException();
     }
+
+    @Override
+    public int hashCode() {
+      int hashCode = 31;
+      if (key != null) {
+        hashCode += key.hashCode();
```
ljacomet: You are missing a multiplication to reduce collisions

chrisdennis: Maybe it makes sense to use the same hashCode also as Map.Entry?

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
lorban: Good point. Let me do this.

107/src/main/java/org/ehcache/jsr107/Eh107CacheWriter.java
lorban: Nice catch and good point. I'll do this too.

lorban:
test this please

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -215,6 +220,8 @@ public Properties getProperties() {
       onHeapStoreServiceConfig.storeByValue(jsr107Config.isStoreByValue());
     }
 
+    builder.addServiceConfig(new EhcacheServiceConfiguration().jsr107CompliantAtomics(ehcacheService.jsr107CompliantAtomics()));
```
ljacomet: This is redundant with the change introduced in `EhcacheManager` that takes care of it also. And following that, passing the additional constructor argument is redundant too.

lorban:
test this please

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -67,7 +67,8 @@
 
   Eh107CacheManager(EhcacheCachingProvider cachingProvider, org.ehcache.CacheManager ehCacheManager, Properties props,
       ClassLoader classLoader, URI uri, Eh107CacheLoaderFactory cacheLoaderFactory,
-      Eh107CacheWriterFactory cacheWriterFactory, org.ehcache.config.Configuration ehConfig, Jsr107Service jsr107Service) {
```
ljacomet: This diff can probably be dropped, as it contains only code formatting changes.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -91,7 +91,7 @@ public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties
 
         Eh107CacheLoaderFactory cacheLoaderFactory = new Eh107CacheLoaderFactory();
         Eh107CacheWriterFactory cacheWriterFactory = new Eh107CacheWriterFactory();
-        Jsr107Service jsr107Service = new DefaultJsr107Service();
+        DefaultJsr107Service jsr107Service = new DefaultJsr107Service();
```
ljacomet: I would drop that change too

ljacomet:
Closing following today's discussions. Needs some updates.

PR #247, opened 2015-01-06
add suppress warnings, unsued imports, etc

timeck:


107/src/test/java/org/ehcache/ParsesConfigExtensionTest.java
```diff
@@ -70,30 +63,9 @@ public void testConfigParse() throws ClassNotFoundException, SAXException, Insta
...
-          .newCacheConfigurationBuilderFromTemplate(name, cfg.getKeyType(), cfg.getValueType());
-    }
-    if(configurationBuilder != null) {
-      configurationBuilder = CacheConfigurationBuilder.newCacheConfigurationBuilder();
-    }
-    // set all attributes from cfg, to complete or override
-    // and finally build our CacheConfig:
-    final CacheConfiguration cacheConfiguration = configurationBuilder.buildConfig(cfg.getKeyType(), cfg.getValueType());
-    // final Cache cache = cacheManager.createCache(name, cacheConfiguration); // throws NPE because of no Store.Provider Service present
   }
```
ljacomet: Is that whole block removed because it is covered by `Eh107XmlIntegrationTest`?

timeck: yeah, and it wasn't really intended to be part of the test. It was just a convenient place for alex to leave the usage code for me to use later.

Issue #217, opened 2014-12-09
Default URI for 107 EhcacheCachingProvider 


timeck:
There was a good deal of discussion about this one. First thing is that we don't want the default URI for 107 `CacheManager`s created through our provider to use classpath resources like `/ehcache.xml` or `/ehcache-failsafe`. 

The default URI for Ehcache 107 cache managers is: ehcache://default. This URI isn't actually resolvable by a user, but we understand what it means. The "default" default config is an empty configuration, this is however overridable with a property (either a system property, or one passed to the provider). Setting say ehcache.config.default to an explicit config (in some form) makes that config resolve when ehcache://default is used. Forms for that config object could include: a URI object, a string URI, a config object, an xml string etc.

A little more discussion happened on the PR for the initial 107 stuff:
https://github.com/ehcache/ehcache3/pull/210

PR #254, opened 2015-01-06
#217 - Default URI for 107 EhcacheCachingProvider

timeck:


alexsnaps:
Other than the actual prop name, I have no remarks (well... I refrained from arguing about the `instanceof` and cast to `XmlConfiguration`... Don't see a clean solution so... I almost stfu!)... Can be merged as far as I'm concerned if the the prop name if not a problem to anyone...

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -38,35 +39,32 @@
  */
 public class EhcacheCachingProvider implements CachingProvider {
 
+  private static final String DEFAULT_URI_STRING = "ehcache://default";
```
cljohnso: I'd recommend a URI of something like `urn:X-ehcache:default-config`.

Unlike the current URI, this conforms to the relevant RFC and standards:
    * [BCP-115/RFC-4395 Guidelines and Registration Procedures for New URI Schemes](https://tools.ietf.org/html/rfc4395)
    * [BCP-66/RFC-3406 Uniform Resource Names (URN) Namespace Definition Mechanisms](https://tools.ietf.org/html/bcp66)
    * [STD-66/RFC-3986 Uniform Resource Identifier (URI): Generic Syntax](https://tools.ietf.org/html/rfc3986)
    * [RFC-2141 URN Syntax](https://tools.ietf.org/html/rfc2141)

We should stay away from using a new, unregistered scheme *and* avoid use of a hierachical designator (`//`) where none is intended (RFC-4395).

Issue #255, opened 2015-01-07
ExpiryPolicy in cache template is not inherited when Cache is created from template


sreekanth-r:
When a Cache is created from a cache template which has expiry policy set, it is not inherited by the caches which are constructed using that template. This is because when new cache configuration builder is constructed, the expiry policy is not passed on to configuration builder from template

sreekanth-r:
Time Spent : 8 Hours

PR #256, opened 2015-01-08
#255 Add Expiry Policy while constructing builder from Cache Template

sreekanth-r:


ljacomet:
Looks good to me, will merge once rebased

Issue #245, opened 2015-01-06
Adapt CacheLoader.loadAll contract


ljacomet:
After discussions triggered by the generics around `CacheLoader`s, there is a proposition to adapt the contract of `org.ehcache.spi.loader.CacheLoader#loadAll` as follows:
* The returned map is only used to find back values mapped to the passed in keys, that is the key/value pair that is going to be installed in the cache is the `key` as found in the iterator passed in as argument and the value returned by `loadAllResult.get(key)`

`Ehcache` code needs to be adapted to reflect this contract update.

timeck:
Happy to wait for the PR, but I don't understand what is changing from the description here. Sounds like it what the methods already does and I don't see how the ```loadAll()``` signature would change or how it affects generics

PR #259, opened 2015-01-09
Issue #245 Ignore keys returned from CacheLoader.loadAll

ljacomet:
Clarifies/Fixes the way results returned from `CacheLoader.loadAll` are processed.

### DIFF
--- a/core/src/main/java/org/ehcache/Ehcache.java
+++ b/core/src/main/java/org/ehcache/Ehcache.java
@@ -412,24 +412,23 @@ public class Ehcache<K, V> implements Cache<K, V>, StandaloneCache<K, V>, Persis
         }
 
         if (cacheLoaderWriter != null) {
-          Map<K, V> loaded = Collections.emptyMap();
+          Map<? super K, ? extends V> loaded = Collections.emptyMap();
           try {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -413,24 +413,19 @@ public void clear() {
...
           try {
-            loaded = (Map<K, V>) cacheLoader.loadAll(computeResult.keySet());
+            Map<? super K, ? extends V> loaded = cacheLoader.loadAll(computeResult.keySet());
+
+            for (K key : computeResult.keySet()) {
+              V value = loaded.get(key);
+              successes.put(key, value);
+              computeResult.put(key, value);
+            }
           } catch (Exception e) {
```
timeck: The scope the try {} block seems a little off reading through this again. You didn't introduce this in this PR so it could be captured in a new issue. 

I think the intention for catching Exception to is to cover the cacheLoader.loadAll() call. That seems okay, but the try block also covers the for loop that populates the computeResult. I suppose it isn't too likely that an exception would be thrown there (lines 419-423 above) but if it did then you could up with keys in both successes and failures I think, which seems wrong.

ljacomet: My bad, will move the populate code out.

Issue #112, opened 2014-10-15
Add logging


alexsnaps:
We need to decide on an abstract logging framework (à la slf4j) and add it to the code base:

 - Add loggers to `core` classes like `Ehcache`, `EhcacheManager` and `ServiceLocator` and choosing some levels to log things in place where required, so they don't act like complete black boxes;
 - Add logging about specific cases, especially around failures, like in lifecycle methods (e.g. `EhcacheManager#close` or `ServiceLocator#startAllServices`;
 - More?

alexsnaps:
This was somewhat discussed here : https://www.youtube.com/watch?v=Rf86ZxtYXaY&list=UU43PVCp2j0b2og2DtxNOU1A#t=2387

ljacomet:
I believe part of the work was done in PR #264. But there still remains works, in `Ehcache` at least.
And to that, the question of how to get the cache name in there still exists. So I am moving this one back to in progress.

alexsnaps:
Could we use MDC for that? see http://www.slf4j.org/apidocs/org/slf4j/MDC.html

AbfrmBlr:
MDC works for Ehcache and Cache names will be added to the logger.
But MDC does not work with sl4j-simple(which we currently use), it will only work with log4j and logback.
Therefore, should we change gradle testRuntime dependency to use log4j (so that tests log it properly) and add an documentation for MDC logging so that users can use it accordingly.

chrisdennis:
Makes sense to me.

ljacomet:
@AbfrmBlr Looks like I let [a typo](https://github.com/ehcache/ehcache3/commit/b02eb0d5adb679beb06fd0c949e0db24ce0739e7#diff-92de9e0237d8b12df0194b049e12722fR132) slip through.
Can you fix it?

ljacomet:
I did fix the typo

PR #264, opened 2015-01-13
Adding logging for Lifecycle operations for issue #112

AbfrmBlr:
Since we don't have any names related to a Ehcache instance, info logging in Ehcache.java will not be much useful.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -157,6 +157,7 @@ subprojects {
 
   dependencies {
     testCompile 'junit:junit:4.11', 'org.hamcrest:hamcrest-library:1.3', 'org.mockito:mockito-core:1.9.5'
+    testRuntime 'org.slf4j:slf4j-simple:1.7.7'
   }
 
   jar {
--- a/core/src/main/java/org/ehcache/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/EhcacheManager.java
@@ -35,6 +35,8 @@ import org.ehcache.spi.service.Service;
 import org.ehcache.spi.service.ServiceConfiguration;
 import org.ehcache.spi.service.ThreadPoolsService;
 import org.ehcache.util.ClassLoading;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -56,6 +58,8 @@ import org.ehcache.spi.loaderwriter.CacheLoaderWriterFactory;
  */
 public class EhcacheManager implements PersistentCacheManager {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheManager.class);
+  
   private final StatusTransitioner statusTransitioner = new StatusTransitioner();
 
   private final ServiceLocator serviceLocator;
@@ -110,6 +114,7 @@ public class EhcacheManager implements PersistentCacheManager {
         }
       }
       closeEhcache(alias, ehcache);
+      LOGGER.info("Cache '{}' is removed from EhcacheManager.", alias);
     }
   }
 
@@ -121,6 +126,7 @@ public class EhcacheManager implements PersistentCacheManager {
     }
     
     ehcache.getRuntimeConfiguration().releaseAllEventListeners(serviceLocator.findService(CacheEventListenerFactory.class));
+    LOGGER.info("Cache '{}' is closed from EhcacheManager.", alias);
   }
 
   @Override
@@ -128,6 +134,8 @@ public class EhcacheManager implements PersistentCacheManager {
     statusTransitioner.checkAvailable();
     Class<K> keyType = config.getKeyType();
     Class<V> valueType = config.getValueType();
+    
+    LOGGER.info("Cache '{}' is getting created in EhcacheManager.", alias);
     
     ClassLoader cacheClassLoader = config.getClassLoader();
     if (cacheClassLoader == null) {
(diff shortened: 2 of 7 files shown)


AbfrmBlr:
I will update the appropriate changes

ljacomet:
Consider adding a test dependency to `org.slf4j:slf4j-simple:1.7.7` as is done in `core` module, this will make logging visible during test execution.
Ideally this is done using a `testRuntime` dependency declared in the root `build.gradle` and clean up of core module `testCompile` dependencies.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -279,19 +291,22 @@ public void init() {
         }
       } catch (RuntimeException e) {
         while (!initiatedCaches.isEmpty()) {
+          String toBeClosed = initiatedCaches.pop();
           try {
-            removeCache(initiatedCaches.pop());
+            removeCache(toBeClosed);
           } catch (Exception exceptionClosingCache) {
-            // todo probably should log these exceptions
+              LOGGER.error("Cache '"+toBeClosed+"' could not be removed due to "+ exceptionClosingCache.getMessage());
```
ljacomet: When logging an exception, pass also the exception as param to the method, so that the stack trace is available in the log file.

ljacomet:
Logging with Slf4j provides an integrated mechanism for building messages - we should rely on it.
See [`Logger` documentation](http://www.slf4j.org/api/org/slf4j/Logger.html).

build.gradle
```diff
@@ -156,7 +156,7 @@ subprojects {
   }
 
   dependencies {
-    testCompile 'junit:junit:4.11', 'org.hamcrest:hamcrest-library:1.3', 'org.mockito:mockito-core:1.9.5'
+    testCompile 'junit:junit:4.11', 'org.hamcrest:hamcrest-library:1.3', 'org.mockito:mockito-core:1.9.5', 'org.slf4j:slf4j-simple:1.7.7'
```
ljacomet: Is adding the dependency to `testRuntime` an issue with IDEs? If not, that's the proper scope to use.

ljacomet:
Could you open the discussion on the mailing list about the consequence of not having the name of the cache available to log statements from the cache itself?

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -129,6 +135,8 @@ void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
     Class<K> keyType = config.getKeyType();
     Class<V> valueType = config.getValueType();
     
+    LOGGER.info("Cache is getting created in EhcacheManager .");
```
ljacomet: I would add the alias to the log statement, multiple caches can be created in parallel.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -168,8 +176,10 @@ void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
     } else {
       caches.remove(alias);
       value.setCache(null);
+      LOGGER.error("Cache '" + alias + "' creation in EhcacheManager failed.");
```
ljacomet: I am in favour of wrapping the failure into another exception with the message and throwing that instead of logging a message and throwing an exception that may not carry the cache alias in its message.

core/src/main/java/org/ehcache/spi/ServiceLocator.java
```diff
@@ -181,13 +185,15 @@ public void startAllServices(final Map<Service, ServiceConfiguration<?>> service
       for (Service service : services.values()) {
         service.start(serviceConfigs.get(service));
         started.push(service);
+        LOGGER.info("All Services successfully started. ");
```
ljacomet: Given the text, this should be outside of the loop

build.gradle
AbfrmBlr: I tested this for eclipse and it works. I will test this for IDEA and Netbeans as well.
All other comments are taken care of.

AbfrmBlr: Same works for intellij as well when projects are imported based on gradle model

ljacomet: What I meant with my request was:
```
    testCompile 'junit:junit:4.11', 'org.hamcrest:hamcrest-library:1.3', 'org.mockito:mockito-core:1.9.5'
    testRuntime 'org.slf4j:slf4j-simple:1.7.7'
```
I tested it in IntelliJ and it is working fine - that is running tests with JUnit runner do proper logging - tried with `EhcacheManagerTest`.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -168,8 +176,9 @@ void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
     } else {
       caches.remove(alias);
       value.setCache(null);
-      throw failure;
+      throw new RuntimeException("Cache '"+alias+"' creation in EhcacheManager failed.", failure);
```
ljacomet: I would pick an `IllegalStateException` here rather than the generic `RuntimeException`.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -276,19 +287,22 @@ public void init() {
         }
       } catch (RuntimeException e) {
         while (!initiatedCaches.isEmpty()) {
+          String toBeClosed = initiatedCaches.pop();
           try {
-            removeCache(initiatedCaches.pop());
+            removeCache(toBeClosed);
           } catch (Exception exceptionClosingCache) {
-            // todo probably should log these exceptions
+              LOGGER.error("Cache '{}' could not be removed due to ",toBeClosed, exceptionClosingCache);
```
ljacomet: Minor detail but your out of alignment here.

ljacomet:
Made some more line remarks, but there is a more general remark with code formatting - mostly around ',' and ' ' (space) after them.

PR #274, opened 2015-01-19
Issue #248

lorban:
Ehcache 2.x open-source disk store.
Missing cache events and statistics, can be added later on if we keep this store implementation.


impl/src/main/java/org/ehcache/internal/store/disk/DiskStorageFactory.java
```diff
@@ -0,0 +1,1206 @@
...
+
+    @Override
+    public String toString() {
+      return "" + value;
+    }
+  }
+
+
+  /**
+   * Path stub used to create unique ehcache directories.
```
ljacomet: What does this comment refer to?

ljacomet:
Overall looks good. Wondering if we should link/merge the `DiskStorePathManager` and the `DefaultLocalPersistenceService`? Or is that something we leave for the future?

alexsnaps:
Merged

PR #283, opened 2015-02-03
Adding Logger injection for Ehcache class  #112

AbfrmBlr:


core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -105,6 +107,7 @@ public EhcacheManager(Configuration config, ServiceLocator serviceLocator, boole
   @Override
   public void removeCache(final String alias) {
     statusTransitioner.checkAvailable();
+    MDC.put("alias", alias);
```
lorban: ehcache might not be the only user of the MDC, and "alias" is quite a generic term. I think this should be prefixed to avoid potential name clashes.

ljacomet:
I must be missing the point here. How is what is proposed going to work as soon as I have more than one cache and concurrent lifecycle operations?

ljacomet:
Could you rename the PR?

impl/src/main/java/org/ehcache/internal/EhcacheProvider.java
```diff
@@ -31,7 +32,7 @@
   @SuppressWarnings("deprecation")
   @Override
   public <K, V> Ehcache<K, V> createCache(Class<K> keyClazz, Class<V> valueClazz, ServiceConfiguration<?>... config) {
-    return new HeapCache<K, V>();
+    return new HeapCache<K, V>(LoggerFactory.getLogger(Ehcache.class + "HeapCache"));
```
ljacomet: This misses the separator used elsewhere '-'

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -53,7 +54,7 @@ public void testWikiExample() {
 
     cacheManager.removeCache("preConfigured"); // <8>
 
-    final StandaloneCache<Long, String> standaloneCache = newCacheBuilder(Long.class, String.class)
+    final StandaloneCache<Long, String> standaloneCache = newCacheBuilder(Long.class, String.class, LoggerFactory.getLogger(Ehcache.class+"GettingStarted"))
```
ljacomet: Misses the separator use elsewhere

impl/src/test/java/org/ehcache/config/CachingTierConfigurationBuilderTest.java
```diff
@@ -50,7 +52,7 @@ public void testNothing() throws CacheAccessException {
     
     
     @SuppressWarnings("unchecked")
-    final Cache<String, String> cache = new HeapCache<String, String>(mock(CacheConfiguration.class), store);
+    final Cache<String, String> cache = new HeapCache<String, String>(mock(CacheConfiguration.class), store, LoggerFactory.getLogger(Ehcache.class+"CachingTierConfigurationBuilderTest"));
```
ljacomet: Missing the separator

impl/src/test/java/org/ehcache/statistics/StatisticsTest.java
```diff
@@ -41,7 +43,7 @@
   public void setup() throws Exception {
     scheduledExecutorService = Executors.newScheduledThreadPool(0);
 
-    cache = StandaloneCacheBuilder.newCacheBuilder(Number.class, String.class)
+    cache = StandaloneCacheBuilder.newCacheBuilder(Number.class, String.class, LoggerFactory.getLogger(Ehcache.class+"StatisticsTest"))
```
ljacomet: Separator again :-)

integration-test/src/test/java/org/ehcache/integration/StandaloneCacheEvictionTest.java
```diff
@@ -32,7 +34,7 @@
 
   @Test
   public void test_eviction_with_specific_eviction_prioritizer() throws Exception {
-    StandaloneCache<Number, String> cache = StandaloneCacheBuilder.newCacheBuilder(Number.class, String.class)
+    StandaloneCache<Number, String> cache = StandaloneCacheBuilder.newCacheBuilder(Number.class, String.class, LoggerFactory.getLogger(Ehcache.class + "StandaloneCacheEvictionTest"))
```
ljacomet: Separator, same for next change

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1064,21 +1079,25 @@ public CacheStatistics getStatistics() {
   }
 
   void create() {
+    logger.info("Cache getting created.");
```
ljacomet: Would have a message like: Creating cache storage.
Cause that is really what's happening there. The Cache is created at that point.

Same holds for the other create log statements and for the destroy ones.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1064,21 +1079,25 @@ public CacheStatistics getStatistics() {
...
     statusTransitioner.checkMaintenance();
     try {
       store.create();
     } catch (CacheAccessException e) {
       throw new RuntimeException("Couldn't create Cache", e);
```
ljacomet: And so updating that exception message is needed

Issue #78, opened 2014-09-24
Add checkstyle javadoc rule on API module


alexsnaps:
We want JavaDoc on all API types... So this task also encompasses adding Javadoc where some is missing!

PR #285, opened 2015-02-03
Issue #78 - Checkstyle + javadoc for api module

ljacomet:
* Configure checkstyle to require javadoc on API module
* Fix checkstyle errors by adding javadoc

### DIFF
--- a/api/src/main/java/org/ehcache/CacheManager.java
+++ b/api/src/main/java/org/ehcache/CacheManager.java
@@ -33,6 +33,9 @@ public interface CacheManager {
    * @param <K> the type of the keys used to access data within this cache
    * @param <V> the type of the values held within this cache
    * @return the created and initialized {@link Cache}
+   *
+   * @throws java.lang.IllegalArgumentException If there is already a cache registered with the given alias.
+   * @throws java.lang.IllegalStateException If the cache creation fails
    */
   <K, V> Cache<K, V> createCache(String alias, CacheConfiguration<K, V> config);
 
--- a/api/src/main/java/org/ehcache/CacheManager.java
+++ b/api/src/main/java/org/ehcache/CacheManager.java
@@ -25,10 +25,38 @@ import org.ehcache.config.CacheConfiguration;
  */
 public interface CacheManager {
 
+  /**
+   * Creates a {@link Cache} in this {@code CacheManager} according to the specified {@link CacheConfiguration}.
+   *
+   * @param alias the alias under which the cache will be created
+   * @param config the configuration of the cache to create
+   * @param <K> the type of the keys used to access data within this cache
+   * @param <V> the type of the values held within this cache
+   * @return the created and initialized {@link Cache}
+   */
   <K, V> Cache<K, V> createCache(String alias, CacheConfiguration<K, V> config);
 
+  /**
+   * Retrieves the {@link Cache} associated with the given alias, if one is known.
+   *
+   * @param alias the alias under which to look the {@link Cache} up
+   * @param keyType the {@link Cache} key class
+   * @param valueType the {@link Cache} value class
+   * @param <K> the type of the keys used to access data within this cache
+   * @param <V> the type of the values held within this cache
+   * @return the {@link Cache} associated with the given alias, {@code null} if no association exists
+   *
+   * @throws java.lang.IllegalArgumentException If the keyType or valueType do not match the ones with which the {@link Cache} was created
+   */
   <K, V> Cache<K, V> getCache(String alias, Class<K> keyType, Class<V> valueType);
 
+  /**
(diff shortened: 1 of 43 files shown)


api/src/main/java/org/ehcache/CacheManager.java
```diff
@@ -25,10 +25,38 @@
...
+  /**
+   * Creates a {@link Cache} in this {@code CacheManager} according to the specified {@link CacheConfiguration}.
+   *
+   * @param alias the alias under which the cache will be created
+   * @param config the configuration of the cache to create
+   * @param <K> the type of the keys used to access data within this cache
+   * @param <V> the type of the values held within this cache
+   * @return the created and initialized {@link Cache}
+   */
   <K, V> Cache<K, V> createCache(String alias, CacheConfiguration<K, V> config);
```
timeck: this method throws an exception if the cache already exists I think

api/src/main/java/org/ehcache/CacheManager.java
```diff
@@ -25,10 +25,38 @@
  */
 public interface CacheManager {
 
+  /**
+   * Creates a {@link Cache} in this {@code CacheManager} according to the specified {@link CacheConfiguration}.
```
timeck: I don't have a lot of javadoc experience really. Is there an argument for/against fully qualifying class names in @link refs?

api/src/main/java/org/ehcache/config/CacheConfiguration.java
```diff
@@ -26,37 +26,43 @@
...
+ * @param <V> the type of the values held within the cache
+ *
  * @author Alex Snaps
  */
 public interface CacheConfiguration<K, V> {
 
   /**
-   * Not sure whether this should be exposed on this interface really
-   * @return unmodifiable collection of service configuration related to this cache
+   * Not sure whether this should be exposed on this interface really.
```
timeck: Not necessarily part of this review but maybe we should take the time to answer this question? @alexsnaps  ?

api/src/main/java/org/ehcache/event/CacheEvent.java
```diff
@@ -19,6 +19,11 @@
 import org.ehcache.Cache;
 
 /**
+ * Class representing an event resulting from a mutative {@link Cache} operation.
```
timeck: doesn't really matter but you could maybe drop the "Class representing" at the beginning

api/src/main/java/org/ehcache/exceptions/SerializerException.java
```diff
@@ -18,24 +18,44 @@
...
  * @author Ludovic Orban
  */
 public class SerializerException extends RuntimeException {
 
   private static final long serialVersionUID = -4008956327217206643L;
 
+  /**
+   * No args constructor for this exception.
+   */
   public SerializerException() {
```
timeck: god kills a kitten every time we write javadoc to satisfy checkstyle for obvious things like this

api/src/main/java/org/ehcache/spi/serialization/SerializationProvider.java
```diff
@@ -19,10 +19,20 @@
 import org.ehcache.spi.service.ServiceConfiguration;
 
 /**
+ * Interface that allows to provide a {@link Serializer} for a given type.
```
timeck: "that allows" confuses me a little, maybe not necessary

api/src/main/java/org/ehcache/CacheManager.java
ljacomet: I believe you do not need the _fqcn_ if the class is in the import list. And I tend to use the _fqcn_ when I want to reference a class that is not in the import list.

api/src/main/java/org/ehcache/spi/serialization/SerializationProvider.java
ljacomet: Replaced with: Contract to provide a {@link Serializer} for a given type.

api/src/main/java/org/ehcache/exceptions/SerializerException.java
ljacomet: At the same time, not sure I want to handle this via a checkstyle exception

api/src/main/java/org/ehcache/event/CacheEvent.java
ljacomet: Done

api/src/main/java/org/ehcache/CacheManager.java
ljacomet: Indeed and another one if the cache fails to init.

ljacomet:
Added a commit to address @timeck's remarks

Issue #271, opened 2015-01-15
StoreSPITest.testClose() is incorrect


lorban:

lorban:
and I mean it.

lorban:
StoreCloseTest.closedStoreCantBeUsed() tries to use the store after it's being closed but doesn't catch the exception nor register it as an expectation which is wrong.

ljacomet:
Might be worth having a look at `Store.close` and `Store.destroy` contracts, as they seem at least unclear.
* Are they related to each other?
 * Does `destroy` implies `close`?

lorban:
After discussing with Alex, it seems that the store contract is fine but the test is the problem.

aurbroszniowski:
Both of your points are correct:
The contract for Store.close is unclear, it doesn't define what happens when the store is accessed after that the store is closed. If it is meant to be so and not to define it explicitely, then it would make sense to remove the SPI test as, like you pointed out, it doesn't define an expectation, and actually it can't with the current contract.

Basically the sentence "The instance can't be used any further upon the method returning." is too vague to be translated into a test.

lorban:
Right, something has to be done either on the test (make it test less) or on the SPI (make it more testable).

BTW, testDestroy() and testProviderReleaseStore() are also failing IMHO due to the same mismatch problem between test and SPI. Check https://github.com/ehcache/ehcache3/pull/274 for details.

PR #290, opened 2015-02-08
issue #271 fixed Store Spi tests

aurbroszniowski:


aurbroszniowski:
One Spi test is still disabled, see https://github.com/ehcache/ehcache3/issues/291

PR #299, opened 2015-02-13
Issue #238 - Update assertions to compare store and writer contents

ljacomet:


core/src/test/java/org/ehcache/EhcacheBasicPutAllTest.java
```diff
@@ -2186,7 +2168,8 @@ public void testPutAllStoreSomeOverlapCacheAccessExceptionAfterWriterFullOverlap
    */
   private static final boolean debugResults;
   static {
-    debugResults = Boolean.parseBoolean(System.getProperty(EhcacheBasicPutAllTest.class.getName() + ".debug", "false"));
```
lorban: I assume you forgot to revert this.

Issue #294, opened 2015-02-10
org.ehcache.internal.store.disk.DiskStoreSPITest : tests are failing on windows


AbfrmBlr:
org.ehcache.internal.store.disk.DiskStoreSPITest > testReplaceKeyValueValue FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testGet FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testPutIfAbsentKey FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testClear FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testClose FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testComputeIfPresent FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testIteratorHasNext FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testContainsKey FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testComputeIfAbsent FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testBulkCompute FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testCompute FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testReplaceKeyValue FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testBulkComputeIfAbsent FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testIteratorNext FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testPutKey FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testIterator FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testRemoveKeyValue FAILED

org.ehcache.internal.store.disk.DiskStoreSPITest > testRemoveKey FAILED

AbfrmBlr:
java.lang.Exception: java.lang.RuntimeException: org.ehcache.exceptions.CacheAccessException: Data file already exists: C:\Users\aab\AppData\Local\Temp\disk%0053tore.data
	at org.ehcache.internal.store.disk.DiskStoreSPITest$1.newStore(DiskStoreSPITest.java:68)
	at org.ehcache.internal.store.disk.DiskStoreSPITest$1.newStore(DiskStoreSPITest.java:58)
[...]

PR #300, opened 2015-02-17
Fixing issue #294 : DiskStoreSpi Tests failing on windows

rishabhmonga:
Tests missing DiskStore.close(), causing failure of file.delete

### DIFF
--- a/core-spi-test/src/main/java/org/ehcache/internal/store/StoreBulkComputeIfAbsentTest.java
+++ b/core-spi-test/src/main/java/org/ehcache/internal/store/StoreBulkComputeIfAbsentTest.java
@@ -69,6 +69,10 @@ public class StoreBulkComputeIfAbsentTest<K, V> extends SPIStoreTester<K, V> {
     } catch (CacheAccessException e) {
       System.err.println("Warning, an exception is thrown due to the SPI test");
       e.printStackTrace();
+    } finally {
+      if(kvStore != null) {
+        kvStore.close();
+      }
     }
   }
 
@@ -103,6 +107,10 @@ public class StoreBulkComputeIfAbsentTest<K, V> extends SPIStoreTester<K, V> {
     } catch (CacheAccessException e) {
       System.err.println("Warning, an exception is thrown due to the SPI test");
       e.printStackTrace();
+    } finally {
+      if(kvStore != null) {
+        kvStore.close();
+      }
     }
   }
 
@@ -143,6 +151,10 @@ public class StoreBulkComputeIfAbsentTest<K, V> extends SPIStoreTester<K, V> {
     } catch (CacheAccessException e) {
       System.err.println("Warning, an exception is thrown due to the SPI test");
       e.printStackTrace();
+    } finally {
+      if(kvStore != null) {
+        kvStore.close();
+      }
     }
   }
 
@@ -175,6 +187,10 @@ public class StoreBulkComputeIfAbsentTest<K, V> extends SPIStoreTester<K, V> {
     } catch (CacheAccessException e) {
       System.err.println("Warning, an exception is thrown due to the SPI test");
       e.printStackTrace();
+    } finally {
+      if(kvStore != null) {
+        kvStore.close();
+      }
     }
   }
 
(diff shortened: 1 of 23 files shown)


ljacomet:
Honestly, this PR is a no go from me.
You end up creating loads of code duplication.

Two options according to me:
1. Improve SPITester to introduce before/after fixtures
2. Bite the bullet and scratch SPITester, replacing it with standard JUnit.

I am in favour of the second option

spi-tester/src/main/java/org/ehcache/spi/test/SPITester.java
```diff
@@ -28,6 +29,25 @@ public Result runTestSuite() {
...
+    ArrayList<Method> afterMethodList = new ArrayList<Method>();
+    for(Method m : testClass.getDeclaredMethods()){
+      if (m.isAnnotationPresent(Before.class)) {
+        beforeMethodList.add(m);
+      }
+      if (m.isAnnotationPresent(After.class)) {
+        afterMethodList.add(m);
+      }
+    }
+    if(!beforeMethodList.isEmpty()) {
```
ljacomet: To match JUnit behavior and be consistent with introduced `@After`, the `@Before` methods should be executed for each test method.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreComputeIfAbsentTest.java
```diff
@@ -33,10 +35,25 @@ public StoreComputeIfAbsentTest(StoreFactory<K, V> factory) {
     super(factory);
   }
 
+  protected Store<K, V> kvStore;
+  protected Store kvStore2;
```
ljacomet: Why a raw type here? Looks like all tests having a kvStore2 always use the raw type.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreBulkComputeIfAbsentTest.java
```diff
@@ -43,9 +45,19 @@ public StoreBulkComputeIfAbsentTest(StoreFactory<K, V> factory) {
...
+    if (kvStore != null) {
+      kvStore.close();
+      kvStore = null;
+    }
+  }
+
   @SPITest
   public void mappingFunctionReturnsIterableOfEntriesForEachInputEntry() throws Exception {
-     Store<K, V> kvStore = factory.newStore(factory.newConfiguration(factory.getKeyType(), factory.getValueType(), null, null, null));
+    kvStore = factory.newStore(factory.newConfiguration(factory.getKeyType(), factory.getValueType(), null, null, null));
```
ljacomet: The introduction of `@Before` allows this code to be refactored into a `setUp` method.

spi-tester/src/main/java/org/ehcache/spi/test/SPITester.java
rishabhmonga: In this part of the code I'm collected all methods with @Before and @After annotations to be invoked later. 
Please see line 43 for @Before and line 63 for @After

core-spi-test/src/main/java/org/ehcache/internal/store/StoreComputeIfAbsentTest.java
rishabhmonga: A raw store type is defined here as was being used in some tests to check behaviour in case of wrong key and/or value type. The definition has been moved out so a common tearDown method can close all stores at the end of each test

core-spi-test/src/main/java/org/ehcache/internal/store/StoreBulkComputeIfAbsentTest.java
rishabhmonga: A setUp method cannot be used in these tests as not all tests are using kvStore and kvStore2 simultaneously.

spi-tester/src/main/java/org/ehcache/spi/test/SPITester.java
```diff
@@ -28,7 +29,27 @@ public Result runTestSuite() {
...
+        beforeMethodList.add(m);
+      }
+      if (m.isAnnotationPresent(After.class)) {
+        afterMethodList.add(m);
+      }
+    }
+
     for (Method m : testClass.getDeclaredMethods()) {
+      if(m.isAnnotationPresent(SPITest.class) && !m.isAnnotationPresent(Ignore.class)) {
+        for (Method bm : beforeMethodList) {
```
ljacomet: Move this inside the `else try {` part instead of having to add conditions.

Issue #216, opened 2014-12-09
Need tests that verify that the 107 config extensions work correctly


timeck:
This mechanism lets one associate 107 caches to either a default cache template or to a specific template for a specific 107 cache name. 

In either case there is no test for this at the moment.

One thing to definitely assert is that all of the overlap between 107 config and ehcache config is applied. Things like expiration, loader/writer, expiry, etc should also be dictated by the 107 (always overriding what might be present in the 107 template config)

sreekanth-r:
JSR 107 Cache Configuration classes i.e. `Eh107CompleteConfiguration` & `Eh107ReverseConfiguration` doesn't expose

* Runtime Configuration (i.e. way to checking configured Capacity)
* Eviction veto
* Eviction Priotirizer

Should these be exposed for unit testing purpose?

ljacomet:
Following the discussions during today's public dev meeting, before looking at the two types you mention, we need to figure out what happens when you have values in the XML template and you pass in a `MutableConfiguration` with minimal things setup on it.
Could you write some code showing that and then revive the discussion?

sreekanth-r:
JSR 107 Configuration with template

```

<ehcache:config
    xmlns:ehcache="http://www.ehcache.org/v3"
    xmlns:jcache="http://www.ehcache.org/v3/jsr107">

  <!-- Defining a default template cache which stores long, string mapping -->
  <ehcache:service>
    <jcache:defaults default-template="defaultTemplate">
      <jcache:cache name="productCache" template="productCacheTemplate"/>
    </jcache:defaults>
  </ehcache:service>

  <ehcache:cache-template name="defaultTemplate" storeByValueOnHeap="true">
    <ehcache:key-type>java.lang.Long</ehcache:key-type>
    <ehcache:value-type>java.lang.String</ehcache:value-type>
    <ehcache:capacity>2</ehcache:capacity>
    <ehcache:expiry>
      <ehcache:ttl unit="millis">1</ehcache:ttl>
    </ehcache:expiry>
    <ehcache:eviction-prioritizer>LRU</ehcache:eviction-prioritizer>
  </ehcache:cache-template>

  <!-- product cache template overriding all properties from default -->
  <ehcache:cache-template name="productCacheTemplate" storeByValueOnHeap="false">
    <ehcache:key-type>java.lang.Long</ehcache:key-type>
    <ehcache:value-type>com.pany.domain.Product</ehcache:value-type>
[...]

PR #301, opened 2015-02-18
Expose Cache loader/Writer Configuration at template

sreekanth-r:
This is required for #216

### DIFF
--- a/xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
+++ b/xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
@@ -296,6 +296,11 @@ public class XmlConfiguration implements Configuration {
         .usingEvictionPrioritizer(getInstanceOfName(cacheTemplate.evictionPrioritizer(), defaultClassLoader, EvictionPrioritizer.class, Eviction.Prioritizer.class))
         .evitionVeto(getInstanceOfName(cacheTemplate.evictionVeto(), defaultClassLoader, EvictionVeto.class))
         .withExpiry(getExpiry(defaultClassLoader, parsedExpiry));
+    final String loaderWriter = cacheTemplate.loaderWriter();
+    if(loaderWriter!= null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -130,4 +132,9 @@ private CacheConfigurationBuilder(final Expiry<? super K, ? super V> expiry, fin
     this.serializationProvider = serializationProvider;
     return this;
   }
+
+  public CacheConfigurationBuilder<K, V> withLoaderWriter(CacheLoaderWriter<? super K, ? super V> loaderWriter) {
```
ljacomet: Why?
It is supposed to be injected through a service

xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -296,6 +296,10 @@ public URL getURL() {
         .usingEvictionPrioritizer(getInstanceOfName(cacheTemplate.evictionPrioritizer(), defaultClassLoader, EvictionPrioritizer.class, Eviction.Prioritizer.class))
         .evitionVeto(getInstanceOfName(cacheTemplate.evictionVeto(), defaultClassLoader, EvictionVeto.class))
         .withExpiry(getExpiry(defaultClassLoader, parsedExpiry));
+    final String loaderWriter = cacheTemplate.loaderWriter();
```
ljacomet: This should be done the same way it is done in `org.ehcache.config.xml.XmlConfiguration#parseConfiguration`

sreekanth-r: Fixed this, removed the other Cache Configuration stuff, as loader and writer should be managed by services.

ljacomet:
Failed build because of test failures - please have a look @sreekanth-r

PR #304, opened 2015-02-25
Adding in Cache loader/writer overriding test case for #216

sreekanth-r:
Adding in test case which verifies following condition

Template | Conf | isRead/WriteThrough | Applicable
------------ | ------------- | ------------- | -------------
Yes|No|No| None
Yes|Yes|Yes| Conf
Yes|No|Yes| Template

With this in #216 only pending task will be writing a brand new MutuableConfiguration for Ehcache.

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
+++ b/107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
@@ -39,15 +39,20 @@ import org.ehcache.Ehcache;
 import org.ehcache.EhcacheHackAccessor;
 import org.ehcache.config.CacheConfiguration;
 import org.ehcache.config.CacheConfigurationBuilder;
+import org.ehcache.config.loaderwriter.DefaultCacheLoaderWriterConfiguration;
 import org.ehcache.config.xml.XmlConfiguration;
 import org.ehcache.internal.store.service.OnHeapStoreServiceConfig;
 import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
 import org.ehcache.spi.service.ServiceConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * @author teck
  */
 class Eh107CacheManager implements CacheManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(Eh107CacheManager.class);
 
   private static MBeanServer MBEAN_SERVER = ManagementFactory.getPlatformMBeanServer();
 
@@ -201,8 +206,11 @@ class Eh107CacheManager implements CacheManager {
     if (builder == null) {
       builder = CacheConfigurationBuilder.newCacheConfigurationBuilder();
     }
-
-    builder = builder.withExpiry(expiry);
+    if(expiry != null) {
+      builder = builder.withExpiry(expiry);
+    } else {
+      LOG.warn("JSR107 Expiry Policy not set for cache {} using expiry policy from template", cacheName);
+    }
 
     OnHeapStoreServiceConfig onHeapStoreServiceConfig = builder.getExistingServiceConfiguration(OnHeapStoreServiceConfig.class);
     if (onHeapStoreServiceConfig == null) {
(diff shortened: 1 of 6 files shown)


107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -219,6 +224,22 @@ public Properties getProperties() {
       cacheLoaderWriterFactory.registerJsr107Loader(cacheName, cacheLoaderWriter);
     }
 
+    //If cache is not read or write thro' remove the loader writer configuration from the JSR 107 conf
+    if(!jsr107Config.isReadThrough() && !jsr107Config.isWriteThrough()) {
+      DefaultCacheLoaderWriterConfiguration conf = builder.getExistingServiceConfiguration(DefaultCacheLoaderWriterConfiguration.class);
+      if(conf != null) {
+        LOG.warn("Removing the loader/writer service configuration from the JSR107 cache {}", cacheName);
```
ljacomet: Do not comment on what you are doing, but why. Something like: "Ignoring template provided CacheLoaderWriter configuration"

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -219,6 +224,22 @@ public Properties getProperties() {
...
+      }
+    }
+    if((jsr107Config.isReadThrough() || jsr107Config.isWriteThrough()) && cacheLoaderWriter == null) {
+      DefaultCacheLoaderWriterConfiguration conf = builder.getExistingServiceConfiguration(DefaultCacheLoaderWriterConfiguration.class);
+      if(conf == null) {
+        throw new InstantiationException("Unable to construct (read/write)through cache without either a templated loader/writer or configured loader/writer");
+      }
+      LOG.warn("Using the templated cache loader/writer {} for the cache {}", conf.getClazz().getName(), cacheName);
+    }
+
```
ljacomet: (general comment on the whole block above) I don't like that we end up with 3 independent if blocks (1 from before + 2 added here). Can this be refactored in if/else rather?
Also I believe we should log the case where a config ends up overriding a template provided `CacheLoaderWriter`.

107/src/test/java/org/ehcache/jsr107/Eh107JSRTemplateOverridingTest.java
```diff
@@ -0,0 +1,109 @@
...
+    MutableConfiguration<Long, Product> conf = new MutableConfiguration<Long, Product>();
+    conf.setTypes(Long.class, Product.class);
+
+    javax.cache.Cache<Long, Product> c1 = cm.createCache("productCache", conf);
+    ProductCacheLoaderWriter.written.clear();
+    c1.put(1L, new Product(1L));
+    assertThat(ProductCacheLoaderWriter.written.isEmpty(), is(true));
+    cm.destroyCache("productCache");
+
+    conf.setReadThrough(true);
```
ljacomet: I would extract the base config creation in a method and call it every time.
Here, if we decided to piggy back on the MutableConfiguration and modify it inside our code, the test could break.

107/src/test/java/org/ehcache/jsr107/Eh107JSRTemplateOverridingTest.java
```diff
@@ -0,0 +1,109 @@
...
+    conf.setReadThrough(false);
+    conf.setWriteThrough(true);
+    c1 = cm.createCache("productCache", conf);
+    c1.put(1L, new Product(1L));
+    assertThat(ProductCacheLoaderWriter.written.isEmpty(), is(false));
+    ProductCacheLoaderWriter.written.clear();
+    cm.destroyCache("productCache");
+  }
+
+  @Test(expected = MultiCacheException.class)
```
ljacomet: Given the non specific exception, I would prefer seeing it handled in a catch block with some assertion making sure it is thrown for the reason we expect.

107/src/test/java/org/ehcache/jsr107/Eh107JSRTemplateOverridingTest.java
sreekanth-r: Are we talking about Modification of mutuable configuration within testcase or within the cachemanger? AFAIK, we cannot modify the configuration within CacheManager as it will cause tck tests fail.

ljacomet: It's a mutable object and this is test code. I would shield it from potential side effects by creating a new one each time through a method doing the equivalent of lines 58 and 59.

107/src/test/java/org/ehcache/jsr107/Eh107JSRTemplateOverridingTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    assertThat(cc1.getExpiryPolicyFactory(), is(EternalExpiryPolicy.factoryOf()));
+    Factory<ExpiryPolicy> factory = CreatedExpiryPolicy.factoryOf(Duration.ONE_MINUTE);
+    conf.setExpiryPolicyFactory(factory);
+    javax.cache.Cache<Long, String> c2 = cm.createCache("c2", conf);
+    Eh107CompleteConfiguration<Long, String> cc2 = c2.getConfiguration(Eh107CompleteConfiguration.class);
+    assertThat(cc2.getExpiryPolicyFactory(), is(factory));
+  }
+
+  private MutableConfiguration<Long, Product> createProductCacheConf() {
```
ljacomet: Can you do the same for the test above? Possibly reusing this method with type params.

Issue #182, opened 2014-11-28
Add Serializer to XML config


alexsnaps:
The `ehcache-core.xsd` in the xml module doesn't let you configure a `org.ehcache.spi.serialization.Serializer` for a given `Cache` configured in XML. We probably want to support this a both `<cache/>` and `<cache-template/>` levels

AbfrmBlr:
SerializationProvider should not be the part of the CacheConfiguration but as of now it needs to be part of the Store.Configuration.
Reason being OnHeapStore needs key and value serializers but DiskStore uses different serializers(for Element and Object).
Would it make sense to retain SerializationProvider as part of Store.Configuration so that the Stores can get the serializers as required?

AbfrmBlr:
To enable Serializers to be configured from xml requires extension of ehcache-core.xsd with a different namespace.(just like jsr107 service)
DefaultSerializationProvider(JavaSerializationProvider in this case) will fall in a different namespace in xsd with a parser implemented. This requires a new project, as having it in impl creates a cyclic dependency among projects.
If the ehcache user wants to use a different SerializationProvider, he will have to create another namepsace and write the parser again(or may be he can reuse the one implemented for the DefaultSerializationProvider).This will hold true for all the Services with ServiceConfiguration. Is it the right understanding ?

PR #305, opened 2015-02-25
Adding Serializer for xml config #182

AbfrmBlr:
Lets discuss this.
This PR contains changes required for cleaning up Cache and Store configs , exposing xml to add custom serializers for default serialization provider.

### DIFF
--- a/core/src/main/java/org/ehcache/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/EhcacheManager.java
@@ -30,7 +30,6 @@ import org.ehcache.events.DisabledCacheEventNotificationService;
 import org.ehcache.exceptions.StateTransitionException;
 import org.ehcache.spi.ServiceLocator;
 import org.ehcache.spi.cache.Store;
-import org.ehcache.spi.serialization.SerializationProvider;
 import org.ehcache.spi.service.Service;
 import org.ehcache.spi.service.ServiceConfiguration;
 import org.ehcache.spi.service.ThreadPoolsService;
@@ -176,18 +175,14 @@ public class EhcacheManager implements PersistentCacheManager {
    *  adjusts the config to reflect new classloader & serialization provider
    */
   private <K, V> CacheConfiguration<K, V> adjustConfigurationWithCacheManagerDefaults(CacheConfiguration<K, V> config) {
-    SerializationProvider serializationProvider = config.getSerializationProvider();
-    if (serializationProvider == null) {
-      serializationProvider = serviceLocator.findService(SerializationProvider.class);
-    }
     ClassLoader cacheClassLoader = config.getClassLoader();
     if (cacheClassLoader == null) {
       cacheClassLoader = cacheManagerClassLoader;
     }
-    if (cacheClassLoader != config.getClassLoader() || serializationProvider != config.getSerializationProvider()) {
+    if (cacheClassLoader != config.getClassLoader() ) {
       config = new BaseCacheConfiguration<K, V>(config.getKeyType(), config.getValueType(), config.getCapacityConstraint(),
           config.getEvictionVeto(), config.getEvictionPrioritizer(), cacheClassLoader, config.getExpiry(),
-          serializationProvider, config.isPersistent(), config.getServiceConfigurations().toArray(
+          config.isPersistent(), config.getServiceConfigurations().toArray(
           new ServiceConfiguration<?>[config.getServiceConfigurations().size()]));
     }
     return config;
@@ -197,7 +192,6 @@ public class EhcacheManager implements PersistentCacheManager {
                                         final Class<K> keyType, final Class<V> valueType, Deque<Releasable> releasables) {
     Collection<ServiceConfiguration<?>> adjustedServiceConfigs = new ArrayList<ServiceConfiguration<?>>(config.getServiceConfigurations());
     ServiceConfiguration[] serviceConfigs = adjustedServiceConfigs.toArray(new ServiceConfiguration[adjustedServiceConfigs.size()]);
-
 
     final Store.Provider storeProvider = serviceLocator.findService(Store.Provider.class);
     final Store<K, V> store = storeProvider.createStore(new StoreConfigurationImpl<K, V>(config), serviceConfigs);
--- a/core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
+++ b/core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
@@ -16,6 +16,7 @@
 
 package org.ehcache;
 
+import java.util.HashMap;
 import java.util.concurrent.ScheduledExecutorService;
 
 import org.ehcache.config.BaseCacheConfiguration;
@@ -30,7 +31,8 @@ import org.ehcache.expiry.Expiry;
 import org.ehcache.spi.ServiceLocator;
 import org.ehcache.spi.cache.Store;
 import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
-import org.ehcache.spi.serialization.SerializationProvider;
+import org.ehcache.spi.service.Service;
+import org.ehcache.spi.service.ServiceConfiguration;
 import org.ehcache.util.ClassLoading;
 import org.slf4j.Logger;
 
@@ -48,7 +50,6 @@ public class StandaloneCacheBuilder<K, V, T extends StandaloneCache<K, V>> {
   private EvictionVeto<? super K, ? super V> evictionVeto;
   private EvictionPrioritizer<? super K, ? super V> evictionPrioritizer;
   private CacheLoaderWriter<? super K, V> cacheLoaderWriter;
-  private SerializationProvider serializationProvider;
   private ScheduledExecutorService statisticsExecutor;
   private CacheEventNotificationService<K, V> cacheEventNotificationService;
   private boolean persistent;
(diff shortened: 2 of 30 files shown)


core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -176,18 +175,15 @@ void closeEhcache(final String alias, final Ehcache<?, ?> ehcache, final Deque<R
    *  adjusts the config to reflect new classloader & serialization provider
    */
   private <K, V> CacheConfiguration<K, V> adjustConfigurationWithCacheManagerDefaults(CacheConfiguration<K, V> config) {
-    SerializationProvider serializationProvider = config.getSerializationProvider();
-    if (serializationProvider == null) {
-      serializationProvider = serviceLocator.findService(SerializationProvider.class);
-    }
+//    SerializationProvider serializationProvider = serviceLocator.findService(SerializationProvider.class);
```
alexsnaps: Do we need to keep these comments? (see line 198 as well)

core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
```diff
@@ -61,16 +59,16 @@ public StandaloneCacheBuilder(final Class<K> keyType, final Class<V> valueType,
 
   T build(ServiceLocator serviceLocator) {
     Store.Provider storeProvider = serviceLocator.findService(Store.Provider.class);
-    if (serializationProvider == null) {
-      this.serializationProvider = serviceLocator.findService(SerializationProvider.class);
-    }
+//    if (serializationProvider == null) {
+//      this.serializationProvider = serviceLocator.findService(SerializationProvider.class);
+//    }
```
alexsnaps: Comments?

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -55,10 +55,26 @@
...
-      <xs:any namespace="##other" minOccurs="1" maxOccurs="1"/>
+      <xs:element name="serializer-default" type="ehcache:serializer-type" minOccurs="0" maxOccurs="1"/>
+      <xs:any namespace="##other" minOccurs="0" maxOccurs="1"/>
     </xs:sequence>
   </xs:complexType>
-
+  
+   <xs:complexType name="serializer-type">
+    <xs:sequence>
+      <xs:element name="clazz" maxOccurs="unbounded" minOccurs="0">
```
alexsnaps: Why not class?

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -55,10 +55,26 @@
 
   <xs:complexType name="service-type">
     <xs:sequence>
-      <xs:any namespace="##other" minOccurs="1" maxOccurs="1"/>
+      <xs:element name="serializer-default" type="ehcache:serializer-type" minOccurs="0" maxOccurs="1"/>
```
alexsnaps: I _think_ this is wrong... doesn't this allow for both elements to be present? It should be either or, not both

core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
```diff
@@ -61,16 +59,13 @@ public StandaloneCacheBuilder(final Class<K> keyType, final Class<V> valueType,
...
     Store.Provider storeProvider = serviceLocator.findService(Store.Provider.class);
-    if (serializationProvider == null) {
-      this.serializationProvider = serviceLocator.findService(SerializationProvider.class);
-    }
 
     final StoreConfigurationImpl<K, V> storeConfig = new StoreConfigurationImpl<K, V>(keyType, valueType,
-        capacityConstraint, evictionVeto, evictionPrioritizer, classLoader, expiry, serializationProvider);
-    final Store<K, V> store = storeProvider.createStore(storeConfig);
+        capacityConstraint, evictionVeto, evictionPrioritizer, classLoader, expiry);
+    final Store<K, V> store = storeProvider.createStore(serviceLocator, storeConfig);
```
ljacomet: There should not be a need to pass in the serviceLocator here. Upon starting a service receives the `ServiceProvider` it is supposed to use.

core/src/main/java/org/ehcache/spi/cache/Store.java
```diff
@@ -424,7 +424,7 @@
      * @param <V> value type
      * @return the Store honoring the configurations passed in
      */
-    <K, V> Store<K, V> createStore(Configuration<K, V> storeConfig, ServiceConfiguration<?>... serviceConfigs);
+    <K, V> Store<K, V> createStore(ServiceLocator serviceLocator, Configuration<K, V> storeConfig, ServiceConfiguration<?>... serviceConfigs);
```
ljacomet: There should not be a need to pass in the serviceLocator here. Upon starting a service receives the `ServiceProvider` it is supposed to use.

impl/src/main/java/org/ehcache/config/serializer/DefaultSerializationProviderConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+import org.ehcache.internal.serialization.JavaSerializationProvider;
+import org.ehcache.spi.serialization.Serializer;
+import org.ehcache.spi.service.ServiceConfiguration;
+
+public class DefaultSerializationProviderConfiguration implements ServiceConfiguration<JavaSerializationProvider> {
+
+  private final Map<String, TypeSerializerConfig> serializerConfig = new HashMap<String, TypeSerializerConfig>();
+  
+  @Override
+  public Class<JavaSerializationProvider> getServiceType() {
```
ljacomet: I believe this should return a `Class<SerializationProvider>`

alexsnaps: Actually this _may well_ be fine... I need to check. I wonder if we don't walk up the chain. I'll write a test for that stuff in order to document this properly once & for all!

AbfrmBlr: DefaulltSerializationProviderConfiguration is in line with DefaultCacheLoaderWriterFactoryConfiguration, if we follow the type system.

impl/src/main/java/org/ehcache/internal/store/disk/DiskStore.java
```diff
@@ -995,7 +1007,7 @@ public void releaseStore(final Store<?, ?> resource) {
 
     @Override
     public void start(final ServiceConfiguration<?> config, final ServiceProvider serviceProvider) {
-      // nothing to do
+      this.serviceLocator = (ServiceLocator)serviceProvider;
```
ljacomet: No need to cast, a `ServiceProvider` has the method you need

impl/src/main/java/org/ehcache/internal/store/OnHeapStore.java
```diff
@@ -777,7 +788,7 @@ public void releaseStore(final Store<?, ?> resource) {
 
     @Override
     public void start(ServiceConfiguration<?> cfg, final ServiceProvider serviceProvider) {
-      // nothing to do
+      this.serviceLocator = (ServiceLocator)serviceProvider;
```
ljacomet: No need to cast, a `ServiceProvider` has the method you need

impl/src/main/java/org/ehcache/internal/store/OnHeapStore.java
```diff
@@ -777,7 +788,7 @@ public void releaseStore(final Store<?, ?> resource) {
 
     @Override
     public void start(ServiceConfiguration<?> cfg, final ServiceProvider serviceProvider) {
-      // nothing to do
+      this.serviceLocator = serviceProvider;
```
ljacomet: I would have renamed the variable as well

107/src/main/java/org/ehcache/jsr107/Eh107MutableConfiguration.java
```diff
@@ -36,4 +52,20 @@ public Eh107MutableConfiguration(CompleteConfiguration<K, V> configuration) {
   public boolean isExpiryPolicyOverridden() {
     return this.expiryPolicyOverridden;
   }
+
```
lorban: What's the point of overriding equals and hashcode just to call the super methods?

core-spi-test/src/main/java/org/ehcache/internal/store/StoreProviderCreateStoreTest.java
```diff
@@ -16,6 +16,7 @@
 
 package org.ehcache.internal.store;
 
+import org.ehcache.spi.ServiceLocator;
```
lorban: Is that import needed?

107/src/main/java/org/ehcache/jsr107/Eh107MutableConfiguration.java
AbfrmBlr: to fix finbugs failure

core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
```diff
@@ -60,17 +61,19 @@ public StandaloneCacheBuilder(final Class<K> keyType, final Class<V> valueType,
   }
 
   T build(ServiceLocator serviceLocator) {
-    Store.Provider storeProvider = serviceLocator.findService(Store.Provider.class);
-    if (serializationProvider == null) {
-      this.serializationProvider = serviceLocator.findService(SerializationProvider.class);
+    try {
+      serviceLocator.startAllServices(new HashMap<Service, ServiceConfiguration<?>>());
+    } catch (Exception e) {
+      logger.error("StandaloneCacheBuilder failed due to {}.", e.getMessage());;
```
lorban: The exception should be logged as such, not its message (ie: logger.error("some message", exception) instead of  logger.error("some message: ", exception.getMessage) ) to make sure the logger can print the full stack trace.

core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
```diff
@@ -60,17 +61,19 @@ public StandaloneCacheBuilder(final Class<K> keyType, final Class<V> valueType,
...
     }
```
lorban: I'm not convinced just going on when starting the services failed is the right thing to do here. What's going to happen down the line?

impl/src/main/java/org/ehcache/internal/store/OnHeapStore.java
```diff
@@ -777,7 +788,7 @@ public void releaseStore(final Store<?, ?> resource) {
 
     @Override
     public void start(ServiceConfiguration<?> cfg, final ServiceProvider serviceProvider) {
-      // nothing to do
+      this.serviceProvider = serviceProvider;
```
lorban: It's probably wise to set the serviceProvider back to null in stop().

impl/src/test/java/org/ehcache/config/CachingTierConfigurationBuilderTest.java
```diff
@@ -44,6 +46,11 @@
   public void testNothing() throws CacheAccessException {
 
     final ServiceLocator serviceLocator = new ServiceLocator();
+    try {
+      serviceLocator.startAllServices(new HashMap<Service, ServiceConfiguration<?>>());
+    } catch (Exception e) {
+      throw new CacheAccessException(e);
```
lorban: Don't do that in test code, change the throws clause of the test method itself.

core/src/main/java/org/ehcache/StandaloneCacheBuilder.java
AbfrmBlr: Right. This will throw NullPointer while creating store. This means build() should throw Exception or return null.

xml/src/test/resources/configs/ehcache-serializer.xml
```diff
@@ -0,0 +1,32 @@
+<!-- ~ Copyright Terracotta, Inc. ~ ~ Licensed under the Apache License,
```
lorban: You probably want to reformat this file.

107/src/main/java/org/ehcache/jsr107/Eh107MutableConfiguration.java
ljacomet: Don't fix others' mess in your PR, risk making things harder if their fix ends up going first.

Issue #251, opened 2015-01-06
Introduce Tiering


alexsnaps:
Basically port the "Montreal design" from the 2.x line: `CacheStore`, `CachingTier` and `AuthoritativeTier`

Issue #122, opened 2014-10-24
Implement Store.ValueHolder.hitRate() in all implementations


lorban:

lorban:
Some work will be needed in the DiskStore too: because of tiering, the hit rate has to be copied to the authority when flushing happens.

alexsnaps:
Need to verify `LFU` works properly too (i.e. add tests or enable them if they exist!)

PR #306, opened 2015-02-25
Issue #251

lorban:
Wiring hasn't been done (I'll open a new task for that) and issue #122 also is relevant (unimplemented hitRate should be copied on flushing).


impl/src/main/java/org/ehcache/internal/store/disk/DiskStore.java
```diff
@@ -211,7 +214,11 @@ private static int hash(int hash) {
   }
 
   @Override
-  public ValueHolder<V> get(final K key) throws CacheAccessException {
+  public ValueHolder<V> get(K key) throws CacheAccessException {
+    return get(key, false);
+  }
+
+  ValueHolder<V> get(final K key, boolean markFaulted) throws CacheAccessException {
```
ljacomet: I would give this method a more explicit name like: `internalGetAndFault`

impl/src/main/java/org/ehcache/internal/store/disk/DiskStore.java
```diff
@@ -484,6 +491,22 @@ public void disableStoreEventNotifications() {
...
+
+  @Override
+  public ValueHolder<V> computeIfAbsentAndFault(K key, Function<? super K, ? extends V> mappingFunction) throws CacheAccessException {
+    return internalComputeIfAbsent(key, mappingFunction, true);
+  }
+
+  @Override
+  public boolean flush(K key, ValueHolder<V> valueHolder, CachingTier<K, V> cachingTier) {
+    int hash = hash(key.hashCode());
+    return segmentFor(hash).flush(key, hash, (DiskValueHolder<V>) valueHolder, cachingTier);
```
ljacomet: How can the blind cast to `DiskValueHolder` be safe on a public method?

impl/src/main/java/org/ehcache/internal/store/disk/Segment.java
```diff
@@ -821,16 +825,24 @@ boolean cleanUpFailedMarker(final K key, final int hash) {
    *
    * @param key     the key
    * @param hash    they hash
-   * @param element the expected element
+   * @param valueHolder   the expected value holder
    * @return true if succeeded
    */
-  boolean flush(final K key, final int hash, final DiskStorageFactory.Element<K, V> element) {
+  boolean flush(final K key, final int hash, DiskValueHolder<V> valueHolder, CachingTier<K, V> cachingTier) {
```
ljacomet: I do not think it needs to be a `DiskValueHolder` which would solve the blind cast mentioned earlier.

impl/src/main/java/org/ehcache/internal/store/tiering/AuthoritativeTier.java
```diff
@@ -0,0 +1,47 @@
...
+import org.ehcache.exceptions.CacheAccessException;
+import org.ehcache.function.Function;
+import org.ehcache.spi.cache.Store;
+
+/**
+ * Authoritative tier, according to Montreal design.
+
+ * @author Ludovic Orban
+ */
+public interface AuthoritativeTier<K, V> extends Store<K, V> {
```
ljacomet: Can't this interface live in the `core` module?

impl/src/main/java/org/ehcache/internal/store/tiering/CachingTier.java
```diff
@@ -0,0 +1,110 @@
...
+import org.ehcache.exceptions.CacheAccessException;
+import org.ehcache.function.Function;
+import org.ehcache.spi.cache.Store;
+
+/**
+ * Caching tier, according to Montreal design.
+ *
+ * @author Ludovic Orban
+ */
+public interface CachingTier<K, V> {
```
ljacomet: Can't this interface live in the `core` module?

impl/src/main/java/org/ehcache/internal/store/tiering/AuthoritativeTier.java
lorban: Yes, it should certainly be there.

impl/src/main/java/org/ehcache/internal/store/tiering/CachingTier.java
lorban: Yes, it should certainly be there.

impl/src/main/java/org/ehcache/internal/store/disk/DiskStore.java
lorban: Agreed. Will do and make it private.

impl/src/main/java/org/ehcache/internal/store/disk/DiskStore.java
lorban: it's actually a mistake and a bug (that also probably indicates a lack in the test coverage): this should be a ValueHolder of the CachingTier, not of the authoritative tier.

That's the reason why there also is a CachingTier object passed in: the ValueHolder interface doesn't expose expiration methods that are needed to copy the expiration back to the authoritative tier, hence the authoritative tier asks the caching tier for that data to copy it in its own ValueHolder.

We may want to revisit the ValueHolder interface, or at least the way I implemented this expiration-copying mechanism.

impl/src/main/java/org/ehcache/internal/store/disk/Segment.java
lorban: See the related comment above on the DiskStore.

ljacomet:
Test failures when running on something else than JDK 1.6 - @lorban please have a look

ljacomet:
test this please

lorban:
test this please

core-spi-test/src/main/java/org/ehcache/internal/store/StoreFactory.java
```diff
@@ -51,6 +52,8 @@
 
   ServiceConfiguration<?>[] getServiceConfigurations();
 
+  ServiceProvider getServiceProvider();
```
alexsnaps: Minor detail, but this feels it shouldn't be there. More so all implementations are actually _instantiating_ a new `ServiceLocator` on invocation. I'd rather add the ServiceProvider to the `SPIStoreTester` constructor, just like the `StoreFactory` instance is injected in the tests

core/src/main/java/org/ehcache/spi/cache/tiering/AuthoritativeTier.java
```diff
@@ -0,0 +1,47 @@
...
+   * Marks the mapping as not evictable and performs computeIfAbsent() atomically.
+   * @return the value holder.
+   */
+  ValueHolder<V> computeIfAbsentAndFault(K key, Function<? super K, ? extends V> mappingFunction) throws CacheAccessException;
+
+  /**
+   * This marks the entry as evictable again.
+   * @return true if something was flushed, false otherwise.
+   */
+  boolean flush(K key, ValueHolder<V> valueHolder, CachingTier<K, V> cachingTier);
```
alexsnaps: I am slightly surprised we're injecting the `CachingTier` instance back into the `AuthoritativeTier` here, can't we just add the expiry stuff required to the `ValueHolder` ? There may well be a good reason for it being the way it is though...

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
```diff
@@ -0,0 +1,350 @@
...
+public class CacheStore<K, V> implements Store<K, V> {
+
+  private final CachingTier<K, V> cachingTier;
+  private final AuthoritativeTier<K, V> authoritativeTier;
+
+  public CacheStore(CachingTier<K, V> cachingTier, AuthoritativeTier<K, V> authoritativeTier) {
+    this.cachingTier = cachingTier;
+    this.authoritativeTier = authoritativeTier;
+
+    this.cachingTier.addInvalidationListener(new CachingTier.InvalidationListener<K, V>() {
```
alexsnaps: I think this code path is racy. But for it being called from the constructor. 
So I'd either make it thread safe, or support a single listener to ever be registered (and can't be changed). But right now this `add` method (and the remove too, though it isn't used) scare me, given their current implementation.

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
```diff
@@ -0,0 +1,350 @@
...
+  public static class Provider implements Store.Provider {
+
+    private ServiceProvider serviceProvider;
+    private ServiceConfiguration<?> serviceConfiguration;
+    private OnHeapStore.Provider onHeapStoreProvider;
+    private DiskStore.Provider diskStoreProvider;
+
+    @Override
+    public <K, V> Store<K, V> createStore(Configuration<K, V> storeConfig, ServiceConfiguration<?>... serviceConfigs) {
+      //todo use the storeConfig to figure out what providers to use
```
alexsnaps: We need to make sure that's clearly explained in the associated falloff issue (it may well already be the case, just thought I'd highlight it)

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
```diff
@@ -0,0 +1,350 @@
...
+
+      diskStoreProvider = new DiskStore.Provider();
+      diskStoreProvider.start(serviceConfiguration, serviceProvider);
+
+      return new CacheStore<K, V>(onHeapStoreProvider.createStore(storeConfig, serviceConfigs), diskStoreProvider.createStore(storeConfig, serviceConfigs));
+    }
+
+    @Override
+    public void releaseStore(Store<?, ?> resource) {
+      resource.close();
```
alexsnaps: This `close()` is already invoked in `org.ehcache.Ehcache#close`, may not be an issue... not sure.

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
```diff
@@ -0,0 +1,350 @@
...
+      cachingTier.remove(key);
+    }
+  }
+
+  @Override
+  public ValueHolder<V> computeIfPresent(final K key, final BiFunction<? super K, ? super V, ? extends V> remappingFunction, final NullaryFunction<Boolean> replaceEqual) throws CacheAccessException {
+    try {
+      return authoritativeTier.computeIfPresent(key, remappingFunction, replaceEqual);
+    } finally {
+      // XXX: it's not possible to discriminate if computation has been performed or not by looking at the computeIfPresent return value
```
alexsnaps: Do these two `XXX` imply you want to try to do something smarter about it? If not, I'd just leave the comment in, without it... Not sure what all people mean with that prefix in comments

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
lorban: I will add such note to the wiring issue.

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
lorban: Well, I'm not sure either. Apparently, Cache.close() is only called when the cache is being removed or destroyed by the cache manager, and Store.Provider.releaseStore() is called by means of the releasables queues in the cache manager when the latter is being shut down.

For now this seems to be the way to do it, and The DiskStore provider also closes the store in its releaseStore() method.

I'm rather confused by this contract.

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
lorban: These XXX mean that there may be room for improvement that I don't know would be worth it or not.

Currently, our Store.computeIfXyz() contracts are strict copies of the CHMv8 ones but we could enhance them by returning a special ValueHolder constant instead of null when the function returned null for instance, to help us discriminate no computation vs computation that returned null.

I'm not saying this is something we should do, I'm just explaining what I had in mind when I wrote that comment.

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
lorban: You're entirely right: only a single listener should be registered or nasty things may happen. I'll make the necessary changes.

core/src/main/java/org/ehcache/spi/cache/tiering/AuthoritativeTier.java
lorban: It would be a lot cleaner to add the expiry stuff to the ValueHolder. Unfortunately, the way expiring vs eternal value holders are implemented in the OnHeapStore means that this would require a non-trivial amount of work, so I consider it another task that I will add to the Waffle board.

Louis made the exact same remark (in one of the outdated comments), and here is what I responded:

That's the reason why there also is a CachingTier object passed in: the ValueHolder interface doesn't expose expiration methods that are needed to copy the expiration back to the authoritative tier, hence the authoritative tier asks the caching tier for that data to copy it in its own ValueHolder.

We may want to revisit the ValueHolder interface, or at least the way I implemented this expiration-copying mechanism.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreFactory.java
lorban: I'm not entirely convinced by your arguments but as you said, this is rather minor. The StoreFactory already requires some cleanups, let's leave that one for that time too.

lorban:
test this please

alexsnaps:
I didn't verify, but I take you created all falloff tasks... merging and closing this. 
Thanks!

Issue #284, opened 2015-02-03
Review StoreEventListener related unit tests in BaseOnHeapStoreTest


aurbroszniowski:
The following list of tests in **org.ehcache.internal.store.BaseOnHeapStoreTest**
testGetExpired()
testContainsKeyExpired()
testPutIfAbsentExpired
testRemoveTwoArgExpired
testReplaceTwoArgExpired
testReplaceThreeArg
testComputeExpired
testComputeIfAbsentExpired
testComputeIfPresentExpired

is duplicating a lot the logic from the tests in **org.ehcache.internal.store.StoreEventListenerTest** 
testGetOnExpiration()
testContainsKeyOnExpiration
testPutIfAbsentOnExpiration
testRemoveOnExpiration
testReplaceTwoArgsOnExpiration
testReplaceThreeArgsOnExpiration
testComputeOnExpiration
testComputeIfAbsentOnExpiration
testComputeIfPresentOnExpiration

They should be reviewed and probably unit tests of the StoreEventListener should be improved (starting by adding faulty test cases)

PR #307, opened 2015-02-26
Fixing issue #284 : Reviewed unit tests related tests in BaseOnHeapStoreTest

rishabhmonga:
I have reviewed the tests in StoreEventListenerTest that are related to tests in BaseOnHeapStoreTest and to me, it makes sense to retain these tests as they can validate the listener if behaviour of Store is changed. I've also added a few negative tests for OnEviction and OnExpiration events. However I did not add tests for each and every store operation as eviction and expiration events would be same for each of them. 
Please provide comments/suggestions on the same

### DIFF
--- a/core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
+++ b/core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
@@ -28,7 +28,10 @@ import org.ehcache.spi.test.After;
 import org.ehcache.spi.test.SPITest;
 import org.ehcache.internal.TimeSource;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import static org.hamcrest.MatcherAssert.assertThat;
--- a/impl/src/test/java/org/ehcache/internal/store/BaseOnHeapStoreTest.java
+++ b/impl/src/test/java/org/ehcache/internal/store/BaseOnHeapStoreTest.java
@@ -32,6 +32,7 @@ import org.ehcache.spi.cache.Store.Iterator;
 import org.ehcache.spi.cache.Store.ValueHolder;
 import org.junit.Test;
 import org.mockito.ArgumentMatcher;
+import org.mockito.Matchers;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -44,6 +45,8 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.nullValue;
 import static org.mockito.Matchers.argThat;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 public abstract class BaseOnHeapStoreTest {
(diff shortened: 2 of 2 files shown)


core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -223,9 +256,47 @@ public void testPutOnEviction() throws Exception {
...
+  }
+
+  @SPITest
+  public void testListenerOnEvictionForMultipleEntries() throws Exception {
+    Store<K, V> kvStore = factory.newStore(factory.newConfiguration(factory.getKeyType(), factory.getValueType(),
+        1L, null, null));
+    TestStoreEventListener listener = new TestStoreEventListener();
+    kvStore.enableStoreEventNotifications(listener);
+
+    Set<K> inputKeys = new HashSet<K>();
```
lorban: inputKeys variable looks useless. Should be removed.

core-spi-test/src/main/java/org/ehcache/internal/store/StoreEventListenerTest.java
```diff
@@ -223,9 +256,47 @@ public void testPutOnEviction() throws Exception {
...
+  @SPITest
+  public void testEventListenerInAbsenceOfEvictionEvent() throws Exception {
+    TestTimeSource timeSource = new TestTimeSource();
+    Store<K, V> kvStore = factory.newStore(factory.newConfiguration(factory.getKeyType(), factory.getValueType(),
+        2L, null, null));
+    TestStoreEventListener listener = new TestStoreEventListener();
+    kvStore.enableStoreEventNotifications(listener);
+    kvStore.put(k, v);
+    kvStore.put(k2, v2);
+    timeSource.advanceTime(1);
```
lorban: why are you calling advanceTime() here? I fail to see what its usefulness is: isn't the store configured to not expire?

ljacomet:
@aurbroszniowski Can you have a look at this PR as it relates to an issue you created?

aurbroszniowski:
@rishabhmonga Sorry if the issue wasn't described clearly enough... the goal was not to review SPI tests, but the unit tests in BaseOnHeapStoreTest (as indicated in the title)
for instance, the negative test case 'no expiration' that you added is missing in the unit tests, e.g. adding in the BaseOnHeapStoreTest some cases like:

```  @Test
  public void testGetNotExpired() throws Exception {
    TestTimeSource timeSource = new TestTimeSource();
    OnHeapStore<String, String> store = newStore(timeSource,
        Expirations.timeToLiveExpiration(new Duration(2, TimeUnit.MILLISECONDS)));
    StoreEventListener<String, String> listener = addExpiryListener(store);
    store.put("key", "value");
    timeSource.advanceTime(1);
    assertThat(store.get("key"), not(nullValue()));
    verify(listener, never()).onExpiration(Matchers.<Cache.Entry<String, String>>any());
  }
```

aurbroszniowski:
And to be more explicit, I would be quite strict NOT to include the tests you are proposing in the SPI tests as they do not verify any explicit contract, but rather adapt the cases to the unit tests in the BaseOnHeapStoreTest unit tests.

rishabhmonga:
I have reviewed the unit tests in BaseOnHeapStore and included a few more to add coverage to StoreEventListener. Additionally I have reverted the changes in StoreEvntListenerTest as these were not required as part of the issue. The PR is updated and needs to reviewed

rishabhmonga:
@aurbroszniowski: would you please review the changes to the PR

PR #324, opened 2015-03-19
Issue 310

lorban:
Still a bit drafty, still lacks XML config capabilities, but nevertheless usable.


ljacomet:
Build failure is for genuine test failures - please check

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -69,6 +73,27 @@ public void testWikiExample() {
   }
 
   @Test
+  public void testTieredStore() throws Exception {
+    CacheConfiguration<Long, String> tieredCacheConfiguration = newCacheConfigurationBuilder()
+        .persistenceMode(CacheConfiguration.PersistenceMode.SWAP)
+        .withResourcePools(newResourcePoolsBuilder().with("heap").with("disk").build())
```
ljacomet: Would use the following names:
`.withCapacityConstraints(newCapacityConstraintsBuilder().heap(10).disk(100).build())`

xml/src/test/resources/configs/resources-templates.xml
```diff
@@ -0,0 +1,53 @@
...
+    <ehcache:value-type>java.lang.String</ehcache:value-type>
+    <ehcache:resources>
+      <ehcache:heap value="15"/>
+    </ehcache:resources>
+  </ehcache:cache-template>
+
+  <ehcache:cache alias="templatedExplicitHeapResource" usesTemplate="explicitHeapResourceTemplate">
+  </ehcache:cache>
+
+  <ehcache:cache-template name="implicitHeapResourceTemplate">
```
ljacomet: Was it not commented yesterday that you would need at least the `<ehcache:heap value="5"/>` element if you do not have a `<ehcache:resources>` one?

As we said no resource configured means no resource available.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -69,6 +73,27 @@ public void testWikiExample() {
...
+        .addServiceConfig(new OnHeapStoreServiceConfig().storeByValue(true))
```
ljacomet: I would leave the storeByValue at the `CacheConfiguration` level. This then makes declaring the services not needed and thus the cache building more straightforward.

api/src/main/java/org/ehcache/Maintainable.java
```diff
@@ -34,4 +34,9 @@
    * @throws java.lang.IllegalStateException if data structures don't exist
    */
   void destroy();
+
+  /**
+   * Exit maintenance mode
+   */
+  void exit();
```
chrisdennis: I thought Maintainable was closable?  Appears it isn't, but it probably should be.

api/src/main/java/org/ehcache/config/CacheConfiguration.java
```diff
@@ -99,5 +99,22 @@
...
    */
-  boolean isPersistent();
+  PersistenceMode getPersistenceMode();
+
+  /**
+   * Get the {@link ResourcePools resource pools} the {@link Cache} can make use of.
+   *
+   * @return the {@link ResourcePools}
+   */
+  ResourcePools getResourcePools();
```
chrisdennis: If they're exclusive to this cache, should we just call them "Resources"?  And should we use a collection type here?

api/src/main/java/org/ehcache/config/ResourcePool.java
```diff
@@ -0,0 +1,44 @@
...
+   * @return the type.
+   */
+  public String getType();
+
+  /**
+   * Get the unit in which the resource is measured.
+   *
+   * @return the unit.
+   */
+  public String getUnit();
```
chrisdennis: Unit should be a proper type right?

api/src/main/java/org/ehcache/config/ResourcePool.java
```diff
@@ -0,0 +1,44 @@
...
+ * @author Ludovic Orban
+ */
+public interface ResourcePool {
+
+  /**
+   * Get the type of the tracked resource.
+   *
+   * @return the type.
+   */
+  public String getType();
```
chrisdennis: Should be a proper type no?  Maybe an interface, and an enum extending it for the built in ones?

xml/src/test/resources/configs/resources-templates.xml
lorban: Yes, I missed it. I'll add it in a subsequent push.

api/src/main/java/org/ehcache/config/ResourcePool.java
```diff
@@ -0,0 +1,44 @@
...
+   * @return the unit.
+   */
+  public String getUnit();
+
+  /**
+   * Get the value measuring the pool size.
+   *
+   * @return the value.
+   */
+  public String getValue();
```
chrisdennis: Would the value ever be anything other than numeric?  Probably always an integer (non fractional) quantity too.

api/src/main/java/org/ehcache/Maintainable.java
lorban: I can replace that exit() method with a extends Closeable if you prefer.

api/src/main/java/org/ehcache/config/ResourcePools.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface ResourcePools {
+
+  /**
+   * Get a specific {@link ResourcePool} based on its type.
+   *
+   * @param resourceType the type of resource the pool is tracking.
+   * @return the {@link ResourcePool}, or null if there is no pool tracking the requested type.
+   */
+  ResourcePool getPoolForResource(String resourceType);
```
chrisdennis: This precludes multiple resource pools of the same type... I understand thats impossible in XML... but might be desirable at a later date.  Plus wouldn't that limitation be better represented as a map against type?

api/src/main/java/org/ehcache/config/CacheConfiguration.java
lorban: I'm not going to start renaming anything unless everyone agrees on what those names should be.

core/src/main/java/org/ehcache/config/ResourcePoolsBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+  private final Map<String, ResourcePool> resourcePools = new HashMap<String, ResourcePool>();
+
+  private ResourcePoolsBuilder() {
+  }
+
+  public static ResourcePoolsBuilder newResourcePoolsBuilder() {
+    return new ResourcePoolsBuilder();
+  }
+
+  public ResourcePoolsBuilder with(String type) {
```
chrisdennis: Why does this builder exist?

impl/src/main/java/org/ehcache/internal/store/DefaultStoreProvider.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @author Ludovic Orban
+ */
+public class DefaultStoreProvider implements Store.Provider {
+  private ServiceProvider serviceProvider;
+
+  @Override
+  public <K, V> Store<K, V> createStore(Store.Configuration<K, V> storeConfig, ServiceConfiguration<?>... serviceConfigs) {
+    ResourcePools resourcePools = storeConfig.getResourcePools();
```
chrisdennis: What would a null resource pool mean?

api/src/main/java/org/ehcache/config/ResourcePool.java
ljacomet: I agree with non fractional, but we probably want long for future compatibility.

api/src/main/java/org/ehcache/Maintainable.java
```diff
@@ -19,7 +19,7 @@
 /**
  * @author Alex Snaps
  */
-public interface Maintainable {
+public interface Maintainable extends AutoCloseable {
```
ljacomet: `AutoCloseable` is not a Java 1.6 type

impl/src/main/java/org/ehcache/internal/store/DefaultStoreProvider.java
lorban: My understanding is that a null resource pool indicates that the kind of resource you're interested in is not available.

A null ResourcePools on the other hand, may be something undesired to prevent null checks at the end of the call stack.

PR #340, opened 2015-03-24
Issue 253

AbfrmBlr:
ThreadProviderService is not part of this request. May we can create a different issue for that.


AbfrmBlr:
We may need to create two issues in backlog for ThreadProviderService and adding corresponding XML config for write beind

ljacomet:
I have nothing against a stable PR that fixes only part of the issue. However, I am not in favor of spinning off other issues - just clearly indicate this PR does not solve all of #253 and work on completing the work through later PRs.

api/src/main/java/org/ehcache/exceptions/CacheWritingException.java
```diff
@@ -27,20 +27,36 @@
 public class CacheWritingException extends RuntimeException {
 
   private static final long serialVersionUID = -3547750364991417531L;
-
-  CacheWritingException() {
+  
+  /**
```
ljacomet: While this will please check style, it is not acceptable. Please add a basic sentence - see other exceptions.
Applies to all the constructors.

api/src/main/java/org/ehcache/spi/loaderwriter/CacheLoaderWriterConfiguration.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+package org.ehcache.spi.loaderwriter;
+
+import org.ehcache.spi.service.ServiceConfiguration;
+
+/**
+ * @author Abhilash
+ *
+ */
+public interface CacheLoaderWriterConfiguration extends ServiceConfiguration<CacheLoaderWriterFactory> {
```
ljacomet: Confused by the name here. Why not `WriteBehindConfiguration`?

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -242,8 +245,16 @@ public void release() {
...
     final CacheLoaderWriter<? super K, V> loaderWriter;
+    CacheLoaderWriter<? super K, V> decorator = null;
     if(cacheLoaderWriterFactory != null) {
       loaderWriter = cacheLoaderWriterFactory.createCacheLoaderWriter(alias, config);
+      CacheLoaderWriterConfiguration cacheLoaderWriterConfiguration = ServiceLocator.findSingletonAmongst(CacheLoaderWriterConfiguration.class, config.getServiceConfigurations().toArray());
+      if(cacheLoaderWriterConfiguration != null && cacheLoaderWriterConfiguration.isWriteBehind()) {
+        final WriteBehindDecoratorLoaderWriterProvider factory = serviceLocator.findService(WriteBehindDecoratorLoaderWriterProvider.class);
+        decorator = factory.createWriteBehindDecoratorLoaderWriter((CacheLoaderWriter<K, V>)loaderWriter, cacheLoaderWriterConfiguration); 
+      }
+      else decorator = loaderWriter;
```
ljacomet: Code style is to use `{ }` even for one liners

api/src/main/java/org/ehcache/spi/loaderwriter/CacheLoaderWriterConfiguration.java
ljacomet: And the same for the `extends ServiceConfiguration<CacheLoaderWriterFactory>` - why is it needed to intersect at this level with the `CacheLoaderWriter` configuration system?

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -242,8 +245,16 @@ public void release() {
...
+
```
ljacomet: Don't you need to also release the decorator if constructed?

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -242,8 +245,16 @@ public void release() {
 
     final CacheLoaderWriterFactory cacheLoaderWriterFactory = serviceLocator.findService(CacheLoaderWriterFactory.class);
     final CacheLoaderWriter<? super K, V> loaderWriter;
+    CacheLoaderWriter<? super K, V> decorator = null;
     if(cacheLoaderWriterFactory != null) {
       loaderWriter = cacheLoaderWriterFactory.createCacheLoaderWriter(alias, config);
+      CacheLoaderWriterConfiguration cacheLoaderWriterConfiguration = ServiceLocator.findSingletonAmongst(CacheLoaderWriterConfiguration.class, config.getServiceConfigurations().toArray());
+      if(cacheLoaderWriterConfiguration != null && cacheLoaderWriterConfiguration.isWriteBehind()) {
+        final WriteBehindDecoratorLoaderWriterProvider factory = serviceLocator.findService(WriteBehindDecoratorLoaderWriterProvider.class);
+        decorator = factory.createWriteBehindDecoratorLoaderWriter((CacheLoaderWriter<K, V>)loaderWriter, cacheLoaderWriterConfiguration);
```
sreekanth-r: Can we avoid type casting of the decorator? Shouldn't decorator also have same type as the CacheLoaderWriter?

impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehindConfig.java
```diff
@@ -0,0 +1,406 @@
...
+  private boolean writeBatching = DEFAULT_WRITE_BATCHING;
+  private int writeBatchSize = DEFAULT_WRITE_BATCH_SIZE;
+  private int retryAttempts = DEFAULT_RETRY_ATTEMPTS;
+  private int retryAttemptDelaySeconds = DEFAULT_RETRY_ATTEMPT_DELAY_SECONDS;
+  private int writeBehindConcurrency = DEFAULT_WRITE_BEHIND_CONCURRENCY;
+  private int writeBehindMaxQueueSize = DEFAULT_WRITE_BEHIND_MAX_QUEUE_SIZE;
+  
+  private String alias;
+  
+  public WriteBehindConfig getWriteBehindConfig(CacheLoaderWriterConfiguration cacheLoaderWriterConfiguration) {
```
ljacomet: Confused by the pattern here. Class has default constructor and non static method that ends up constructing a new instance.
Either have this signature become a constructor signature or move the method to be static and make the constructor private.

ljacomet: At the same time, following my comments on `CacheLoaderWriterConfiguration` changes, I do not see why this is not this config type that is exposed to end users.

api/src/main/java/org/ehcache/spi/loaderwriter/CacheLoaderWriterConfiguration.java
AbfrmBlr: I thought WriteBehind is one special case of CacheLoaderWriter. By default, CacheLoaderWriter provided by the user becomes WriteThrough. So the user just needs to tell in CacheLoaderWriterConfiguration itself that use the provided LoaderWriter as WriteBehind.
And I added this config at api level , so even if we have a CustomLoaderWriterFactory(other than DefaultLoaderWriterFactory) , they need to stick to this ServiceConfig. I have drawn this analogy from CacheEventListeners. Although the name can be changed to whatever that makes sense.

core/src/main/java/org/ehcache/EhcacheManager.java
lorban: If checkstyle doesn't catch that, it's a bug that should be fixed.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehindConfig.java
lorban: I agree about the confusing pattern. My guts tell me the write behind config should have its own builder.

And the class should be renamed to WriteBehindConfiguration like other configuration classes.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehindConfig.java
```diff
@@ -0,0 +1,406 @@
...
+    WriteBehindConfig config = new WriteBehindConfig();
+    // TODO get Alias
+    config.setAlias("");
```
lorban: Indeed, you need something to uniquely identify the queue here. You can leave the todo and open a new task to wire that in later on, but in the meantime I'd generate unique aliases in some way (using a AtomicInteger for instance) to make sure WriteBehind is usable when configure on multiple caches.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehindConfig.java
```diff
@@ -0,0 +1,406 @@
...
+  /**
+   * @return this configuration instance
+   * @see #setWriteCoalescing(boolean)
+   */
+  public boolean getWriteCoalescing() {
+      return writeCoalescing;
+  }
+
+  /**
+   * Sets whether to batch write operations. If set to {@code true}, {@link net.sf.ehcache.writer.CacheWriter#writeAll} and {@code CacheWriter#deleteAll}
```
lorban: Some ugly copy/pated javadoc straight from ehcache 2. That must be cleaned up.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/operations/DeleteOperation.java
```diff
@@ -0,0 +1,101 @@
...
+              getKey().equals(((DeleteOperation<K, V>) other).getKey());
+    } else {
+      return false;
+    }
+  }
+
+  @Override
+  public void throwAway(CacheLoaderWriter<K, V> cacheLoaderWriter,
+      RuntimeException e) {
+    // TODO Do we need this ?
```
lorban: I'm sure we do need this, but I'm not sure what we should do. Maybe calling some method of the resilience strategy?

impl/src/main/java/org/ehcache/loaderwriter/writebehind/operations/WriteOperation.java
```diff
@@ -0,0 +1,129 @@
...
+          ((WriteOperation) other).getKey());
+    } else {
+      return false;
+    }
+  }
+  
+  @Override
+  public void throwAway(CacheLoaderWriter<K, V> cacheLoaderWriter,
+      RuntimeException e) {
+    // TODO Do we need this ?
```
lorban: Same remark as above.

impl/src/test/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindTestBase.java
```diff
@@ -0,0 +1,122 @@
...
+    Map<String, String> entries = new HashMap<String, String>(10);
+    Set<String> keys = new HashSet<String>(10);
+    for(int i=10 ; i<20; i++) {
+      entries.put("test"+i, "test"+i);
+      keys.add("test"+i);
+    }
+    
+    testCache.putAll(entries);
+    try {
+      Thread.sleep(500);
```
lorban: Tests should never have a Thread.sleep() call. This must be reworked.

impl/src/test/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindTestBase.java
```diff
@@ -0,0 +1,122 @@
...
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
lorban: e.printStackTrace() is one of the worst things you can do in a test. If you cannot handle the exception, let it flow out of the method to make the test framework report it. But in this case, there shouldn't be a Thread.sleep() so this comment is only for pedantry.

lorban:
On a general note, there's something wrong with your formatting. Sometimes the lines are broken around 80 characters, sometimes they're much wider. And sometimes you're using 4 spaces indents and sometimes 2 spaces.

impl/src/test/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindTestBase.java
AbfrmBlr: How would I wait to make sure that the updates in WriteBehind Queue finally reaches the SOR ?

impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehindConfig.java
AbfrmBlr: Alias is just for logging purpose , may be we can ignore it and just log the cacheloaderwriter class instead

core/src/main/java/org/ehcache/EhcacheManager.java
AbfrmBlr: Missed that. Will add the same

impl/src/main/java/org/ehcache/loaderwriter/writebehind/operations/DeleteOperation.java
AbfrmBlr: But resilience strategy is something which is supposed to provide resilience in case of failures of internal components of Cache. It excludes the scope of CacheLoaderWriter

ljacomet:
That's a real test failure you have there ... only on Java 1.8 - so either ordering specific or possible a race.

chrisdennis:
You might want to have a look at what was discussed in the public meeting here to see if it encourages any renaming in the config builder aspects of this https://www.youtube.com/watch?v=NEwlUbxSezA

alexsnaps:
Couple of random thing:
 - Please leave "some" reference to the original author(s) of the code you ported from 2.x
 - Get rid of the public modifier on interface methods
 - `org.ehcache.loaderwriter.writebehind.operations.KeyBasedOperation` doesn't need `V`
 - `org.ehcache.loaderwriter.writebehind.operations.SingleOperation#throwAway` would delegate to the `ResilienceStrategy`
 - I'm confused about the `WriteBehindDecoratorLoaderWriter`, why can't we just call into, e.g.,  the `CacheLoaderWriter#writeAll` on `Cache.putAll` ?

xml/src/test/resources/configs/writebehind-cache.xml
```diff
@@ -0,0 +1,59 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<ehcache:config
+    xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
+    xmlns:foo='http://www.example.com/foo'
```
alexsnaps: You don't use that `foo` ns

xml/src/test/java/org/ehcache/config/xml/IntegrationConfigTest.java
```diff
@@ -58,4 +59,33 @@ public void testLoaderWriter() throws ClassNotFoundException, SAXException, Inst
...
+    assertThat(configuration.getCacheConfigurations().containsKey("bar"), is(true));
+    final CacheManager cacheManager = CacheManagerBuilder.newCacheManager(configuration);
+    cacheManager.init();
+    final Cache<Number, String> cache = cacheManager.getCache("bar", Number.class, String.class);
+    assertThat(cache, notNullValue());
+    assertThat(cache.get(1), notNullValue());
+    final Number key = new Long(42);
+    TestCacheLoaderWriter.latch = new CountDownLatch(1);
+    cache.put(key, "Bye y'all!");
+    TestCacheLoaderWriter.latch.await();
```
alexsnaps: These need to timeout... if this ever break, this would lock up for ever...

AbfrmBlr:
Some xs:int 's are still required because their value defaults to 0

alexsnaps:
I will let @ljacomet merge then... but the type you are looking for @AbfrmBlr is `xs:nonNegativeInteger`, not `xs:int`. So, if I'm not missing anything, make them that...

Issue #323, opened 2015-03-18
Fix key and value instance creation in Store SPI tests


ljacomet:
A number of Store SPI tests use the following constructs:
``` java
K key = factory.getKeyType().newInstance();
V value = factory.getValueType().newInstance();
```

But should instead use:
``` java
K key = factory.createKey(42);
V value = factory.createValue(42);
```

Otherwise SPI tests can only run with key and value types having a default constructor.

PR #341, opened 2015-03-24
issue #323 - Fixed key and value instance creation in SPI tests

aurbroszniowski:


ljacomet:
I believe we should introduce some randomness in the value of the parameter passed for key and value creation.

aurbroszniowski:
You want to avoid randomness in parameters otherwise you introduce randomness in the test success/failures

Issue #345, opened 2015-03-26
JSR 107 Cache created from template with offheap does not have offheap


ljacomet:
``` xml
  <ehcache:service>
    <jcache:defaults>
      <jcache:cache name="offheapCache" template="offheapTemplate"/>
    </jcache:defaults>
  </ehcache:service>
  <ehcache:cache-template name="offheapTemplate">
    <ehcache:key-type>java.lang.String</ehcache:key-type>
    <ehcache:value-type>java.lang.String</ehcache:value-type>
    <ehcache:resources>
      <ehcache:heap size="10"/>
      <ehcache:offheap size="10" unit="mb"/>
    </ehcache:resources>
  </ehcache:cache-template>
```

The above does not result in a cache with an off heap tier when invoked as follows:
``` java
    MutableConfiguration<String, String> configuration = new MutableConfiguration<String, String>();
    configuration.setTypes(String.class, String.class);
    Cache<String, String> cache = cacheManager.createCache("offheapCache", configuration);
    cache.put("42", "That's for sure the answer!");
```

PR #358, opened 2015-04-03
Issue #345 : Fixing JSR 107 Cache created from template with offheap doe...

rishabhmonga:
...s not have offheap

Added resourcePools to CacheConfigurationBuilder constructor


AbfrmBlr:
Adding a test case in CacheConfigurationBuilderTest would be a good idea

Issue #303, opened 2015-02-24
Missing Ehcache.getAll tests coverage


ljacomet:
In `EhcacheBasicGetAllTest` there is currently no coverage of `org.ehcache.spi.loaderwriter.CacheLoaderWriter#loadAll` failures with a `BulkCacheLoadingException`.
As these are handled differently, it is worth adding these test cases.

ljacomet:
Actually, there is no handling of `BulkCacheLoadingException` in getAll at all - so this needs to be implemented and covered by tests.

PR #359, opened 2015-04-06
Issue #303: Added test case for BulkCacheLoadingException

palmanojkumar:
Update documentation for Cache with OnHeapStore and CacheLoaderWriter


palmanojkumar:
more test case needs to be added

Issue #278, opened 2015-01-28
Implement default `CacheEventListenerFactory`


alexsnaps:
As stated initially when the feature was implemented:

```
        // XXX this assumes a new instance returned for each call - yet args are always the same. Is this okay?
```

Obviously this isn't right like so... We need to either have the `CacheEventListenerFactory` configure all listeners at once, or pass each `CacheEventListenerConfiguration` to the service.

We need to:
 - Have a a `DefaultCacheEventListenerFactory` à la `DefaultCacheLoaderWriterFactory`
 - Solve the problem around multiple configs per Cache & what they return
 - Enable `CacheEventListener` in XML

ljacomet:
From last public dev meeting: we will support multiple listeners through configuration

lorban:
I added a few more remarks to PR #361. Hopefully you can sort them out, add the required commits to the PR and then get them merged in again.

PR #361, opened 2015-04-07
Issue #278 : Renaming and Improving DefaultCacheEventListenerBuilder functionality

rishabhmonga:
Issue #278 : Incorporated changes as per review comments

Issue #278 : Rebasing

Issue #278 : Fixing test failures from rebasing

Issue #278 : Narrowing ServiceConfiguration type in DefaultCacheEventListenerFactory + cleanUp

Issue #278 : Changes wrt review comments

Issue #278 : Added sample listener to GettingStarted.java

Issue #278 : Renaming and Improving DefaultCacheEventListenerBuilder functionality

### DIFF
--- a/core/src/main/java/org/ehcache/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/EhcacheManager.java
@@ -293,6 +293,8 @@ public class EhcacheManager implements PersistentCacheManager {
       evtService = new DisabledCacheEventNotificationService<K, V>();
     }
 
+    Ehcache<K, V> ehCache = new Ehcache<K, V>(config, store, decorator, evtService, statisticsExecutor,
+        useLoaderInAtomics, LoggerFactory.getLogger(Ehcache.class + "-" + alias));
 
     final CacheEventListenerFactory evntLsnrFactory = serviceLocator.findService(CacheEventListenerFactory.class);
     if (evntLsnrFactory != null) {
@@ -300,9 +302,9 @@ public class EhcacheManager implements PersistentCacheManager {
       ServiceLocator.findAmongst(CacheEventListenerConfiguration.class, config.getServiceConfigurations().toArray());
       for (CacheEventListenerConfiguration lsnrConfig: evtLsnrConfigs) {
         // XXX this assumes a new instance returned for each call - yet args are always the same. Is this okay?
-        final CacheEventListener<K, V> lsnr = evntLsnrFactory.createEventListener(alias, config);
-        evtService.registerCacheEventListener(lsnr, lsnrConfig.orderingMode(), lsnrConfig.firingMode(),
-            lsnrConfig.fireOn());
+        final CacheEventListener<K, V> lsnr = evntLsnrFactory.createEventListener(alias, lsnrConfig);
+          ehCache.getRuntimeConfiguration().registerCacheEventListener(lsnr, lsnrConfig.orderingMode(), lsnrConfig.firingMode(),
+          lsnrConfig.fireOn());
         if (lsnr != null) {
           releasables.add(new Releasable() {
             @Override
@@ -314,7 +316,7 @@ public class EhcacheManager implements PersistentCacheManager {
       }
     }
 
-    return new Ehcache<K, V>(config, store, decorator, evtService, statisticsExecutor, useLoaderInAtomics, LoggerFactory.getLogger(Ehcache.class + "-" + alias));
+    return ehCache;
   }
 
   public void registerListener(CacheManagerListener listener) {
--- a/impl/src/test/java/org/ehcache/GettingStarted.java
+++ b/impl/src/test/java/org/ehcache/GettingStarted.java
@@ -19,14 +19,23 @@ package org.ehcache;
 import org.ehcache.config.CacheConfiguration;
 import org.ehcache.config.CacheConfigurationBuilder;
 import org.ehcache.config.ResourcePoolsBuilder;
+import org.ehcache.config.event.CacheEventListenerBuilder;
+import org.ehcache.config.event.DefaultCacheEventListenerConfiguration;
 import org.ehcache.config.persistence.PersistenceConfiguration;
 import org.ehcache.config.units.EntryUnit;
 import org.ehcache.config.units.MemoryUnit;
+import org.ehcache.event.CacheEvent;
+import org.ehcache.event.CacheEventListener;
+import org.ehcache.event.EventFiring;
+import org.ehcache.event.EventOrdering;
+import org.ehcache.event.EventType;
 import org.ehcache.internal.store.heap.service.OnHeapStoreServiceConfig;
 import org.junit.Test;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.util.HashSet;
+import java.util.Set;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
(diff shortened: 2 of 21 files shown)


impl/src/main/java/org/ehcache/config/event/DefaultCacheEventListenerBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+    return this;
+  }
+
+  public DefaultCacheEventListenerBuilder firingMode(EventFiring eventFiringMode){
+    this.eventFiringMode = eventFiringMode;
+    return this;
+  }
+
+  public DefaultCacheEventListenerBuilder synchronous(){
+    this.eventFiringMode = EventFiring.SYNCHRONOUS;
```
AbfrmBlr: Are we fine with ordered/unordered and synchrous/asynchronous naming or we want to go with firingMode and eventOrdering ?

lorban: IIRC this matches exactly what we agreed upon.

impl/src/main/java/org/ehcache/config/event/DefaultCacheEventListenerBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+import org.ehcache.event.EventType;
+
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author rism
+ */
+public class DefaultCacheEventListenerBuilder {
```
lorban: Why the "Default" prefix?

impl/src/main/java/org/ehcache/config/event/DefaultCacheEventListenerBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+        case EXPIRED:
+          eventSetToFireOn.add(EventType.EXPIRED);
+          break;
+        case UPDATED:
+          eventSetToFireOn.add(EventType.UPDATED);
+          break;
+        case REMOVED:
+          eventSetToFireOn.add(EventType.REMOVED);
+          break;
+        default:
```
lorban: I think default should throw IllegalArgumentException

xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -175,6 +185,40 @@ private void parseConfiguration()
...
+                eventSetToFireOn.add(org.ehcache.event.EventType.EXPIRED);
+                break;
+              case UPDATED:
+                eventSetToFireOn.add(org.ehcache.event.EventType.UPDATED);
+                break;
+              case REMOVED:
+                eventSetToFireOn.add(org.ehcache.event.EventType.REMOVED);
+                break;
+              default:
+                break;
```
lorban: same as above: I think default should throw IllegalArgumentException

xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -302,6 +346,41 @@ public URL getURL() {
...
+            case EXPIRED:
+              eventSetToFireOn.add(org.ehcache.event.EventType.EXPIRED);
+              break;
+            case UPDATED:
+              eventSetToFireOn.add(org.ehcache.event.EventType.UPDATED);
+              break;
+            case REMOVED:
+              eventSetToFireOn.add(org.ehcache.event.EventType.REMOVED);
+              break;
+            default:
```
lorban: Same as above.

lorban:
This is worth adding a GettingStarted sample IMHO.

AbfrmBlr:
findbugs is failing in EhcacheManager. That needs to be fixed.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -300,9 +302,9 @@ public void release() {
       ServiceLocator.findAmongst(CacheEventListenerConfiguration.class, config.getServiceConfigurations().toArray());
       for (CacheEventListenerConfiguration lsnrConfig: evtLsnrConfigs) {
         // XXX this assumes a new instance returned for each call - yet args are always the same. Is this okay?
-        final CacheEventListener<K, V> lsnr = evntLsnrFactory.createEventListener(alias, config);
-        evtService.registerCacheEventListener(lsnr, lsnrConfig.orderingMode(), lsnrConfig.firingMode(),
-            lsnrConfig.fireOn());
+        final CacheEventListener<K, V> lsnr = evntLsnrFactory.createEventListener(alias, lsnrConfig);
+          ehCache.getRuntimeConfiguration().registerCacheEventListener(lsnr, lsnrConfig.orderingMode(), lsnrConfig.firingMode(),
+          lsnrConfig.fireOn());
         if (lsnr != null) {
```
lorban: There is something fishy about this null check. If for whatever reason the CacheEventListenerFactory creates a null listener, we do register it in the cache's runtime config but not in the releasables stack.

Either creating a null listener makes sense and the null check should also encompass the registration in the runtime config, or that doesn't make sense and the null check should be done right after creation and throw an exception.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -212,4 +236,17 @@ private void performAssertions(Cache<Long, String> cache, boolean same) {
     assertThat(s2 == s3, is(same));
   }
 
+  public static class ListenerObject implements CacheEventListener<Object, Object> {
+    private static final Object object = new Object() {
```
lorban: What's the point of that object?

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -212,4 +236,17 @@ private void performAssertions(Cache<Long, String> cache, boolean same) {
...
+      @Override
+      public String toString() {
+        return "class "+ org.ehcache.config.event.DefaultCacheEventListenerConfiguration.class.getName();
+      }
+    };
+
+    @Override
+    public void onEvent(CacheEvent<Object, Object> event) {
+      //noop
```
lorban: Logging something here would make the sample a lot more visual IMHO.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -202,6 +211,21 @@ public void testStoreByValue() {
...
+    DefaultCacheEventListenerConfiguration cacheEventListenerConfiguration = CacheEventListenerBuilder
+        .newEventListenerConfig(ListenerObject.class, EventType.CREATED, EventType.UPDATED)
+        .unordered().asynchronous().build();
+    
+    final CacheManager manager = CacheManagerBuilder.newCacheManagerBuilder()
+        .withCache("foo",
+            CacheConfigurationBuilder.newCacheConfigurationBuilder()
+                .addServiceConfig(cacheEventListenerConfiguration)
+                .buildConfig(Object.class, Object.class)).build(true);
+
```
lorban: You should probably add a few puts and removes to demonstrate that CREATED and UPDATED but not REMOVED events are getting fired.

xml/src/main/java/org/ehcache/config/xml/ConfigurationParser.java
```diff
@@ -266,6 +269,43 @@ public String loaderWriter() {
           }
 
           @Override
+          public List<Listener> listener() {
```
lorban: The method name should be "listeners" and not "listener": use a plural noun as you may return more than one.
Then why a List and not an Iterable backed by a Set? The order in which event types are registered does not matter and you probably don't want twice the same type to be registered, even if that would be harmless since the event types eventually end up in a Set in XmlConfiguration.

xml/src/main/java/org/ehcache/config/xml/ConfigurationParser.java
```diff
@@ -394,6 +434,39 @@ public boolean storeByValueOnHeap() {
...
```
lorban: Same remarks as above.

Issue #339, opened 2015-03-24
Exception when using a persistent cache but not indicating a persistenceMode 


aurbroszniowski:
The follwing will create an exception
```
    PersistentCacheManager persistentCacheManager = CacheManagerBuilder.newCacheManagerBuilder()
        .with(new PersistenceConfiguration(new File(System.getProperty("java.io.tmpdir") + "/myData"))) 
        .withCache("persistent-cache", CacheConfigurationBuilder.newCacheConfigurationBuilder()
            .withResourcePools(newResourcePoolsBuilder()
                .heap(10, EntryUnit.ENTRIES)
                .disk(100, EntryUnit.ENTRIES) // <2>
                .build())
            .buildConfig(Long.class, String.class))
        .build();
```

However it is working when indicating a persistenceMode
```
.persistenceMode(CacheConfiguration.PersistenceMode.CREATE_IF_ABSENT)
```

```
org.ehcache.exceptions.StateTransitionException: Cache 'persistent-cache' creation in EhcacheManager failed.
	at org.ehcache.EhcacheManager.init(EhcacheManager.java:356)
	at org.ehcache.CacheManagerBuilder.newCacheManager(CacheManagerBuilder.java:59)
	at org.ehcache.CacheManagerBuilder.build(CacheManagerBuilder.java:48)
	at org.ehcache.GettingStarted.persistentCacheManager(GettingStarted.java:93)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
	at java.lang.reflect.Method.invoke(Method.java:597)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
[...]

PR #364, opened 2015-04-09
Fixing #339 & adding persistence mode in xml

AbfrmBlr:


impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -178,6 +178,29 @@ public void testPersistentDiskCache() {
...
+    PersistentCacheManager persistentCacheManager = CacheManagerBuilder.newCacheManagerBuilder()
+        .with(new PersistenceConfiguration(new File(System.getProperty("java.io.tmpdir") + "/persistent-cache-data")))
+        .withCache("persistent-cache", cacheConfiguration)
+        .build(true);
+
+    Cache<Long, String> cache = persistentCacheManager.getCache("persistent-cache", Long.class, String.class);
+
+    // Comment the following line on subsequent run and see the test pass
+    cache.put(42L, "That's the answer!");
+    assertThat(cache.get(42L), is("That's the answer!"));
```
lorban: The assertion is comparing strings here. While is() actually works (it does ==), it's more correct to use equalTo() (which calls equals()).

xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -156,7 +157,21 @@ private void parseConfiguration()
...
+      if(cacheDefinition.persistenceMode() != null) {
+        CacheConfiguration.PersistenceMode mode = null;
+        switch (cacheDefinition.persistenceMode()) {
+        case CREATE_IF_ABSENT: 
+          mode = CacheConfiguration.PersistenceMode.CREATE_IF_ABSENT ;
+          break;
+        case EXPECT_EXISTS:
+          mode = CacheConfiguration.PersistenceMode.EXPECT_EXISTS;
+          break;
+        default:
```
lorban: I'd prefer to see "case SWAP:" and have the default throw IllegalArgumentException.

xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -292,6 +307,21 @@ public URL getURL() {
...
+    if(cacheTemplate.persistenceMode() != null) {
+      CacheConfiguration.PersistenceMode mode = null;
+      switch (cacheTemplate.persistenceMode()) {
+      case CREATE_IF_ABSENT: 
+        mode = CacheConfiguration.PersistenceMode.CREATE_IF_ABSENT ;
+        break;
+      case EXPECT_EXISTS:
+        mode = CacheConfiguration.PersistenceMode.EXPECT_EXISTS;
+        break;
+      default:
```
lorban: same remark as above

xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
AbfrmBlr: But already xsd takes care of all the validations, so in any case default maps to only one that is SWAP. And SWAP is anyways the default mode of persistence.

lorban: Call me paranoid, but I think an XML config file that doesn't specify the XSD is going to be accepted by the code. Plus if the XSD ever gets out of sync with the code sometime in the future (think of a bug in a future code change) then the bug would be a lot easier to track down.
Or do you think I'm being unreasonably paranoid?

ljacomet: I agree with Ludovic here

ljacomet:
Needs to be reworked - closing this

PR #370, opened 2015-04-09
Issue 346: Serviceify CacheEventListenerService

palmanojkumar:


lorban:
This needs rebasing, otherwise I think it's ready to be merged in.

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationServiceProviderConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+package org.ehcache.internal.events;
+
+import org.ehcache.events.CacheEventNotificationListenerServiceProvider;
+import org.ehcache.spi.service.ServiceConfiguration;
+
+/**
+ * @author palmanojkumar
+ *
+ */
+public class CacheEventNotificationServiceProviderConfiguration implements ServiceConfiguration<CacheEventNotificationListenerServiceProvider> {
```
alexsnaps: Is this class even useful? I could be missing something, but it doesn't configure anything, and I don't see it used neither... or is it?

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationListenerServiceProviderImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import org.ehcache.spi.service.ThreadPoolsService;
+
+/**
+ * @author palmanojkumar
+ *
+ */
+public class CacheEventNotificationListenerServiceProviderImpl implements CacheEventNotificationListenerServiceProvider {
+
+  private ServiceConfiguration<?> config;
+  private ServiceProvider serviceProvider;
```
alexsnaps: Don't these fields need to be `volatile`?

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationListenerServiceProviderImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                                                         threadPoolsService.getEventsUnorderedDeliveryExecutor());
+    } else {
+      return new DisabledCacheEventNotificationService<K, V>();
+    }
+  }
+
+  @Override
+  public <K, V> void releaseCacheEventNotificationService(CacheEventNotificationService<K, V> cenlService) {
+    if (cenlService != null) {
+      cenlService.releaseAllListeners();
```
alexsnaps: I think this call renders `org.ehcache.config.CacheRuntimeConfiguration#releaseAllEventListeners` obsolete... even though it's still invoked from `org.ehcache.EhcacheManager#closeEhcache`, effectively rendering this `org.ehcache.events.CacheEventNotificationService#releaseAllListeners` invoked twice, or am I reading this code wrong?

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationServiceProviderConfiguration.java
alexsnaps: A test to prove either way would probably be in order...

palmanojkumar: Currently this class is not used anywhere. This config class is intended for service  lookup using ServiceProvider#findServiceFor method.

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationListenerServiceProviderImpl.java
palmanojkumar: I looked up service usage pattern and dont find specific reason to make it volatile. ServiceLocator  acquires lock for starting or stopping services. Let me know if I missed necessity of volatile here

alexsnaps: Right, but what provides the visibility guarantees to `. createCacheEventNotificationService` that access the `serviceProvider` field? 

hint:I actually think there _is_ one, but I have not verified whether it's entirely correct and further more, if you do see which it is, you'd probably agree it's _very_ brittle.

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationServiceProviderConfiguration.java
alexsnaps: Right, it _could_ be used that way... but why _would_ it? and more so, I don't think it _is_ used that way at all...

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationListenerServiceProviderImpl.java
palmanojkumar: You are correct here.. currently CacheEventNotificationService#releaseAllListeners is invoked twice which is wrong and needs to be fixed.
Ideally provider which creates resource instances should handle resource closure as well. Currently CacheRuntimeConfiguration#releaseAllEventListeners is used only for closing ehcache. If there are no more (in future as well) consumer of this contract then provider should handle this resource closure.
your thought?

I will upload next change based on decision on this.

alexsnaps: I'd get rid of that call in `.close()` indeed. If you want to do reference counting to eagerly close things down... Sure why not, but I guess it'd only make sense if you also only lazily "start" the other resources. I wouldn't do this as part of this though... Feel free to create an issue for it, if you think it's worthwhile!

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationServiceProviderConfiguration.java
lorban: If you believe keeping this class makes sense, maybe you should come up with a sample use-case and add it to GettingStarted?

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationListenerServiceProviderImpl.java
lorban: I think there's still some confusion around the services layer. My original understanding was that all the services are to be initialized eagerly during cache manager initialization and not used anymore once everything is up and running.

I now realize this is wrong and that the services could be called at any time after the cache manager is initialized, simply by creating new caches with CacheManager.createCache(). I therefore agree that marking such fields as volatile is the best thing to do, but there are other services with such problem that would need to be fixed too. I'll file a bug about them.

Regarding this PR, I agree with Alex that those fields should be volatile.

alexsnaps: Oh, right... yes. Indeed, the `Service` could be accessed anytime. During a `.createCache` or actually a `Service` using another, may invoke methods on it at "any" time. I guess we should add a one liner to the `Service` javadoc, mentioning implementation _have to_ be threadsafe

core/src/test/java/org/ehcache/EhcacheManagerTest.java
```diff
@@ -414,6 +456,70 @@ public void testDoesNotifyAboutLifecycle() {
     cacheManager.close();
     verify(listener).stateTransition(Status.AVAILABLE, Status.UNINITIALIZED);
   }
+  
+  @Test
+  public void testCloseNoLoaderWriterAndCacheEventListener() throws Exception {
```
palmanojkumar: Test case to ensure that EhcacheManager.close() works properly

core/src/test/java/org/ehcache/EhcacheManagerTest.java
```diff
@@ -414,6 +456,70 @@ public void testDoesNotifyAboutLifecycle() {
...
+      <K, V> Ehcache<K, V> createNewEhcache(final String alias, final CacheConfiguration<K, V> config, final Class<K> keyType, final Class<V> valueType, Deque<Releasable> releasables) {
+        final Ehcache<K, V> ehcache = super.createNewEhcache(alias, config, keyType, valueType, releasables);
+        return spy(ehcache);
+      }
+    };
+    cacheManager.init();
+    Ehcache<Object, Object> testCache = (Ehcache<Object, Object>) cacheManager.getCache("foo", Object.class, Object.class);
+    cacheManager.close();
+    verify(testCache).close();
+    // verify(mockStore, times(1)).close();
```
palmanojkumar: currently store.close is invoked twice for EhcacheManager.close() and hence above verify is commented out.
once store.close invocation is fixed. we should enable this verification as well.

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationServiceProviderConfiguration.java
palmanojkumar: As discussed, Removed CacheEventNotificationServiceProviderConfiguration.java

api/src/main/java/org/ehcache/config/CacheRuntimeConfiguration.java
```diff
@@ -64,9 +64,4 @@ void registerCacheEventListener(CacheEventListener<? super K, ? super V> listene
    */
   void deregisterCacheEventListener(CacheEventListener<? super K, ? super V> listener);
   
-  /**
-   * Remove all registered event listeners.
-   * Invoked by {@link org.ehcache.CacheManager} when a {@link org.ehcache.Cache} is being removed from it.
-   */
-  void releaseAllEventListeners();
```
palmanojkumar: As discussed, removed this method

core/src/test/java/org/ehcache/EhcacheManagerTest.java
ljacomet: Do we really care that close is invoked only once? If not, you could leave the verify without specifying a `times(1)`

palmanojkumar: i thin we should track store.close call. Closing DiskStore multiple times can be problematic. Currently store.close is invoked twice, which remain undetected for some time.

ljacomet: Then I would use `verify(mockStore, atLeastOnce()).close(); // TODO Remove once 346 is resolved`
I really dislike having commented code part of a commit, and even more without something to attract attention to the fact it needs to be fixed later.

core/src/test/java/org/ehcache/EhcacheManagerTest.java
```diff
@@ -414,6 +457,70 @@ public void testDoesNotifyAboutLifecycle() {
...
+      <K, V> Ehcache<K, V> createNewEhcache(final String alias, final CacheConfiguration<K, V> config, final Class<K> keyType, final Class<V> valueType, Deque<Releasable> releasables) {
+        final Ehcache<K, V> ehcache = super.createNewEhcache(alias, config, keyType, valueType, releasables);
+        return spy(ehcache);
+      }
+    };
+    cacheManager.init();
+    Ehcache<Object, Object> testCache = (Ehcache<Object, Object>) cacheManager.getCache("foo", Object.class, Object.class);
+    cacheManager.close();
+    verify(testCache).close();
+    verify(mockStore, atLeastOnce()).close(); //TODO : remove atLeastOnce() once issue-346 is resolved
```
ljacomet: Issue is 348, my bad for the confusion

core/src/test/java/org/ehcache/EhcacheManagerTest.java
```diff
@@ -414,6 +457,70 @@ public void testDoesNotifyAboutLifecycle() {
...
+      <K, V> Ehcache<K, V> createNewEhcache(final String alias, final CacheConfiguration<K, V> config, final Class<K> keyType, final Class<V> valueType, Deque<Releasable> releasables) {
+        final Ehcache<K, V> ehcache = super.createNewEhcache(alias, config, keyType, valueType, releasables);
+        return spy(ehcache);
+      }
+    };
+    cacheManager.init();
+    Ehcache<Object, Object> testCache = (Ehcache<Object, Object>) cacheManager.getCache("foo", Object.class, Object.class);
+    cacheManager.close();
+    verify(testCache).close();
+    verify(mockStore, atLeastOnce()).close(); //TODO : remove atLeastOnce() once issue-348 is resolved
```
palmanojkumar: fixed the todo comment.

PR #371, opened 2015-04-14
Getting started example for write behind #253

AbfrmBlr:
This needs to go in for #253 to complete

### DIFF
--- a/core/src/main/java/org/ehcache/config/writebehind/WriteBehindConfiguration.java
+++ b/core/src/main/java/org/ehcache/config/writebehind/WriteBehindConfiguration.java
@@ -36,19 +36,50 @@ public class WriteBehindConfiguration implements ServiceConfiguration<WriteBehin
   private int writeBehindMaxQueueSize;
   
   public WriteBehindConfiguration(int minWriteDelay, int maxWriteDelay, int rateLimitPerSecond, boolean writeCoalescing, 
-      boolean writeBatching, int writeBatchSize, int retryAttempts,int retryAttemptDelaySeconds,
+      int writeBatchSize, int retryAttempts,int retryAttemptDelaySeconds,
       int writeBehindConcurrency, int writeBehindMaxQueueSize) {
+    
+    validateParams(minWriteDelay, maxWriteDelay, rateLimitPerSecond, writeBatchSize, retryAttempts, 
+        retryAttemptDelaySeconds, writeBehindConcurrency, writeBehindMaxQueueSize);
     
     this.minWriteDelay = minWriteDelay;
     this.maxWriteDelay = maxWriteDelay;
     this.rateLimitPerSecond = rateLimitPerSecond;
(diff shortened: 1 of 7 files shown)


lorban:
This sample is good but not very visual as the WriteBehindTestLoaderWriter loader/writer does all its magic internally without logging anything.

Maybe using an inline CacheLoaderWriter that extends from WriteBehindTestLoaderWriter to add some INFO logs would be more useful for a GettingStarted example? Or do you guys think that would be pointless / over the top?

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -243,4 +302,82 @@ public void onEvent(CacheEvent<Object, Object> event) {
...
+    @Override
+    public Map<K, V> loadAll(Iterable<? extends K> keys) throws Exception {
+      throw new UnsupportedOperationException("Implement me!");
+    }
+
+    @Override
+    public void write(K key, V value) throws Exception {
+      lock.writeLock().lock();
+      try {
+        LOGGER.info("Key - '{}' ,Value - '{}' successfully wriiten", key, value);
```
ljacomet: Typo: double `i`

core/src/main/java/org/ehcache/config/writebehind/WriteBehindConfigurationBuilder.java
```diff
@@ -58,19 +58,31 @@ public WriteBehindConfigurationBuilder segment(int concurrency) {
...
   }
   
-  public WriteBehindConfigurationBuilder batch(int batchSize) {
+  public WriteBehindConfigurationBuilder disableCoalescing() {
+    this.writeCoalescing = false;
+    
+    return this;
+  }
+  
+  public WriteBehindConfigurationBuilder enableBatching(int batchSize) {
```
ljacomet: I would revert the change with regards to batching. `batch(1)` was a good way of saying no watching. It would require the logic to be improved though - including refusing values 0 or below.

AbfrmBlr:
test this please

core/src/main/java/org/ehcache/config/writebehind/WriteBehindConfiguration.java
```diff
@@ -36,21 +36,52 @@
...
+    if(writeBehindMaxQueueSize < 0) {
+      throw new IllegalArgumentException("WriteBehind queue size cannot be less than 0.");
+    }
+    if(retryAttempts < 0) {
+      throw new IllegalArgumentException("RetryAttempts cannot be less than 0.");
+    }
+    if(retryAttemptDelaySeconds < 1) {
+      throw new IllegalArgumentException("RetryAttemptDelaySeconds cannot be less than 1.");
+    }
+    if(writeBatchSize < 0) {
```
ljacomet: Can't even be less than one

PR #374, opened 2015-04-15
Issue 312: Tiering SPI test suite

aurbroszniowski:


core-spi-test/src/main/java/org/ehcache/internal/tier/AuthoritativeTierComputeIfAbsentAndFault.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+      });
+
+      for (long seed = 2L; seed < 15000; seed++) {
+        tier.put(factory.createKey(seed), factory.createValue(seed));
+      }
+      assertThat(tier.get(key), is(nullValue()));
+
+    } catch (CacheAccessException e) {
+      System.err.println("Warning, an exception is thrown due to the SPI test");
```
lorban: That's a sad error handling. Is there a reason why the test method cannot be marked "throws Exception"?

PR #375, opened 2015-04-16
Issue 311

lorban:
The expiration concept has been moved down to the ValueHolder interface and a new AbstractValueHolder class has been introduced to centralize all the ValueHolder's timestamp-handling logic.

This simplifies the current ValueHolder implementations, future ones as well as the CachingTier / AuthoritativeTier contracts of the CacheStore since the AuthoritativeTier doesn't have to query the CachingTier anymore to get the expiration timestamp of a ValueHolder.


impl/src/main/java/org/ehcache/internal/store/offheap/OffHeapStore.java
```diff
@@ -141,7 +138,7 @@ public OffHeapStore(final Configuration<K, V> config, Serializer<K> keySerialize
         public OffHeapValueHolder<V> apply(K mappedKey, OffHeapValueHolder<V> mappedValue) {
           long now = timeSource.getTimeMillis();
 
-          if (mappedValue == null || mappedValue.isExpired(now)) {
+          if (mappedValue == null || mappedValue.isExpired(now, OffHeapValueHolder.TIME_UNIT)) {
```
ljacomet: Using the `OffHeapValueHolder.TIME_UNIT` here seems fishy without also converting now into it. Why don't you simply pass `TimeUnit.MILLISECONDS`?
This remark applies to all calls like this and also to equivalent ones in `OnHeapStore`.

ljacomet:
I am confused by the concept of defining a native time unit in the `AbstractValueHolder` and also by its uses. Can you elaborate on what it is supposed to achieve?

impl/src/main/java/org/ehcache/internal/store/offheap/OffHeapStore.java
lorban: Indeed, in this particular case, the should be TimeUnit.MILLISECONDS and not OffHeapValueHolder.TIME_UNIT since the 'now' timestamp is coming from the timesource and not from the offheap store's value holder.

If you're speaking only about the isExpired() method, you're right that the time unit should always be TimeUnit.MILLISECONDS which makes me believe that we could get rid of this extra parameter.

lorban:
The AbstractValueHolder is useful when the value holder of a store fits its timestamps in long primitives. Those timestamps don't necessarily have to store milliseconds, they could store nanos or full seconds, hence the AbstractValueHolder stores the timestamps but not their unit.

I agree that this logic is a rather convoluted and/or over the top. The idea was to have a central place for all the timestamp comparison logic while still allowing the flexibility for stores to use different time units. I may have failed at that task...

lorban:
test this please

ljacomet:
test this please

ljacomet:
test this please

PR #379, opened 2015-04-21
issue-261: Cleaning up commit set for clarity..

sunnydas:
..Basic intention of testing event mechanisms in some of the EhcacheBasic*Tests


sunnydas:
Still inheriting from EhcacheCrudBase as it seemed the easier and simpler thing to do.

Issue #349, opened 2015-04-01
Improve `DefaultLocalPersistenceProvider`


alexsnaps:
NPE on null arg to constructor
add logging

rishabhmonga:
can IllegalArgumentException be used in this case?

ljacomet:
No, use an NPE when null is passed in. `IllegalArgumentException` is when you get a real value but invalid in the context.

PR #383, opened 2015-04-23
Issue #349 : Improve `DefaultLocalPersistenceProvider`

rishabhmonga:
Throwing IllegalArgumentException when PersistenceConfiguration is null and added logging

### DIFF
--- a/impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
+++ b/impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
@@ -22,12 +22,13 @@ import org.ehcache.internal.store.disk.DiskStorePathManager;
 import org.ehcache.spi.ServiceProvider;
 import org.ehcache.spi.service.LocalPersistenceService;
 import org.ehcache.spi.service.ServiceConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileLock;
-import java.util.UUID;
 
 /**
  * @author Alex Snaps
@@ -39,6 +40,7 @@ public class DefaultLocalPersistenceService implements LocalPersistenceService {
   private FileLock lock;
   private File lockFile;
   private RandomAccessFile rw;
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultLocalPersistenceService.class);
 
   private boolean started;
 
@@ -46,8 +48,7 @@ public class DefaultLocalPersistenceService implements LocalPersistenceService {
     if(persistenceConfiguration != null) {
       rootDirectory = persistenceConfiguration.getRootDirectory();
     } else {
-      // todo: this probably isn't smart... null shouldn't mean that!
-      rootDirectory = new File(System.getProperty("java.io.tmpdir"), UUID.randomUUID() + File.separator);
+      throw new NullPointerException("PersistenceConfiguration cannot be null");
     }
     diskStorePathManager = new DiskStorePathManager(rootDirectory.getAbsolutePath());
   }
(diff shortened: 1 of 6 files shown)


impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -16,18 +16,20 @@
 
 package org.ehcache.internal.persistence;
 
+import org.ehcache.Ehcache;
```
ljacomet: Why the `Ehcache` import?

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -46,8 +49,7 @@ public DefaultLocalPersistenceService(final PersistenceConfiguration persistence
     if(persistenceConfiguration != null) {
       rootDirectory = persistenceConfiguration.getRootDirectory();
     } else {
-      // todo: this probably isn't smart... null shouldn't mean that!
-      rootDirectory = new File(System.getProperty("java.io.tmpdir"), UUID.randomUUID() + File.separator);
+      throw new IllegalArgumentException("PersistenceConfiguration cannot be null");
```
ljacomet: Should be a NPE

ljacomet:
Genuine test failures - please fix

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -39,6 +40,7 @@
   private FileLock lock;
   private File lockFile;
   private RandomAccessFile rw;
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultLocalPersistenceService.class + "-" + "LocalPersistenceService");
```
ljacomet: Why do you add a string to the logger name?

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -77,12 +79,13 @@ public synchronized void stop() {
         // passes on windows
         rw.close();
         if (!lockFile.delete()) {
-          // todo log something?;
+          LOGGER.info("Lock file was not deleted {}.", lockFile.getPath());
```
ljacomet: I am in favour of `WARN` level for this

impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreSPITest.java
```diff
@@ -220,7 +220,12 @@ public float hitRate(TimeUnit unit) {
 
       @Override
       public Store.Provider newProvider() {
-        return new CacheStore.Provider();
+        Store.Provider service = new CacheStore.Provider();
```
ljacomet: Why the service variable name? This is a provider

impl/src/test/java/org/ehcache/internal/store/offheap/OffHeapStoreSPITest.java
```diff
@@ -97,7 +97,9 @@ private ResourcePools getOffHeapResourcePool(Comparable<Long> capacityConstraint
 
       @Override
       public Store.Provider newProvider() {
-        return new OffHeapStore.Provider();
+        Store.Provider service = new OffHeapStore.Provider();
```
ljacomet: Why the service variable name? This is a provider

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
rishabhmonga: I agree with using WARN here, but the dev-guidelines specifically discourage using WARN

lorban: Since this method can throw exceptions on error, why shouldn't this also throw an exception too instead of logging the error?

ljacomet: @lorban and I looked into this a bit closer. Looks like failing to delete the file will not prevent getting a lock when starting. Thus the log level could even stay at DEBUG.

PR #384, opened 2015-04-23
issue-372: fix field visibility in services

palmanojkumar:


lorban:
This is all good, but for completeness sake I would add to the Store.Provider interface's javadoc that implementations have to be thread-safe.

lorban:
This now needs to be rebased.

PR #343, opened 2015-03-25
Issue #261 : Changes in EhcacheBasic*Test to include events.

sunnydas:
Modifications to EhcacheBasic*Test to include event validations.

Issue #261, opened 2015-01-09
Add event notification verification to Ehcache unit tests


cljohnso:
The `Ehcache` unit tests in the `EhcacheBasic*Test` classes do not validate event notification.  This checking needs to be added.

sunnydas:
Have added separate events test for the ehcachebasic*Test(s) , did not revert the original basicTests to exclude events. Please let me know if that is Ok.

sunnydas:
I found it easier to inlcude EhcacheCrudbase in the eventing tests as a few things were reused in the test,

PR #397, opened 2015-05-04
Issue #261

ljacomet:
Take over work from @sunnydas on #261 and incorporate all review remarks from PR #343


core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
```diff
@@ -0,0 +1,216 @@
...
+    final Ehcache<String, String> ehcache = this.getEhcache();
+    try {
+      ehcache.putAll(entries);
+      fail();
+    } catch (NullPointerException e) {
+      // Expected
+    }
+    verify(cacheEventListener, never()).onEvent(Matchers.<CacheEvent<String, String>>any());
+    try {
+      ehcache.getRuntimeConfiguration().deregisterCacheEventListener(cacheEventListener);
```
aurbroszniowski: do we really need to test the failure of the deregistering on every tests? it seems that testing the failure of deregistering a cacheEventListener should be a separate test case

ljacomet: Indeed. Also, I really do not like the fact that this ends up throwing an exception. Will look into this.

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+  @Test
+  public void testPutAll() throws Exception {
+    final Map<String, String> originalStoreContent = getEntryMap(KEY_SET_A, KEY_SET_B);
+    final FakeStore fakeStore = new FakeStore(originalStoreContent);
+    this.store = spy(fakeStore);
+    final Ehcache<String, String> ehcache = this.getEhcache();
+    final Map<String, String> contentUpdates = getAltEntryMap("new_", fanIn(KEY_SET_A, KEY_SET_C, KEY_SET_D));
+    ehcache.putAll(contentUpdates);
+    verify(cacheEventListener, times(18)).onEvent(argThat(isCreatedOrUpdated));
```
aurbroszniowski: maybe it would be clearer to use KEY_SET_A.size() + KEY_SET_C.size() + KEY_SET_D.size() instead of 18

core/src/test/java/org/ehcache/EhcacheBasicPutEventsTests.java
```diff
@@ -0,0 +1,194 @@
...
+    verify(cacheEventListener, never()).onEvent(Matchers.<CacheEvent<String, String>>any());
+    try {
+      ehcache.getRuntimeConfiguration().deregisterCacheEventListener(cacheEventListener);
+      fail();
+    } catch (UnsupportedOperationException e) {
+      //expected
+    }
+  }
+
+  private Ehcache<String, String> getEhcache(final CacheLoaderWriter<String, String> cacheLoaderWriter) {
```
aurbroszniowski: this could be extracted to EhcacheBasicCrudBase in order to avoid code duplication

core/src/test/java/org/ehcache/EhcacheBasicPutIfAbsentEventsTest.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  @Test
+  public void testPutIfAbsentNoStoreEntryCacheAccessExceptionHasCacheLoaderWriterEntry() throws Exception {
+    final FakeStore fakeStore = new FakeStore(Collections.<String, String>emptyMap());
+    this.store = spy(fakeStore);
+    doThrow(new CacheAccessException("")).when(this.store).computeIfAbsent(eq("key"), getAnyFunction());
+    final Ehcache<String, String> ehcache = this.getEhcache(this.cacheLoaderWriter);
+    ehcache.putIfAbsent("key", "value");
+    verify(cacheEventListener,times(1)).onEvent(argThat(isCreated));
```
aurbroszniowski: the missing space makes me think that the formatting of the whole PR should be double-checked

core/src/test/java/org/ehcache/EhcacheBasicRemoveAllEventsTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+  @Test
+  public void testRemoveAll() throws Exception {
+    final Map<String, String> originalStoreContent = getEntryMap(KEY_SET_A, KEY_SET_B);
+    final FakeStore fakeStore = new FakeStore(originalStoreContent, Collections.singleton("keyA3"));
+    this.store = spy(fakeStore);
+    final Ehcache<String, String> ehcache = this.getEhcache(null);
+    final Set<String> contentUpdates = fanIn(KEY_SET_A, KEY_SET_C);
+    ehcache.removeAll(contentUpdates);
+    verify(cacheEventListener, times(5)).onEvent(argThat(isRemoved));
```
aurbroszniowski: same comment as above, in general, explaining the expectation rather than a number would improve the understanding of the tests

core/src/test/java/org/ehcache/EhcacheBasicRemoveAllEventsTest.java
```diff
@@ -0,0 +1,208 @@
...
+    try {
+      ehcache.getRuntimeConfiguration().deregisterCacheEventListener(cacheEventListener);
+      fail();
+    } catch (UnsupportedOperationException e) {
+      //expected
+    }
+  }
+
+  @Test
+  public void testRemoveAllStoreSomeOverlapCacheAccessExceptionBeforeWriterNoOverlapSomeFail() throws Exception {
```
aurbroszniowski: I don't understand this method name

core/src/test/java/org/ehcache/EhcacheBasicRemoveAllEventsTest.java
```diff
@@ -0,0 +1,208 @@
...
+    final Map<String, String> originalStoreContent = getEntryMap(KEY_SET_A, KEY_SET_B);
+    final FakeStore fakeStore = new FakeStore(originalStoreContent);
+    this.store = spy(fakeStore);
+    doThrow(new CacheAccessException("")).when(this.store)
+        .bulkCompute(getAnyStringSet(), getAnyEntryIterableFunction());
+
+    final Map<String, String> originalWriterContent = getEntryMap(KEY_SET_B, KEY_SET_D);
+    final FakeCacheLoaderWriter fakeLoaderWriter = new FakeCacheLoaderWriter(originalWriterContent, KEY_SET_C);
```
aurbroszniowski: why can't we use a mock here?

core/src/test/java/org/ehcache/EhcacheBasicReplaceValueEventsTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+  @Mock
+  protected CacheEventListener<String, String> cacheEventListener;
+
+  protected CacheEventNotificationService<String, String> cacheEventNotificationService;
+
+  protected IsUpdated<String, String> isUpdated = new IsUpdated<String, String>();
+
+  @Test
+  public void testReplaceValueNullNullNull() {
```
aurbroszniowski: maybe something named more like replaceForNullKeyAndValueByNullValue() ?

core/src/test/java/org/ehcache/EhcacheBasicRemoveAllEventsTest.java
ljacomet: Cause we need the loader to perform some work while invoked before getting to the resilience strategy.

ljacomet:
All review remarks have been taken into consideration

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
```diff
@@ -0,0 +1,173 @@
...
+package org.ehcache;
+
+import static org.ehcache.EhcacheBasicBulkUtil.KEY_SET_A;
+import static org.ehcache.EhcacheBasicBulkUtil.KEY_SET_B;
+import static org.ehcache.EhcacheBasicBulkUtil.KEY_SET_C;
+import static org.ehcache.EhcacheBasicBulkUtil.KEY_SET_D;
+import static org.ehcache.EhcacheBasicBulkUtil.fanIn;
+import static org.ehcache.EhcacheBasicBulkUtil.getAltEntryMap;
+import static org.ehcache.EhcacheBasicBulkUtil.getEntryMap;
+import static org.hamcrest.CoreMatchers.is;
```
aurbroszniowski: one of several unused import

core/src/test/java/org/ehcache/EhcacheBasicPutEventsTests.java
```diff
@@ -0,0 +1,152 @@
...
+import org.junit.Test;
+import org.mockito.Matchers;
+import org.mockito.Mock;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The class tries to test some of the eventing mechanism for basic PUT operations.
+ */
+public class EhcacheBasicPutEventsTests extends EhcacheEventsTestBase {
```
aurbroszniowski: I just realized that the name of the class is inconsistent, the 's' at the end should be removed

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+public class EhcacheBasicPutAllEventsTest extends EhcacheEventsTestBase {
+
+  protected IsCreatedOrUpdated<String, String> isCreatedOrUpdated = new IsCreatedOrUpdated<String, String>();
+
+  @Test
+  public void testPutAllMapNull() throws Exception {
+    final Map<String, String> originalStoreContent = getEntryMap(KEY_SET_A, KEY_SET_B);
+    final FakeStore fakeStore = new FakeStore(originalStoreContent);
+    this.store = spy(fakeStore);
```
aurbroszniowski: I'm a bit picky here, but the tests would be clearer by replacing those 3 lines by a call to a method, such as
```
this.store = getStore(KEY_SET_A, KEY_SET_B);

private Store getStore(Set<String>... sets) {
    Map<String, String> originalStoreContent = getEntryMap(sets);
    FakeStore fakeStore = new FakeStore(originalStoreContent);
    return spy(fakeStore);
}
```

especially since there's a getCache() method to encapsulate the cache creation

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
```diff
@@ -0,0 +1,173 @@
...
+    final Ehcache<String, String> ehcache = getEhcache();
```
aurbroszniowski: Again being picky, it seems that 'store' shouldn't be a class field since it isn't shared amongst multiple methods, so it should a local variable passed as an argument to getEhcache(...)
Otherwise, it is not clear why we need to init a Store

aurbroszniowski:
Is testing the 'deregisterCacheEventListener' method out of scope for this PR?

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
ljacomet: My bad, fixing all

core/src/test/java/org/ehcache/EhcacheBasicPutEventsTests.java
ljacomet: Consider it done

ljacomet:
Yes because it already is done in `org.ehcache.EhcacheEventTest#testLazyStoreEventListening`

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
ljacomet: Doing that, can't use a setup method unfortunately because of initialisation ordering issues - @Mock annotations happening after the class setup is done.

core/src/test/java/org/ehcache/EhcacheBasicPutAllEventsTest.java
ljacomet: As it is inherited from the parent abstract test class, I am going to leave this as-is.

ljacomet:
With regards to the store initialisation, I ended up creating methods along the getEhcache ones in the parent class.
All remarks have been addressed.

Issue #391, opened 2015-04-29
Clean up builders according to design guildelines


ljacomet:
See https://github.com/ehcache/ehcache3/wiki/design.basics#builder-design-guidelines

lorban:
CacheManagerBuilder, CacheConfigurationBuilder and ResourcePoolsBuilder are good candidates to start with. There probably are others too.

ljacomet:
[Relevant conversation](https://www.youtube.com/watch?v=NEwlUbxSezA) during one of our dev meetings

ljacomet:
Key points from the conversation to be generalised in current code:
* All methods calls on builder that return the builder type are expected to make a copy instead of modifying in place. This allows to reuse intermediate builder states for different final states,
* Allow to skip `.build()` call when the result is consumed by another builder,
* Provide named methods for boolean settings or settings based on Enums. To decide against this, look at what kind of method explosion is causes on the builder as whole.

PR #401, opened 2015-05-05
Issue #391 Clean up CacheConfigurationBuilder

ljacomet:
* Always copy when modifying
* Rename addServiceConfig to add and removeServiceConfig to remove
* Overload add to take a Builder<? extends ServiceConfiguration<?>>, introducing that new interface
* WriteBehindConfigurationBuilder and CacheEventListenerBuilder now implements Builder
* Fix all CacheConfigurationBuilder usage to take method return into account

### DIFF
--- a/core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
+++ b/core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
@@ -21,10 +21,12 @@ import org.ehcache.spi.service.ServiceConfiguration;
 
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.Set;
 
 import org.ehcache.expiry.Expirations;
 import org.ehcache.expiry.Expiry;
 
+import static java.util.Collections.emptySet;
 import static org.ehcache.config.ResourcePoolsBuilder.newResourcePoolsBuilder;
 
 /**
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -106,16 +114,17 @@ private CacheConfigurationBuilder(final Expiry<? super K, ? super V> expiry, fin
   }
   
   public CacheConfigurationBuilder<K, V> withClassLoader(ClassLoader classLoader) {
-    this.classLoader = classLoader;
-    return this;
+    return new CacheConfigurationBuilder<K, V>(expiry, classLoader, evictionPrioritizer, evictionVeto, resourcePools, serviceConfigurations);
   }
   
   public CacheConfigurationBuilder<K, V> withResourcePools(ResourcePools resourcePools) {
```
lorban: Do we really want to keep this method?

core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -59,9 +61,14 @@ private CacheConfigurationBuilder(final Expiry<? super K, ? super V> expiry, fin
     this.serviceConfigurations.addAll(serviceConfigurations);
   }
 
-  public CacheConfigurationBuilder<K, V> addServiceConfig(ServiceConfiguration<?> configuration) {
-    serviceConfigurations.add(configuration);
-    return this;
+  public CacheConfigurationBuilder<K, V> add(ServiceConfiguration<?> configuration) {
+    CacheConfigurationBuilder<K, V> otherBuilder = new CacheConfigurationBuilder<K, V>(expiry, classLoader, evictionPrioritizer, evictionVeto, resourcePools, serviceConfigurations);
```
lorban: I would personally prefer to add a 'copy constructor' that takes a CacheConfigurationBuilder as parameter. Not a huge problem but I find that approach a bit simpler.

core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
ljacomet: Yes, it allows users to share ResourcePools instances instead of a mutable builder in case construction of caches happens in different parts of their code.

core/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
ljacomet: Done

ljacomet:
Duh, pushed too fast - need to fix code that was added after my refactoring

PR #407, opened 2015-05-08
Issue #391 Builder alignment

ljacomet:
Work on WriteBehind and CacheEventListener configuration builders.

This concludes issue #391


ljacomet:
PR updated following remarks discussed with @lorban

Issue #320, opened 2015-03-16
Add capacity constraint to `RuntimeConfig`


alexsnaps:
Do it!

rishabhmonga:
As per discussion with Louis following things were to be included with this : 
- add store parameter to CacheEventNotificationListenerServiceProvider
- add validations to updateResourcePools() (for negative values, new tier, disk size less than heap size if both defined with ENTRIES)
- add listener to heap so updateResourcePools() for heap is functional
- add resourcePoolType to events generated from updateResourcePools()
- find a way to allow ehcache to register itself to the store listener if possible. Else leave a comment about the setSource method workaround in CacheEventNotificationService interface

rishabhmonga:
•	create a ResourcePoolMerger which merges all validated ResourcePool(s) and returns a resourcePools object

ljacomet:
PR #408 has been merged.
However, the code as it stands silently applies a heap mutation with a `MemoryUnit` taking the size and interpreting is as a count based change. This should be fixed to consider this task closed. I would simply refuse it at the validation.

PR #408, opened 2015-05-08
Issue #320 : Adding capacity constraint to RuntimeConfiguration

rishabhmonga:


core/src/main/java/org/ehcache/RuntimeConfiguration.java
```diff
@@ -0,0 +1,184 @@
...
+      ResourceType currentResourceType = resourceTypeIterator.next();
+      if (config.getResourcePools().getPoolForResource(currentResourceType) != null
+          && pools.getPoolForResource(currentResourceType) != null
+          && config.getResourcePools().getPoolForResource(currentResourceType).getType()
+          == pools.getPoolForResource(currentResourceType).getType()) {
+        Long oldValue = config.getResourcePools()
+            .getPoolForResource(currentResourceType)
+            .getSize();
+        Long newValue = pools.getPoolForResource(currentResourceType).getSize();
+        fireCacheConfigurationChange(CacheConfigurationProperties.UPDATESIZE, oldValue, newValue);
```
rishabhmonga: if the listener for this event throws an exception, what should our response be to that. Although it depends on how the listener will be implemented, do we try to set the value back to the previous one? Because in that case we will require a setter for resource pool size.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -105,7 +94,6 @@
   private final ResilienceStrategy<K, V> resilienceStrategy;
   private final RuntimeConfiguration runtimeConfiguration;
```
ljacomet: Do not use a raw type here since you made the type generic.

core/src/main/java/org/ehcache/InternalRuntimeConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache;
+
+/**
+ * @author rism
+ */
+public interface InternalRuntimeConfiguration <K, V> {
```
ljacomet: What are `K` and `V` used for?

core/src/main/java/org/ehcache/InternalRuntimeConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * adds listeners for cache configuration changes
+   */
+  boolean addCacheConfigurationListener(CacheConfigurationListener listener);
```
ljacomet: I can't find these methods used anywhere but in test code. What are they for?

core/src/main/java/org/ehcache/RuntimeConfiguration.java
ljacomet: While I understand your question, I have no clue how this code is supposed to work even when there are no exceptions.
What will be the difference for an event updating the heap resource and one the disk resource?

core/src/main/java/org/ehcache/RuntimeConfiguration.java
```diff
@@ -0,0 +1,184 @@
...
+  private final CacheConfiguration<? super K, ? super V> config;
+  private final Class<K> keyType;
+  private final Class<V> valueType;
+  private final EvictionVeto<? super K, ? super V> evictionVeto;
+  private final EvictionPrioritizer<? super K, ? super V> evictionPrioritizer;
+  private final ClassLoader classLoader;
+  private final Expiry<? super K, ? super V> expiry;
+  private final ResourcePools resourcePools;
+  private final CacheEventNotificationService<K, V> eventNotificationService;
+  private Store<K, V> store;
```
ljacomet: Not happy with the `Store` being a field here. From its current use, it could be hidden inside the `CacheEventNotificationService`. And doing that provides a better abstraction for the enabling / disabling of `Store` events.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -340,6 +341,24 @@ public void writeBehindCache() throws ClassNotFoundException {
...
+            .heap(10, EntryUnit.ENTRIES)
+            .disk(10, EntryUnit.ENTRIES).build())
+        .buildConfig(Long.class, String.class);
+    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder().withCache("tieredCache", tieredCacheConfiguration)
+        .with(new PersistenceConfiguration(new File(System.getProperty("java.io.tmpdir") + "/tiered-cache-data")))
+        .build(true);
+
+    Cache<Long, String> tieredCache = cacheManager.getCache("tieredCache", Long.class, String.class);
+    ResourcePools pools = ResourcePoolsBuilder.newResourcePoolsBuilder().heap(20L, EntryUnit.ENTRIES).disk(2L, EntryUnit.ENTRIES).build();
+    tieredCache.getRuntimeConfiguration().updateResourcePools(pools);
```
ljacomet: From what I see in the code, this call will have no effects.
Doing `tieredCache.getRuntimeConfiguration().getResourcePools().getPoolForResource(ResourceType.Core.HEAP).getSize()` after it will still return `10`.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -340,6 +341,24 @@ public void writeBehindCache() throws ClassNotFoundException {
     // end::writeBehindCache[]  
   }
 
+  @Test
+  public void testUpdateResourcesInRuntime() {
```
ljacomet: The point of a GettingStarted example is that it is documented. Please create the necessary call outs in documentation to indicate what you are illustrating.

core/src/main/java/org/ehcache/RuntimeConfiguration.java
ljacomet: Also, no validation whatsoever? I can resize off heap, set a negative value, size my disk smaller than heap, ...

rishabhmonga:
test this please

ljacomet:
test this please

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -100,6 +113,11 @@ public void releaseAllListeners() {
   }
 
   @Override
+  public void setStoreListenerSource(Cache<K, V> source) {
+    storeListener.setSource(source);
+  }
+
```
rishabhmonga: Even though this is unadvised, due to the absence of a better way to register ehcache instance to storeListener, a setSource method has been used here.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Oct 01 10:52:23 CEST 2014
+#Mon May 18 11:26:37 IST 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.1-bin.zip
```
ljacomet: Drop changes on this file and the one on gradle-wrapper from this PR please.

api/src/main/java/org/ehcache/config/CacheRuntimeConfiguration.java
```diff
@@ -63,5 +56,12 @@ void registerCacheEventListener(CacheEventListener<? super K, ? super V> listene
    * @throws java.lang.IllegalStateException if the listener isn't already registered
    */
   void deregisterCacheEventListener(CacheEventListener<? super K, ? super V> listener);
-  
+
```
ljacomet: Drop that blank line to align on code formatting

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
```diff
@@ -312,6 +313,12 @@ public void disableStoreEventNotifications() {
     }
   }
 
+  @Override
+  public CacheConfigurationChangeListener getConfigurationListener() {
+//    throw new UnsupportedOperationException("TODO Implement me!");
+    return null;
```
ljacomet: I see no general comment on this PR indicating it is a partial fix although this code will end up causing exceptions to happen, notably an NPE when trying to update resources on a tiered cache.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -339,6 +342,44 @@ public void writeBehindCache() throws ClassNotFoundException {
     // end::writeBehindCache[]  
   }
 
+  @Test
+  public void testUpdateResourcesInRuntime() throws InterruptedException {
```
ljacomet: I do not see this example being used in the documentation.

impl/src/test/java/org/ehcache/GettingStarted.java
```diff
@@ -339,6 +342,44 @@ public void writeBehindCache() throws ClassNotFoundException {
...
+    CacheEventListenerConfigurationBuilder cacheEventListenerConfiguration = CacheEventListenerConfigurationBuilder
+        .newEventListenerConfiguration(ListenerObject.class, EventType.EVICTED).unordered().asynchronous();
+
+    CacheConfiguration<Long, String> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
+        .add(cacheEventListenerConfiguration)
+        .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
+            .heap(10L, EntryUnit.ENTRIES).build()).buildConfig(Long.class, String.class);
+
+    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder().withCache("cache", cacheConfiguration)
+        .with(new PersistenceConfiguration(new File(System.getProperty("java.io.tmpdir") + "/cache-data")))
```
ljacomet: Why the persistence config? It is noise in this context.

core/src/main/java/org/ehcache/config/ResourcePoolsBuilder.java
```diff
@@ -38,6 +38,23 @@ public static ResourcePoolsBuilder newResourcePoolsBuilder() {
     return new ResourcePoolsBuilder();
   }
 
+  public static ResourcePoolsBuilder newResourcePoolsBuilder(ResourcePools pools) {
+    ResourcePoolsBuilder poolsBuilder = new ResourcePoolsBuilder();
+    if (pools.getPoolForResource(ResourceType.Core.HEAP) != null) {
```
ljacomet: Why hardcode the resources you consider? You should leverage the `org.ehcache.config.ResourcePools#getResourceTypeSet` method that you added.

impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
ljacomet: Also, when doing partial commits, they should not result in causing issues , having no-ops is acceptable but ending up with exceptions is not.

rishabhmonga: As decided, the scope of this commit only permits updating onheap. However the exceptions on other cases will be removed.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -335,12 +338,14 @@ public void close() {
...
         }
       }
+      evtService.setStoreListenerSource(ehCache);
     }
 
     for (LifeCycled lifeCycled : lifeCycledList) {
       ehCache.addHook(lifeCycled);
     }
 
+
```
ljacomet: Change not needed - please clean up

core/src/main/java/org/ehcache/config/ResourcePoolsBuilder.java
```diff
@@ -38,6 +38,15 @@ public static ResourcePoolsBuilder newResourcePoolsBuilder() {
     return new ResourcePoolsBuilder();
   }
 
+  public static ResourcePoolsBuilder newResourcePoolsBuilder(ResourcePools pools) {
+    ResourcePoolsBuilder poolsBuilder = new ResourcePoolsBuilder();
+    for (ResourceType currentResourceType : pools.getResourceTypeSet()) {
+      poolsBuilder = poolsBuilder.with(currentResourceType, pools.getPoolForResource(currentResourceType).getSize(),
+          pools.getPoolForResource(currentResourceType).getUnit(), currentResourceType.isPersistable());
```
ljacomet: It's not because the resource type is persistable that it has been declared persistent. Read the value from the `ResourcePool` - which you should extract to a local var to not have to call it three times.

core/src/main/java/org/ehcache/spi/cache/Store.java
```diff
@@ -495,6 +496,8 @@
 
   }
 
+  public CacheConfigurationChangeListener getConfigurationListener();
```
ljacomet: Adding a method on an SPI interface requires adding javadoc to it.

core/src/main/java/org/ehcache/util/ResourcePoolMerger.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.ehcache.config.ResourcePools;
+import org.ehcache.config.ResourcePoolsBuilder;
+import org.ehcache.config.ResourceType;
+import org.ehcache.config.units.EntryUnit;
+
+/**
+ * @author rism
+ */
+public class ResourcePoolMerger {
```
ljacomet: This class deserves unit tests

core/src/test/java/org/ehcache/CacheConfigurationChangeListenerTest.java
```diff
@@ -0,0 +1,102 @@
...
+    cache.close();
+  }
+
+  @Test
+  public void testCacheConfigurationChangeFiresEvent () {
+    Listener configurationListener = new Listener();
+    this.runtimeConfiguration.addCacheConfigurationListener(configurationListener);
+    this.cache.getRuntimeConfiguration().updateResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
+        .heap(10L, EntryUnit.ENTRIES).build());
+    assertThat(configurationListener.eventSet.size(), is(lessThan(2)) );
```
ljacomet: Why `lessThan(2)`? How is `0` a valid state in this test?

docs/user/index.adoc
```diff
@@ -78,6 +78,20 @@ include::../../impl/src/test/java/org/ehcache/GettingStarted.java[tag=persistent
...
 <2> You define a resource pool for the disk.
 
+==== Update-Resource-Pools
+
+[source,java,indent=0]
+----
+include::../../impl/src/test/java/org/ehcache/GettingStarted.java[tag=updateResourcesInRuntime]
+----
+
+<1> If you wish to update capacity of resource pools, 'updateResourcePools()' method in runtime configuration can be of help.
```
ljacomet: None of these callouts exist in the GettingStarted code - please add them

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -83,13 +86,21 @@
   private final Serializer<V> valueSerializer;
   private final Serializer<K> keySerializer;
 
-  private final long capacity;
+  private long capacity;
```
ljacomet: Shouldn't this be volatile now?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -126,7 +137,7 @@ public OnHeapStore(final Configuration<K, V> config, TimeSource timeSource, bool
       this.valueSerializer = null;
       this.keySerializer = null;
     }
-    
+
```
ljacomet: Please revert all these formatting changes, they clutter this change set

core/src/main/java/org/ehcache/config/ResourcePoolsBuilder.java
```diff
@@ -38,6 +38,19 @@ public static ResourcePoolsBuilder newResourcePoolsBuilder() {
     return new ResourcePoolsBuilder();
   }
 
+  public static ResourcePoolsBuilder newResourcePoolsBuilder(ResourcePools pools) {
+    ResourcePoolsBuilder poolsBuilder = new ResourcePoolsBuilder();
+    for (ResourceType currentResourceType : pools.getResourceTypeSet()) {
+      Boolean persistence = false;
+      if(currentResourceType.isPersistable()) {
```
ljacomet: Why the check? Trust that the pools you get is valid.

If it was not, the proper handling would be to throw as otherwise you actually change the pools definition you are copying.

core/src/test/java/org/ehcache/util/ResourcePoolsMergerTest.java
```diff
@@ -0,0 +1,91 @@
...
+    existingSet.add(ResourceType.Core.HEAP);
+    existingSet.add(ResourceType.Core.OFFHEAP);
+    toBeUpdatedSet.add(ResourceType.Core.OFFHEAP);
+    when(existing.getResourceTypeSet()).thenReturn(existingSet);
+    when(toBeUpdated.getResourceTypeSet()).thenReturn(toBeUpdatedSet);
+    ResourcePoolMerger merger = new ResourcePoolMerger();
+    try {
+      merger.validateAndMerge(existing, toBeUpdated);
+    } catch (Exception e) {
+      assertThat(e instanceof UnsupportedOperationException, is(true));
```
ljacomet: Catch the proper exception type directly and assert that the errors indeed relates to off heap.

core/src/test/java/org/ehcache/util/ResourcePoolsMergerTest.java
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      merger.validateAndMerge(existing, toBeUpdated);
+    } catch (Exception e) {
+      assertThat(e instanceof IllegalArgumentException, is(true));
+    }
+  }
+
+  @Test
+  public void testUpdatingOffHeap() {
+    ResourcePools existing = mock(ResourcePools.class);
```
ljacomet: Mocking is overkill in this context, just use the builder directly.

ljacomet:
test this please

rishabhmonga:
test this please

rishabhmonga:
test this please

PR #411, opened 2015-05-11
Issue #216 CacheConfiguration to JSR-107 Configuration bridging

ljacomet:
This covers part of issue #216.

The use of Ehcache 3 templates will be in another pull request.


ljacomet:
test this please

PR #415, opened 2015-05-21
Issue #216 Finalise Ehcache - JSR-107 integration

ljacomet:
* Support for template  override of JSR-107 configuration
* Various clean ups to get there


ljacomet:
Looks like I missed running the TCK locally - investigating

ljacomet:
Rebased and passing TCK - all good for review

palmanojkumar:
Looks good to me.

rishabhmonga:
looks good to me too

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,139 @@
...
+
+  String getCacheTemplateName(String cacheName) {
+    String template = jsr107Service.getTemplateNameForCache(cacheName);
+    if (template != null) {
+      return template;
+    }
+    return jsr107Service.getDefaultTemplate();
+  }
+
+  static class ConfigHolder<K, V> {
```
chrisdennis: Can we make these final?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    this.jsr107Service = jsr107Service;
+    this.cacheLoaderWriterFactory = cacheLoaderWriterFactory;
+    this.log = log;
+  }
+
+  <K, V> ConfigHolder<K, V> mergeConfigurations(String cacheName, Configuration<K, V> configuration) {
+    ConfigHolder<K, V> result = new ConfigHolder<K, V>();
+    CacheConfigurationBuilder<K, V> builder = newCacheConfigurationBuilder();
+    boolean useEhcacheExpiry = true;
```
chrisdennis: Can we move these declaration closer to their usage?  Putting them this far up makes me automatically assume the usage pattern is really complex (but it isn't).  Also I think we can avoid initing them when declaring, which means finding the 'true' condition doesn't require me to walk all the execution paths.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,139 @@
...
+    } catch (Throwable throwable) {
+      MultiCacheException mce = new MultiCacheException(throwable);
+      if (result.cacheResources != null) {
+        result.cacheResources.closeResources(mce);
+      }
+      throw mce;
+    }
+  }
+
+  String getCacheTemplateName(String cacheName) {
```
chrisdennis: Why is this logic part of this class and not part of the Jsr107Service impl?  That would mean we could eliminate the getDefaultTemplate method entirely right?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      result.cacheResources = new CacheResources<K, V>(cacheName, result.jsr107Configuration);
+
+      String templateName = getCacheTemplateName(cacheName);
+      if (xmlConfiguration != null && templateName != null) {
+        CacheConfigurationBuilder<K, V> templateBuilder = xmlConfiguration.newCacheConfigurationBuilderFromTemplate(templateName,
+                        result.jsr107Configuration.getKeyType(), result.jsr107Configuration.getValueType());
+        if (templateBuilder != null) {
+          builder = templateBuilder;
+          log.debug("Configuration of cache {} will be supplemented by template {}", cacheName, templateName);
```
chrisdennis: Supplemented doesn't seem like quite the right word any more.  Maybe "combined" or "merged"?

chrisdennis: The logging level choices in this method seem weird.  The presence of a template is debug, but the overriding of many properties is info?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,139 @@
...
+        expiryPolicy = result.cacheResources.getExpiryPolicy();
+      }
+
+      result.cacheResources = new CacheResources<K, V>(cacheName, result.cacheResources.getCacheLoaderWriter(), expiryPolicy,
+          result.cacheResources.getListenerResources());
+      result.jsr107Configuration = new Eh107CompleteConfiguration<K, V>(result.jsr107Configuration, result.cacheConfiguration, useEhcacheExpiry, useEhcacheLoaderWriter);
+      return result;
+    } catch (Throwable throwable) {
+      MultiCacheException mce = new MultiCacheException(throwable);
+      if (result.cacheResources != null) {
```
chrisdennis: Can we move the try catch around and eliminate the null check here?

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -157,36 +151,35 @@ public Properties getProperties() {
...
-        cacheResources.closeResources(mce);
+        configHolder.cacheResources.closeResources(mce);
         throw mce;
       }
 
       Eh107Cache<K, V> cache = null;
       try {
-        cache = new Eh107Cache<K, V>(cacheName, new Eh107CompleteConfiguration<K, V>(config, ehCache.getRuntimeConfiguration()), cacheResources, ehCache, this, expiry);
+        if (configHolder.useEhcacheLoaderWriter) {
+          configHolder.cacheResources = new CacheResources<K, V>(cacheName, EhcacheHackAccessor.getCacheLoaderWriter((Ehcache<K, V>)ehCache), configHolder.cacheResources.getExpiryPolicy(), configHolder.cacheResources.getListenerResources());
```
chrisdennis: Could there be an old loader writer that we need to close here?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: I am using this object as a result object from a package private class. Also the logic currently forces me to redefine some of these fields during execution. Unless I manage to clean up code to get rid of second reason, I don't think it is worth it.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Makes sense, gives me idea to clean up that logic.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Correct - doing that

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Changing logging of template to info

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
ljacomet: Current logic says so, will do it or refactor logic so that it is not a concern (not initialised)

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Consider it done

ljacomet:
I pushed a relatively large update to this PR which addresses comments from @chrisdennis:
* Move init logic from `CacheResources` to `ConfigurationMerge` so that the former focuses on runtime responsibilities.
* All `CacheResources` and `ConfigHolder` are now final

107/src/main/java/org/ehcache/jsr107/CacheResources.java
```diff
@@ -40,47 +31,23 @@
  */
 class CacheResources<K, V> {
 
+  private final CompleteConfiguration<K, V> configuration;
```
chrisdennis: This variable is unused.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  <K, V> ConfigHolder<K, V> mergeConfigurations(String cacheName, Configuration<K, V> configuration) {
+    final Eh107CompleteConfiguration<K, V> jsr107Configuration = new Eh107CompleteConfiguration<K, V>(configuration);
+
+    Eh107Expiry<K, V> expiryPolicy = null;
+    CacheLoaderWriter<? super K, V> loaderWriter = null;
+    try {
+      CacheConfigurationBuilder<K, V> builder = newCacheConfigurationBuilder();
+      String templateName = jsr107Service.getTemplateNameForCache(cacheName);
```
chrisdennis: We only build off this templateName if the xmlConfiguration is non null... could we move this assignment (and declaration) inside the `if(xmlConfiguration != null)` conditional?  This causes cascading issues with the logging further down... but if the conditional on those logging calls are aligned with the conditional on the creation of the builder then thats not immediately obvious.  It's also not clear to me that they aren't either brittle now, or candidates for becoming brittle later.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -0,0 +1,186 @@
...
+            jsr107Configuration.getKeyType(), jsr107Configuration.getValueType());
+        if (templateBuilder != null) {
+          builder = templateBuilder;
+          log.info("Configuration of cache {} will be supplemented by template {}", cacheName, templateName);
+        }
+      }
+
+      builder = handleStoreByValue(jsr107Configuration, builder);
+
+      final boolean useEhcacheExpiry = !builder.hasDefaultExpiry();
```
chrisdennis: Holy Negation, Batman!

107/src/main/java/org/ehcache/jsr107/CacheResources.java
ljacomet: Fixing

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: This will stay as-is, there is no easy way of conveying that the template related logging will happen only if we actually get a template.

Issue #40, opened 2014-09-04
Wire & honor ARC constraints into OnHeapStore


alexsnaps:

chrisdennis:
This is likely obsolete. Byte-based sizing of heap tiers is being "strongly deprecated" due to the increasing difficulty of sizing in Java 17 and beyond.

PR #417, opened 2015-05-25
Issue #320 : Add capacity constraint to `RuntimeConfig`

rishabhmonga:
Adding ResourceUnit validation

### DIFF
--- a/core/src/main/java/org/ehcache/util/ResourcePoolMerger.java
+++ b/core/src/main/java/org/ehcache/util/ResourcePoolMerger.java
@@ -40,6 +40,10 @@ public class ResourcePoolMerger {
           existing.getPoolForResource(currentResourceType).isPersistent()) {
         throw new IllegalArgumentException("Persistence configuration cannot be updated");
       }
+      if(!toBeUpdated.getPoolForResource(currentResourceType).getUnit()
+          .equals(existing.getPoolForResource(currentResourceType).getUnit())) {
+        throw new UnsupportedOperationException("Updating ResourceUnit is not supported");
+      }
     }
 
     ResourcePoolsBuilder mergedPoolsBuilder = ResourcePoolsBuilder.newResourcePoolsBuilder(existing);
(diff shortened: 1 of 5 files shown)


rishabhmonga:
This PR provides a validation for `ResourceUnit` so as to disallow changes if the resource has `MemoryUnit` as unit of measure and not allow changes to `ResourceUnit` for a resource. Also to be noted, if the `ResourcePools` instance being provided to the `updateResourcePools()` does not contain a resource present in the configuration, the existing configuration for that resource will be retained.

core/src/main/java/org/ehcache/util/ResourcePoolMerger.java
```diff
@@ -40,6 +41,9 @@ public ResourcePools validateAndMerge(ResourcePools existing, ResourcePools toBe
           existing.getPoolForResource(currentResourceType).isPersistent()) {
         throw new IllegalArgumentException("Persistence configuration cannot be updated");
       }
+      if(toBeUpdated.getPoolForResource(currentResourceType).getUnit() instanceof MemoryUnit) {
```
ljacomet: While this test makes sense as long as #40 is not implemented, I believe we can have one expressed in the same way the persistent check is made - disallow changing the resource unit altogether.

core/src/test/java/org/ehcache/util/ResourcePoolsMergerTest.java
```diff
@@ -76,4 +73,24 @@ public void testUpdatingHeapSizeToLargerThanDisk() {
...
   }
+  
+  @Test
+  public void testUpdateResourceUnit() {
+    ResourcePoolsBuilder existingPoolsBuilder = ResourcePoolsBuilder.newResourcePoolsBuilder();
+    existingPoolsBuilder = existingPoolsBuilder.heap(20L, EntryUnit.ENTRIES).disk(500, MemoryUnit.MB);
+    ResourcePools existing = existingPoolsBuilder.build();
+
+    ResourcePoolsBuilder toBeUpdatedPoolsBuilder = ResourcePoolsBuilder.newResourcePoolsBuilder();
+    toBeUpdatedPoolsBuilder = toBeUpdatedPoolsBuilder.disk(2, MemoryUnit.GB);
```
ljacomet: This test fails today but should pass in the future once ARC is implemented.
So I would replace it with a test checking that we do not try to switch a count based resource to a memory based one.

rishabhmonga:
test this please

Issue #392, opened 2015-04-29
Define and apply naming conventions around providers and factories


ljacomet:

lorban:
Have a look at the org.ehcache.spi.service.ServiceFactory file, as it contains:

*ProviderFactory
*FactoryProvider
*ServiceFactory
*ServiceProviderFactory

Those classes (and probably the related ones too) do not respect the Provider / Factory naming convention which adds quite some confusion. They should all be renamed consistently.

ljacomet:
1. `ClassInstanceProvider` subclasses to be name `XxxProvider` and the class creating them `XxxProviderFactory`
2. Ehcache 3 services to be named `YyyService` unless they fall under point 1. Their factory classes to be named `YyyServiceFactory`.
3. Look at existing classes and reuse naming conventions. Signal any discrepancies.

PR #419, opened 2015-05-26
Issue #392 Consistent naming around providers and services

ljacomet:
* Subclasses of ClassInstanceProvider to be named *Provider
* Factories for the above to be named *ProviderFactory
* Factories for other services declared in ServiceFactory file: *ServiceFactory
* ClassInstanceProviderConfiguration renamed to ClassInstanceConfiguration as subtypes are passed in the ClassInstanceProvider when getting an instance from it.
* ClassInstanceProviderFactoryConfiguration renamed to ClassInstanceProviderConfiguration as subtypes are passed in when the ClassInstanceProvider service is started.
* Applied the two renames above to their respective hierarchy


AbfrmBlr:
Should we rename CacheEventListenerFactory and CacheLoaderWriterFactory interfaces as well ?
They don't extend ClassInstanceProvider, only there subclasses extend it.

ljacomet:
@AbfrmBlr Just pushed a commit that takes care of this, good catch

Issue #413, opened 2015-05-18
WriteBehind writer blocks cache put operation


aurbroszniowski:
I define a CacheLoaderWriter that simulate a long write operation, and couple it with a write behind config:

```
public class LongCacheLoaderWriter<K, V> implements CacheLoaderWriter<K, V> {

 ...

 public void write(final K k, final V v) throws Exception {
    try {
      Thread.sleep(4000);
    } catch (InterruptedException e) {
      throw new RuntimeException(e);
    }
  }
...
```

The following code, just doing a cache.put(...) is then blocked by the writer

```
   CacheManagerBuilder<CacheManager> cacheManagerBuilder = CacheManagerBuilder.newCacheManagerBuilder();
    cacheManagerBuilder
        .withCache("writebehindcache", newCacheConfigurationBuilder()
            .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
                .heap(100, EntryUnit.ENTRIES).build())
            .buildConfig(Long.class, String.class));

    cacheManagerBuilder
        .using(WriteBehindConfigurationBuilder.newWriteBehindConfiguration()
            .concurrencyLevel(3).batchSize(4)
            .queueSize(10)
            .build());
    cacheManagerBuilder
        .using(new LongCacheLoaderWriterFactory<Long, String>(datasource));
    CacheManager cacheManager = null;
    try {
      cacheManager = cacheManagerBuilder.build(true);

      Cache<Long, String> cache = cacheManager.getCache("writebehindcache", Long.class, String.class);
      assertThat(datasource.size(), is(equalTo(0)));
      long start = System.currentTimeMillis();
      cache.put(1L, "value");
      long elapsedTime = System.currentTimeMillis() - start;
      assertThat("the cache.put should not be blocked by the write behind cachewriter", elapsedTime, is(lessThan(3000L)));
    } finally {
      if (cacheManager != null) {
        cacheManager.close();
      }
    }
```

rishabhmonga:
I tried recreating the bug by using the following CacheLoaderWriter for simulating write operation:


  public class LongCacheLoaderWriter<K, V> implements CacheLoaderWriter {

    private static final Logger LOGGER = LoggerFactory.getLogger(LongCacheLoaderWriter.class);
    
    @Override
    public Object load(Object key) throws Exception {
      throw new UnsupportedOperationException("Implement me!");
    }

    @Override
    public Map loadAll(Iterable keys) throws Exception {
[...]

PR #427, opened 2015-06-02
Issue #413 : Fixing WriteBehind writer blocks cache put operation

rishabhmonga:
Added validations to disallow cache specific configurations to be added at cache manager level

### DIFF
--- a/impl/src/main/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactory.java
+++ b/impl/src/main/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactory.java
@@ -40,5 +40,4 @@ public class DefaultCacheLoaderWriterFactory extends ClassInstanceProvider<Cache
   public void releaseCacheLoaderWriter(final CacheLoaderWriter<?, ?> cacheLoaderWriter) {
     // noop
   }
-
 }
--- a/impl/src/test/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactoryTest.java
+++ b/impl/src/test/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactoryTest.java
@@ -16,19 +16,28 @@
 
 package org.ehcache.spi.loaderwriter;
 
+import org.ehcache.Cache;
 import org.ehcache.CacheManager;
 import org.ehcache.CacheManagerBuilder;
 import org.ehcache.config.CacheConfiguration;
 import org.ehcache.config.CacheConfigurationBuilder;
 import org.ehcache.config.DefaultConfiguration;
+import org.ehcache.config.ResourcePoolsBuilder;
 import org.ehcache.config.loaderwriter.DefaultCacheLoaderWriterConfiguration;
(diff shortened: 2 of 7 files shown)


impl/src/main/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactory.java
```diff
@@ -41,4 +43,10 @@ public void releaseCacheLoaderWriter(final CacheLoaderWriter<?, ?> cacheLoaderWr
     // noop
   }
 
+  @Override
+  public void start(ServiceConfiguration<?> config, ServiceProvider serviceProvider) {
+    if (config != null) {
+      throw new IllegalArgumentException("CacheLoaderWriter configuration must not be provided at CacheManager level");
```
rishabhmonga: Please note that this validation will fail if user implement their own factory

impl/src/main/java/org/ehcache/spi/event/DefaultCacheEventListenerFactory.java
```diff
@@ -42,4 +43,11 @@ public DefaultCacheEventListenerFactory() {
   public void releaseEventListener(CacheEventListener<?, ?> cacheEventListener) {
     //noop
   }
+
+  @Override
+  public void start(ServiceConfiguration<?> config, ServiceProvider serviceProvider) {
```
ljacomet: You can't just override the start method from `ClassInstanceProvider` without calling super ...
And in the case of services extending `ClassInstanceProvider` the check needs to be a bit smarter.

impl/src/main/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactory.java
```diff
@@ -41,4 +43,10 @@ public void releaseCacheLoaderWriter(final CacheLoaderWriter<?, ?> cacheLoaderWr
     // noop
   }
 
+  @Override
+  public void start(ServiceConfiguration<?> config, ServiceProvider serviceProvider) {
```
ljacomet: Same comment as for other start overridde

impl/src/test/java/org/ehcache/spi/loaderwriter/DefaultCacheLoaderWriterFactoryTest.java
```diff
@@ -44,38 +44,6 @@ public void testCacheConfigUsage() {
...
-    final DefaultConfiguration configuration = new DefaultConfiguration(caches, null, new DefaultCacheLoaderWriterFactoryConfiguration()
-        .addLoaderFor("foo", MyLoader.class));
-    final CacheManager manager = CacheManagerBuilder.newCacheManager(configuration);
-    manager.init();
-    final Object foo = manager.getCache("foo", Object.class, Object.class).get(new Object());
-    assertThat(foo, is(MyLoader.object));
-  }
-
-  @Test
-  public void testCacheConfigOverridesCacheManagerConfig() {
```
ljacomet: Why did you remove these two tests?

Issue #353, opened 2015-04-01
Cache.unwrap should support deep unwrap


alexsnaps:
Currently it only _casts_ `this`, we should let someone "deeper", e.g. on Cache: `unwrap(Eh107Cache.class)` as well as `unwrap(org.ehcache.Cache)`

Obviously for all 107 types declaring the `unwrap` method

ljacomet:
Some of that work has been done with regards to configuration in the context of #216

PR #429, opened 2015-06-04
Issue #353 : Allowing unwrap to provide deep unwrap

rishabhmonga:


rishabhmonga:
test this please

ljacomet:
Apart from some test remarks, the changes here look good (especially fixing the mess of unwrap in the config class :smile: )
However, I now realise that except for configuration, all these changes are kind of useless. Most unwrap change are done to enable unwrapping `this` to the Ehcache JSR-107 concrete type ... but all these but the configuration one are package private and thus users can never get to them explicitly anyway. And they also do not offer any useful methods. So I am now wondering if all changes should be dropped but the configuration related bit (although the tests are flawed  I think).

Does this make sense to others? @alexsnaps @AbfrmBlr

alexsnaps:
I don't think the unwrapping `this` is an issue. It's an orthogonal concern. The changes, while not immediately useful, remain valid as far as I'm concerned.

107/src/test/java/org/ehcache/jsr107/UnwrapTest.java
```diff
@@ -0,0 +1,129 @@
...
+    Cache<String, String> cache = cacheManager.createCache("cache", configuration);
+    org.ehcache.event.CacheEvent<String, String> ehEvent = new EhEvent();
+    Eh107CacheEntryEvent<String, String> cacheEntryEvent = new Eh107CacheEntryEvent<String, String>(cache, EventType.CREATED, ehEvent, false);
+    assertThat(cacheEntryEvent.unwrap(org.ehcache.event.CacheEvent.class), is(instanceOf(CacheEvent.class)));
+    assertThat(cacheEntryEvent.unwrap(cacheEntryEvent.getClass()), is(instanceOf(Eh107CacheEntryEvent.class)));
+  }
+
+  @SuppressWarnings("unchecked")
+  @Test
+  public void testCompleteConfigurationUnwrap() {
```
ljacomet: This test makes little sense. Why do you pass through an Ehcache cache instance to create the complete configuration?
Also, I believe it is redundant with what exists in `org.ehcache.docs.EhCache107ConfigurationIntegrationDocTest`
Same remark applies to the next test.

So either fix - by getting the config from the JSR-107 cache and then testing the unwrap or delete.

ljacomet:
Ok, fine by me then. Commented on the tests that I find bogus.

107/src/main/java/org/ehcache/jsr107/Unwrap.java
```diff
@@ -20,13 +20,15 @@
  */
 final class Unwrap {
 
-  static <T> T unwrap(Class<T> clazz, Object obj) {
+  static <T> T unwrap(Class<T> clazz, Object... obj) {
     if (clazz == null || obj == null) {
```
rishabhmonga: It was possible in some cases that one of the objects is null, causing NPE. Thus added a check here.

rishabhmonga:
test this please

PR #430, opened 2015-06-08
Its NOT PR..  Review of EhcacheExecutorService API

palmanojkumar:
This PR is reaised to review new api and related changes only.

The changeset addresses:
1. New api developed
2. Getting started example
3. Some code refactoring showing changes required with use of new api

Most of the classes from package org.ehcache.statistics and org.ehcache.statistics.extended use ScheduleExecutorService but these classes are not refactored as of now.


palmanojkumar:
Newly added classed can move across packages if required.

palmanojkumar:
Review comments from Chris D and Alex.
1.	maxThreadsAllowed is not good way to limit resources in system. There are  associated complexity and race conditions to check threads state as well.
2.	The abstraction should decide type of pool to return based on context and config. 
3.	We don’t need own EhcacheExecutorService. As long as, lifecycle methods are concerned , the implementation should decide the action to take when such lifecycle method is invoked by consumer.

palmanojkumar:
Review comments:
Remove ThreadFactoryProvider configuration from service configuration.

Issue #355, opened 2015-04-01
Wire lifecycle concerns into `PersistentService`


alexsnaps:
Methods of `Persistent*` types don't handle lifecycle concerns (e.g. `PersistentCacheManager.destroyCache`), that needs to be properly fixed (i.e. delegate to the appropriate `PersistenceProvider`)

ljacomet:
Pushed out an initial PR: #434

Regarding persistent user managed caches, I have two approaches and would like feedback on these.
The first one hides the handling of persistent storage lifecycle inside the `PersistentUserManagedCache`:
```java
    PersistentUserManagedCache<Long, String> cache = UserManagedCacheBuilder.newUserManagedCacheBuilder(Long.class, String.class, "persistent-user-managed") // <1>
        .with(new UserManagedPersistenceConfiguration<Long, String>(new File(getStoragePath(), "myUserData"))) // <2>
        .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
            .heap(10L, EntryUnit.ENTRIES)
            .disk(100L, EntryUnit.ENTRIES, true)) // <3>
        .build(true);

    cache.close(); // <4>
    cache.toMaintenance().destroy(); // <5>

```
1. Create a `UserManagedCacheBuilder`, the identifier is mandatory if the cache needs to become persistent
2. Pass it the persistence configuration, which also makes the builder returns a `PersistentUserManagedCache`
3. Define disk resources, persistent or not
4. Closing the cache does not destroy storage if it was made persistent in 3.
5. On a closed cache, put it in maintenance and destroy its storage.

The second one forces the user to manage the lifecycle of the data himself:
```java
    LocalPersistenceService persistenceService = getPersistenceService(); // <1>
    persistenceService.start(new PersistenceConfiguration(new File(getStoragePath(), "myUserData"))); // <2>
    FileBasedPersistenceContext persistenceContext = persistenceService.createPersistenceContext("cache1"); // <3>

    UserManagedCache<Long, String> cache = UserManagedCacheBuilder.newUserManagedCacheBuilder(Long.class, String.class)
        .withPersistenceContext(persistenceContext) // <4>
        .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
            .heap(10L, EntryUnit.ENTRIES)
            .disk(100L, EntryUnit.ENTRIES, true)) // <5>
        .build(true);

    cache.close(); // <6>
[...]

PR #434, opened 2015-06-11
Issue #355 Lifecycle concerns for persistent data

ljacomet:
Now controlled through LocalPersistenceService and PersistentCacheManager in place of Store level.

* Change LocalPersistenceService API
* Adapt DiskStore creation
* Change LocalPersistenceService API
* Wire destruction of persistence context into PersistentCacheManager.destroyCache
* Enable destruction of persistence context without binding a cache
* Deleted StoreDestroyTest as this is no longer a Store concern

This is the first step in issue #355 `UserManagedCache` still needs to be addressed.


ljacomet:
Pushed updated commit with minor clean up.

ljacomet:
Still not a single comment ... 
Can someone please merge then? Or ping me to do a drive through one more time?

AbfrmBlr:
I understand that we have decoupled the lifecycle concern of cache with the persistent store(context) and now creation/destruction is taken care by the persistence service itself.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1075,28 +1074,12 @@ public CacheStatistics getStatistics() {
...
-    try {
-      if (store instanceof Persistable) {
-        logger.info("Creating Cache storage.");
-        ((Persistable) store).create();
-        logger.info("Cache storage successfully created.");
-      }
-    } catch (Exception e) {
-      throw new RuntimeException("Couldn't create Cache storage ", e);
-    }
+    // TODO figure out persistence and user managed caches
```
alexsnaps: Where or when is that todo to be addressed? Is there an other issue for that?

AbfrmBlr: If I recall it right , this is just the partial PR which talks about API stuff. There will be further PRs for the same issue

alexsnaps: Duh! Y...

PR #437, opened 2015-06-19
fixed NPE message

palmanojkumar:


impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehindDecoratorLoaderWriterProviderFactory.java
```diff
@@ -50,6 +50,9 @@ public void start(ServiceConfiguration<?> config, ServiceProvider serviceProvide
       
       @Override
       public <K, V> WriteBehindDecoratorLoaderWriter<K, V> createWriteBehindDecoratorLoaderWriter(CacheLoaderWriter<K, V> cacheLoaderWriter, WriteBehindConfiguration configuration) {
+        if (cacheLoaderWriter == null) {
+          throw new RuntimeException("WriteBehind requires non null CacheLoaderWriter.");
```
ljacomet: A `NullPointerException` here is better than a simple runtime one - the message being the added information.

impl/src/test/java/org/ehcache/WriteBehindDecoratorLoaderWriterProviderFactoryTest.java
```diff
@@ -81,6 +85,14 @@ public void testAddingWriteBehindConfigurationAtCacheLevel() {
     assertThat(serviceConfiguration, IsCollectionContaining.<ServiceConfiguration<?>>hasItem(instanceOf(WriteBehindConfiguration.class)));
     cacheManager.close();
   }
+  
+  @SuppressWarnings("unchecked")
+  @Test(expected=RuntimeException.class)
+  public void testWriteBehindWithoutCacheLoaderWriter() {
+    WriteBehindConfiguration writeBehindConfiguration = WriteBehindConfigurationBuilder.newWriteBehindConfiguration().build();
+    WriteBehindDecoratorLoaderWriterProviderFactory factory = new WriteBehindDecoratorLoaderWriterProviderFactory();
+    factory.create(mock(ServiceConfiguration.class), null).createWriteBehindDecoratorLoaderWriter(null, writeBehindConfiguration);
```
ljacomet: Most of the setup in this test is superfluous as calling into the last line with nulls only will result in the same exception.
However, doing it so would require to catch the exception and make sure we are getting the exact one we expect.

palmanojkumar: hmm. removed mock service configuration and doing specific exception test.

ljacomet:
Content good, commit history + commit comments not good.

Can you please rebase in a single commit that references the issue?

palmanojkumar:
Rebased all commits to single commit. test this please.

ljacomet:
Format for referencing issues is #<issue> - see https://guides.github.com/features/issues/#notifications especially the references section.

Issue #431, opened 2015-06-08
Check and refuse inverted pyramid in sizing


ljacomet:
Example: disk store entry count must be greater than heap entry count

ljacomet:
Implement a generic solution that is not limited to the current constraints

PR #440, opened 2015-06-22
issue #431: added check for inverted pool size

palmanojkumar:


impl/src/main/java/org/ehcache/internal/store/DefaultStoreProvider.java
```diff
@@ -60,6 +60,11 @@
       if (heapPool == null) {
         throw new IllegalArgumentException("Cannot store to disk without heap resource");
       }
+      if (diskPool.getSize() == heapPool.getSize()) {
+        LOG.info("Disk pool size should be greater than Heap pool size");
+      } else if (diskPool.getSize() < heapPool.getSize()) {
+        throw new IllegalArgumentException("Disk pool size must be greater than Heap pool size");
+      }
```
alexsnaps: I'm fairly confident this could already be implemented agnostically from the existing limitation imposed on the tiering model, i.e. in a way that the pyramid gets validated so that higher tiers are always smaller than lower ones that share the common `ResourceUnit`... 
Doing it that way, would make sure this constraint remains enforced when we do ARC or introduce the new `DiskStore`.

impl/src/test/java/org/ehcache/internal/store/DefaultStoreProviderTest.java
```diff
@@ -0,0 +1,91 @@
...
+  
+  @Rule
+  public ExpectedException expectedEx = ExpectedException.none();
+  
+  
+  @SuppressWarnings("unchecked")
+  @Test
+  public void testCreateStoreDiskPoolLessThanHeapPool() {
+    expectedEx.expect(IllegalArgumentException.class);
+    expectedEx.expectMessage("Disk pool size must be greater than Heap pool size");
```
chrisdennis: Personally I'm not a fan of this kind of exception assertion in junit.  The scoping of the `@Rule` field is wrong - it's applying to all methods when thats not really what you want.  From my point of view a simple try catch is simpler and just as clear, and can also be extremely specific about which line is expected to throw.

PR #444, opened 2015-06-29
Issue #355 Lifecycle concerns for persistent data in user managed caches

ljacomet:


core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -83,9 +83,9 @@
 /**
  * @author Alex Snaps
  */
-public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V>, PersistentUserManagedCache<K, V> {
+public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V> {
 
-  private final StatusTransitioner statusTransitioner;
+  protected final StatusTransitioner statusTransitioner;
```
ljacomet: Not really happy with having to do that, however I have not found another solution and really want to have user managed persistent related code in a subclass.

ljacomet:
test this please

core/src/main/java/org/ehcache/Ehcache.java
AbfrmBlr: Rather than making the variable protected, we can use checkMaintenance method and introduce other methods to do maintenance status transition. Just a thought !

ljacomet:
test this please

ljacomet:
Update the PR and introduced package private method to hit the `StatusTransitioner` from the introduced `Ehcache` subclass.

core/src/main/java/org/ehcache/PersistentUserManagedEhcache.java
```diff
@@ -0,0 +1,101 @@
...
+
+        @Override
+        public void close() {
+          internalExitMaintenance().succeeded();
+        }
+      };
+      st.succeeded();
+      return maintainable;
+    } catch (RuntimeException e) {
+      st.failed(e); // this throws
```
chrisdennis: Can we not just change the contract of failed(t) to return the `StatusTranstionException` rather than throwing it?  That way this becomes:

```java
    return maintainable;
    throw st.failed(e);
  }
}
```

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1026,42 +1026,16 @@ public void close() {
...
-  void create() {
+  void checkMaintenance() {
     statusTransitioner.checkMaintenance();
-    // TODO figure out persistence and user managed caches
   }
 
-  void destroy() {
-    statusTransitioner.checkMaintenance();
-    // TODO figure out persistence and user managed caches
+  StatusTransitioner.Transition internalToMaintenance() {
```
chrisdennis: I think we talked about this offline, but I'll reiterate here.  This feels really funky, it seems wrong that the Ehcache class is exposing methods that allow for the direct manipulation of an internal state.  Also it seems wrong that the Ehcache class can be transitioned to a "Maintainable" state when it isn't even capable (on it's own) of being persistent.  It feels like we're bleeding details of the lifecycle of a persistent cache in to the non-persistent version.  Maybe the answer is an abstract superclass which contains methods to query the lifecycle but that the persistent and non persistent subclasses implement their own independent lifecycles?  That idea probably sucks... maybe this is the best solution.  The status quo seems weird to me though.

core/src/main/java/org/ehcache/PersistentUserManagedEhcache.java
ljacomet: Agreed, I have reviewed the usages. Making it return and the consumer explicitly call `throw` will make the code cleaner / clearer.

core/src/main/java/org/ehcache/Ehcache.java
ljacomet: I am trying out with having the `PersistentUserManagedEhcache` delegate to `Ehcache` instead of extending it.

ljacomet:
Please test this

ljacomet:
Please test this

ljacomet:
test this please

ljacomet:
please test this

ljacomet:
please test this

ljacomet:
please test this

ljacomet:
test this please

ljacomet:
hodor

ljacomet:
test this please

ljacomet:
test this please (damn unstable tests)

ljacomet:
test this please

ljacomet:
Fixing test requires #464 to be merged in

ljacomet:
test this please (need to fix that plugin!)

ljacomet:
test this please

ljacomet:
test this please (getting annoyed here)

Issue #380, opened 2015-04-21
SPI Tester should report exceptions


aurbroszniowski:
Many SPI tests catch exceptions (such as CacheAccessException) because they are valid according to the contracts, it is however important to report them, and currently they are only reported through a printStackTrace, we should have the SPI Test runner to gather those exceptions and report them in a way instead.
SPI tests would of course be updated to sending the exceptions to the SPI Test runner instead of printing the stack trace

PR #445, opened 2015-07-01
Issue #380 : Fixing SPI Tester should report exceptions

rishabhmonga:


rishabhmonga:
test this please

rishabhmonga:
test this please

rishabhmonga:
test this please

core-spi-test/src/main/java/org/ehcache/internal/store/StoreClearTest.java
```diff
@@ -67,10 +69,12 @@ public void removesAllOfTheMappings()
     try {
       kvStore.clear();
     } catch (CacheAccessException e) {
-      System.err.println("Warning, an exception is thrown due to the SPI test");
-      e.printStackTrace();
+      validException = true;
     }
 
     assertThat(kvStore.containsKey(key), is(false));
+    if(validException) {
```
aurbroszniowski: I think that using a boolean and throw the exception lazily is dangerous, if an exception is thrown, due to a failure, even valid, it is likely that the assertions are going to fail thus marking the test as fail while it would be valid to mark it as passing.
This comment apply to all other test cases that use a boolean

spi-tester/src/main/java/org/ehcache/spi/test/ValidSPITesterException.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.ehcache.spi.test;
+
+/**
+ * Can be used to report valid exceptions in SPI test so that the said exceptions don't cause a failure
+ * 
+ * @author rism
+ */
+public class ValidSPITesterException extends Exception {
```
aurbroszniowski: better to double check with a native english speaker but using 'valid' for an exception sounds a bit strange

Issue #252, opened 2015-01-06
Replace "old" DiskStore with "new" DiskStore


alexsnaps:
Wire it _all_ in and have CRUD ops implemented (and wired) at the `Store` level

alexsnaps:
Wire it _all_ in and have CRUD ops implemented (and wired) at the `Store` level

ljacomet:
May require some work in offheap library

ljacomet:
As part of this, `StoreEventListenerTest` StoreSPITest should be split between expiry and eviction.
This would allow all store impls to pass the expiry ones while the count based eviction remains an issue for stores where the natural unit is memory.

PR #456, opened 2015-07-09
Issue #252 : Initial (mostly testless) cut of the OffHeapDiskStore integration

chrisdennis:
This PR represents an initial cut of the OffHeapDiskStore integration.  Currently I'm a little suspicious of the recovery/persistence integration code, and there is a dearth of tests.  I'm going to be working on both of those aspects, but since this is such a large PR I wanted to give people the option of getting started now so that I can get feedback as early as possible.


lorban:
You better start by fixing the failing tests?

impl/src/test/java/org/ehcache/PersistentCacheManagerTest.java
```diff
@@ -21,24 +21,26 @@
...
+  @Rule
+  public final TemporaryFolder folder = new TemporaryFolder();
+  
   @Test
-  public void testInitializesLocalPersistenceService() {
-    final File rootDirectory = mock(File.class);
-    when(rootDirectory.exists()).thenReturn(true);
-    when(rootDirectory.getAbsolutePath()).thenReturn("CRAP!");
+  public void testInitializesLocalPersistenceService() throws IOException {
+    final File rootDirectory = folder.newFolder("testInitializesLocalPersistenceService");
```
ljacomet: With these changes, the test should fail as no exception is thrown anymore. The only reason this still passes is because there is no `fail` call before the catch block. Can you adapt the test to make sure it tests what was tested before?

impl/src/main/java/org/ehcache/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -0,0 +1,375 @@
...
+        source,
+        storageEngineFactory,
+        config.getInitialSegmentTableSize(),
+        evictionVeto,
+        mapEvictionListener, true);
+    return new EhcachePersistentConcurrentOffHeapClockCache<K, OffHeapValueHolder<V>>(factory, config.getConcurrency());
+
+  }
+
+  private static <K, V> Predicate<Map.Entry<K, OffHeapValueHolder<V>>> wrap(EvictionVeto<? super K, ? super V> delegate) {
```
ljacomet: This method is identical to `OffHeapStore.wrap` and so can probably be extracted to the `AbstractOffHeapStore`

Issue #447, opened 2015-07-02
Problem with FIFO Eviction


alanzplus:
Hey currently I' using the ehcache3 lib and I had a problem with FIFO eviction prioritizer.

The following is my part of cache config xml.

```xml
 <ehcache:eviction-prioritizer>FIFO</ehcache:eviction-prioritizer>
 <ehcache:resources>
     <ehcache:heap size="5" unit="entries"/>
  </ehcache:resources>
```

So if I have list of key like `1, 2, 3, 4, 5, 6`.
Then I put these key into the cache in order (from 1 to 6). So when I put the key 6,
the cache should evict key 1. However, the key the cache evicted is quite randomly.
Sometimes, it just evicted key 6! (this what i just inserted)

Does anyone know about this ?

alexsnaps:
@alanzplus FIFO is based on the `creationTime` (which is in millisecond precision). Further more, a sample population is chosen and the eviction victim is chosen from there... 
With that in mind, could it be you're creating things that quickly that they _all_ qualify as "eldest" as far as the Cache is concerned?

alanzplus:
@alexsnaps according to your comment, I tried to insert things into cache by an "large interval like 1 second.
But the problem cannot be solved. And also according to http://ehcache.org/documentation/2.8/apis/cache-eviction-algorithms,(and if ehcache3 use the same implementation as ehcache2), 15 random sample will be selected, so in my case, since only have 5 elements, the first key should be evicted. BTW, in the ehcach 3, what kind of strategy is used for sampling, and what is the sample size?

rishabhmonga:
@alanzplus would you please confirm topology you are using. Also can you please share the complete ehcache config xml

alexsnaps:
@alanzplus Best would be to share a complete reproducible test case... @rishabhmonga took a crack at reproducing the issue, but "only" got it to expose the behavior you describe in the `OffHeapStore` case if I got him right... Which isn't surprising, but probably still undocumented.

rishabhmonga:
@alanzplus following is a test that demonstrates the use-case you mentioned when using an on heap cache. Since the cache size here is smaller than the size of the sample chosen for selecting eviction candidates, the outcome is predictable and confirms expected behavior.

````java
    TestTimeSource timeSource = new TestTimeSource();

    CacheConfiguration<String, String> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
[...]

PR #459, opened 2015-07-10
Issue #447 : Adding test to confirm FIFO

rishabhmonga:


rishabhmonga:
test this please

AbfrmBlr:
It would be a good idea to add another test with a larger sample depicting and asserting the true behavior of FIFO eviction, i.e. it is not a strict FIFO as it seems to be from the current test case.

impl/src/test/java/org/ehcache/EhcacheFIFOEvictionTest.java
```diff
@@ -0,0 +1,126 @@
...
+//    Adding elements after time intervals
+    for(int i = 0; i < 100; i++) {
+      cache.put(i, i);
+      timeSource.advanceTime(1);
+    }
+    
+    cache.put(100, 100);
+
+//    asserting last element is not evicted
+    assertThat(cache.get(99), Is.<Number>is(99));
```
chrisdennis: Should we be stronger here?  Given that we're doing grey box testing of this - can't we assert that the `sampleSize - 1` most recently added entries should be there?

impl/src/test/java/org/ehcache/EhcacheFIFOEvictionTest.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  static class TestTimeSource implements TimeSource {
```
chrisdennis: Don't we already have an implementation of this class somewhere else?

impl/src/test/java/org/ehcache/EhcacheFIFOEvictionTest.java
```diff
@@ -0,0 +1,112 @@
...
+//    Adding elements after time intervals
+    for(int i = 0; i < 100; i++) {
+      cache.put(i, i);
+      timeSource.advanceTime(1);
+    }
+
+    cache.put(100, 100);
+
+//    asserting last-in elements are never evicted
+    for(int i = 94; i < 101; i++) {
```
ljacomet: I do not like the magic number used as lower bound here. Maybe we could have the sample size as a constant and use it here? And by changing this "integration" test to a "unit" test on `OnHeapStore`, it could probably remain package private.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -86,6 +86,7 @@
 
   private static final int ATTEMPT_RATIO = 4;
   private static final int EVICTION_RATIO = 2;
+  final int SAMPLE_SIZE = 8;
```
ljacomet: This should be a static

impl/src/test/java/org/ehcache/internal/store/heap/OnHeapStoreFIFOEvictionTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.io.Serializable;
+
+import static org.ehcache.config.ResourcePoolsBuilder.newResourcePoolsBuilder;
+import static org.junit.Assert.assertThat;
+
+/**
+ * @author rism
+ */
+public class OnHeapStoreFIFOEvictionTest {
```
ljacomet: Your test with smaller number of entries should still exist here.

PR #463, opened 2015-07-15
Issue #355 rework persistence service

chrisdennis:
* a directory for each context, and no assumptions about index and data files.
* a reworked naming scheme that avoids escaping capital letters and instead relies on a SHA1 suffix to prevent FS limitation induced collisions.

### DIFF
--- a/impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
+++ b/impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
@@ -31,15 +31,24 @@ import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileLock;
+import java.nio.charset.Charset;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+
+import static java.lang.Integer.toHexString;
+import static java.nio.charset.Charset.forName;
 
 /**
  * @author Alex Snaps
  */
 public class DefaultLocalPersistenceService implements LocalPersistenceService {
 
+  private static final Charset UTF8 = forName("UTF8");
   private static final int DEL = 0x7F;
   private static final char ESCAPE = '%';
   private static final Set<Character> ILLEGALS = new HashSet<Character>();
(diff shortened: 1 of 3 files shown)


chrisdennis:
test this please

chrisdennis:
test this please you pile of shitty integration crap

ljacomet:
test this please (this is getting annoying)

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -127,123 +136,110 @@ static void createLocationIfRequiredAndVerify(final File rootDirectory) {
...
-    File file = new File(rootDirectory, name);
+  
+  public File getDirectoryFor(String identifier) {
+    File directory = new File(rootDirectory, safeIdentifier(identifier));
     
-    for (File parent = file.getParentFile(); parent != null; parent = parent.getParentFile()) {
+    for (File parent = directory.getParentFile(); parent != null; parent = parent.getParentFile()) {
       if (rootDirectory.equals(parent)) {
-        return file;
+        return directory;
```
ljacomet: Could you add a test that fails this condition?

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -127,123 +136,110 @@ static void createLocationIfRequiredAndVerify(final File rootDirectory) {
...
-          .getIndexFile()
-          .getAbsolutePath());
-    } else if (!dataFileCreated && !indexFileCreated) {
-      LOGGER.info("Reusing " + fileBasedPersistenceContext.getDataFile().getAbsolutePath() + " and " + fileBasedPersistenceContext
-          .getIndexFile().getAbsolutePath());
+    File persistenceDirectory = fileBasedPersistenceContext.getDirectory();
+    
+    if (persistenceDirectory.isDirectory()) {
+      LOGGER.info("Reusing " + persistenceDirectory.getAbsolutePath());
+    } else if (persistenceDirectory.mkdir()) {
```
ljacomet: We probably need to call `File.mkdirs()` to be safe

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
chrisdennis: No.

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
chrisdennis: No we don't.

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
ljacomet: Indeed, our escaping actually protects us

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
ljacomet: We only create a single dir under the root dir - so no need indeed

PR #465, opened 2015-07-17
Issue #122 : Enabling hitRate() for Store.ValueHolder

rishabhmonga:


rishabhmonga:
`OffHeapDiskStoreTest.testFlushUpdatesAccessStats` seems to randomly fail sometimes during build. I am trying to figure out the cause but I've created this PR to get some initial reviews and may be a second set of eyes for the test as well

core/src/main/java/org/ehcache/spi/cache/AbstractValueHolder.java
```diff
@@ -29,6 +30,8 @@
   private final long creationTime;
   private volatile long lastAccessTime;
   private volatile long expirationTime;
+  private volatile long hits;
+  private static final AtomicLongFieldUpdater<AbstractValueHolder> ATOMIC_UPDATER = AtomicLongFieldUpdater.newUpdater(AbstractValueHolder.class, "hits");
```
chrisdennis: HITS_UPDATER?

core/src/main/java/org/ehcache/spi/cache/AbstractValueHolder.java
```diff
@@ -129,9 +133,20 @@ public boolean equals(Object obj) {
...
-    return 0.0f;
+  public float hitRate(long now, TimeUnit unit) {
+    final long endTime = TimeUnit.NANOSECONDS.convert(now, TimeUnit.MILLISECONDS);
+    final long startTime = TimeUnit.NANOSECONDS.convert(creationTime, nativeTimeUnit());
+    float duration = (endTime - startTime)/(float)TimeUnit.NANOSECONDS.convert(1, unit);
+    return (hits/duration);
+  }
+
+  @Override
+  public long getHits() {
```
chrisdennis: I know that @alexsnaps already blurred the lines on this one, but could we stick to one naming convention within this class?  Currently only `getId` and `getHits` have a get prefix.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Oct 01 10:52:23 CEST 2014
+#Fri Jul 17 15:47:11 IST 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.1-bin.zip
```
chrisdennis: Huh?  How is this related?

impl/src/main/java/org/ehcache/internal/store/offheap/OffHeapValueHolder.java
```diff
@@ -33,14 +33,15 @@
   private final V value;
   private final WriteContext writeContext;
 
-  public OffHeapValueHolder(long id, V value, long creationTime, long expireTime) {
-    this(id, value, creationTime, expireTime, 0, null);
+  public OffHeapValueHolder(long id, V value, long creationTime, long expireTime, long hits) {
```
chrisdennis: If this is always called with zero (since new values have no hits) then we can eliminate the param and hardcode the zero, right?

chrisdennis:
`OffHeapValueHolderPortabilityTest.testWriteBackSupport()` should verify that the hits is written back.

core/src/main/java/org/ehcache/spi/cache/AbstractValueHolder.java
```diff
@@ -129,9 +133,20 @@ public boolean equals(Object obj) {
...
+    return this.hits;
+  }
+
+  public void setHits(long hits) {
+    this.hits = hits;
```
ljacomet: This method should be protected and most likely use the updater instead of direct field access.

core/src/main/java/org/ehcache/events/StoreEventListener.java
```diff
@@ -25,14 +26,16 @@
    *
    * @param key the <code>key</code> of the mapping being evicted
    * @param valueHolder the {@link Store.ValueHolder} being evicted
+   * @param timeSource the associated {@link TimeSource}
    */
-   void onEviction(K key, Store.ValueHolder<V> valueHolder);
+  void onEviction(K key, Store.ValueHolder<V> valueHolder, TimeSource timeSource);
```
ljacomet: Let's work in order to not have to do this API change, which makes little sense.
From usage, I see that if the `CacheEventNotificationServiceImpl` had access to a `TimeSource` the param is no longer needed. My IDE tells me this is the only usage.

impl/src/test/java/org/ehcache/internal/store/offheap/OffHeapValueHolderPortabilityTest.java
```diff
@@ -56,16 +59,23 @@ public void testEncodeDecode() {
...
     WriteContext writeContext = mock(WriteContext.class);
     OffHeapValueHolder<String> decoded = valueHolderPortability.decode(encoded, writeContext);
 
+    Class<?> abstractValueHolder = AbstractValueHolder.class;
+    Method setHits = abstractValueHolder.getDeclaredMethod("setHits", long.class);
+    setHits.setAccessible(true);
+    
     decoded.setExpirationTime(4L, TimeUnit.MILLISECONDS);
     decoded.setLastAccessTime(6L, TimeUnit.MILLISECONDS);
+    setHits.invoke(decoded, 8L);
```
rishabhmonga: The test had to be slightly changes as `AbstractValueHolder.setHits()` was made protected. Also, with one of the set methods made protected should the others be made protected as well, namely, `setExpirationTime()` and `setLastAccessTime()`

rishabhmonga:
test this please

impl/src/main/java/org/ehcache/internal/events/CacheEventNotificationListenerServiceProviderImpl.java
```diff
@@ -44,9 +51,11 @@ public void stop() {
 
   public <K, V> CacheEventNotificationService<K, V> createCacheEventNotificationService(Store<K, V> store) {
     ThreadPoolsService threadPoolsService = serviceProvider.findService(ThreadPoolsService.class);
+    TimeSourceConfiguration timeSourceConfig = findSingletonAmongst(TimeSourceConfiguration.class, config);
```
ljacomet: This is never going to work. The config you get will never be a `TimeSourceConfiguration` and so this always returns null.
I believe you found a shortcut in the service system where we just use a configuration for the time source and not a service. I do not see any way of making this work without fixing that issue.
I agree that the `TimeSourceService` will be very shallow and similar to the `TimeSourceConfiguration` but we have a system in place to access services not service configs.
One good way of testing what I am saying is an integration test that tries to configure a different time source. You will see it never makes it here.

rishabhmonga:
test this please

core-spi-test/src/main/java/org/ehcache/internal/store/StoreValueHolderHitRateTest.java
```diff
@@ -41,11 +41,10 @@ public StoreValueHolderHitRateTest(final StoreFactory<K, V> factory) {
   }
 
   @SPITest
-  @Ignore(reason = "Failing test to fix")
   public void hitRateCanBeReturned()
       throws IllegalAccessException, InstantiationException {
     Store.ValueHolder<V> valueHolder = factory.newValueHolder(factory.createValue(1));
 
-    assertThat(valueHolder.hitRate(TimeUnit.MILLISECONDS), is(notNullValue()));
+    assertThat(valueHolder.hitRate(TimeUnit.MILLISECONDS.convert(Time.time(), TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS), is(notNullValue()));
```
ljacomet: Why that `Time.time()` usage instead of a simpler `System.currentTimeMillis()`? Especially as the former actually returns `System.nanoTime()`.

core/src/test/java/org/ehcache/EhcacheEventsTestBase.java
```diff
@@ -61,7 +62,8 @@
   protected Ehcache<String, String> getEhcache(final CacheLoaderWriter<String, String> cacheLoaderWriter, String name) {
     ExecutorService orderedExecutor = Executors.newSingleThreadExecutor();
     ExecutorService unorderedExecutor = Executors.newCachedThreadPool();
-    cacheEventNotificationService = new CacheEventNotificationServiceImpl<String, String>(orderedExecutor, unorderedExecutor, store);
+    TestTimeSource timeSource = new TestTimeSource();
```
ljacomet: Why do you introduce a `TestTimeSource` here? Is it because the `SystemTimeSource` is not in scope as it lives in `impl` module?
Especially as the `TestTimeSource` provides the advanceTime method which is not used.
Two options:

* Move `SystemTimeSource from `impl` to `core`
* Use an anonymous inner class that wraps `System.currentTimeMillis()`

core/src/test/java/org/ehcache/events/CacheEventNotificationServiceImplTest.java
```diff
@@ -64,7 +65,8 @@ public void setUp() {
     orderedExecutor = Executors.newSingleThreadExecutor();
     unorderedExecutor = Executors.newCachedThreadPool();
     store = mock(Store.class);
-    eventService = new CacheEventNotificationServiceImpl<Number, String>(orderedExecutor, unorderedExecutor, store);
+    TestTimeSource timeSource = new TestTimeSource();
```
ljacomet: Same question as with the previous `TestTimeSource` inclusion - which reinforces the point about moving `SystemTimeSource`

impl/src/test/java/org/ehcache/internal/store/offheap/AbstractOffHeapStoreTest.java
```diff
@@ -112,12 +129,12 @@ public void testFlushUpdatesAccessStats() throws CacheAccessException {
   private static class TestStoreEventListener<K, V> implements StoreEventListener<K, V> {
 
     @Override
-    public void onEviction(final K key, final Store.ValueHolder<V> valueHolder) {
+    public void onEviction(K key, Store.ValueHolder<V> valueHolder) {
```
ljacomet: Can you revert these changes and adapt your IDE settings? Same for next change

ljacomet:
The following changes classes should be moved to their own commit as they do not relate to the issue but are code formatting / imports clean up:

* `StoreExpiryEventListenerTest`
* `StoreEventListener`
* `OffHeapDiskStoreSPITest`

core-spi-test/src/main/java/org/ehcache/internal/store/StoreValueHolderHitRateTest.java
chrisdennis: This test can never fail right?  `hitRate` returns a float, and floats can't autobox in to null values.

rishabhmonga:
test this please

impl/src/test/java/org/ehcache/internal/store/disk/OffHeapDiskStoreSPITest.java
```diff
@@ -22,6 +22,7 @@
 import org.ehcache.config.ResourcePoolsBuilder;
 import org.ehcache.config.StoreConfigurationImpl;
 import org.ehcache.config.persistence.CacheManagerPersistenceConfiguration;
+import org.ehcache.config.persistence.DefaultPersistenceConfiguration;
```
chrisdennis: This import is unused.

rishabhmonga:
test this please

Issue #457, opened 2015-07-09
ServiceProvider should support retrieving without implicit creation


chrisdennis:
Some requests for services rely on the assumption that there is an existing configured service ready to be retrieved.  Currently this kind of request cannot be performed against a ServiceProvider, instead they must call `findService(Class<T> klazz)` which if the service is absent, which in these cases should be an exceptional condition, will instead try to instantiate the service with a null configuration, which at best will trigger a exception (if a null config is not supported by the service), or at worst will instantiate and configure a service instance that was not expected/wanted.

At a minimum this applies for the case of persistent storage trying to locate the persistence service, but it may also apply to other services as well.

ljacomet:
I started looking into our `Service` usage and I realise that the place in which I code access to a service is not the place where I want / should know if that service is supposed to be there always, only when configured or ...

We seem to have different categories of services:

* Always available - example: `CacheLoaderWriterProvider`
* Available if configured - example: `LocalPersistenceService`

So what I am thinking is to offer marker interfaces / annotations that would indicate if a service is supposed to be instantiated if not found when queried. It would then become the responsibility of the service creator to decide that.Note that this would remain an internal thing only as there is no way it a user provided service could benefit from that.

Thoughts on this?

ljacomet:
Initial feedback on the PR + some experiments made me realise there are other issues that should be addressed and may enable a consistent view over the whole system.

And that includes an explanation for that `FIX ME` lurking inside the `ServiceLocator` which relates to possible different end states when discovering services in a started locator versus doing the same in a non started one and starting it afterward - at least depending on the way the `ServiceFactory` handles the config it receives through create.

Next experiment will be to remove all parameters from `ServiceFactory.create` and see where that leads.

ljacomet:
After one more discussion during the Ehcache dev meeting of July 30th, find a number of proposals for resolving the issue around: _To load or not to load a service when there is no config for it_

1. ~~`ServiceProvider.findService` becomes `ServiceProvider.getService` + `ServiceProvider.getOrCreateService`~~
    * ~~This pushes the responsibility of deciding wether a service can be loaded without config in the hands of its consumer.~~
    * I believe the above has been ruled out as wrong responsibility
2. Annotate services to allow them to be resolved even when no configuration was provided.
    * This pushes the responsibility of deciding wether a service can be loaded without config to the `Service` developer
    * Services are marked as _default_ without any consideration for the context - `CacheManager` or `UserManagedCacheBuilder`
[...]

PR #470, opened 2015-07-22
Issue #457

ljacomet:
* Introduction of `MandatoryService` annotation to allow indication of `Service`s that must be loaded even when not explicitly configured
* Improve error reporting when no `LocalPersistenceService` can be found but a disk store is configured
* Misc cleanups in `ServiceProvider` and `ServiceLocator`
* One line delete on an old obsolete comment


ljacomet:
Closing - needs rework

PR #473, opened 2015-07-22
Issue 467

chrisdennis:
Two tweaks here to improve the robustness of these tests.

1. Make sure we use discrete persistence contexts for each test.
2. Correctly lifecycle the stores used in the tests.


impl/src/test/java/org/ehcache/internal/store/disk/OffHeapDiskStoreSPITest.java
```diff
@@ -63,16 +63,15 @@
 
   private AuthoritativeTierFactory<String, String> authoritativeTierFactory;
   private final Map<Store<String, String>, String> createdStores = new ConcurrentHashMap<Store<String, String>, String>();
-  private DefaultLocalPersistenceService persistenceService;
 
   @Rule
   public final TemporaryFolder folder = new TemporaryFolder();
```
ljacomet: This is no longer required I believe

PR #476, opened 2015-07-23
Issue 442 Initial cut of offheap disk store byte based capacity eviction

chrisdennis:
This is an initial cut.  I'd like some review of the changes here... but I plan on adding some more test coverage before anyone merges it.

### DIFF
--- a/impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
+++ b/impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
@@ -37,11 +37,11 @@ import java.security.NoSuchAlgorithmException;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 
 import static java.lang.Integer.toHexString;
 import static java.nio.charset.Charset.forName;
+import java.util.concurrent.ConcurrentMap;
 
 /**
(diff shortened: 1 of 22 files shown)


impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -138,34 +138,32 @@ static void createLocationIfRequiredAndVerify(final File rootDirectory) {
...
-    FileBasedPersistenceContext persistenceContext = knownPersistenceContexts.get(stringIdentifier);
-    if (persistenceContext != null) {
-      try {
-        destroy(stringIdentifier, persistenceContext, true);
-      } finally {
-        knownPersistenceContexts.remove(stringIdentifier);
-      }
+    FileBasedPersistenceContext persistenceContext = knownPersistenceContexts.remove(stringIdentifier);
+    if (persistenceContext == null) {
+      throw new CachePersistenceException("Identifier " + identifier + " not known to this persistence service");
```
ljacomet: We cannot do that, because we want the following scenario to be possible:

1. Create a persistent cache "foo" and fill it
2. Shut down VM
3. Restart VM, initialise cache manager without initialising caches
4. Destroy cache "foo"'s persistence context before initialising it.

And I believe at that point, the context is not marked as known.

chrisdennis:
Tests are added - everything should be working.  Okay to merge if others are.

PR #478, opened 2015-07-24
Issue 365

lorban:
Added compute / computeIfAbsent / computeIfPresent to Cache with signatures comparable to CHMv8 ones. Unfortunately there is no way to make the 107 EntryProcessor mechanism fit into the ehcache one without bastardizing it some way or another.


lorban:
Postponed - we won't include these changes into Milestone 2.

PR #479, opened 2015-07-24
Issue #431 Implement generic tiering size inversion testing

chrisdennis:


chrisdennis:
I'm not amazingly happy with the way this turned out... but doing anything cleaner would, I think, require an overhaul of the units system.  Figured I can let others comment on this as it stands before going down that route.

api/src/main/java/org/ehcache/config/units/MemoryUnit.java
```diff
@@ -117,6 +82,31 @@ static long x(long d, long m, long over) {
...
+  
+  @Override
+  public String toString() {
+    return stringForm;
+  }
+
+  @Override
+  public int compareTo(long thisSize, long thatSize, ResourceUnit thatUnit) {
+    if (thatUnit instanceof MemoryUnit) {
+      return Long.signum(convert(thisSize, (MemoryUnit) thatUnit) - thatSize);
```
ljacomet: convert is misused here, according to its contract. Maybe a test? :smiley:

core/src/test/java/org/ehcache/config/ResourcePoolsImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+import org.junit.Test;
+
+/**
+ *
+ * @author cdennis
+ */
+public class ResourcePoolsImplTest {
```
lorban: I would have expected a few tests that make sure mixing entries and memory doesn't throw an exception and that everything succeeds when there's no pool inversion

api/src/main/java/org/ehcache/config/ResourceUnit.java
```diff
@@ -19,4 +19,20 @@
...
+   * equal to, or greater than {@code thatSize} of {@code thatUnit}
+   * respectively.
+   * 
+   * @param thisSize size in {@code this} unit
+   * @param thatSize size in {@code thatUnit}
+   * @param thatUnit other unit
+   * @return as per the conventional compare contract
+   * @throws IllegalArgumentException if the units are not comparable
+   */
+  public int compareTo(long thisSize, long thatSize, ResourceUnit thatUnit) throws IllegalArgumentException;
```
lorban: no need to use the 'public' keyword here

api/src/main/java/org/ehcache/config/units/MemoryUnit.java
```diff
@@ -27,87 +27,52 @@
...
 
   /**
    * Scale d by m, checking for overflow.
    * This has a short name to make above code more readable.
    */
-  static long x(long d, long m, long over) {
-      if (d >  over) return Long.MAX_VALUE;
-      if (d < -over) return Long.MIN_VALUE;
-      return d * m;
+  private static long doConvert(int delta, long amount) {
```
lorban: such non-obvious logic begs for a unit test

lorban:
IMHO such logic will do for M2. We may need/want to revisit the unit system for ARC so I wouldn't do that much more for now, except maybe hint QA that they can now test for such behavior.

Issue #220, opened 2014-12-09
Outcomes for put() of an immediated expired value, and for a remove() that removes nothing


timeck:
At the moment there is no `end()` call on the observer in these scenarios. Do we want a discreet outcome for these paths? The code contains XXX comments asking this same question.

lorban:
My personal logic tells me that there shouldn't be special cases for these two use-cases but the 107 TCK dictates otherwise: there are two tests checking for those specific cases:

 - put that immediately expires does not get reported as a put in the stats
 - remove on a non-existent key does not get reported as a remove in the stats

The goal of this issue is now to figure out a way to make our code behave as the 107 TCK mandates while not encumbering it with 107-specific logic.

PR #480, opened 2015-07-27
 #220: remove special-case stat counting

lorban:
IMHO there should not be special stats for the two special cases: putting an immediately expiring value is counted as a normal put, and removing a key that's not present should be counted as a plain normal remove.

Agree? Disagree?


lorban:
test this please

AbfrmBlr:
following tck tests are failing for jdk-7 and jdk-8 -
org.jsr107.tck.management.CacheMBStatisticsBeanTest > testCacheStatistics 

org.jsr107.tck.management.CacheMBStatisticsBeanTest > testExpiryOnCreation

Issue #461, opened 2015-07-15
Improve Cache Config XML usability


JSAhuja:
In cache config XML characters like \n(New Line), spaces, \t(tabs) should be handles duringXML parsing.

Example:
```
<ehcache:serializer type="java.lang.Long">
                     serializers.CustomSerializerFloat
</ehcache:serializer>
```
When using the above configuration, getting class not found exception.

alexsnaps:
I had a quick look at possibly fixes... Here's my take for now:

 - add a dedicated FQCN type to our XSD;
 - use that type where classnames are to be expected;
 - add an adapter to XJC for that type, so it trims the text value.

wdyt?

JSAhuja:
I think so this should work.

PR #483, opened 2015-07-28
Adding whitespace and line feeds collapse for class names #461

AbfrmBlr:


AbfrmBlr:
We can discuss this. I have created a new simpletype which does collapse for line feeds and extra spaces, so I guess we don't need to write adapter thing for this issue.

xml/src/test/resources/configs/default-serializer.xml
```diff
@@ -14,6 +14,8 @@
   ~ limitations under the License.
   -->
 
+<!-- Please do not format the file, this is there to test #461   -->
```
chrisdennis: Can we have a test that parses this file in some way that doesn't collapse the whitespace and asserts that the whitespace we expect to see still exists?  (doesn't even have to be an XML parser)

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -403,4 +403,9 @@
       <xs:enumeration value="UPDATED" />
     </xs:restriction>
   </xs:simpleType>
+  <xs:simpleType name="fully-qualified-class-name-type">
```
ljacomet: Minor change: rename the type to `fqcn-type` which is shorter (and I hope meaningful to a java dev)

Issue #474, opened 2015-07-22
Caching tiers should flush all cached entries on shutdown


chrisdennis:
On shutdown caching tiers should flush all cached entry metadata down on to the authoritative tier.  This has two desired effects:

1. It updates all the entry metadata to be correct as of the point of shutdown (cache hits, access times etc.)
2. It means that upon recovery of the authority none of the entries will be pinned which is 'A Good Thing&trade;'.

PR #486, opened 2015-07-29
Issue #474 : Fixing Caching tiers should flush all cached entries…

rishabhmonga:
… on shutdown

### DIFF
--- a/core/src/main/java/org/ehcache/spi/cache/tiering/CachingTier.java
+++ b/core/src/main/java/org/ehcache/spi/cache/tiering/CachingTier.java
@@ -59,7 +59,7 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
    * Empty out the caching store.
    * @throws CacheAccessException
    */
-  void invalidate() throws CacheAccessException;
+  void clear() throws CacheAccessException;
 
   /**
    * Set the caching tier's invalidation listener. The invalidation listener can only be set once.
--- a/impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
+++ b/impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
@@ -366,8 +366,25 @@ public class OnHeapStore<K, V> implements Store<K,V>, CachingTier<K, V> {
     map.clear();
   }
 
-  @Override
-  public void invalidate() throws CacheAccessException {
+  private void invalidate() {
+    if(map.map != null) {
+      for (K k : map.map.keySet()) {
+        try {
+          invalidate(k);
+        } catch (CacheAccessException cae) {
+          LOG.warn("Failed to invalidate mapping for key {}", k, cae);
+        }
+      }
+    }
+    if(map.keyCopyMap != null) {
+      for(OnHeapKey<K> key : map.keyCopyMap.keySet()) {
+        try {
+          invalidate(key.getActualKeyObject());
+        } catch (CacheAccessException cae) {
+          LOG.warn("Failed to invalidate mapping for key {}", key, cae);
+        }
+      }
+    }
     map.clear();
   }
 
@@ -503,6 +520,7 @@ public class OnHeapStore<K, V> implements Store<K,V>, CachingTier<K, V> {
       return null;
     }
 
+    setAccessTimeAndExpiry(key, cachedValue, now);
     return getValue(cachedValue);
   }
 
@@ -1057,6 +1075,7 @@ public class OnHeapStore<K, V> implements Store<K,V>, CachingTier<K, V> {
 
     @Override
     public void releaseCachingTier(CachingTier<?, ?> resource) {
+      ((OnHeapStore)resource).invalidate();
       releaseStore((Store<?, ?>) resource);
     }
 
--- a/impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
+++ b/impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
@@ -213,7 +213,7 @@ public class CacheStore<K, V> implements Store<K, V> {
       authoritativeTier.clear();
     } finally {
       try {
-        realCachingTier.invalidate();
+        realCachingTier.clear();
       } finally {
         if(!cachingTierRef.compareAndSet(noopCachingTier, realCachingTier)) {
           throw new AssertionError("Something bad happened");
@@ -454,7 +454,7 @@ public class CacheStore<K, V> implements Store<K, V> {
     }
 
     @Override
-    public void invalidate() throws CacheAccessException {
+    public void clear() throws CacheAccessException {
       // noop
     }
 
--- a/impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreTest.java
+++ b/impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreTest.java
@@ -19,9 +19,11 @@ import org.ehcache.exceptions.CacheAccessException;
 import org.ehcache.function.BiFunction;
 import org.ehcache.function.Function;
 import org.ehcache.function.NullaryFunction;
+import org.ehcache.spi.ServiceLocator;
 import org.ehcache.spi.cache.Store;
 import org.ehcache.spi.cache.tiering.AuthoritativeTier;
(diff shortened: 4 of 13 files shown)


impl/src/main/java/org/ehcache/internal/store/tiering/CacheStore.java
```diff
@@ -383,13 +383,15 @@ public void disableStoreEventNotifications() {
...
+    @SuppressWarnings("unchecked")
     @Override
-    public void releaseStore(Store<?, ?> resource) {
+    public void releaseStore(Store<?, ?> resource) throws CacheAccessException {
       Map.Entry<CachingTier.Provider, AuthoritativeTier.Provider> entry = providersMap.get(resource);
       if (entry == null) {
         throw new IllegalArgumentException("Given store is not managed by this provider : " + resource);
       }
       CacheStore cacheStore = (CacheStore) resource;
+      cacheStore.flushAll((Store)cacheStore.realCachingTier);
```
AbfrmBlr: With #487 this assumption may not be true anymore.

ljacomet: Had a chat with Rishabh on this, I believe the right path is to make use of `CachingTier.invalidate()`

rishabhmonga:
test this please

impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreTest.java
```diff
@@ -80,8 +83,9 @@ public void testGetHitsAuthoritativeTier() throws Exception {
     when(cachingTier.getOrComputeIfAbsent(any(Number.class), any(Function.class))).then(new Answer<Store.ValueHolder<CharSequence>>() {
       @Override
       public Store.ValueHolder<CharSequence> answer(InvocationOnMock invocation) throws Throwable {
-        Number key = (Number) invocation.getArguments()[0];
```
ljacomet: Can you revert all these formatting changes and tweak the config of your IDE to only reformat code you touched?
They add noise to the PR for no added value. Same holds for the next 4 blocks of changes.

rishabhmonga:
test this please

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -367,6 +367,16 @@ public void clear() throws CacheAccessException {
 
   @Override
   public void invalidate() throws CacheAccessException {
+    if(map.map != null) {
```
ljacomet: The whole invalidation process may take a while. Should we improve this part so that it can happens without interference from other cache operations? I am interested in having other opinions on this.

One idea would be to swap the underlying map while we do the operation as it can be discarded anyway.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -367,6 +367,16 @@ public void clear() throws CacheAccessException {
 
   @Override
   public void invalidate() throws CacheAccessException {
```
ljacomet: With the change of contract, the call from `CacheStore.clear()` seems wrong. We may need to add a `clear()` method on the `CachingTier` interface. Others?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
chrisdennis: Do we ever invalidate when other cache operations are happening?

rishabhmonga:
test this please

impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreTest.java
ljacomet: Looks like this updated changeset revived the unrelated formatting changes :smile:

impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreTest.java
```diff
@@ -270,7 +273,7 @@ public void testClear() throws Exception {
 
     cacheStore.clear();
 
-    verify(cachingTier, times(1)).invalidate();
+//    verify(cachingTier, times(1)).invalidate();
```
ljacomet: Please do not leave commented out code

ljacomet:
Minor remarks otherwise matched what was discussed

rishabhmonga:
test this please

rishabhmonga:
test this please

impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreTest.java
```diff
@@ -349,8 +351,8 @@ public void testComputeIfAbsent_computes() throws Exception {
     when(authoritativeTier.computeIfAbsentAndFault(any(Number.class), any(Function.class))).thenAnswer(new Answer<Store.ValueHolder<CharSequence>>() {
       @Override
       public Store.ValueHolder<CharSequence> answer(InvocationOnMock invocation) throws Throwable {
-        Number key = (Number) invocation.getArguments()[0];
```
ljacomet: This is again unrelated formatting changes - can you get rid of them?

rishabhmonga:
test this please

core/src/main/java/org/ehcache/spi/cache/Store.java
```diff
@@ -441,7 +441,7 @@
      * Informs this Provider, a Store it created is being disposed (i.e. closed)
      * @param resource
      */
-    void releaseStore(Store<?, ?> resource);
+    void releaseStore(Store<?, ?> resource) throws CacheAccessException;
```
lorban: I'm not sure rethrowing CacheAccessException is the right thing to do here. IMHO the releaseStore() implementations should handle those internally.

core/src/main/java/org/ehcache/spi/cache/tiering/CachingTier.java
```diff
@@ -82,7 +82,7 @@
   interface Provider extends Service {
     <K, V> CachingTier<K, V> createCachingTier(Store.Configuration<K, V> storeConfig, ServiceConfiguration<?>... serviceConfigs);
 
-    void releaseCachingTier(CachingTier<?, ?> resource);
+    void releaseCachingTier(CachingTier<?, ?> resource) throws CacheAccessException;
```
lorban: Same remark as above: do we want this exception to be rethrown or handled internally?

core/src/main/java/org/ehcache/spi/cache/Store.java
chrisdennis: Anything thrown here is going to propagate all the the way out the the user wrapped in a `StateTransitionException`.  I'm not averse to having these methods throw when bad things happen, but to me two things need to happen.  Firstly we can't be throwing eagerly, and secondly I don't think we should be throwing CacheAccessException... it's an internal exception that we shouldn't be leaking (even if it is wrapped).  Does this sound sensible?

alexsnaps: Right, and there is logic in those state transitions, so _I think_ in this case, the `.shutdown()` will continue. If anything though, I agree with @chrisdennis and we shouldn't make throwing `CacheAccessException` part of the contract, as it's an internal, implementation specific detail that the `Store` is even accessed further when `releaseStore` is invoked.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -366,8 +366,17 @@ public void clear() throws CacheAccessException {
...
+    if(map.map != null) {
+      for (K k : map.map.keySet()) {
+        invalidate(k);
+      }
+    }
+    if(map.keyCopyMap != null) {
+      for(OnHeapKey<K> key : map.keyCopyMap.keySet()) {
+        invalidate(key.getActualKeyObject());
+      }
+    }
```
alexsnaps: This can't possibly happen! See the discussion on @lorban PR, this method _should_ be a plain clear!

rishabhmonga:
test this please

PR #488, opened 2015-07-30
Issue #457

ljacomet:
Quite a number of things in here, hopefully nice isolated in separate commits:

* Make `ServiceLocator.discoverService` private
* Fix contracts of `ServiceFactory.create` and `Service.start` to prevent different outcomes depending on the started state of `ServiceLocator`
* Change `ServiceProvider` api
* Two small commits to improve / fix unrelated things


ljacomet:
Discussions during public dev meeting led to trying yet another way of indicating which services a `CacheManager` must always provide, whether they have been explicitly configured or not. Back to work ...

Issue #469, opened 2015-07-21
Cache manager with custom classloader produces ClassNotFoundException when deserialize cache entry


exmortal:
Hi,

I found a strange situation with a cache manager classloader. 

I want to create a cache manager for particular classloader. I have a custom class that will be used as a cache entry. Cache manager creates by classloader where this class is located. 

Here is my steps:
1) create a cache manager for selected classloader
2) create a cache instance
3) put a single value to cache
4) try to get the same value, but get ClassNotFoundException ,

Seems that selected classloader doesn't take part in a deserialization process. I have a quick look to echache source code and found that manager classloader is used only for manager's identification by pair URI + classloader. Cache instances use a classloader where cache provider initialized for deserialization.

JSR says "To acquire non-default or alternative configurations of CacheManagers, for example with custom ClassLoaders or caching implementor properties, developers should use one of the overloaded CachingProvider getCacheManager methods."
I believe that this classloader should be used for serialization/deserialization of the cache entries.

I just want to know is it a bug or expected behavior. 

Thanks,

BR,
Vlad

ljacomet:
Hi Vlad,

It would be great if you were able to post sample code exactly illustrating
what you are trying to achieve.
In parallel, we may have a bug in the code, will double check something I
just saw when browsing it right now.

I'll keep you posted.

Regards,
Louis

exmortal:
Hi Louis,

I try to explain. I have 2 OSGi bundles. One of them contains echcach v3 jar and takes care about cache manager initialization. (create and expose it via OSGi service method). Second one needs to get a cache instance from provider initialized in bundle1. Also it contains CacheItem class that should be used as a value type. 

Here is a test code: 
// get an instance of JSR107 cache provider created in bundle1
CacheProvider provider = bundle1CacheProviderFactory.getProvider();
// get an instance of cache manager for the bundle2 classloader
CacheManager manager = provider.getCacheManager(bundle2ClassLoader, null);

// create cache instance with empty configuration
MutableConfiguration<String, CacheItem> configuration = new MutableConfiguration<>();
final Cache<String, CacheItem> cache = manager.createCache(cacheName, configuration);

CacheItem item1 = new CacheItem();
cache.put("1", item1); // OK
cache.get("1"); // ClassCastException. Dezerializer uses classloader where cache provider created (classloader of bundle1) but not one I specified when create a cache manager. 

I know that if I use a own cache provider in each bundle will solve this issue, but I think that it will be a  real overheat. 

[...]

PR #490, opened 2015-07-30
Issue #469 Take ClassLoader into account when using default URI

ljacomet:


ljacomet:
Updated change with a test

chrisdennis:
Looks okay to me... needs rebasing now, but once that is done I'll be happy to merge.

Issue #487, opened 2015-07-29
CacheStore.getConfigurationChangeListeners() incorrectly assumes that caching and authoritative tiers both implement Store


lorban:

lorban:
In CacheStore.getConfigurationChangeListeners() both tiers are cast to Store to fetch their config change listeners.

So far, this assumption is correct but the new 2-tiers caching tier combining onheap and offheap looks like it won't implement Store.

PR #491, opened 2015-07-31
Issue #492 and misc changes

ljacomet:
* delete an obsolete comment
* clean up SerializerTest


core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -341,7 +341,6 @@ public void close() throws Exception {
       Collection<CacheEventListenerConfiguration> evtLsnrConfigs =
       ServiceLocator.findAmongst(CacheEventListenerConfiguration.class, config.getServiceConfigurations().toArray());
       for (CacheEventListenerConfiguration lsnrConfig: evtLsnrConfigs) {
-        // XXX this assumes a new instance returned for each call - yet args are always the same. Is this okay?
         final CacheEventListener<K, V> lsnr = evntLsnrFactory.createEventListener(alias, lsnrConfig);
```
AbfrmBlr: Do we really need a commit just to remove the comment ?

ljacomet: This allows to reference the issue that should have ended up removing the comment.

ljacomet:
Added content for #492 and one more clean up related to #487

PR #495, opened 2015-08-04
Issue #457

ljacomet:
* Make ServiceLocator.discoverService private
* Single path for service initialization through ServiceLocator
* Solution for service initialization without configuration
* Improve error reporting when no persistence service is found

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
+++ b/107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
@@ -37,7 +37,7 @@ import org.ehcache.management.ManagementRegistry;
 import org.ehcache.spi.ServiceLocator;
 import org.ehcache.spi.ServiceProvider;
 import org.ehcache.spi.service.Service;
-import org.ehcache.spi.service.ServiceDependency;
+import org.ehcache.spi.service.ServiceDependencies;
 import org.ehcache.util.ClassLoading;
 
 /**
@@ -118,7 +118,7 @@ public class EhcacheCachingProvider implements CachingProvider {
     return cacheManager;
   }
 
-  @ServiceDependency(services = ManagementRegistry.class)
+  @ServiceDependencies(ManagementRegistry.class)
   static class ManagementRegistryCollectorService implements Service {
 
     public volatile ManagementRegistry managementRegistry;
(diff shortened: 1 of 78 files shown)


107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -118,13 +118,14 @@ public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties
     return cacheManager;
   }
 
+  @ServiceDependency(services = ManagementRegistry.class)
```
chrisdennis: This (and other instances) can be done without the `services = ` right?

core/src/main/java/org/ehcache/spi/ServiceLocator.java
```diff
@@ -145,8 +136,17 @@ void addService(final Service service, final boolean expectsAbstractRegistration
         }
       }
 
+      if (service.getClass().isAnnotationPresent(ServiceDependency.class)) {
```
chrisdennis: Could we just to this with the `getAnnotation` call and a null check?

core/src/main/java/org/ehcache/spi/ServiceLocator.java
```diff
@@ -164,25 +164,28 @@ void addService(final Service service, final boolean expectsAbstractRegistration
...
   }
 
   @Override
-  public <T extends Service> T findService(Class<T> serviceType) {
-    return findService(serviceType, null);
+  public <T extends Service> T getService(Class<T> serviceType) {
+    return findService(serviceType, null, false);
+  }
+
+  public <T extends Service> T getOrCreateService(Class<T> serviceType) {
```
chrisdennis: Crazy idea... could we eliminate this method entirely by adding services that list the mandatory services as their dependencies?

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
ljacomet: Only by renaming `services()` to `value()` but as we have only one annotation parameter, this seems to be the convention. Will do that.

core/src/main/java/org/ehcache/spi/ServiceLocator.java
ljacomet: Yep, especially given the implementation of `isAnnotationPresent`. Still, I prefer the current style - feels more "correct" and less `null` tied.

core/src/main/java/org/ehcache/spi/ServiceLocator.java
ljacomet: I do not think so. Using `getOrCreateService` still allows the `Service` instance to be resolved by finding a proper `ServiceFactory` or just be a no-op when a `Service` instance has been provided at configuration.

ljacomet:
All comments have been addressed, including the crazy idea ... see added commit for details

Issue #494, opened 2015-08-03
Persistence Context not working as expected


rishabhmonga:
destroy() does not destroy non-persistent context

rishabhmonga:
when `OffHeapDiskStore.Provider.releaseStore(store)` is called for a non-persistent store the `fileBasedPersistenceContext` should be cleaned up

ljacomet:
@rishabhmonga Not at all, that was the whole point of #355 in order to decouple this. Let's have a chat to better describe this issue.

rishabhmonga:
`EhcacheManager.destroy()` should clean up all `persistenceContext`. This will require starting the `LocalPersistenceService` so that the cleanup can be done.

PR #505, opened 2015-08-06
Issue #494 : destroy() should clean all file based persistence context

rishabhmonga:


core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -496,13 +498,25 @@ public void create() {
 
         @Override
         public void destroy() {
+          startPersistenceService();
```
ljacomet: As starting locks the persistence location and failing to do so indicates it is used by someone else, it must be part of "going to maintenance" because you can't go into maintenance unless that succeeds.
So starting and stopping must happen as part of going and leaving maintenance.

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -213,6 +222,20 @@ static void destroy(String identifier, FileBasedPersistenceContext fileBasedPers
     }
   }
 
+  private static boolean recursiveDeleteDirectoryContent(File file) {
+    if (file.isDirectory()) {
+      File[] contents = file.listFiles();
+      if (contents.length > 0) {
+        for (File f : contents) {
+          if(!recursiveDelete(f)) {
+            return false;
```
lorban: I wouldn't abort the deletion process if one file/folder could not be deleted. I'd store that failure in a boolean flag, go on with the deletion and return that flag at the end of the method.

impl/src/main/java/org/ehcache/internal/persistence/DefaultLocalPersistenceService.java
```diff
@@ -166,6 +166,15 @@ public void destroyPersistenceContext(Object identifier) throws CachePersistence
     }
   }
 
+  @Override
+  public void destroyAllPersistenceContext() {
+    if(recursiveDeleteDirectoryContent(rootDirectory)){
+      LOGGER.info("Destroying all file based persistence context");
```
lorban: The log should rather say "Destroyed all file based persistence context" instead of "Destroying..." as the logging happens after deletion was done.

Issue #496, opened 2015-08-04
EXPIRED events aren't fired for mappings that are in the caching tier


lorban:
Here's a crude test that helps reproducing the issue:

````java
  @Test
  public void testExpiredEvents() throws Exception {
    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
        .withCache("tieredCache",
            CacheConfigurationBuilder.newCacheConfigurationBuilder()
                .withExpiry(Expirations.timeToLiveExpiration(new Duration(1, TimeUnit.SECONDS)))
                .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
                        .heap(3, EntryUnit.ENTRIES)
                        .offheap(10, MemoryUnit.MB)
                )
                .buildConfig(Long.class, String.class)).build(true);

    final List<Long> expiredKeys = new CopyOnWriteArrayList<Long>();

    Cache<Long, String> tieredCache = cacheManager.getCache("tieredCache", Long.class, String.class);

    tieredCache.getRuntimeConfiguration().registerCacheEventListener(new CacheEventListener<Long, String>() {
      @Override
      public void onEvent(CacheEvent<Long, String> event) {
        expiredKeys.add(event.getEntry().getKey());
      }
    }, EventOrdering.ORDERED, EventFiring.SYNCHRONOUS, EnumSet.of(EventType.EXPIRED));

    tieredCache.put(1L, "one");
    tieredCache.put(2L, "two");
    tieredCache.put(3L, "three");
    tieredCache.put(4L, "four");
    tieredCache.put(5L, "five");

    tieredCache.get(1L);
    tieredCache.get(2L);
    tieredCache.get(3L);

    try {
      Thread.sleep(1100);
    } catch (InterruptedException e) {
      e.printStackTrace();
    }

    tieredCache.get(1L);
    tieredCache.get(2L);
    tieredCache.get(3L);
    tieredCache.get(4L);
    tieredCache.get(5L);

    cacheManager.close();

    assertThat(expiredKeys, containsInAnyOrder(1L, 2L, 3L, 4L, 5L));
  }
````

PR #511, opened 2015-08-07
issue #496 Fire expired events when entries are flushed from Authoritative tier

albinsuresh:


integration-test/src/test/java/org/ehcache/integration/ExpiryEventsTest.java
```diff
@@ -0,0 +1,98 @@
...
+    tieredCache.put(3L, "three");
+    tieredCache.put(4L, "four");
+    tieredCache.put(5L, "five");
+
+    tieredCache.get(1L);
+    tieredCache.get(2L);
+    tieredCache.get(3L);
+
+    try {
+      Thread.sleep(1100);
```
lorban: No. We can't have tests with Thread.sleep() in them. The time source is pluggable, so make use of that.

PR #522, opened 2015-08-12
Issue #521 Switch to annotations based Findbugs supression and cleanup what can be

chrisdennis:


build.gradle
```diff
@@ -241,10 +246,12 @@ subprojects {
   findbugs {
     ignoreFailures = false
     sourceSets = [sourceSets.main]
-    excludeFilter = file("$projectDir/config/findbugs-excludes.xml")
-    toolVersion = '2.0.3'
+    if (JavaVersion.current().compareTo(JavaVersion.VERSION_1_7) >= 0) {
+      findbugs.toolVersion = '3.0.1'
```
ljacomet: 3.0.1 here, 3.0.0 above, is that correct?

ljacomet: Checked maven central and indeed no 3.0.1 for the annotations. All good

Issue #297, opened 2015-02-10
Document tiering system


ljacomet:

ljacomet:
This is about documenting the implementation of the tiering system for Ehcache developers on the wiki.

PR #528, opened 2015-08-19
Adding diagram for faulting #297

AbfrmBlr:


ljacomet:
* I do not like the vertical orientation of the text, makes it hard to process the information
* There should not be text on top of text
* Terminology can be improved:
  * "The returning thread always removes the value from the caching tier ..." -> "On return, the mapping in the caching tier is invalidated"
  * "Whenever CT evicts or expires an entry it flushes the same to authority so that it can be marked as evictable" -> "On eviction or expiry in the caching tier, the mapping is flushed to the authority"
  * "The returning thread always returns with the read value from authority, but the value is installed in caching tier ff it is the correct value at that time" -> "On return, the mapping faulted from the authority is installed in the caching tier. The install may fail under heavy contention to preserve correctness."

Another way of addressing the issue, is to have the diagram carry only numbers next to the arrows and edit the wiki to contain the description for each of these.

Issue #512, opened 2015-08-07
Build failure in docs project with PermGen space error on Java 6


albinsuresh:
Environment:
java version "1.6.0_65" on Mac

Snippet from the build output:

:docs:asciidoctor FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':docs:asciidoctor'.
> PermGen space

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

albinsuresh:
Build passes on Java versions 7 and 8.

ljacomet:
This is a consequence of PR #424 which resolved issue #394. It probably should be documented on the dev wiki.

ljacomet:
Found a general solution for this:

* Bump gradle version to 2.6 (and some plugins that need an update)
* Configure the build to use the "gradle daemon" which can be configured with JVM args

Unless someone has a strong argument against it, this is the way to go for "portable" builds.

PR #529, opened 2015-08-19
Issue #512 Use gradle daemon

ljacomet:
* Bump gradle to version 2.6
* Bump shadow plugin to 1.2.2
* Configure gradle daemon with max memory and perm gen memory


chrisdennis:
The Gradle documentation doesn't call it out explicitly, but it does always refer to enabling the daemon in an environment, not in a project.  I'd rather just add some documentation to the developer wiki about how to get around it rather than enabling the daemon in the project itself.  That way we also don't have to worry about remembering to disable it in the CI system(s).

https://docs.gradle.org/current/userguide/gradle_daemon.html

ljacomet:
Good point indeed.

So of the two properties:
````
org.gradle.daemon=true
org.gradle.jvmargs=-Xmx512m -XX:MaxPermSize=128m 
````
I remove the first one, keep the second one and add a wiki section documenting that you need to either use the daemon or apply `GRADLE_OPTS`.
Better?

chrisdennis:
Yes, I think that makes sense.

ljacomet:
Udpated PR and find below the proposed [wiki](https://github.com/ehcache/ehcache3/wiki) addition:

> ## Developer be welcome
> ...
>
> ## Building Ehcache 3
> The build system is [Gradle](http://gradle.org). Ehcache 3 is built by default using Java 1.6 and remains compatible with that version.
> The project is configured to use the [Gradle daemon](https://docs.gradle.org/current/userguide/gradle_daemon.html). Simply add a `gradle.properties` file in your `$HOME/.gradle/` folder with the following content:
> ````
> org.gradle.daemon=true
> ````
> If that is not an option, make sure you define the following environment variable:
> `GRADLE_OPTS=GRADLE_OPTS="-Xmx512m -XX:MaxPermSize=128m"`
>
> ## How to contribute?
> ...

Issue #158, opened 2014-11-18
Implement CopyProvider that abstracts SerializationProvider away in the by-value on-heap store


lorban:

ljacomet:
Offer an API to do by value caching without relying on `Serialization`

albinsuresh:
Here is a proposal for the user-facing APIs for store-by-value caching.

When store-by-value is enabled for a cache the user can configure the cache with custom `Copier` classes for keys and values to be used for copying. The user can create a copier by implementing the `Copier<T>` interface.

```java
public interface Copier<T> {
  /**
   * Creates a copy of the instance passed in.
   *
   * @param obj the instance to copy
   * @return the copy of the {@code obj} instance
   */
  T copy(T obj);
}

```

And the user would configure the copiers as follows:

```java

    CacheConfiguration<Description, Person> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
        .add(new OnHeapStoreServiceConfiguration().storeByValue(true)) // <1>
        .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).build())
        .add(new DefaultCopierConfiguration<Description>(DescriptionCopier.class,
            CopierConfiguration.Type.KEY)) //// <2>
        .add(new DefaultCopierConfiguration<Person>(PersonCopier.class,
            CopierConfiguration.Type.VALUE)) //// <3>
        .buildConfig(Description.class, Person.class);

    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
        .withCache("cache", cacheConfiguration)
        .build(true);

    Cache<Long, String> cache = cacheManager.getCache("cache", Long.class, String.class);

    cache.put(1L, "one");
    assertThat(cache.get(1L), equalTo("one"));

    cacheManager.close();

```

1. By default, objects stored on-heap are stored by reference. If you wish to store them by value (which requires
    copying them) then you have to set storeByValue to true.
2. Configure the copier to be used for keys. *DescriptionCopier* in this case  as the keys are *Description* objects.
3. Configure the copier to be used for values. *PersonCopier* in this case as the values are *Person* objects.

The custom types and the corresponding copiers used in the above code snippet are the following(just for reference):

```java

  private static class Description {
    int id;
    String alias;

    Description(Description other) {
      this.id = other.id;
      this.alias = other.alias;
    }
  }

  private static class Person {
    String name;
    int age;

[...]

Issue #527, opened 2015-08-19
Default Serializer at Cache manager level for jsr107 Cache


JSAhuja:
Required to throw an expection, when Default Serilizer defined at cache manager level for jsr107 configuration.

Sample config:
````
CachingProvider provider=Caching.getCachingProvider();
CacheManager cacheManager = provider.getCacheManager();
DefaultSerializationProviderConfiguration defaultSerialization =
				new DefaultSerializationProviderConfiguration().addSerializerFor(Number.class, CustomSerializerNumber.class);
CacheConfiguration<Float, Long> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
				.withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder().heap(20, EntryUnit.ENTRIES))
				.add(defaultSerialization)
				.buildConfig(Float.class, Long.class);
		Cache<Float, Long> cache = cacheManager.createCache("Heapcache",Eh107Configuration.fromEhcacheCacheConfiguration(cacheConfiguration));
````

JSAhuja:
As more generalized, jsr107 Cache configuration should not allow any Cache manager level ServiceConfiguration.

ljacomet:
As this problem has been raised multiple times and seems to be a source of confusion, it seems time has come to introduce marker interfaces that will prevent the builders from taking a configuration from the wrong level.

Issue #455, opened 2015-07-07
Add support for Serializer instances with durable state, and add serializer implementations that eliminate OSC headers


chrisdennis:
In order to enable the efficient storage of keys and values in persistent datasets we will need to support the existence of Serializers with durable state, and support the recovery of that state at initialization time.  See the PersistentSerializablePortability in Terracotta-OSS/offheap-store for the motivation behind this.

PR #530, opened 2015-08-20
Issue #158 Implementation of CopyProvider

albinsuresh:
This is the initial implementation of CopyProvider service. The following are the features that I have implemented as part of this PR:

1. Copier and CopyProvider APIs and the helper implementations like ReadWriteCopier, IdentityCopier
2. SerializingCopier implementation
3. Default copiers at the cache manager level.

Things *TO BE DONE*:

1. JSR integration
2. XML integration
3. Removing `OnHeapStoreServiceConfiguration`. Could not remove it yet due to the dependencies in JSR and XML modules.


albinsuresh:
The test failures are around store-by-value tests in TCK. Since I haven't done the JSR integration these were bound to fail. So please go ahead with the review.

Now I'm surprised to see that TCK tests are skipped for Java 1.6. Any reason for that?

ljacomet:
JSR-107 TCK requires Java 1.7, that's the reason

impl/src/main/java/org/ehcache/internal/copy/SerializingCopier.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private final Serializer<T> serializer;
+
+  public SerializingCopier(Serializer<T> serializer) {
+    this.serializer = serializer;
+  }
+
+  @Override
+  public T copyOnRead(final T obj) {
+    throw new AssertionError("This method should never be used.");
```
ljacomet: Why? We will do our best to not use it, and I can see that behaviour in a test class. But not in a product class.
Same of `copyOnWrite`

impl/src/main/java/org/ehcache/spi/copy/DefaultCopyProvider.java
```diff
@@ -0,0 +1,104 @@
...
+  public DefaultCopyProvider(DefaultCopyProviderConfiguration configuration) {
+    super(configuration, (Class) DefaultCopierConfiguration.class);
+  }
+
+
+  @Override
+  public <T> Copier<T> createKeyCopier(final Class<T> clazz, ServiceConfiguration<?>... configs) {
+    DefaultCopierConfiguration<T> conf = find(CopierConfiguration.Type.KEY, configs);
+    if(conf != null && conf.getClazz().isAssignableFrom(SerializingCopier.class)) {
+      SerializationProvider serializationProvider = serviceProvider.getService(SerializationProvider.class);
```
ljacomet: You should declare the `SerialisationProvider` as a service dependency.

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -133,9 +137,9 @@ public void threeTiersCacheManager() throws Exception {
         .withCache("threeTieredCache",
             CacheConfigurationBuilder.newCacheConfigurationBuilder()
                 .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
-                        .heap(10, EntryUnit.ENTRIES) // <2>
```
ljacomet: Please do not include irrelevant formatting changes in your commit

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -147,7 +151,6 @@ public void threeTiersCacheManager() throws Exception {
   public void defaultSerializers() throws Exception {
     // tag::defaultSerializers[]
     CacheConfiguration<Long, String> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
-        .add(new OnHeapStoreServiceConfiguration().storeByValue(true)) // <1>
```
ljacomet: You cannot just remove lines that have a callout, the corresponding entry in doc must be adapted and subsequent numbers too.

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -302,6 +304,176 @@ public void updateResourcesAtRuntime() throws InterruptedException {
...
+      this.alias = alias;
+    }
+
+    @Override
+    public boolean equals(final Object other) {
+      if(this == other) return true;
+      if(other == null || this.getClass() != other.getClass()) return false;
+
+      Description that = (Description)other;
+      if(id != that.id || alias != that.alias) return false;
```
ljacomet: That's wrong, don't compare `String` using `==`

Same for `Person`

ljacomet:
I've had a quick look at the number of Copier / Serializer calls and it is pretty bad - not that it's cause specially by the changes here.

One possible improvement is to make sure methods such as `org.ehcache.internal.store.heap.OnHeapStore#newUpdateValueHolder` do not cause an invocation - signature could be change to make sure it uses `V oldValue` instead of the holder.

107/src/test/java/org/ehcache/jsr107/EhCachingProviderTest.java
```diff
@@ -64,7 +64,7 @@ public void testCacheUsesCacheManagerClassLoaderForDefaultURI() {
 
     CacheManager cacheManager = cachingProvider.getCacheManager(cachingProvider.getDefaultURI(), limitedClassLoader);
 
-    MutableConfiguration<Object, Object> configuration = new MutableConfiguration<Object, Object>();
+    MutableConfiguration<Object, Object> configuration = new MutableConfiguration<Object, Object>().setStoreByValue(true);
```
ljacomet: This change brings nothing as it matches the default. Can you get rid of it?

107/src/test/resources/ehcache-107-integration.xml
```diff
@@ -9,8 +9,7 @@
     </jcache:defaults>
   </ehcache:service>
 
-  <ehcache:cache alias="productCache"
-                 storeByValueOnHeap="true"> <!-- OPTIONAL: storing by value on heap, will result in keys and values being copied in and out of the Cache -->
```
ljacomet: Your changes make this attribute obsolete. But could you update the test so that it still "creates" a cache consider byValue in 107 world?

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -270,26 +270,32 @@ private boolean isDiskTransient(Ehcache<?, ?> ehcache) {
...
+        }
+      }
+      try {
+        valueSerializer = serialization.createValueSerializer(valueType, config.getClassLoader(), serviceConfigs);
         lifeCycledList.add(closerFor(valueSerializer));
       } catch (UnsupportedTypeException e) {
-        Set<ResourceType> resources = config.getResourcePools().getResourceTypeSet();
         if (resources.contains(DISK) || resources.contains(OFFHEAP)) {
+          try {
+            if(keySerializer != null) {
```
ljacomet: `keySerializer` was already added to the `lifeCycledList`, so no need to take it into account here.

albinsuresh:
Changes from the previous iteration:

1. CopyProvider service uses the serializers from the cache configuration directly instead of finding the Serialization service on its own. So SerializationProvider is not a dependency of CopyProvider anymore. But as a side effect every CopyProvider implementations need to take in a Serializer for key/value irrespective of the fact whether they use it or not.
2. XML configuration integration of CopyProvider service
3. JSR-107 integration of CopyProvider service
4. Cache manager level configuration of default Copiers. Modified this code to comply with the configuration level changes introduced in #527.
5. Fixed a bug introduced as part of #455 with respect to the creation of serializers for keys and values.

api/src/main/java/org/ehcache/spi/copy/Copier.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public interface Copier<T> {
+
+  /**
+   * Creates a copy of the instance passed in.
+   *
+   * @param obj the instance to copy
+   * @return the copy of the {@code obj} instance
+   */
+  T copyOnRead(T obj);
```
chrisdennis: copyForRead?

api/src/main/java/org/ehcache/spi/copy/Copier.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Creates a copy of the instance passed in.
+   *
+   * @param obj the instance to copy
+   * @return the copy of the {@code obj} instance
+   */
+  T copyOnWrite(T obj);
```
chrisdennis: copyForWrite

api/src/main/java/org/ehcache/spi/copy/Copier.java
albinsuresh: Wouldn't it be better if we stick to the ehcache 2.x terminologies??? And personally I feel that copyOnRead sounds better than copyForRead.

107/src/test/java/org/ehcache/jsr107/Eh107XmlIntegrationTest.java
```diff
@@ -179,12 +182,15 @@ public void testXmlExampleIn107() throws Exception {
     Eh107ReverseConfiguration<Long, Product> eh107ReverseConfiguration = productCache.getConfiguration(Eh107ReverseConfiguration.class);
     assertThat(eh107ReverseConfiguration.isReadThrough(), is(true));
     assertThat(eh107ReverseConfiguration.isWriteThrough(), is(true));
-    assertThat(eh107ReverseConfiguration.isStoreByValue(), is(true));
```
ljacomet: With the changes done to XML, this test assertion can remain and pass, right?

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
```diff
@@ -89,8 +89,9 @@ public void mergeConfigNoTemplateNoLoaderWriter() {
     boolean storeByValue = false;
     Collection<ServiceConfiguration<?>> serviceConfigurations = configHolder.cacheConfiguration.getServiceConfigurations();
     for (ServiceConfiguration<?> serviceConfiguration : serviceConfigurations) {
-      if (serviceConfiguration instanceof OnHeapStoreServiceConfiguration) {
-        storeByValue = ((OnHeapStoreServiceConfiguration) serviceConfiguration).storeByValue();
+      if (serviceConfiguration instanceof CopierConfiguration) {
+        storeByValue = true;
```
ljacomet: This condition is broken. If there is a `CopierConfiguration`, then it must be `storeByValue = true`. And changing that fails the test, showing that a former `storeByRef` template no longer overrides the default `storeByValue` of JSR-107.
So both test and code needs to be adapted to restore behaviour.

107/src/test/resources/ehcache-107-serializer.xml
```diff
@@ -12,13 +12,13 @@
...
 
-  <cache alias="cache" storeByValueOnHeap="true">
+  <cache alias="cache">
     <key-type>java.lang.Long</key-type>
     <value-type>java.lang.String</value-type>
     <heap size="20" unit="entries"/>
   </cache>
 
-  <cache alias="cache1" storeByValueOnHeap="true">
+  <cache alias="cache1">
```
ljacomet: With your changes, the serialisers are not used. Can you see with @rishabhmonga what was the objective of this test and adapt / remove?

impl/src/main/java/org/ehcache/internal/store/heap/SerializedOnHeapValueHolder.java
```diff
@@ -15,24 +15,23 @@
...
+class SerializedOnHeapValueHolder<V> extends OnHeapValueHolder<V> {
   private final ByteBuffer buffer;
-  private final int hash;
   private final Serializer<V> serializer;
 
-  protected ByValueOnHeapValueHolder(V value, long creationTime, Serializer<V> serializer) {
-    this(value, creationTime, NO_EXPIRE, serializer);
-  }
+  @SuppressFBWarnings("SE_BAD_FIELD")
+  private SoftReference<V> cachedValue;
```
ljacomet: I understand the goal here but want to draw attention to this construct.
Given the places where `getCurrentValue` is called, it will most likely be _always_ populated, which may have a negative impact on the memory footprint of the `OnHeapStore` with a `SerializingCopier`.
Opinions?

impl/src/main/java/org/ehcache/internal/store/heap/SerializedOnHeapValueHolder.java
```diff
@@ -59,12 +65,24 @@ public final V value() {
     }
   }
 
+  public V getCurrentValue() {
+    if(cachedValue == null || cachedValue.get() == null) {
```
ljacomet: Make this thread safe.

api/src/main/java/org/ehcache/spi/copy/Copier.java
chrisdennis: Ehcache 2.x naming is irrelevant here as far as I'm concerned.  I prefer "copyForRead" because it puts the focus on the method, and not on the caller of the method - your vote carries though unless someone else weighs in.

lorban: I like copyForRead better, and it was the name used in ehcache 2 code too: http://ehcache.org/apidocs/2.6.9/net/sf/ehcache/store/compound/ReadWriteCopyStrategy.html

api/src/main/java/org/ehcache/spi/copy/Copier.java
lorban: Same as above: copyForWrite sounds better.

impl/src/main/java/org/ehcache/internal/store/heap/SerializedOnHeapValueHolder.java
lorban: SoftReference's have very evil side-effects. I'm very much against any usage of them anywhere.

api/src/main/java/org/ehcache/spi/copy/Copier.java
albinsuresh: So that's settled. It's 2 against 1. So "copyForRead" it is.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
albinsuresh: @ljacomet I'm not quite clear on what you meant here. The old code was checking if a cache is store-by-value based on the storeByValue field of the OnHeapStoreServiceConfiguration if one such config is found. But with my changes it is checked on the basis of the presence of an explicit CopierConfiguration. So I don't understand why that fails this test.

ljacomet: Sorry that comment was meant for line 210

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -72,10 +74,10 @@
           builder = templateBuilder;
           log.info("Configuration of cache {} will be supplemented by template {}", cacheName, templateName);
         }
+      } else {
+        builder = handleStoreByValue(jsr107Configuration, builder);
```
ljacomet: I agree with what you have done here. If you use a template, you want to benefit from Ehcache specifics without touching JSR-107 config, but we know our _default_ support of _by value_ is inefficient. So indeed, we ignore 107 and simply do what you told us from an Ehcache perspective.

However the call and the method (at least name) should be changed to reflect this better.

107/src/main/java/org/ehcache/jsr107/Eh107Configuration.java
```diff
@@ -78,8 +78,8 @@ private Eh107ConfigurationWrapper(CacheConfiguration<K, V> cacheConfiguration) {
 
     @Override
     public boolean isStoreByValue() {
-      OnHeapStoreServiceConfiguration heapConfig = findSingletonAmongst(OnHeapStoreServiceConfiguration.class, cacheConfiguration.getServiceConfigurations());
-      return heapConfig != null && heapConfig.storeByValue();
+      CopierConfiguration copierConfig = findSingletonAmongst(CopierConfiguration.class, cacheConfiguration.getServiceConfigurations());
+      return copierConfig == null;
```
ljacomet: Condition is inverted here: It must be when there is no `copierConfig` then it is `storeByRef` and if there is one then it is `storeByValue`.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
```diff
@@ -195,18 +196,19 @@ public void jsr107LoaderGetsOverriddenByTemplate() throws Exception {
...
-        newCacheConfigurationBuilder().add(new OnHeapStoreServiceConfiguration().storeByValue(false))
+        newCacheConfigurationBuilder()
     );
 
     MutableConfiguration<Object, Object> configuration = new MutableConfiguration<Object, Object>();
 
     ConfigurationMerger.ConfigHolder<Object, Object> configHolder = merger.mergeConfigurations("cache", configuration);
 
-    boolean storeByValue = true;
+    boolean storeByValue = false;
```
ljacomet: A cleaner way of initialising this var is to use `configuration.isStoreByValue()`

Issue #165, opened 2014-11-21
Decouple event creation from event firing (as a way to resolve current issues)


vfunshteyn:
As described here, the currently chosen model of delivering events (at cache level, within lambda compute functions) may fire mutative events too optimistically: https://groups.google.com/forum/#!topic/ehcache-dev/N_UTJKIlCNI

While this may be undetectable by 107's TCK, in a longer term we may want to do something about it.

ljacomet:
Timeboxed investigation: 1 day then discuss with team

AbfrmBlr:
One obvious way to solve this is to push all the event firing responsibility at Stores. This will essentially require modifying the internal data structures used by respective Stores. In this way we control that an event is only fired when the mutation in Store is successful, maintaining contract for event listeners provided by JSR spec.

The pros is that it works, but the cons is that it will require extensive testing to make sure that behavior of event firing remains same across the Stores(which is not the case as of now) as every Store gets to manipulate the event firing mechanism.

For the local stores, current event firing mechanism is not an issue actually but in case of clustered stores, this is require a change.

For all the local stores, even if the event is fired before the actual mutation  is done in data structure, the failure at store can only be attributed to reasons which cause system failure, and recovering back from it is anyways not possible like OOME, disk failure etc.

AbfrmBlr:
Just clarification in case it is not properly conveyed :- The idea is to modify code flow in Authoritative tiers as those are the places which source of truth for all mutations.

ljacomet:
From dev meeting on July 30th, we discussed that the end result of the event subsystem should be a decoupling between event creation and event firing.

Issue title and type updated to reflect this.

rishabhmonga:
Please follow the link to find proposed approach :
https://groups.google.com/d/msg/ehcache-dev/EJNFrstWM8c/tk8bKUeiAwAJ

rishabhmonga:
As per discussion, the contract of event notification will change, depending on what kind of listeners are registered. In the case where any of the listeners is `ORDERED`, `EventOrdering` for all listeners will become `ORDERED`

rishabhmonga:
[...]

PR #532, opened 2015-08-21
Issue #165 : Decouple event creation from event firing

rishabhmonga:


api/src/main/java/org/ehcache/event/CacheEvent.java
```diff
@@ -69,4 +69,72 @@
    */
   @Deprecated
   Cache<K, V> getSource();
+
+  /**
+   * This marks events fireable atomically
+   */
+  void markFireable();
```
ljacomet: That's a no go - all these methods can't make it to a pubic API. I believe the `ThreadLocal` approach is going to make some not needed, but they really need to all go - use a wrapper where you need these.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -252,4 +249,48 @@ public void setSource(Cache<K, V> source) {
...
+            }
+            if(!orderedListenersSet.isEmpty()) {
+              if(wrapper.config.firingMode() == EventFiring.SYNCHRONOUS) {
+                if(waitForEventToGetFireable(eventQueue.peek())) {
+                  CacheEvent cacheEvent = eventQueue.remove();
+                  wrapper.getListener().onEvent(cacheEvent);
+                }
+              } else {
+                eventQueue.peek().signalProcessedCondition();
+                CacheEvent cacheEvent = eventQueue.remove();
```
ljacomet: How will multiple listeners process a single event?

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -252,4 +249,48 @@ public void setSource(Cache<K, V> source) {
...
+  private boolean waitForEventToGetFireable(final CacheEvent<K, V> event) {
+    event.lockEvent();
+    try {
+      while (!event.isFireable() && !event.hasFailed()) {
+        event.awaitFireableCondition();
+      }
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    } finally {
+      event.markProcessed();
```
ljacomet: How can we mark an event as processed here??

ljacomet:
From the code I have seen in `CacheEventNotificationServiceImpl` a lot of work remains to be done. It will probably be very useful for the development to find a way to represent the different conditions when a single listener of each type exist. And then see how these would compose when multiple listeners of different or same type are declared.

rishabhmonga:
test this please

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -327,8 +335,10 @@ public V apply(final K key, final V previousValue) {
       store.compute(key, remappingFunction);
       if (modified.get()) {
         removeObserver.end(RemoveOutcome.SUCCESS);
+        markAllCacheEventsFireable();
       } else {
         removeObserver.end(RemoveOutcome.NOOP);
+        markAllCacheEventsFireable();
```
ljacomet: No returns here, so call can be moved outside of the if

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -556,6 +568,7 @@ public void putAll(final Map<? extends K, ? extends V> entries) throws BulkCache
           resilienceStrategy.putAllFailure(entries, e, new BulkCacheWritingException(failures, successes));
         }
       }
+      markAllCacheEventsFailed();
```
ljacomet: This will never be applied if line 554 is executed. Consider moving it to a `finally` block.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -690,6 +704,7 @@ public void removeAll(final Set<? extends K> keys) throws BulkCacheWritingExcept
           resilienceStrategy.removeAllFailure(keys, e, new BulkCacheWritingException(failures, successes));
         }
       }
+      markAllCacheEventsFailed();
```
ljacomet: This will never be applied if line 692 is executed. Consider moving it to a `finally` block.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -841,6 +860,7 @@ public V apply(final K k, V inCache) {
       store.compute(key, remappingFunction, REPLACE_FALSE);
       if (removed.get()) {
         conditionalRemoveObserver.end(ConditionalRemoveOutcome.SUCCESS);
+        markAllCacheEventsFireable();
```
ljacomet: The more I think about the pattern of marking events fireable only on certain path, the more I fear consequences. I am afraid you are _silencing_ potential expiry and / or eviction events.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1355,6 +1392,20 @@ public void removeAll() {
     }
   }
   
+  private void markAllCacheEventsFireable() {
+    while (!EventThreadLocal.get().isEmpty()) {
```
ljacomet: I do not like the use of static here. Everything could be hidden behind the `EventNotificationService` api.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -59,13 +60,16 @@
   private final StoreListener<K, V> storeListener = new StoreListener<K, V>();
   private final Store<K, V> store;
   private final TimeSource timeSource;
+  private final ConcurrentHashMap<K, ConcurrentLinkedQueue<CacheEventWrapper>> keyBasedEventQueueMap = new ConcurrentHashMap<K, ConcurrentLinkedQueue<CacheEventWrapper>>();
```
ljacomet: Not an option to have a map with one queue per key. You need to have something that spreads key over different queues indeed but no a 1-1 mapping.

core/src/main/java/org/ehcache/events/EventThreadLocal.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * @author rism
+ */
+public class EventThreadLocal {
+  private static final ThreadLocal<ConcurrentLinkedQueue<CacheEventWrapper>> threadLocal = new ThreadLocal<ConcurrentLinkedQueue<CacheEventWrapper>>() {
+    @Override
+    protected ConcurrentLinkedQueue<CacheEventWrapper> initialValue() {
+      return new ConcurrentLinkedQueue<CacheEventWrapper>();
```
chrisdennis: Why do you need a concurrent linked queue?

core/src/main/java/org/ehcache/events/EventThreadLocal.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.ehcache.event.EventType;
+
+import java.util.HashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * @author rism
+ */
+public class EventThreadLocal {
```
chrisdennis: We should probably have a way of clearing this thread local when a thread "leaves" the cache.

core/src/test/java/org/ehcache/EhcacheEventTest.java
```diff
@@ -82,12 +82,6 @@ public void setUp() throws Exception {
     cache.init();
   }
   
-  @After
-  public void tearDown() {
-    // Make sure no more events have been sent
-    verify(eventNotifier, new NoMoreInteractions()).onEvent(any(CacheEvent.class));
```
rishabhmonga: Removing this check as this essentially checks for any interaction with `eventNotifier` which is not true anymore as `eventNotifier` might be invoked for `fireEvents()` method

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
rishabhmonga: This map is populated only in case of `ORDERED` event firing and is cleaned up as and when events are fired and the event queues, containing events per key, are emptied. But this is still essentially a 1-1 mapping. Is this acceptable?

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -252,4 +282,28 @@ public void setSource(Cache<K, V> source) {
...
+      @Override
+      protected List<CacheEventWrapper> initialValue() {
+        return new ArrayList<CacheEventWrapper>();
+      }
+    };
+
+    public List<CacheEventWrapper> get() {
+      HashMap<Object, CacheEventWrapper> cacheEventMap = new HashMap<Object, CacheEventWrapper>();
+      for (CacheEventWrapper cacheEventWrapper : threadLocal.get()) {
+        if((cacheEventWrapper.cacheEvent.getType() == EventType.CREATED)
```
ljacomet: What are you trying to achieve here?
Because you are creating a map, iterating over elements, all for removing duplicates ... but at each call to `get()`!!

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -125,43 +133,46 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+          interrupted = true;
+        } catch (ExecutionException ee) {
+          LOGGER.error("Cache Event Listener failed for event type {} due to ", eventWrapper.cacheEvent.getType(), ee);
+        } 
+      } else {
+        eventDispatcher.dispatch(eventWrapper, wrapper.getListener());
       }
+      
+    }
+    if (eventWrapperQueue != null && !eventWrapperQueue.isEmpty() && eventWrapperQueue.peek() == eventWrapper.cacheEvent) {
```
ljacomet: How do you respect per key ordering across threads if you do not even check this var before now?

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,57 +185,74 @@ public boolean hasListeners() {
...
+  
+  private boolean addOrderedListenerToSet(CacheEventListener<? super K, ? super V> listener, EventOrdering ordering) {
+    eventDispatcher = eventDispatchers.get(1);
+    return ordering != EventOrdering.ORDERED || orderedListenersSet.add(listener);
+  }
+  
+  private boolean removeOrderedListenerFromSet(CacheEventListener<? super K, ? super V> listener) {
+    orderedListenersSet.remove(listener);
+    if (orderedListenersSet.isEmpty()) {
+      eventDispatcher = eventDispatchers.get(0);
```
ljacomet: Given the switch done here, the field needs to be volatile

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
rishabhmonga: This is here to cater to the scenario where a key ends up having more than one `CacheEvent`(`CREATE`, `UPDATE`, `REMOVE`). With each new event added to the thread local we make sure that if it is a `CacheEvent`, the key for that event does not have any other `CacheEvent` prior to it. If it does the older `CacheEvent` is marked `Failed`.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
rishabhmonga: I think in our previous discussion, we both missed this point. If a 'per key based order' needs to be maintained, we will require some worker thread waiting on the head of the queue for the event to become fireable as events in this queue might come from multiple threads and not all of them may still be in a fireable condition. This logic was present/working in the previous commit.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,48 +147,92 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+    ConcurrentLinkedQueue<CacheEventWrapper> eventWrapperQueue = getEventQueue((K)cacheEventWrapper.cacheEvent.getKey());
+    orderedEventWait(cacheEventWrapper, eventWrapperQueue);
+    if (!orderedListenersSet.isEmpty()) {
+      while (!eventWrapperQueue.isEmpty() && eventWrapperQueue.peek().isFireable()) {
+        CacheEventWrapper event = eventWrapperQueue.peek();
+        actualDispatch(event);
+        cacheEventWrapper.markFired();
+        if (event != eventThreadLocal.get().get(0)) {
+          event.lockEvent();
+          event.firedEventCondition.signal();
```
rishabhmonga: This logic ensures that only the thread that created the event at the front of the `eventQueue` processes `FIREABLE` events, while the other threads wait for their events to get `FIRED`. Thus no matter what order the events become `FIREABLE` (i.e. threads reach the dispatch logic) the event firing order will be maintained. Understandably, this also means that the calling thread that created the event at the front of the `eventQueue` will be the last one to return after event dispatch.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,60 +243,89 @@ public boolean hasListeners() {
...
-    @Override
-    public boolean equals(Object other) {
-      if (!(other instanceof EventListenerWrapper)) {
-        return false;
+    fireEvents();
+  }
+  
+  private ConcurrentLinkedQueue<CacheEventWrapper> getEventQueue (K key) {
+    long mapKeyForQueue = key.hashCode() % NUMBER_OF_EVENT_QUEUES;
+    keyBasedEventQueueMap.putIfAbsent(mapKeyForQueue, new ConcurrentLinkedQueue<CacheEventWrapper>());
```
ljacomet: Initialise in constructor

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,60 +243,89 @@ public boolean hasListeners() {
...
       }
-      EventListenerWrapper l2 = (EventListenerWrapper)other;
-      return listener.equals(l2.listener);
+    }
+    if (getEventQueue(key).isEmpty()) {
+      removeQueue(key);
     }
   }
 
+  private void removeQueue (K key) {
```
ljacomet: That is not needed now that you have a limited number of queues. Also the current impl is not thread safe.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,60 +243,89 @@ public boolean hasListeners() {
...
+  }
+  
+  private ConcurrentLinkedQueue<CacheEventWrapper> getEventQueue (K key) {
+    long mapKeyForQueue = key.hashCode() % NUMBER_OF_EVENT_QUEUES;
+    keyBasedEventQueueMap.putIfAbsent(mapKeyForQueue, new ConcurrentLinkedQueue<CacheEventWrapper>());
+    return keyBasedEventQueueMap.get(mapKeyForQueue);
+  }
+  
+  private void removeMapEntry(K key) {
+    for (CacheEventWrapper cacheEventWrapper : getEventQueue(key)) {
```
ljacomet: get the queue once and store in local var

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,60 +243,89 @@ public boolean hasListeners() {
...
+
+  @Override
+  public void fireEvents() {
+    while (!eventThreadLocal.get().isEmpty()) {
+      CacheEventWrapper cacheEventWrapper = eventThreadLocal.get().get(0); 
+      if (!cacheEventWrapper.hasFailed()) {
+        dispatchEvent(cacheEventWrapper);
+      }
+      eventThreadLocal.get().remove(0);
+      removeMapEntry((K)cacheEventWrapper.cacheEvent.getKey());
```
ljacomet: I do not understand why you iterate over the whole event queue each time you process a single element.
According to the logic, you only need to remove events that are marked as failed. So you should add a single removal to an `else` on the `if` above.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,60 +243,89 @@ public boolean hasListeners() {
...
+  private final Set<EventListenerWrapper> aSyncListenersSet = new CopyOnWriteArraySet<EventListenerWrapper>();
+
+  private boolean addOrderedListenerToSet(CacheEventListener<? super K, ? super V> listener, EventOrdering ordering) {
+    eventDispatcher = eventDispatchers.get(1);
+    return ordering != EventOrdering.ORDERED || orderedListenersSet.add(listener);
+  }
+  
+  private boolean removeOrderedListenerFromSet(CacheEventListener<? super K, ? super V> listener) {
+    orderedListenersSet.remove(listener);
+    if (orderedListenersSet.isEmpty()) {
```
ljacomet: Given this is the condition for interacting with the ordered queues, what happens when you have elements in the queues and this becomes true?

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -174,60 +243,89 @@ public boolean hasListeners() {
...
+      eventDispatcher = eventDispatchers.get(0);
     }
-    
-    @SuppressWarnings("unchecked")
-    <K, V> CacheEventListener<K, V> getListener() {
-      return (CacheEventListener<K, V>) listener;
+    return true;
```
ljacomet: Fishy to return a constant here. Either you need to know something, or you can just return void from the method.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -245,6 +248,7 @@ public V apply(final K key, final V previousValue) {
       ValueHolder<V> computed = store.compute(key, remappingFunction);
       if (computed != null) {
         putObserver.end(PutOutcome.ADDED);
+        eventNotificationService.fireEvents();
```
ljacomet: Why doing this only in one branch? What happens otherwise if expiry or evictions were triggered by the logic?

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -263,6 +267,7 @@ public V apply(final K key, final V previousValue) {
         }
       } finally {
         putObserver.end(PutOutcome.FAILURE);
+        eventNotificationService.fireSuccessfulEvents();
```
ljacomet: This `finally` is inside the `catch` block. So there are ways to exit this method without cleaning / firing events. This feels very brittle to me. We should _ALWAYS_ handle events one way or another.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,48 +147,92 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+      ConcurrentLinkedQueue<CacheEventWrapper> eventWrapperQueue = getEventQueue(event.getKey());
+      eventWrapperQueue.add(cacheEventWrapper);
+    }
+  }
+  
+  private void dispatchEvent(CacheEventWrapper cacheEventWrapper) {
+    cacheEventWrapper.markFireable();
+    ConcurrentLinkedQueue<CacheEventWrapper> eventWrapperQueue = getEventQueue((K)cacheEventWrapper.cacheEvent.getKey());
+    orderedEventWait(cacheEventWrapper, eventWrapperQueue);
+    if (!orderedListenersSet.isEmpty()) {
```
ljacomet: Already hinted at this elsewhere, but what happens when that value changes in the middle of processing? when it becomes true? when it becomes false?

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -82,9 +99,20 @@ public CacheEventNotificationServiceImpl(ExecutorService orderedDelivery, Execut
   public void registerCacheEventListener(CacheEventListener<? super K, ? super V> listener,
                                   EventOrdering ordering, EventFiring firing, EnumSet<EventType> forEventTypes) {
     boolean doRegister = forEventTypes.contains(EventType.EVICTED) || forEventTypes.contains(EventType.EXPIRED);
-    if (!registeredListeners.add(new EventListenerWrapper(listener, firing, ordering, forEventTypes))) {
+    EventListenerWrapper wrapper = new EventListenerWrapper(listener, firing, ordering, forEventTypes);
+    if(wrapper.config.orderingMode() == EventOrdering.ORDERED) {
+      orderedListenerCount.incrementAndGet();
+    }
+    
+    if(aSyncListenersSet.contains(wrapper) || syncListenersSet.contains(wrapper)) {
```
ljacomet: No need to do that here, just observe the result of calling add below and throw when `false`

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,112 +174,162 @@ public void setStoreListenerSource(Cache<K, V> source) {
   }
 
   @Override
-  public void onEvent(final CacheEvent<K, V> event) {
+  public void onEvent(CacheEvent<K, V> event) {
```
ljacomet: We need to find a way to bypass completely the creation of the thread locals when there are no registered listeners.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
rishabhmonga: Then it would not be able to check for the scenario where we try to add the same listener with  `ASYNCHRONOUS` configuration when its already present in the `SYNCHRONOUS` set.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
rishabhmonga: would it be acceptable to directly mark the event `FIRED` if no listeners are found at this point?

rishabhmonga: And we don't call `dispatchEvent()` for events that are `FIRED`

rishabhmonga:
test this please

ljacomet:
test this please

rishabhmonga:
test this please

rishabhmonga:
test this please

rishabhmonga:
test this please

core/src/main/java/org/ehcache/events/CacheEventWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * /**
+   * This marks events as fired atomically
+   *
+   * @param signal whether {@code firedEventCondition} needs to be signalled
+   * 
+   * <p> May cause spurious signals </p>
+   */
+  void markFiredAndSignalCondition(Boolean signal) {
```
rishabhmonga: mark event as `FIRED` but only signal if its not at the front of the threadLocal event list

impl/hs_err_pid3433.log
```diff
@@ -0,0 +1,534 @@
+#
```
ljacomet: This log file needs to be removed from the PR

xml/src/test/java/org/ehcache/config/xml/XmlConfigurationTest.java
```diff
@@ -574,6 +577,26 @@ public void testDefaulSerializerXmlsSerializersValueHasWhitespaces() throws Exce
...
+  @Test
+  public void testCacheEventNotificationServiceConfigurationTest() throws Exception {
+    final URL resource = XmlConfigurationTest.class.getResource("/configs/ehcache-cacheEventNotification.xml");
+    XmlConfiguration xmlConfig = new XmlConfiguration(resource);
+    assertThat(xmlConfig.getCacheConfigurations().size(), is(2));
+
+    Collection<?> configuration = xmlConfig.getCacheConfigurations().get("bar").getServiceConfigurations();
+    checkEventNotificationConfigurationExists(configuration);
+    Collection<?> templateConfiguration = xmlConfig.getCacheConfigurations().get("foo").getServiceConfigurations();
+    checkEventNotificationConfigurationExists(templateConfiguration);
```
ljacomet: You can replace that whole logic / checks up to the `assertThat` by using `org.ehcache.spi.ServiceLocator#findSingletonAmongst`

xml/src/test/java/org/ehcache/config/xml/XmlConfigurationTest.java
```diff
@@ -127,7 +129,7 @@ public void testPrioritizerCache() throws Exception {
     assertThat(xmlConfig.getCacheConfigurations().get("foo").getEvictionPrioritizer(), sameInstance((EvictionPrioritizer)Eviction.Prioritizer.LFU));
 
     CacheConfigurationBuilder<Object, Object> example = xmlConfig.newCacheConfigurationBuilderFromTemplate("example");
-    assertThat(example.buildConfig(Object.class, Object.class).getEvictionPrioritizer(), sameInstance((EvictionPrioritizer) Eviction.Prioritizer.LFU));
+    assertThat(example.buildConfig(Object.class, Object.class).getEvictionPrioritizer(), sameInstance((EvictionPrioritizer)Eviction.Prioritizer.LFU));
```
ljacomet: Please get rid of this formatting change, same for the next two changes

rishabhmonga:
test this please

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -82,12 +108,28 @@ public CacheEventNotificationServiceImpl(ExecutorService orderedDelivery, Execut
...
+
+    if (firing == EventFiring.ASYNCHRONOUS) {
+      aSyncListenersSet.add(wrapper);
+    } else {
+      syncListenersSet.add(wrapper);
+    }
     if (doRegister) {
       store.enableStoreEventNotifications(storeListener);
     }
+    if (eventThreadLocal instanceof NoOpEventThreadLocalImpl) {
```
ljacomet: You need to invert the if and the synchronized calls, otherwise changes could have happened before you reach the exclusive block. Or you need to perform the check inside again.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -99,21 +141,50 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
+          removed = true;
+          if(listenerWrapper.config.orderingMode() == EventOrdering.ORDERED) {
+            orderedListenerCount.decrementAndGet();
+          }
+          break;
+        }
+      }
     }
     if (!hasListeners()) {
+      synchronized (this) {
```
ljacomet: Same remark as for registration

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -99,21 +141,50 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
 
   // TODO this should be really the shutdown method for the service
   @Override
   public void releaseAllListeners() {
-    for (EventListenerWrapper wrapper: registeredListeners) {
-      registeredListeners.remove(wrapper);
-    }
+    orderedListenerCount.set(0);
+    aSyncListenersSet.clear();
+    syncListenersSet.clear();
```
ljacomet: This should take care of `eventThreadLocal` also.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,44 +193,81 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
     final EventType type = event.getType();
     LOGGER.trace("Cache Event notified for event type {}", type);
-    Map<EventListenerWrapper, Future<?>> notificationResults = 
-        new HashMap<CacheEventNotificationServiceImpl.EventListenerWrapper, Future<?>>(registeredListeners.size());
-    
-    for (final EventListenerWrapper wrapper: registeredListeners) {
-      if (!wrapper.config.fireOn().contains(type)) {
-        continue;
+    CacheEventWrapper<K, V> cacheEventWrapper = new CacheEventWrapper<K, V>(event);
+    if (!(syncListenersSet.isEmpty() && aSyncListenersSet.isEmpty())) {
```
ljacomet: Replace with `if (hasListeners()) {`

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,44 +193,81 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+      if(eventThreadLocal.get().isEmpty()) {
+        eventThreadLocal.setOrdered(orderedListenerCount.get() != 0);
+      }
+      eventThreadLocal.get().add(cacheEventWrapper);
```
ljacomet: This will throw if the listener is removed concurrently and you replace the real impl with the no-op one

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,44 +193,81 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+      markFailedCacheEventsInThreadLocal();
```
ljacomet: This should move inside the `eventThreadLocal` and be a post processing of an `add` operation

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,44 +193,81 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
     LOGGER.trace("Cache Event notified for event type {}", type);
-    Map<EventListenerWrapper, Future<?>> notificationResults = 
-        new HashMap<CacheEventNotificationServiceImpl.EventListenerWrapper, Future<?>>(registeredListeners.size());
-    
-    for (final EventListenerWrapper wrapper: registeredListeners) {
-      if (!wrapper.config.fireOn().contains(type)) {
-        continue;
+    CacheEventWrapper<K, V> cacheEventWrapper = new CacheEventWrapper<K, V>(event);
+    if (!(syncListenersSet.isEmpty() && aSyncListenersSet.isEmpty())) {
+      if(eventThreadLocal.get().isEmpty()) {
```
ljacomet: I would abstract the whole logic of setting the ordered flag inside the impl.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -170,64 +278,63 @@ public void run() {
...
+    });
+    return configurationChangeListenerList;
+  }
 
-    @Override
-    public int hashCode() {
-      return listener.hashCode();
+  @Override
+  public void fireAllEvents() {
+    if (eventThreadLocal != null) {
```
ljacomet: Not needed given the new way of interacting with `eventThreadLocal`

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -170,64 +278,63 @@ public void run() {
...
-    
-    @Override
-    public boolean equals(Object other) {
-      if (!(other instanceof EventListenerWrapper)) {
-        return false;
+  }
+
+  @Override
+  public void processAndFireRemainingEvents() {
+    if (eventThreadLocal != null) {
```
ljacomet: Same as for `fireAllEvents`

ljacomet:
@rishabhmonga I added quite some comments that are again linked to having proper abstractions between the different objects interacting.
Could you do one more pass using the remarks I made and extending them to parts I did not go over?

rishabhmonga:
sure, already I've taken care of a few of them

rishabhmonga:
test this please

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -99,21 +139,51 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
+        }
+      }
+    }
+    synchronized (this) {
+      if (!hasListeners()) {
+        if (eventThreadLocal instanceof EventThreadLocalImpl) {
+          eventThreadLocal = new NoOpEventThreadLocalImpl();
+        }
+      }
       store.disableStoreEventNotifications();
```
ljacomet: This is now outside of the `hasListeners()` check which means the first deregister will cause store notifications to be disable. Not what we want.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -99,21 +139,51 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
   // TODO this should be really the shutdown method for the service
   @Override
   public void releaseAllListeners() {
-    for (EventListenerWrapper wrapper: registeredListeners) {
-      registeredListeners.remove(wrapper);
-    }
+    orderedListenerCount.set(0);
+    aSyncListenersSet.clear();
+    syncListenersSet.clear();
+    eventThreadLocal.cleanUp();
```
ljacomet: Not what I meant with the comment on the line above. After this call, there are no listeners left so ... ??

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -82,12 +107,28 @@ public CacheEventNotificationServiceImpl(ExecutorService orderedDelivery, Execut
   public void registerCacheEventListener(CacheEventListener<? super K, ? super V> listener,
```
rishabhmonga: Objective – register and classify the listeners so as to maintain sync/async sets and enable threadlocal

What it internally does:
•	Listener will be eligible for registration iff it listens to atleast expirations or evictions
•	If listener is ordered, increament atomicInteger count
•	Pigeonhole listeners as sync/async and add it to the respective sets. Exception if already present
•	Finally, enable EventThreadLocal if not already enabled.

Why it should work:
Registering of a listener makes sure that EventThreadLocal gets enabled and adds to the orderedListener count so the eventDispatcher can be configured accordingly. Former is done under lock to prevent conflicting assignments. Maintaining sync/async listener sets helps in dispatch later.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -99,20 +140,54 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
    */
   @Override
   public void deregisterCacheEventListener(CacheEventListener<? super K, ? super V> listener) {
```
rishabhmonga: Objective – deregister listener, decrement ordered count (if necessary). Remove listener from respective sync/async set and disable threadlocal if no listener left

What it internally does:
•	Create a listener wrapper that has all necessary information of the listener.
•	Check if listener present in either of the sync/async set and remove, also reduce orderedListener count 
•	Finally disable EventThreadLocal to a noop if no listeners are left.

Why it should work:
A `removed` boolean is maintained that helps to track if the listener did get removed from the concerned listener set. Along with the removal from set, if the listener was ordered, a decrement in the ordered listener count is also done. Also if no listeners are left, the threadLocal is switched to a no-op under lock.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -99,20 +140,54 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
+      }
+    }
+    if(!removed) {
+      throw new IllegalStateException("Unknown cache event listener: " + listener);
     }
   }
 
   // TODO this should be really the shutdown method for the service
   @Override
   public void releaseAllListeners() {
```
rishabhmonga: Objective – clear everything maintained about the listeners

What it internally does:
•	Reset orderedListeners count
•	Clear sync/async sets
•	Disable EventThreadLocal to a noop
Why it should work:
A reassignment of the threadLocal to a noop implementation ensures that an unnecessary threadLocal does not exist when there are no listeners are available.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
rishabhmonga: Objective – Process event so its in-line for dispatch depending on ordered/unordered dispatch.
What it internally does:
•	Add event to threadLocal.eventList and to the correct event queue if events are ordered.

Why it should work:
The premise here is that each thread should fire the events it creates. This is done by maintaining a list of events created by the thread in a list in its threadLocal. At the time of dispatch, if ordered, thread checks its list of events, visits each concerned queue to dispatch the events. Else, in case of unordered, a dispatch task needs to be created directly.
Thus in this method, in preparation for the dispatch, the event is added to the threadLocal.eventList and the eventQueues in case of ordered dispatch.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -170,64 +264,61 @@ public void run() {
...
     }
-    
-    @Override
-    public boolean equals(Object other) {
-      if (!(other instanceof EventListenerWrapper)) {
-        return false;
+  }
+
+  @Override
+  public void processAndFireRemainingEvents() {
```
rishabhmonga: Objective – To be used when the operation failed and only the associated eviction and expiration events get fired.

What it internally does:
•	Mark all cache events (such as create, update, remove) as failed
•	Call fireAllEvents()
•	Finally cleanUp the threadLocal

Why it should work:
This is called in case the operation has failed. The the threadLocal eventList contains all the events created during the operation so all operations that are not marked as failed, will be eligible for dispatch using the fireAllEvents() itself.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -170,64 +264,61 @@ public void run() {
...
+      }
+    });
+    return configurationChangeListenerList;
+  }
 
-    @Override
-    public int hashCode() {
-      return listener.hashCode();
+  @Override
+  public void fireAllEvents() {
```
rishabhmonga: Objective – Indicates that the events created by the concerned thread are eligible for firing. And should now be fired

What it internally does:
•	Goes through the threadLocal’s eventList, beginning with the first and call the dispatchEvent() method on the event
•	And then, remove this (first element) from the eventList
•	Continues until the eventList is empty i.e. all events created by the thread have been sent to the dispatch method

Why it should work:
Continuing with the premise that each thread fires the events it created throughout the operation. Each thread goes through the eventList it possesses, processing events and then removing them from the list thus going over every event it created. Finally clearing up the threadLocal.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,46 +197,65 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
-        } catch (InterruptedException e) {
-          interrupted = true;
-        } finally {
-          if (interrupted) {
-            Thread.currentThread().interrupt();
-          }
+    }
+  }
+
+  private void dispatchEvent(CacheEventWrapper cacheEventWrapper, EventThreadLocal currentEventThreadLocal) {
```
rishabhmonga: Objective – receives the event that needs to be dispatched, processes them based on ordering and trigger dispatch

What it internally does:
•	Mark the event `FIREABLE`
•	In case of ordered event dispatch
o	Get the concerned queue that carries the event
o	Initiate wait if the event is not at the front of the queue
o	If event is at the front of the queue then for all events that are `FIREABLE`, 
•	Get the event from the front of queue
•	Send it for `actualDispatch()`
•	If event doesn’t belongs to the thread, mark it FIRED and signal FIRED-CONDITION
•	Else if it does belong to the thread, just mark the event FIRED
•	Remove the event from the event queue
•	Repeat until eventQueue empty or event at the front of the queue is not `FIREABLE` yet.
•	Send event for `actualDispatch()`

Why it should work:
The thread that created the event in the front of the queue is responsible for firing it. Others wait for their event to be fired (waiting on the FIRED_CONDITION), no matter sync or async. This helps to ensure that ordering is maintained and events that are `FIREABLE` can be processed as soon as the front of the concerned queue is `FIREABLE`. The approach of letting the thread with the event on the front of the queue fire events is preferred because having the approach of firing the event at the front and letting the next event know it’s the front now can cause conflicts and a requirement for additional conditions.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,46 +197,65 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+        } else {
+          event.markFiredAndSignalCondition(false);
         }
       }
+    } else {
+      actualDispatch(cacheEventWrapper, unOrderedEventDispatcher);
+    }
+  }
+
+  private void orderedEventWait(CacheEventWrapper cacheEventWrapper, Queue<CacheEventWrapper> eventWrapperQueue, EventThreadLocal currentEventThreadLocal) {
```
rishabhmonga: Objective – make threads that don’t have event at the front of the queue (but want to fire their event) wait until they are notified hem that their event if FIRED

What it internally does:
•	Check the dispatch is ordered
•	Confirm event is not at the front of the queue
•	Wait for the signal on FIRED_CONDITION, that should be provided by the thread with event at the front

Why it should work:
Having other threads to wait while a single thread fires their events allows for a simple approach to carry out dispatch for all events that are `FIREABLE`. Also since the thread that fires all events signals for all the events that it fires but are not created by that thread makes sure that no thread is kept in a perpetual wait.

core/src/main/java/org/ehcache/events/CacheEventNotificationServiceImpl.java
```diff
@@ -122,46 +197,65 @@ public void setStoreListenerSource(Cache<K, V> source) {
...
+    if (currentEventThreadLocal.isOrdered()) {
+      if (cacheEventWrapper != eventWrapperQueue.peek()) {
+        cacheEventWrapper.waitTillFired();
+      }
+    }
+  }
+
+  private void actualDispatch(CacheEventWrapper event, EventDispatcher<K, V> eventDispatcher) {
```
rishabhmonga: Objective: send event to the dispatcher for all sync/async listeners and facilitate synchronous dispatch

What it internally does:
•	If the event is not marked `FAILED`. It is sent to the eventDispatcher along with the set of syncListenerSet and then again with asyncListenerSet. 
•	The Future received form the syncListenerSet dispatcher is used to help with synchronous event firing

Why it should work:
In absence of sync listeners, threads would just come in, send their events to the eventDispatcher and continues without waiting for the event firing to be completed.
If synchronous listeners are present, threads that try to fire events will need to wait for the event firing to be completed. Future returned by the eventDispatcher aids in this regard.

impl/src/main/java/org/ehcache/internal/executor/DefaultThreadPoolsService.java
```diff
@@ -67,7 +67,7 @@ public void start(final ServiceProvider serviceProvider) {
   @Override
   public void stop() {
     statisticsExecutor.shutdownNow();
-    eventsOrderedDeliveryExecutor.shutdownNow();
-    eventsUnorderedDeliveryExecutor.shutdownNow();
+    eventsOrderedDeliveryExecutor.shutdown();
+    eventsUnorderedDeliveryExecutor.shutdown();
```
rishabhmonga: In case of Asynchronous Listeners, this might be required so that a call to `cacheManager.close()` does not cause on going events to stop.

ljacomet:
Closing as more work needs to happen to finalise - see #165

Issue #525, opened 2015-08-14
Persistence directory path String Trim


JSAhuja:
If persistence path is defined like as below, throws an exception
```
org.ehcache.exceptions.StateTransitionException: Directory couldn't be created
```
```
	<ehcache:service>
		<ehcache:persistence>
			<ehcache:directory>
				target/customSerializerTest/EhcacheDiskStore
			</ehcache:directory>
		</ehcache:persistence>
	</ehcache:service>
```

Directory path string, trim required.

PR #535, opened 2015-08-26
Issue #525 Switch to using an attribute for disk path

chrisdennis:
@cschanck wisely recommended that the simplest solution to the disk path trimming problem would be to move to using an attribute rather than an element body to hold the disk path.  This is my attempt at doing exactly that.

### DIFF
--- a/107/src/test/resources/ehcache-107-stats.xml
+++ b/107/src/test/resources/ehcache-107-stats.xml
@@ -14,9 +14,7 @@
   </service>
 
   <service>
-    <persistence>
-      <directory>build/ehcache-107-stats</directory>
-    </persistence>
+    <persistence directory="build/ehcache-107-stats"/>
   </service>
   
   <cache-template name="diskCache">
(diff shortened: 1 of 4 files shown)


AbfrmBlr:
Can we do this by using regex patterns in xs restrictions ? if its possible

ljacomet:
What exactly do you want to achieve with a regexp?

````xml
<persistence>
  <directory>
    /some/dir
  </directory>
</persistence>
````

The above would be illegal then?
I like the proposed solution as anyone with a bit of XML knowledge will realise that the content of the attribute _will be_ the path. And so if they want special chars, they can have them. Including leading and trailing whitespace.

AbfrmBlr:
Why do leading and trailing whitespaces are legal in directory path ?

ljacomet:
Why? No idea
But they are a possibility, and as such, why should we trim them? Think that the Ehcache XML config could be generated for example.
So given this, why would we _change_ what the user entered.

chrisdennis:
Most modern filesystems support pretty much any character in a filesystem path: https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

Issue #534, opened 2015-08-25
Writebehind Configuration, maxWriteDelay & minWriteDelay Validation


JSAhuja:
A validation required, to verify maxWriteDelay value >= minWriteDelay value

Sample:
````
Class<CacheLoaderWriter<?, ?>> klazz = (Class<CacheLoaderWriter<?, ?>>)  (Class) (LoaderWriter.class);
Cache<Long, String> writeBehindCache = cacheManager.createCache("writeBehindCache",
	            CacheConfigurationBuilder.newCacheConfigurationBuilder()
	                    .add(new DefaultCacheLoaderWriterConfiguration(klazz))
	                    .add(WriteBehindConfigurationBuilder.newWriteBehindConfiguration()
	                            .delay(22, 1)) // delay(int minWriteDelay, int maxWriteDelay)
	                    .buildConfig(Long.class, String.class));
````

PR #541, opened 2015-08-31
Issue #534 Enforce min write delay smaller or equal to max write delay

ljacomet:
* Enforce the contract
* Improve javadocs
* Change default min write delay to 0
  * As a consequence, build is faster as write behind related tests run faster

### DIFF
--- a/api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
+++ b/api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
@@ -73,28 +73,28 @@ public interface WriteBehindConfiguration extends ServiceConfiguration<WriteBehi
   int getWriteBatchSize();
 
   /**
-   * The number of times the write of a mapping is retried in case it fails.
+   * The number of times the write of a mapping will be retried in the case of failure.
    *
    * @return Retrieves the number of times the write of element is retried.
    */
   int getRetryAttempts();
 
   /**
-   * The number of seconds to wait before retrying an failed operation.
+   * A number of seconds to wait before retrying an failed operation.
    *
    * @return Retrieves the number of seconds to wait before retrying an failed operation.
    */
   int getRetryAttemptDelaySeconds();
 
   /**
-   * The amount of bucket/thread pairs configured for this cache's write behind.
+   * A number of bucket/thread pairs configured for this cache's write behind.
    *
    * @return Retrieves the amount of bucket/thread pairs configured for this cache's write behind
    */
   int getWriteBehindConcurrency();
 
   /**
-   * The maximum amount of operations allowed on the write behind queue.
+   * The maximum number of operations allowed on the write behind queue.
    *
    * {@code 0} indicates an unbound queue.
    *
--- a/api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
+++ b/api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
@@ -23,70 +23,80 @@ import org.ehcache.spi.service.ServiceConfiguration;
  */
 public interface WriteBehindConfiguration extends ServiceConfiguration<WriteBehindDecoratorLoaderWriterProvider> {
   /**
-   * the minimum number of seconds to wait before writing behind
+   * The minimum number of seconds to wait before writing behind.
+   *
+   * Lower than or equal to {@see getMaxWriteDelay}.
    *
    * @return Retrieves the minimum number of seconds to wait before writing behind
    */
   int getMinWriteDelay();
 
   /**
-   * the maximum number of seconds to wait before writing behind
+   * The maximum number of seconds to wait before writing behind.
+   *
+   * Greater than or equal to {@see getMinWriteDelay}
    *
    * @return Retrieves the maximum number of seconds to wait before writing behind
    */
   int getMaxWriteDelay();
 
   /**
-   * the maximum number of write operations to allow per second.
+   * The maximum number of write operations to allow per second.
+   *
+   * {@code 0} means no limit.
    *
    * @return Retrieves the maximum number of write operations to allow per second.
    */
   int getRateLimitPerSecond();
 
   /**
-   * whether write operations should be batched
+   * Whether write operations should be batched.
    *
    * @return Retrieves whether write operations should be batched
    */
   boolean isWriteBatching();
 
   /**
-   * write coalescing behavior
+   * Whether write operations can be coalesced.
    *
    * @return Retrieves the write coalescing behavior is enabled or not
    */
   boolean isWriteCoalescing();
 
   /**
-   * the size of the batch operation.
+   * The recommended size of a batch of operations.
+   *
+   * Real batch size will be influenced by arrival frequency of operations and max write delay.
    *
    * @return Retrieves the size of the batch operation.
    */
   int getWriteBatchSize();
 
   /**
-   * the number of times the write of element is retried.
+   * The number of times the write of a mapping is retried in case it fails.
    *
    * @return Retrieves the number of times the write of element is retried.
    */
   int getRetryAttempts();
 
   /**
-   * the number of seconds to wait before retrying an failed operation.
+   * The number of seconds to wait before retrying an failed operation.
    *
    * @return Retrieves the number of seconds to wait before retrying an failed operation.
    */
   int getRetryAttemptDelaySeconds();
 
   /**
-   * the amount of bucket/thread pairs configured for this cache's write behind
+   * The amount of bucket/thread pairs configured for this cache's write behind.
    *
    * @return Retrieves the amount of bucket/thread pairs configured for this cache's write behind
    */
   int getWriteBehindConcurrency();
 
   /**
-   * the maximum amount of operations allowed on the write behind queue
+   * The maximum amount of operations allowed on the write behind queue.
+   *
+   * {@code 0} indicates an unbound queue.
    *
    * @return Retrieves the maximum amount of operations allowed on the write behind queue
    */
--- a/core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
+++ b/core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
@@ -27,7 +27,7 @@ import static java.lang.String.format;
  */
 public class DefaultWriteBehindConfiguration implements WriteBehindConfiguration {
 
-  private int minWriteDelay = 1;
+  private int minWriteDelay = 0;
   private int maxWriteDelay = 1;
   private int rateLimitPerSecond = 0;
   private boolean writeCoalescing = false;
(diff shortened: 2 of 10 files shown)


api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
    *
    * @return Retrieves the maximum number of seconds to wait before writing behind
    */
   int getMaxWriteDelay();
 
   /**
-   * the maximum number of write operations to allow per second.
+   * The maximum number of write operations to allow per second.
+   *
+   * {@code 0} means no limit.
```
chrisdennis: Bleagh... nasty Ehcache 2.x wart re-emerging.

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
    *
    * @return Retrieves the write coalescing behavior is enabled or not
    */
   boolean isWriteCoalescing();
 
   /**
-   * the size of the batch operation.
+   * The recommended size of a batch of operations.
+   *
+   * Real batch size will be influenced by arrival frequency of operations and max write delay.
```
chrisdennis: and min write delay?

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
    *
    * @return Retrieves the size of the batch operation.
    */
   int getWriteBatchSize();
 
   /**
-   * the number of times the write of element is retried.
+   * The number of times the write of a mapping is retried in case it fails.
```
chrisdennis: I don't think this sentence means what you want it to:
> The number of times the write of a mapping is retried in case it fails.
versus
> The number of times the write of a mapping will be retried in the case of failure.

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
    *
    * @return Retrieves the number of times the write of element is retried.
    */
   int getRetryAttempts();
 
   /**
-   * the number of seconds to wait before retrying an failed operation.
+   * The number of seconds to wait before retrying an failed operation.
```
chrisdennis: a

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
    *
    * @return Retrieves the number of seconds to wait before retrying an failed operation.
    */
   int getRetryAttemptDelaySeconds();
 
   /**
-   * the amount of bucket/thread pairs configured for this cache's write behind
+   * The amount of bucket/thread pairs configured for this cache's write behind.
```
chrisdennis: number?

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
    *
    * @return Retrieves the amount of bucket/thread pairs configured for this cache's write behind
    */
   int getWriteBehindConcurrency();
 
   /**
-   * the maximum amount of operations allowed on the write behind queue
+   * The maximum amount of operations allowed on the write behind queue.
```
chrisdennis: number

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
```diff
@@ -23,70 +23,80 @@
...
+   *
+   * {@code 0} indicates an unbound queue.
```
chrisdennis: More warts

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -25,7 +27,7 @@
  */
 public class DefaultWriteBehindConfiguration implements WriteBehindConfiguration {
 
-  private int minWriteDelay = 1;
+  private int minWriteDelay = 0;
   private int maxWriteDelay = 1;
```
chrisdennis: I'm not sure it makes sense that these values have defaults... getting déjà vu here.

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -89,17 +91,20 @@ public int getWriteBehindMaxQueueSize() {
     return writeBehindMaxQueueSize;
   }
 
-  public void setMinWriteDelay(int minWriteDelay) {
-    if(minWriteDelay < 1) {
-      throw new IllegalArgumentException("Minimum write delay seconds cannot be less then 1.");
+  public void setWriteDelays(Integer minWriteDelay, Integer maxWriteDelay) {
```
chrisdennis: Eliminating the default would get rid of the nasty null contract here too.

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -89,17 +91,20 @@ public int getWriteBehindMaxQueueSize() {
...
+    maxWriteDelay = maxWriteDelay != null ? maxWriteDelay : 1;
+    if (minWriteDelay < 0) {
+      throw new IllegalArgumentException("Minimum write delay seconds cannot be less then 0.");
     }
-    this.minWriteDelay = minWriteDelay;
-  }
-
-  public void setMaxWriteDelay(int maxWriteDelay) {
-    if(maxWriteDelay < 1) {
+    if (maxWriteDelay < 1) {
```
chrisdennis: Why is the unit seconds here?

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -89,17 +91,20 @@ public int getWriteBehindMaxQueueSize() {
...
       throw new IllegalArgumentException("Maximum write delay seconds cannot be less then 1.");
     }
+    if (maxWriteDelay < minWriteDelay) {
+      throw new IllegalArgumentException(
+          format("Maximum write delay (%d seconds) must be equal or greater than minimum write delay (%d seconds)", maxWriteDelay, minWriteDelay));
```
chrisdennis: Why suddenly using printf style formatting here?

xml/src/main/java/org/ehcache/config/xml/ConfigurationParser.java
```diff
@@ -782,7 +782,7 @@ public int retryAttemptsDelay() {
 
     @Override
     public int minWriteDelay() {
-      return this.writebehind.getWritedelay() != null ? this.writebehind.getWritedelay().getMin().intValue() : 1 ;
```
chrisdennis: Isn't this encoding defaults in two places?  Something we really should try to avoid.

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -210,7 +210,7 @@
             <xs:complexType>
               <xs:simpleContent>
                 <xs:extension base="xs:string">
-                  <xs:attribute type="xs:positiveInteger" name="min" default="1" use="optional"/>
+                  <xs:attribute type="xs:nonNegativeInteger" name="min" default="0" use="optional"/>
```
chrisdennis: Three places!

xml/src/test/resources/configs/writebehind-cache.xml
```diff
@@ -27,9 +27,7 @@
         <ehcache:loaderwriter>
           <ehcache:class>com.pany.ehcache.integration.TestCacheLoaderWriter</ehcache:class>
         </ehcache:loaderwriter>
-        <ehcache:writebehind coalesce="false" concurrency="3" size="10">
-          <ehcache:batchsize>5</ehcache:batchsize>
-        </ehcache:writebehind>
+        <ehcache:writebehind coalesce="false" concurrency="1" size="10"/>
```
chrisdennis: Why did we change these?

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
alexsnaps: Right, it an int, let's make this Integer.MAX_VALUE (which most probably is more than large enough even if considered a "strict" limit)

api/src/main/java/org/ehcache/spi/loaderwriter/WriteBehindConfiguration.java
alexsnaps: Again, use `Integer.MAX_VALUE`, which is inline with JDK types btw...

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -89,17 +91,20 @@ public int getWriteBehindMaxQueueSize() {
...
   }
 
-  public void setMinWriteDelay(int minWriteDelay) {
-    if(minWriteDelay < 1) {
-      throw new IllegalArgumentException("Minimum write delay seconds cannot be less then 1.");
+  public void setWriteDelays(Integer minWriteDelay, Integer maxWriteDelay) {
+    minWriteDelay = minWriteDelay != null ? minWriteDelay : 0;
+    maxWriteDelay = maxWriteDelay != null ? maxWriteDelay : 1;
+    if (minWriteDelay < 0) {
+      throw new IllegalArgumentException("Minimum write delay seconds cannot be less then 0.");
```
albinsuresh: less "than" 0???

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -89,17 +91,20 @@ public int getWriteBehindMaxQueueSize() {
...
     }
-    this.minWriteDelay = minWriteDelay;
-  }
-
-  public void setMaxWriteDelay(int maxWriteDelay) {
-    if(maxWriteDelay < 1) {
+    if (maxWriteDelay < 1) {
       throw new IllegalArgumentException("Maximum write delay seconds cannot be less then 1.");
```
albinsuresh: less "than" 1???

core/src/test/java/org/ehcache/config/writebehind/DefaultWriteBehindConfigurationTest.java
```diff
@@ -0,0 +1,61 @@
...
+  @Test(expected = IllegalArgumentException.class)
+  public void testMaxWriteDelayInvalid() {
+    new DefaultWriteBehindConfiguration().setWriteDelays(null, -1);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testMaxWriteDelaysNotLowerThanMinWriteDelays() {
+    new DefaultWriteBehindConfiguration().setWriteDelays(10, 5);
+  }
+
```
albinsuresh: ExpectedException Rule for the above 3 tests, may be???

ljacomet:
Created issue #543 to address the remarks around defaults and limit values.

core/src/test/java/org/ehcache/config/writebehind/DefaultWriteBehindConfigurationTest.java
ljacomet: Not sure the one liners would benefit from additional setup.

xml/src/test/resources/configs/writebehind-cache.xml
ljacomet: Test makes no use of batching and this actually caused delay for no added value.

PR #546, opened 2015-09-03
Issue 37: (NOT A PR) XAStore open for comments

lorban:


impl/src/main/java/org/ehcache/internal/store/DefaultStoreProvider.java
```diff
@@ -40,6 +41,7 @@
 /**
  * @author Ludovic Orban
  */
+@SupplementaryService
```
chrisdennis: How will the `CacheManager` distinguish between the transactional store provider and the default one if both are available?

transactions/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+  compile project(':impl'), project(':xml')
+  compile group: 'javax.transaction', name: 'jta', version: '1.1'
+  testCompile project(path: ':core-spi-test')
+  compile group: 'org.codehaus.btm', name: 'btm', version: '2.1.4'
```
chrisdennis: Why not testCompile?

transactions/src/main/java/org/ehcache/transactions/EhcacheXAResource.java
```diff
@@ -0,0 +1,227 @@
...
+      sb.append("TMRESUME|");
+    }
+    if ((flags & XAResource.TMRESUME) == XAResource.TMRESUME) {
+      sb.append("TMRESUME|");
+    }
+    if ((flags & XAResource.TMSTARTRSCAN) == XAResource.TMSTARTRSCAN) {
+      sb.append("TMSTARTRSCAN|");
+    }
+    if ((flags & XAResource.TMENDRSCAN) == XAResource.TMENDRSCAN) {
+      sb.append("TMSTARTRSCAN|");
```
chrisdennis: typo

transactions/src/main/java/org/ehcache/transactions/EhcacheXAResource.java
```diff
@@ -0,0 +1,227 @@
...
+      throw new EhcacheXAException("End flag not supported : " + xlat(flag), XAException.XAER_PROTO);
+    }
+    if (currentXid == null) {
+      throw new EhcacheXAException("Not started on : " + xid, XAException.XAER_PROTO);
+    }
+
+    currentXid = null;
+  }
+
+  private static String xlat(int flags) {
```
chrisdennis: Man, I really want to do this with a static map built by reflection...

transactions/src/main/java/org/ehcache/transactions/SerializableXid.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public int getFormatId() {
+        return formatId;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public byte[] getBranchQualifier() {
+        return branchQualifier;
```
chrisdennis: Findbugs will throw a hissy fit over this.

chrisdennis:
The Journal should be Cache scoped right?  How do we know which journal the caller wants?

transactions/build.gradle
lorban: Because BTM (and probably some other TM's too) need some glue code to work with non-JMS / non-JDBC XAResource's.

transactions/src/main/java/org/ehcache/transactions/EhcacheXAResource.java
lorban: Thanks, I will fix it.

transactions/src/main/java/org/ehcache/transactions/EhcacheXAResource.java
lorban: I will do that during a later code cleanup phase.

transactions/src/main/java/org/ehcache/transactions/SerializableXid.java
lorban: Probably, I'll worry about findbugs during a later code cleanup stage.

lorban:
It doesn't have to, the journal is a map of XID -> transaction state. Since XIDs are guaranteed to be universally unique, a single journal can be shared across as many resources as you want.

But we could scope it at the cache level to avoid having the cache manager worry about cache concerns. As long as that won't cause perf headaches when multiple disk-persistent caches have to persist their journal.

impl/src/main/java/org/ehcache/internal/store/DefaultStoreProvider.java
lorban: If both are available, the transactional one will take precedence but its implementation makes it delegate to DefaultStoreProvider when a cache doesn't have a transactional config.

PR #551, opened 2015-09-03
Issue #540 - Added a LifeCycleService + Bugfix when recovering from maintenance mode

mathieucarbou:
@alexsnaps, @ljacomet, @lorban : pull request for #540 (service part), with the modifications we talked about.

I used a `StateChangeListener` to hook into the state transitioner because these listeners are triggered after all init / close hooks are executed, which is what we need to be aware of the initialized and closed states I think (?).

Also fixed a missing `succeeded();` call when closing maintenance mode

Added 2 tests for both the issue and the LifeCycle Service.

Issue #545, opened 2015-09-03
CacheManager Naming Support


mathieucarbou:
Find a way to recover the identification name of a cache manager for the management registry.

mathieucarbou:
not needed anymore

Issue #540, opened 2015-08-28
Provide a way for services to be aware of CacheManager and Cache lifecycle events (init and close)


mathieucarbou:
__use case:__

`ManagementRegistry` is created per cache manager (by a factory) and should start working when the cache manager is fully initialized, thus it needs a way to be aware of this.

Currently only `start()` and `stop()` methods for `Service` lifecycle are available.

This will also solve #350 since `DefaultManagementRegistry` will be able to use the schedule thread pool from the ThreadPoolService.

__proposal:__

- Sort of `LifeCycled` which could be implemented by a Service.
- On an Ehcache lifecycle call, also call services implementing this interface

__drawbacks:__

- It links the created instance of DefaultManagementRegistry (by the factory) to one cache manager.
- will have to create another class (i.e. SharedManagementRegistry) for working across several cache managers (http://ehcache.github.io/docs/user/3.0/management.html)

mathieucarbou:
After a discussion with @alexsnaps and @chrisdennis , I will make the `DefaultManagementRegistry` uses `Lifecycled` and the per-cache manager thread pool factory. Because DefaultManagementRegistry is created per-cache manager with the `DefaultManagementRegistryFactory` (META-INF/services).

Thus, I will change the tests to make sure a DefaultManagementRegistry is only used in 1 cache manager at a time.

If we need a way to interact with all management registries (of all cache managers) from a facade, we will need to think of an implementation with a sort of delegation layer.

mathieucarbou:
not needed anymore

Issue #552, opened 2015-09-03
ManagementRegistry refactoring


mathieucarbou:
- depends on #540 (listen on cache manager initialized state)
- make the DefaultManagementRegistry linked to one cache manager (This will also solve #350 since DefaultManagementRegistry will be able to use the schedule thread pool from the ThreadPoolService.)
- Introduce another class to handle the sharing of statistics across several cache managers (i.e. SharedManagementRegistry). http://ehcache.github.io/docs/user/3.0/management.html
- Might depends on #545 if the registry needs a way to discover the name of a CacheManager.
- Decouple ManagementRegistry from core module to make the management module a standalone optional module

PR #553, opened 2015-09-04
Issue #545 - Added NamingService, responsible of providing the...

mathieucarbou:
- PR for issue #545 
- Depends on Issue #540 (solved by PR #551)

The idea here is to provide a service, configurable by a user, which is responsible for naming the cache manager this service is used into.

The `NamingService` would thus be required by the management service, which need to have a cache manager name. Other services requiring a name would also be able to depend on this service. 

Naming a cache manager is optional, and just needed by some services.

Note:

This PR contains 2 commits from PR #551 and 1 commit for the NamingService (20969f2)

### DIFF
--- a/api/src/main/java/org/ehcache/spi/alias/AliasConfiguration.java
+++ b/api/src/main/java/org/ehcache/spi/alias/AliasConfiguration.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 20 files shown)


albinsuresh:
Shouldn't there be some documentation on this in the GettingStarted doc and samples, since this is a user facing API?

mathieucarbou:
I think so also. 
Actually this pull request is to introduce a more bigger refactoring of the DefaultManagementRegistry (#552) which will use this service for naming the cache managers. 
And before adding documentation and so on, I was much more thinking of preparing all pull requests then discuss with the team all the pull request + DefaultManagementRegistry refactoring.
And if all is good I will update pull requests with the reviews :-)
Does it work for you guys ? 
@ljacomet @lorban @albinsuresh  ?

mathieucarbou:
As per @lorban suggestions:

- Renaming from *Name* to *Alias*
- Added `String AliasService.getCacheAlias(Cache)`

Also:

- Added tests
- Added more javadoc to precise that the implementation could change

mathieucarbou:
update: introducing `CacheBinding` so that listeners can be aware of cache alias at cache init and close. This allows to decouple the alias service from the static classes StatisticsManager and ContextManager used to store the alias of cache and cache managers.

mathieucarbou:
test this please

mathieucarbou:
will create a pull request with several separated commits

Issue #468, opened 2015-07-21
Design and implement bulk operation statistics.


chrisdennis:
We need to decide how the statistics on bulk operations 'work'.  Of particular interest is how bulk operations can be related to their non bulk equivalents.  There is a related issue around how operations in general related to their logically equivalent operations, whether those are bulk to non-bulk breakdowns or compound to non-compound breakdowns.  The needs of JSR-107 are a particular constraining parameter set here.

ljacomet:
From multiple discussions, what needs to happen here is that bulk stats should:
* Offer operation statistics like  getAll count
* Offer logical operations count like hits, misses, puts and removes

Current idea is to implement this inline, as is done for getAll counting hits and defer adding this support to the statistics framework.

albinsuresh:
Some bulk stats tests need to be added to `org.ehcache.jsr107.StatisticsTest` class, right???

ljacomet:
Indeed, added to the PR

PR #556, opened 2015-09-07
Issue #468 bulk ops statistics

ljacomet:
* Added `OperationObserver` for getAll, putAll and removeAll
* Added counter for `getAll` miss
* Replace `AtomicLong` with `LongAdder` for bulk counters

### DIFF
--- a/core/src/main/java/org/ehcache/Ehcache.java
+++ b/core/src/main/java/org/ehcache/Ehcache.java
@@ -40,6 +40,7 @@ import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
 import org.ehcache.statistics.BulkOps;
 import org.ehcache.statistics.CacheOperationOutcomes.CacheLoadingOutcome;
 import org.ehcache.statistics.CacheOperationOutcomes.ConditionalRemoveOutcome;
+import org.ehcache.statistics.CacheOperationOutcomes.GetAllOutcome;
 import org.ehcache.statistics.CacheOperationOutcomes.GetOutcome;
 import org.ehcache.statistics.CacheOperationOutcomes.PutIfAbsentOutcome;
 import org.ehcache.statistics.CacheOperationOutcomes.PutOutcome;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -91,13 +94,16 @@
...
+  private final OperationObserver<RemoveAllOutcome> removeAllObserver = operation(RemoveAllOutcome.class).named("removeAll").of(this).tag("cache").build();
   private final OperationObserver<ConditionalRemoveOutcome> conditionalRemoveObserver = operation(ConditionalRemoveOutcome.class).named("conditionalRemove").of(this).tag("cache").build();
   private final OperationObserver<CacheLoadingOutcome> cacheLoadingObserver = operation(CacheLoadingOutcome.class).named("cacheLoading").of(this).tag("cache").build();
   private final OperationObserver<PutIfAbsentOutcome> putIfAbsentObserver = operation(PutIfAbsentOutcome.class).named("putIfAbsent").of(this).tag("cache").build();
   private final OperationObserver<ReplaceOutcome> replaceObserver = operation(ReplaceOutcome.class).named("replace").of(this).tag("cache").build();  
-  private final ConcurrentMap<BulkOps, AtomicLong> bulkMethodEntries = new ConcurrentHashMap<BulkOps, AtomicLong>();
+  private final ConcurrentMap<BulkOps, LongAdder> bulkMethodEntries = new ConcurrentHashMap<BulkOps, LongAdder>();
```
chrisdennis: EnumMap with initialized mappings to LongAdders?

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1068,15 +1103,15 @@ private void checkNonNullContent(Collection<?> collectionOfThings) {
   }
 
   private void addBulkMethodEntriesCount(BulkOps op, long count) {
-    AtomicLong current = bulkMethodEntries.get(op);
+    LongAdder current = bulkMethodEntries.get(op);
     if (current == null) {
```
chrisdennis: Once we have a pre-inited EnumMap we can do away with this null check.

core/src/main/java/org/ehcache/Ehcache.java
ljacomet: Yes!

core/src/main/java/org/ehcache/Ehcache.java
ljacomet: And others ... will change that

PR #557, opened 2015-09-07
Issue #158 JSR-107 doc update on cache config overriding using templates

albinsuresh:


ljacomet:
@alexsnaps We had a chat on this topic and you have doubts. Can you check and comment?

ljacomet:
@albinsuresh See last comment on issue #158

Issue #558, opened 2015-09-07
Expiration statistics in authoritative tiers


ljacomet:
All authoritative tiers need to have an `OperationObserver` for expirations.

ljacomet:
At first, I thought that we should not track expiration on `CachingTier` specific operations. But then I realised that a tier is either used as an `AuthorityTier` or as a CachingTier` and so from a given tier in a cache, counters will only ever be updated according to the _role_ of the tier.
So in the end, expiration counter is to be invoked in all methods that do expire entries.

PR #560, opened 2015-09-08
Issue #558 Expiration statistic in Stores

ljacomet:


ljacomet:
test this please

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1027,7 +1028,13 @@ private void checkValue(V valueObject) {
       throw new ClassCastException("Invalid value type, expected : " + valueType.getName() + " but was : " + valueObject.getClass().getName());
     }
   }
-  
+
+  private void onExpiration(K mappedKey, ValueHolder<V> mappedValue) {
+    expirationObserver.begin();
+    expirationObserver.end(StoreOperationOutcomes.ExpirationOutcome.SUCCESS);
```
chrisdennis: Not a fan of firing the observers like this, but there's probably nothing else we can do here.

chrisdennis:
Looks okay to me.

Issue #561, opened 2015-09-09
Service registration in ServiceLocator sometimes ignore the provided service


lorban:
Take the following test:

      for (int i = 0; i < 10000; i++) {
        CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
            .withCache("txCache1", CacheConfigurationBuilder.newCacheConfigurationBuilder().buildConfig(Long.class, String.class))
            .using(new DefaultStoreProvider())
            .using(new DefaultTimeSourceService(new TimeSourceConfiguration(new TestTimeSource())))
            .build(true);
        cacheManager.close();
      }

It will fail after a few iterations because the DefaultTimeSourceService is a dependency of DefaultStoreProvider, so sometimes the provided DefaultTimeSourceService is used  and sometimes the DefaultTimeSourceService is loaded via the ServiceLoader.

PR #565, opened 2015-09-10
Issue #561 Fix services dependencies loading

ljacomet:
* defer to when the ServiceLocater starts all services


ljacomet:
test this please

PR #574, opened 2015-09-14
Issue #559 AbstractOffHeapStore iterator to skip expired entries

albinsuresh:


impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -832,44 +833,68 @@ public boolean test(Map.Entry<K, OffHeapValueHolder<V>> argument) {
...
+      advance();
+    }
+
+    private void advance() {
+      next = null;
+      while (next == null && mapIterator.hasNext()) {
+        Map.Entry<K, OffHeapValueHolder<V>> entry = mapIterator.next();
+        final long now = timeSource.getTimeMillis();
+        if (entry.getValue().isExpired(now, TimeUnit.MILLISECONDS)) {
+          onExpiration(entry.getKey(), entry.getValue());
```
albinsuresh: The expired entries are not removed here because OffHeapHashMap iterator does not support remove() operation.

ljacomet: Confused here - you can't remove it but you fire an event?

albinsuresh: Aah missed it. Will remove that. So I'm assuming that deferring the removal of expired entries is fine.

impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -832,44 +833,67 @@ public boolean test(Map.Entry<K, OffHeapValueHolder<V>> argument) {
...
+      advance();
+    }
+
+    private void advance() {
+      next = null;
+      while (next == null && mapIterator.hasNext()) {
+        Map.Entry<K, OffHeapValueHolder<V>> entry = mapIterator.next();
+        final long now = timeSource.getTimeMillis();
+        if (entry.getValue().isExpired(now, TimeUnit.MILLISECONDS)) {
+          continue;
```
ljacomet: So we just skip expired entries but do not remove them, correct?

albinsuresh: Exactly. Removal of those entries will happen in subsequent get/compute calls. That should be fine, right?

ljacomet:
Needs a rebase

Issue #369, opened 2015-04-09
Only minimal (de)serialization between tiers


alexsnaps:
If not the case already, make sure we don't unnecessarily deserialize out of one tier, to only re-serialize back into the next one...

ljacomet:
Adding on the topic: currently Store.containsKey ends up calling into get in order to handle expiry - which means deserialization also. Probably worth optimising away.

lorban:
Depends on #158 as the latter will have an impact on this.

ljacomet:
Started looking at operations and cache configurations.

See https://docs.google.com/spreadsheets/d/1JiS3kp3q1SHR0YrIVb7YOUp3JlKJXSDt_Tndz2V1IDg/edit?usp=sharing

Will push a first PR for taking care of low hanging fruits.

Next step is around making `OffHeapValueHolder` lazy with regards to the deserialize of the value itself and add a method to enable taking a copy of the `ByteBuffer` holding it.

ljacomet:
When faulting from authority to caching tier, while we need to deserialize to compute the new expiry duration, we may win a serialization when the caching tier stores serialized entries.
See `org.ehcache.internal.store.heap.OnHeapStore#importValueFromLowerTier` for example.
However, this would mean making the authority aware that the caching tier uses serialization.

ljacomet:
Idea of making the `OffHeapValueHolder` lazy has been pushed out. Only with real perf number will this option be considered again.

Others improvements have been made though in two PRs.

PR #575, opened 2015-09-14
Issue #369 Reduce serialisations in OnHeapStore

ljacomet:
Low hanging fruits to reduce serialization / deserialization in `OnHeapStore` with a `SerializingCopier`


impl/src/main/java/org/ehcache/internal/store/heap/CopiedOnHeapValueHolder.java
```diff
@@ -16,6 +16,7 @@
 
 package org.ehcache.internal.store.heap;
```
albinsuresh: Shouldn't this be moved to the `holder` package?

impl/src/main/java/org/ehcache/internal/store/heap/SerializedOnHeapValueHolder.java
```diff
@@ -16,6 +16,7 @@
 package org.ehcache.internal.store.heap;
```
albinsuresh: `holder` package???

impl/src/test/java/org/ehcache/integration/SerializerCountingTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private final JavaSerializer<T> serializer;
+
+    public CountingSerializer(ClassLoader classLoader) {
+      serializer = new JavaSerializer<T>(classLoader);
+    }
+
+    @Override
+    public ByteBuffer serialize(T object) throws SerializerException {
+      if (PRINT_STACK_TRACES) {
```
albinsuresh: I fail to understand the use/significance of this if block

impl/src/test/java/org/ehcache/internal/store/heap/CopiedOnHeapKeyTest.java
```diff
@@ -17,6 +17,7 @@
 package org.ehcache.internal.store.heap;
```
albinsuresh: `holders` package???

impl/src/test/java/org/ehcache/integration/SerializerCountingTest.java
ljacomet: By changing the variable on top, all calls to a `Serializer` method will start printing stack traces, to more easily see in which context they are called.
This class was not really a test (no asserts) but more a helper for tracking `Serializer` calls.
Now, I could actually perform assertions, as a protection against number of serialisation regressions.

ljacomet:
Pushed an updated version moving all *Holders into the holders package, including tests.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -939,6 +928,17 @@ private void setAccessTimeAndExpiry(K key, OnHeapValueHolder<V> valueHolder, lon
     return makeValue(value, now, expirationTime, this.valueCopier);
   }
 
+  private OnHeapValueHolder<V> importValueFromLowerTier(K key, ValueHolder<V> valueHolder, long now) {
+    V realValue = valueHolder.value();
+    Duration expiration = expiry.getExpiryForAccess(key, realValue);
+    if(valueCopier instanceof SerializingCopier) {
+      // TODO Pass a copy of the byteBuffer from lower tier and pass it instead of realValue
```
lorban: is that TODO still accurate?

impl/src/test/java/org/ehcache/integration/SerializerCountingTest.java
lorban: Replacing the ```System.out``` calls with assertions is probably a good addition for a later time, when we're confident we can't do any better. For now, I'd leave this test as-is.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
ljacomet: Let me remove it - it may remain a possibility - saving a _serialization_ but at the cost of making the authority aware that the on heap tier is serialiazing ... so not sure at all we want that. I'll add a note instead on the issue for that.

PR #581, opened 2015-09-15
Issue #158 JSR-107 doc update on cache config overriding using templates

albinsuresh:


107/src/test/java/org/ehcache/docs/EhCache107ConfigurationIntegrationDocTest.java
```diff
@@ -166,4 +184,51 @@ public void testWithoutEhcacheExplicitDependencyAndNoCodeChanges() throws Except
...
+    mutableConfiguration.setTypes(Long.class, String.class);
+
+    Cache<Long, String> myCache = null; 
+    myCache = cacheManager.createCache("anyCache", mutableConfiguration);
+    myCache.put(1L, "foo");
+    assertNotSame("foo", myCache.get(1L));
+
+    myCache = cacheManager.createCache("byRefCache", mutableConfiguration);
+    myCache.put(1L, "foo");
+    assertSame("foo", myCache.get(1L));
```
ljacomet: The following assertion should pass but does not:
`assertThat(myCache.getConfiguration(Configuration.class).isStoreByValue(), is(false));`

107/src/test/java/org/ehcache/docs/EhCache107ConfigurationIntegrationDocTest.java
```diff
@@ -166,4 +184,51 @@ public void testWithoutEhcacheExplicitDependencyAndNoCodeChanges() throws Except
...
+
+    myCache = cacheManager.createCache("weirdCache", mutableConfiguration);
+    myCache.put(1L, "foo");
+    assertNotSame("foo", myCache.get(1L));
```
albinsuresh: In reference to your previous comment, what should `myCache.getConfiguration(Configuration.class).isStoreByValue()` return for this *weirdCache* that's neither store-by-value nor store-by-ref but a mix of both?

ljacomet: I would go for the value status in case key and value are not aligned

ljacomet:
Good to go - needs  rebase though

PR #589, opened 2015-09-16
Issue 37: XA Store

lorban:
This PR is about the full XA store. Things that need to be reviewed with a bit more attention:

 - the couple of changes I had to make in api/core/impl
 - the asciidoc "integration"
 - the independency / integration of the "transactions" module
 - the config settings "look and feel"
 - everything else!

### DIFF
--- a/api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
+++ b/api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.exceptions;
+
+/**
+ * Generic exception used when an internal operation fails on a {@link org.ehcache.Cache} but shouldn't be
+ * handled by a resilience strategy but rather rethrown to the caller.
+ *
+ * @deprecated This mechanism is a stop-gap solution until {@link org.ehcache.resilience.ResilienceStrategy}
+ * instances can be plugged-in.
+ *
+ * @author Ludovic Orban
+ */
+@Deprecated
+public class RethrowingCacheAccessException extends CacheAccessException {
+
+  /**
+   * Create an instance of RethrowingCacheAccessException.
+   * @param cause the cause RuntimeException that will be rethrown.
+   */
+  public RethrowingCacheAccessException(RuntimeException cause) {
+    super(cause);
+  }
+
+  @Override
+  public synchronized RuntimeException getCause() {
+    return (RuntimeException) super.getCause();
+  }
+}
--- a/impl/src/main/java/org/ehcache/spi/copy/DefaultCopyProvider.java
+++ b/impl/src/main/java/org/ehcache/spi/copy/DefaultCopyProvider.java
@@ -23,11 +23,8 @@ import org.ehcache.internal.classes.ClassInstanceProvider;
 import org.ehcache.internal.copy.IdentityCopier;
 import org.ehcache.internal.copy.SerializingCopier;
 import org.ehcache.spi.ServiceLocator;
-import org.ehcache.spi.ServiceProvider;
-import org.ehcache.spi.serialization.SerializationProvider;
 import org.ehcache.spi.serialization.Serializer;
 import org.ehcache.spi.service.ServiceConfiguration;
-import org.ehcache.spi.service.ServiceDependencies;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -58,20 +55,24 @@ public class DefaultCopyProvider extends ClassInstanceProvider<Copier<?>> implem
   private <T> Copier<T> createCopier(CopierConfiguration.Type type, Class<T> clazz,
                                      Serializer<T> serializer, ServiceConfiguration<?>... configs) {
     DefaultCopierConfiguration<T> conf = find(type, configs);
-    if(conf != null && conf.getClazz().isAssignableFrom(SerializingCopier.class)) {
-      return new SerializingCopier<T>(serializer);
+    Copier<T> copier;
+    if (conf != null && conf.getInstance() != null) {
+      copier = conf.getInstance();
+    } else if (conf != null && conf.getClazz().isAssignableFrom(SerializingCopier.class)) {
+      copier = new SerializingCopier<T>(serializer);
+    } else {
+      copier = createCopier(clazz, conf);
     }
-    return createCopier(clazz, conf);
+    LOG.info("Copier for <{}> : {}", clazz.getName(), copier);
+    return copier;
   }
 
   private <T> Copier<T> createCopier(Class<T> clazz, DefaultCopierConfiguration<T> config) {
     String alias = (config != null ? null : clazz.getName());
     Copier<T> copier = (Copier<T>) newInstance(alias, config);
     if (copier == null) {
-      LOG.info("No registered copier found for for <{}>. Using the default Identity copier.", clazz.getName());
       copier = new IdentityCopier<T>();
     }
-    LOG.info("Copier for <{}> : {}", clazz.getName(), copier);
     return copier;
   }
 
--- a/transactions/config/checkstyle-suppressions.xml
+++ b/transactions/config/checkstyle-suppressions.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!DOCTYPE suppressions PUBLIC
+    "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+    "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+</suppressions>
--- a/transactions/config/checkstyle-suppressions.xml
+++ b/transactions/config/checkstyle-suppressions.xml
@@ -5,4 +5,5 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
+  <suppress files="^((?!.*test[\\/]java[\\/]org[\\/]ehcache[\\/]docs[\\/].*).)*$" checks="AvoidStaticImport"/>
 </suppressions>
--- a/transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
+++ b/transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
@@ -1,0 +1,886 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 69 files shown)


lorban:
Things that have been left aside:
 - Integration with different transaction managers. Currently only BTM is supported and has been tested. Though, the XA mechnisms used should work with other impls too and the TX mgr lookup code has been made pluggable.
 - XAStore's putIfAbsent(), replace(2 args), replace(3 args), remove(2 args) because they currently are uneeded.
 - Non-strict XA, i.e.: using a JTA Synchronization to sync the XA store with a TM.

impl/src/main/java/org/ehcache/spi/copy/DefaultCopyProvider.java
```diff
@@ -58,20 +55,24 @@ public DefaultCopyProvider(DefaultCopyProviderConfiguration configuration) {
...
-    return createCopier(clazz, conf);
+    LOG.info("Copier for <{}> : {}", clazz.getName(), copier);
+    return copier;
   }
 
   private <T> Copier<T> createCopier(Class<T> clazz, DefaultCopierConfiguration<T> config) {
     String alias = (config != null ? null : clazz.getName());
     Copier<T> copier = (Copier<T>) newInstance(alias, config);
     if (copier == null) {
-      LOG.info("No registered copier found for for <{}>. Using the default Identity copier.", clazz.getName());
```
ljacomet: Might be interesting to keep that one as debug

transactions/src/docs/asciidoc/user/xa.adoc
```diff
@@ -0,0 +1,207 @@
+include::common.adoc[]
```
ljacomet: Why is this file here and not in the docs module?

transactions/config/checkstyle-suppressions.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!DOCTYPE suppressions PUBLIC
+    "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+    "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+  <suppress files="^((?!.*test[\\/]java[\\/]org[\\/]ehcache[\\/]transactions[\\/]xa[\\/]docs[\\/].*).)*$" checks="AvoidStaticImport"/>
```
ljacomet: Seems to indicate that you want samples for doc in a different package. Why?

ljacomet:
Mostly made comments around the module setup and integration with the project.

On that front, it is missing integration in building the ehcache shaded jar. I am assuming we want that jar to contain transaction stuff from now on.

impl/src/main/java/org/ehcache/spi/copy/DefaultCopyProvider.java
lorban: That particular location wasn't ideal, so I've moved that INFO log a few lines above which makes the logger output less confusing.

transactions/config/checkstyle-suppressions.xml
lorban: For no special reason. I wondered how independent the transactions modules has to be, do I bundled everything including the doc in it. I could move all that to the docs module if you believe this makes more sense.

transactions/src/docs/asciidoc/user/xa.adoc
lorban: Same answer as below: I can move the XA doc to the docs module if you believe it makes more sense.

impl/src/main/java/org/ehcache/spi/copy/DefaultCopyProvider.java
ljacomet: Right, no need to log twice, the other log entry indicates which copier is used always.

api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Create an instance of RethrowingCacheAccessException.
+   * @param cause the cause RuntimeException that will be rethrown.
+   */
+  public RethrowingCacheAccessException(RuntimeException cause) {
+    super(cause);
+  }
+
+  @Override
+  public synchronized RuntimeException getCause() {
```
ljacomet: Why `synchronized`?

transactions/src/main/java/org/ehcache/transactions/xa/configuration/XAStoreConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+import org.ehcache.transactions.xa.XAStore;
+
+/**
+ * @author Ludovic Orban
+ */
+public class XAStoreConfiguration implements ServiceConfiguration<XAStore.Provider> {
+
+  private final String uniqueXAResourceId;
+
+  //TODO: ideally, the uniqueXAResourceIdSuffix should be the cache alias, injected by the CM
```
ljacomet: Is that still a _TODO_? Cause it mentions something that no longer seem to exist

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
```diff
@@ -0,0 +1,908 @@
...
+
+        // check if the store is persistent, and get the PersistenceSpaceIdentifier if it is
+        LocalPersistenceService.PersistenceSpaceIdentifier persistenceSpaceId = null;
+        ResourcePools resourcePools = storeConfig.getResourcePools();
+        for (ResourceType resourceType : resourcePools.getResourceTypeSet()) {
+          ResourcePool pool = resourcePools.getPoolForResource(resourceType);
+          if (pool.isPersistent()) {
+            try {
+              LocalPersistenceService persistenceService = serviceProvider.getService(LocalPersistenceService.class);
+              persistenceSpaceId = persistenceService.getOrCreatePersistenceSpace(uniqueXAResourceId);
```
ljacomet: Are you sure you need a different storage _space_? That is one disconnected from the _space_ used by this specific cache?
Why don't you look up the persistenceSpaceId passed in the service configurations?
Then the data would end up in the same location as the rest of the data for the cache, and so if the cache is destroyed then this data is destroyed as well.

ljacomet:
PR needs a rebase, for some reason it is based on the last real commit and not the merge commit.

api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
lorban: Because the overridden implementation is synchonized too.

transactions/src/main/java/org/ehcache/transactions/xa/configuration/XAStoreConfiguration.java
lorban: First: uniqueXAResourceIdSuffix should read uniqueXAResourceId, I'll fix that

Then, this can be discussed; we may still want the unique XAResource ID to default to the cache name at a later stage, or close that door forever.

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
lorban: Our understandings aren't aligned: mine is that this code creates a subspace into the cache's space. I'll investigate and if I'm wrong, I'll change this code accordingly.

lorban:
test this please

api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
ljacomet: What?

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
ljacomet: For that you need to invoke `org.ehcache.internal.persistence.DefaultLocalPersistenceService#createPersistenceContextWithin` which takes a persistenceSpaceId as first param and a name (the subfolder) as second.

api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
ljacomet: So not `synchronised` in 1.6, well in 1.7 and 1.8 ... ok, good for me then

lorban: Still weird...

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
lorban: The createPersistenceContextWithin() call is done in DefaultJournalProvider. But you're right that there is no need to create a new persistence space, I can just lookup the space ID that is in the service configs instead and reuse the cache's space.

lorban:
test this please

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
```diff
@@ -0,0 +1,897 @@
...
+      final Transaction transaction = transactionManagerWrapper.getTransactionManager().getTransaction();
+      if (transaction == null) {
+        throw new XACacheAccessException(new XACacheException("Cannot access XA cache outside of XA transaction scope"));
+      }
+      EhcacheXAResource<K, V> xaResource = xaResources.get(transaction);
+      if (xaResource == null) {
+        xaResource = new EhcacheXAResource<K, V>(underlyingStore, journal, transactionContextFactory);
+        transactionManagerWrapper.registerXAResource(uniqueXAResourceId, xaResource);
+        transactionManagerWrapper.getTransactionManager().getTransaction().enlistResource(xaResource);
+        xaResources.put(transaction, xaResource);
```
ljacomet: Unless a transaction can only ever be bound to one thread, this should be a `putIfAbsent` guarding the synchronization registration.

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
```diff
@@ -0,0 +1,897 @@
...
+      advance();
+      return rc;
+    }
+  }
+
+  @Override
+  public ValueHolder<V> compute(K key, BiFunction<? super K, ? super V, ? extends V> mappingFunction, NullaryFunction<Boolean> replaceEqual) throws CacheAccessException {
+    checkKey(key);
+    XATransactionContext<K, V> currentContext = getCurrentContext();
+    if (currentContext.touched(key)) {
```
ljacomet: The whole if block can be extracted to a method `updateCommandForKey` to clarify flow and is duplicated in `computeIfPresent` so the method can be used there too.

AbfrmBlr:
test this please

api/src/main/java/org/ehcache/exceptions/RethrowingCacheAccessException.java
ljacomet: Thinking of it some more, no need for your impl to be synchronized since you call into super which will be (or not) depending on the VM running.

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
lorban: A transaction can only ever be bound/visible by one thread at a time, though it can migrate from one thread to another by the means of explicit calls to TransactionManager.suspend() / .resume(). And we create a new EhcacheXAResource instance for every transaction.

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
lorban: Huh? AFAICT compute(), computeIfAbsent() and computeIfPresent() behave slightly differently in almost every aspect of their logic.
What am I overlooking?

ljacomet: Inside the `currentContext.touched(key)` the code is the same. Maybe you need a condition before invoking the passed in function in `computeIfPresent`?

AbfrmBlr:
test this please

ljacomet:
test this please

AbfrmBlr:
test this please

Issue #497, opened 2015-08-04
Finalize & doc serializers


alexsnaps:

ljacomet:
Depends on #155, #455 and #369

PR #591, opened 2015-09-17
Issue #497 Finalize Serializer documentation and samples

ljacomet:


impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -339,7 +342,7 @@ public void cacheSerializingCopiers() throws Exception {
     CacheConfiguration<Long, Person> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
         .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).build())
         .add(new DefaultCopierConfiguration<Long>((Class)SerializingCopier.class,   //<1>
```
albinsuresh: `DefaultCopierConfiguration<Person>` right??? Since you changed the copier config type.

ljacomet: Good catch - fixing that!

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -217,15 +223,29 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
 <2> Configure the value's serializer.
 
-=== Store-By-Value Caches with Copiers
+===== Implementation of a `Serializer`
+
+When implementing a `org.ehcache.spi.serialization.Serializer`, you obviously need to respect the interface contract.
+In addition, there are a few implicit contracts:
+
+* Your implementation must have a public constructor taking a `ClassLoader` as parameter.
+* There are two annotations that you have to use to mark your `Serializer` - a single implementation can use both:
```
albinsuresh: "have to use" or "may use"? The annotations are not mandatory when configured at the cache level, right? It is mandatory for the cache manager level config, though.

ljacomet: Well, _have to_ covers both :smiley:

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -185,15 +188,18 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
-multi-tiered cache, serialization and deserialization happens when the data is moved between the tiers. Ehcache provides
-a serializer that can serialize any `Serializable` type. So if you are using standard `Serializable` types as your keys
-and values, no extra serializers need to be configured as the inbuilt serializer would handle those cases. But if you
-use your own custom types in ehcache that are not serializable, the stock serializer won't be able to perform the
-serialization/deserialization on its own. So you need to provide the serialization logic as well. Inorder to do this you
-can register your own custom serializers to the cache/cache manager. So if you register a serializer for some type of
-data, that serializer would be used to perform all the serialization/deserialization for that type.
-You can configure a set of serializers on the cache manager:
+Serialization is primarily used in ehcache to store the data in the non-heap tiers.
+So whenever the user uses a multi-tiered cache, serialization and deserialization happens when the data is moved to, from and between the tiers.
```
AbfrmBlr: This shouldn't be always true after #369 is done

ljacomet: Maybe, but it remains an implementation detail while indicating that here hints that you want a performant serialisation solution.

lorban: True, I'd reword that as '… serialization and deserialization *can happen* when …'

Issue #594, opened 2015-09-18
Configuring a cache with a disk pool without configuring a CacheManagerPersistenceConfiguration throws NPE


lorban:
Here's a simple test that reproduces the problem:

```
    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
        .withCache("aCache", CacheConfigurationBuilder.newCacheConfigurationBuilder()
                .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
                        .heap(10, EntryUnit.ENTRIES)
                        .disk(20, MemoryUnit.MB)
                )
                .buildConfig(Long.class, String.class)
        )
        .build(true);
```

lorban:
And here's the stack trace:

```
SEVERE: Initialize failed.
Exception in thread "main" org.ehcache.exceptions.StateTransitionException: Cache 'aCache' creation in EhcacheManager failed.
	at org.ehcache.StatusTransitioner$Transition.failed(StatusTransitioner.java:223)
	at org.ehcache.EhcacheManager.init(EhcacheManager.java:485)
	at org.ehcache.CacheManagerBuilder.build(CacheManagerBuilder.java:48)
	at sample.Main.main(Main.java:37)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
	at java.lang.reflect.Method.invoke(Method.java:597)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:140)
Caused by: java.lang.IllegalStateException: Cache 'aCache' creation in EhcacheManager failed.
	at org.ehcache.EhcacheManager.createCache(EhcacheManager.java:223)
	at org.ehcache.EhcacheManager.init(EhcacheManager.java:470)
	... 7 more
Caused by: java.lang.NullPointerException
	at org.ehcache.EhcacheManager.createNewEhcache(EhcacheManager.java:269)
	at org.ehcache.EhcacheManager.createCache(EhcacheManager.java:204)
	... 8 more
```

PR #596, opened 2015-09-18
Issue #594 Explicit error message when LocalPersistenceService is req…

ljacomet:
…uired but missing

### DIFF
--- a/core/src/main/java/org/ehcache/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/EhcacheManager.java
@@ -264,6 +264,10 @@ public class EhcacheManager implements PersistentCacheManager {
     if (config.getResourcePools().getResourceTypeSet().contains(ResourceType.Core.DISK)) {
       LocalPersistenceService persistenceService = serviceLocator.getService(LocalPersistenceService.class);
       
+      if (persistenceService == null) {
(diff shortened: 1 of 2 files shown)


lorban:
I'd consider an integration test as well.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -264,6 +264,10 @@ private boolean isDiskTransient(Ehcache<?, ?> ehcache) {
     if (config.getResourcePools().getResourceTypeSet().contains(ResourceType.Core.DISK)) {
       LocalPersistenceService persistenceService = serviceLocator.getService(LocalPersistenceService.class);
       
+      if (persistenceService == null) {
+        throw new IllegalStateException("No LocalPersistenceService could be found - did you configure it at the CacheManager level?");
+      }
+
```
AbfrmBlr: Can't we validate in builder ?

lorban: I don't think we want our builders to have any kind of intelligence w.r.t the logical validity of what they're building.

ljacomet: We cannot and should not validate in builder.

Issue #350, opened 2015-04-01
Add/fix services for threading


alexsnaps:
ThreadPools, Executor, ... other Services?
Have the proper dependency tree and make sure these services are used everywhere (e.g. DiskStore, stats, eventNotification, writeBehind, BootstrapCacheLoader, ... )

PR #601, opened 2015-09-18
PR for #540, #545, #552: LifeCycleService, AliasService and ManagementRegistry refactoring

mathieucarbou:
The goals / use cases of this PR:

See issues: #540, #545, #552

All 3 commits matches these 3 issues and are tested and works _BUT_ #552 depends on #545 and #540, and #545 depends on #540.

- A `ManagementRegistry` should be linked to one and only one cache manager, and re-use its services (`ThreadPoolService`, etc) and comply to the cache manager lifecycle - cf issue #350 
=> Refactored `DefaultManagementRegistry` to use cache manager services

- Provide a way to manage several cache managers from one entry point 
=> Introduced `SharedMangementService`, which uses `ManagementRegistry`, `AliasService` and `LifeCycleService`. This service is not a registry, but just a management service which is able to work across several cache managers.

- Introduce a new method on `ManagementRegistry` to be able to retrieve in 1 call several statistics from several contexts at once. This is a requirement we have from a management perspective, to avoid looping several times.
=> added `collectManyStatistics()`

- Decoupling of `EhcacheManager` class from all statistics and management classes.

- Management services should be "started" once cache and cache managers are initialized, not at service start/stop. Provided a way so that services can listen on the completion of init and close for some objects (caches, cache managers, etc) 
=> done thanks to the `LifeCycleService`

- Cache manager alias is hard-coded and auto-generated at the moment. This alias should be configurable or provided because it is required by the management services and we have to expose it / use it to create context to get the stats. 
=> So introduced an `AliasService` which is responsible of providing cache manager alias and cache aliases within a cache manager. This service is used by the `ManagementRegistry` and the `SharedManagementService` 

- Updated JSR 107 classes to use the `AliasService` instead of creating a final context at constructor time to get the stats: this context could be created in the ctor because the name was auto-generated statically at cache manager creation. But it is not the case anymore.

- Updated tests and doc.

Note: the code behind the `ManagementRegistry` has been left as is and is not refactored (all the capabilities classes, statistic provider classes, etc). As @lorban explained me, this is something that should probably be done a day but this is not part of this PR. The goal was much more to have a working management service that we can start using and we will be able to rewrite the code behind afterwards.


ljacomet:
As discussed with @mathieucarbou and others, this PR will not be merged. Work is ongoing to remove the added `AliasService` and `EhcacheBinding`.

mathieucarbou:
Yep. I nearly finished :-) I think I have a solution, but I'll need to see with you because chacheAdded and cacheRemoved from CacheManagerListeenr are not fired at init / close.

PR #602, opened 2015-09-21
Custom serializers documentation

albinsuresh:
With custom `Transient` only and `Persistent` only serializer implementations.


ljacomet:
@albinsuresh Could you isolate the fix to getting-started and open a PR for just that?

Issue #617, opened 2015-09-29
Persistent Cache with cache manager level serializer


JSAhuja:
Persistent Cache is not picking up the serializer defined at the Cache Manager level.
````
DefaultSerializationProviderConfiguration serializer =    new DefaultSerializationProviderConfiguration().addSerializerFor(java.lang.Long .class, CustomSerializerLong.class);
PersistentCacheManager persistentCacheManager = CacheManagerBuilder.newCacheManagerBuilder()
        .with(new CacheManagerPersistenceConfiguration(new File(getClass().getClassLoader().getResource(".").toURI().getPath(), "myData")))
        .withCache("persistent-cache", CacheConfigurationBuilder.newCacheConfigurationBuilder()
            .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
                .heap(10, EntryUnit.ENTRIES)
                .disk(10L, MemoryUnit.MB, true))
            .buildConfig(Long.class, String.class))
        .using(serializer)
        .build(true);
    Cache<Long, String> cache = persistentCacheManager.getCache("persistent-cache", Long.class, String.class);
    persistentCacheManager.close();
````
Console log for the serializer registered
````
15:33:54.578 [main] INFO  org.ehcache.EhcacheManager - Cache 'persistent-cache' is getting created in EhcacheManager.
15:33:54.583 [main] INFO  o.e.i.p.DefaultLocalPersistenceService - Reusing C:\Users\jia\Documents\SourceTree\ehcachetest\copier\target\test-classes\myData\persistent-cache_4773b8fc9608d7ffb5160821d40afbfe5cec777f
15:33:54.589 [main] INFO  o.e.i.p.DefaultLocalPersistenceService - Reusing C:\Users\jia\Documents\SourceTree\ehcachetest\copier\target\test-classes\myData\persistent-cache_4773b8fc9608d7ffb5160821d40afbfe5cec777f\DefaultSerializationProvider-Key
15:33:54.612 [main] INFO  o.e.s.s.DefaultSerializationProvider - Serializer for <java.lang.Long> : org.ehcache.internal.serialization.CompactPersistentJavaSerializer@3252ac20
[...]

PR #618, opened 2015-09-30
Issue #617 Removing serializer annotations and instead enforcing the …

albinsuresh:
…constructor requirements

### DIFF
--- a/api/src/main/java/org/ehcache/spi/serialization/Serializer.java
+++ b/api/src/main/java/org/ehcache/spi/serialization/Serializer.java
@@ -16,33 +16,31 @@
 package org.ehcache.spi.serialization;
 
 import java.io.Closeable;
-import java.lang.annotation.Retention;
-import java.lang.annotation.Target;
 import java.nio.ByteBuffer;
 import org.ehcache.exceptions.SerializerException;
 
-import static java.lang.annotation.ElementType.TYPE;
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
-
 /**
  * Interface defining the contract used to transform types in a serial form.
- * Implementations must be thread-safe and must define a constructor accepting a single ClassLoader parameter.
+ * Implementations must be thread-safe.
+ * 
+ * When used within the default serialization provider, there is an additional requirement:
+ * The implementations must define either or both of the two constructors:
+ * <ul>
+ *   <li>
+ *     Accepting a single {@link ClassLoader} parameter: This constructor will be used to initialize the serializer for
+ *     transient caches
+ *   </li>
+ *   <li>
+ *     Accepting a {@link ClassLoader} parameter and a file based persistence context: This constructor will be used to
+ *     initialize the serializer for persistent caches 
+ *   </li>
+ * </ul> .
  *
  * @param <T> the type of the instances to serialize
  *
  * @author cdennis
  */
 public interface Serializer<T> extends Closeable {
-
-  /**
-   * Marks a {@code Serializer} that is suitable for operation within a persistent cache.
-   */
-  @Target(TYPE) @Retention(RUNTIME) @interface Persistent {}
-  
-  /**
-   * Marks a {@code Serializer} that is suitable for operation within a transient cache.
-   */
-  @Target(TYPE) @Retention(RUNTIME) @interface Transient {}
 
   /**
    * Transforms the given instance into its serial form.
--- a/docs/src/docs/asciidoc/user/getting-started.adoc
+++ b/docs/src/docs/asciidoc/user/getting-started.adoc
@@ -223,11 +223,16 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
 <1> Configure the key's serializer.
 <2> Configure the value's serializer.
 
+Serializer instances are created per cache for keys and values separately.
+That is, even if you configure the same serializer type for multiple caches, different instances are created for each cache.
+The same applies to the serializers configured at the cache manager level as well.
+Similarly, even within a cache, if you configure the same serializer type for both key and value, different instances are created. 
+
 ===== Implementation of a `Serializer`
 
 When implementing a `org.ehcache.spi.serialization.Serializer`, you obviously need to respect the interface contract.
 
-In addition, if you are using the `DefaultSerializationProvider` service provided by ehcache; which would be the case
+In addition, if you are using the `DefaultSerializationProvider` service provided by Ehcache; which would be the case
 for most of the users, unless they are using their own implementation of `SerializationProvider` service; there are a
 few implicit contracts:
 
@@ -235,10 +240,12 @@ few implicit contracts:
 * A serializer implementation to be used with persistent caches must have a public two-argument constructor taking a `ClassLoader` and a `FileBasedPersistenceContext` as parameters.
  
 You are free to have as many extra constructors as you want. But these are the only constructors that will be used by
-ehcache to instantiate these serializers. For a transient cache the single-argument constructor will be used and for a
+Ehcache to instantiate these serializers. For a transient cache the single-argument constructor will be used and for a
 persistent cache the two-argument constructor will be used. Even though the user is enforced to have either or both the
-constructors in his serializer implementation, the usage of these arguments is upto the user. The user can decide to
-completely ignore these parameters. But the presence of these constructors is a strict requirement.
+constructors in his serializer implementation, the usage of these arguments is up to the user. The user can decide to
+completely ignore these parameters. But the presence of these constructors is a strict requirement. Now, even though the
+usage of the constructor arguments is not enforced, it is highly recommended to use the `FileBasedPersistenceContext` to
+persist the state of the serializers, if there is any. 
 
 ==== Store-By-Value Caches with Copiers
 
--- a/docs/src/docs/asciidoc/user/getting-started.adoc
+++ b/docs/src/docs/asciidoc/user/getting-started.adoc
@@ -226,15 +226,19 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
 ===== Implementation of a `Serializer`
 
 When implementing a `org.ehcache.spi.serialization.Serializer`, you obviously need to respect the interface contract.
-In addition, there are a few implicit contracts:
 
-* Your implementation must have a public constructor taking a `ClassLoader` as parameter.
-* There are two annotations that you have to use to mark your `Serializer` - a single implementation can use both:
-** `org.ehcache.spi.serialization.Serializer.Transient`: With this annotation you indicate that the `Serializer`
-  implementation is fit for use inside a transient cache.
-** `org.ehcache.spi.serialization.Serializer.Persistent`: With this annotation you indicate that the `Serializer`
-  implementation is fit for use inside a persistent cache. The constructor can take an additional parameter, a
-  `org.ehcache.spi.service.FileBasedPersistenceContext` which can then be used to persist serialization metadata.
+In addition, if you are using the `DefaultSerializationProvider` service provided by ehcache; which would be the case
+for most of the users, unless they are using their own implementation of `SerializationProvider` service; there are a
+few implicit contracts:
+
+* A serializer implementation to be used with transient caches must have a public constructor taking a `ClassLoader` as parameter.
+* A serializer implementation to be used with persistent caches must have a public two-argument constructor taking a `ClassLoader` and a `FileBasedPersistenceContext` as parameters.
+ 
+You are free to have as many extra constructors as you want. But these are the only constructors that will be used by
+ehcache to instantiate these serializers. For a transient cache the single-argument constructor will be used and for a
+persistent cache the two-argument constructor will be used. Even though the user is enforced to have either or both the
+constructors in his serializer implementation, the usage of these arguments is upto the user. The user can decide to
+completely ignore these parameters. But the presence of these constructors is a strict requirement.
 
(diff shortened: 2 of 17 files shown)


api/src/main/java/org/ehcache/spi/serialization/Serializer.java
```diff
@@ -26,7 +26,17 @@
 
 /**
  * Interface defining the contract used to transform types in a serial form.
- * Implementations must be thread-safe and must define a constructor accepting a single ClassLoader parameter.
+ * Implementations must be thread-safe and must define either or both of the two constructors:
```
chrisdennis: Admittedly I made the same mistake here, but the constructor details here are part of the DefaultSerializationProvider contract and not an inate part of the Serializer contract.

impl/src/main/java/org/ehcache/config/serializer/DefaultSerializationProviderConfiguration.java
```diff
@@ -41,37 +44,38 @@
     if (serializerClass == null) {
       throw new NullPointerException("Serializer class cannot be null");
     }
+    
     String alias = serializableClass.getName();
+    Constructor<? extends Serializer<T>> transientConstructor = null;
+    Constructor<? extends Serializer<T>> persistentConstructor = null;
     
-    if (serializerClass.isAnnotationPresent(Serializer.Transient.class)) {
```
chrisdennis: This set of changes would be simpler if you extracted an isConstructorPresent(...) method which could do the exception handling.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
albinsuresh: This piece of information should be there in some class/interface that the user would directly use, right? The user never really uses/interacts with the `DefaultSerializationProvider` class directly. He only uses the `DefaultSerializerConfiguration` class to configure and of course the `Serializer` interface to do the actual implementation. So I'm not sure if moving it to the `DefaultSerializationProvider` class would be a good idea because the user may never see that.

ljacomet: I agree with @albinsuresh on the location. However, we should make that specific link explicit by referencing here the fact that it applies to usage with the `DefaultSerializationProvider`.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -228,13 +228,19 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
-* There are two annotations that you have to use to mark your `Serializer` - a single implementation can use both:
-** `org.ehcache.spi.serialization.Serializer.Transient`: With this annotation you indicate that the `Serializer`
-  implementation is fit for use inside a transient cache.
-** `org.ehcache.spi.serialization.Serializer.Persistent`: With this annotation you indicate that the `Serializer`
-  implementation is fit for use inside a persistent cache. The constructor can take an additional parameter, a
-  `org.ehcache.spi.service.FileBasedPersistenceContext` which can then be used to persist serialization metadata.
+* A serializer implementation to be used with transient caches must have a public constructor taking a `ClassLoader` as parameter.
+* A serializer implementation to be used with persistent caches must have a public two-argument constructor taking a `ClassLoader` and a `FileBasedPersistenceContext` as parameters.
+ 
+You are free to have as many extra constructors as you want. But these are the only constructors that will be used by
```
ljacomet: Here also I would indicate the constraints apply to usage with the default Ehcache serialization provider.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
chrisdennis: Okay, I can probably go with that.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
```diff
@@ -26,7 +26,20 @@
 
 /**
  * Interface defining the contract used to transform types in a serial form.
- * Implementations must be thread-safe and must define a constructor accepting a single ClassLoader parameter.
+ * 
+ * When used within the DefaultSerializationProvider which is the default implementation of {@link SerializationProvider} 
+ * service provided by ehcache, there are a few additional requirements:
+ * The implementations must be thread-safe and must define either or both of the two constructors:
```
ljacomet: The thread-safe bit remains true always, so that should be indicated before the special integration with default serialization provider - which I would reference like in my sentence and not a class name given the module differences.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
```diff
@@ -26,7 +26,20 @@
...
+ * <ul>
+ *   <li>
+ *     Accepting a single ClassLoader parameter: This constructor will be used to initialize the serializer for
+ *     transient caches
+ *   </li>
+ *   <li>
+ *     Accepting a ClassLoader parameter and a FileBasedPersistenceContext: This constructor will be used to initialize
+ *     the serializer for persistent caches
```
albinsuresh: I am assuming that the previous comment is applicable to to the `FileBasedPersistenceContext ` reference as well

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -228,13 +228,23 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
+constructors in his serializer implementation, the usage of these arguments is upto the user. The user can decide to
+completely ignore these parameters. But the presence of these constructors is a strict requirement.
+ 
+If you decide to use the `FileBasedPersistenceContext` parameter there are some extra perks. The cache manager will keep
+the serializer's persistent data at the same space as the cache's data is stored. There is no added advantage in doing
+so. It is just an option. If you do not want the data to co-exist for some reason then you can choose to ignore the 
+`FileBasedPersistenceContext` altogether and persist the data wherever you want.
+
+The constructor requirements detailed above are applicable only if you are using the `DefaultSerializationProvider`
+service provided by ehcache(That would be the case for most of the users, unless you are using your own implementation
```
ljacomet: You can drop parenthesis here and have the bit between them as the last sentence.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -228,13 +228,23 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
+persistent cache the two-argument constructor will be used. Even though the user is enforced to have either or both the
+constructors in his serializer implementation, the usage of these arguments is upto the user. The user can decide to
+completely ignore these parameters. But the presence of these constructors is a strict requirement.
+ 
+If you decide to use the `FileBasedPersistenceContext` parameter there are some extra perks. The cache manager will keep
+the serializer's persistent data at the same space as the cache's data is stored. There is no added advantage in doing
+so. It is just an option. If you do not want the data to co-exist for some reason then you can choose to ignore the 
+`FileBasedPersistenceContext` altogether and persist the data wherever you want.
+
+The constructor requirements detailed above are applicable only if you are using the `DefaultSerializationProvider`
```
ljacomet: I would move this paragraph above the description of the implicit contracts

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -228,13 +228,23 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
+* A serializer implementation to be used with transient caches must have a public constructor taking a `ClassLoader` as parameter.
+* A serializer implementation to be used with persistent caches must have a public two-argument constructor taking a `ClassLoader` and a `FileBasedPersistenceContext` as parameters.
+ 
+You are free to have as many extra constructors as you want. But these are the only constructors that will be used by
+ehcache to instantiate these serializers. For a transient cache the single-argument constructor will be used and for a
+persistent cache the two-argument constructor will be used. Even though the user is enforced to have either or both the
+constructors in his serializer implementation, the usage of these arguments is upto the user. The user can decide to
+completely ignore these parameters. But the presence of these constructors is a strict requirement.
+ 
+If you decide to use the `FileBasedPersistenceContext` parameter there are some extra perks. The cache manager will keep
```
ljacomet: I would drop this paragraph altogether, as it relates to implementation details. And I would certainly NOT encourage devs to create persistent serializers that go write on random places.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -226,15 +226,19 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
+
+In addition, if you are using the `DefaultSerializationProvider` service provided by ehcache; which would be the case
+for most of the users, unless they are using their own implementation of `SerializationProvider` service; there are a
+few implicit contracts:
+
+* A serializer implementation to be used with transient caches must have a public constructor taking a `ClassLoader` as parameter.
+* A serializer implementation to be used with persistent caches must have a public two-argument constructor taking a `ClassLoader` and a `FileBasedPersistenceContext` as parameters.
+ 
+You are free to have as many extra constructors as you want. But these are the only constructors that will be used by
+ehcache to instantiate these serializers. For a transient cache the single-argument constructor will be used and for a
```
ljacomet: We capitalise Ehcache

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -226,15 +226,19 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
+persistent cache the two-argument constructor will be used. Even though the user is enforced to have either or both the
+constructors in his serializer implementation, the usage of these arguments is upto the user. The user can decide to
```
ljacomet: typo: upto -> up to

impl/src/main/java/org/ehcache/config/serializer/DefaultSerializationProviderConfiguration.java
```diff
@@ -41,40 +42,42 @@
...
-          } else {
-            persistentSerializers.put(alias, serializerClass);
-          }
-        } else if (persistentSerializers.containsKey(alias)) {
-          transientSerializers.put(alias, serializerClass);
-        } else {
-          persistentSerializers.put(alias, serializerClass);
-          transientSerializers.put(alias, serializerClass);
-        }
+    String alias = serializableClass.getName();
```
chrisdennis: I understand that this new code is logically equivalent to what was there before, however that isn't entirely clear when looking at the changeset.  The only thing that needs to change to accomplish the task is to change the conditions in the if statements to `isConstructorPresent(...)`.

albinsuresh: Okay, I blame this diff engine for messing up the whole thing. My change set was just this much:

```java
    boolean transientConstructorPresent = false;
    boolean persistentConstructorPresent = false;
    
    if(transientConstructorPresent = isConstructorPresent(serializerClass, ClassLoader.class)) {
      if (transientSerializers.containsKey(alias)) {
        throw new IllegalArgumentException("Duplicate transient serializer for class : " + alias);
      } else {
        transientSerializers.put(alias, serializerClass);
      }
    }
    
    if(persistentConstructorPresent = isConstructorPresent(serializerClass, ClassLoader.class, FileBasedPersistenceContext.class)) {
      if (persistentSerializers.containsKey(alias)) {
        throw new IllegalArgumentException("Duplicate persistent serializer for class : " + alias);
      } else {
        persistentSerializers.put(alias, serializerClass);
      }
    }

    if(!transientConstructorPresent && !persistentConstructorPresent) {
      throw new IllegalArgumentException("The serializer: " + serializerClass.getName()
                                         + " does not meet the constructor requirements for either transient or persistent caches.");
    }
```

And what you are suggesting is this(as it'd simplify the diff set):

```java
    if (isConstructorPresent(serializerClass, ClassLoader.class)) {
      if (isConstructorPresent(serializerClass, ClassLoader.class, FileBasedPersistenceContext.class)) {
        if (transientSerializers.containsKey(alias)) {
          if (persistentSerializers.containsKey(alias)) {
            throw new IllegalArgumentException("Duplicate transient & persistent serializer for class : " + alias);
          } else {
            persistentSerializers.put(alias, serializerClass);
          }
        } else if (persistentSerializers.containsKey(alias)) {
          transientSerializers.put(alias, serializerClass);
        } else {
          persistentSerializers.put(alias, serializerClass);
          transientSerializers.put(alias, serializerClass);
        }
      } else {
        if (transientSerializers.containsKey(alias)) {
            throw new IllegalArgumentException("Duplicate transient serializer for class : " + alias);
        } else {
          transientSerializers.put(alias, serializerClass);
        }
      }
    } else if (isConstructorPresent(serializerClass, ClassLoader.class, FileBasedPersistenceContext.class)) {
      if (persistentSerializers.containsKey(alias)) {
          throw new IllegalArgumentException("Duplicate persistent serializer for class : " + alias);
      } else {
        persistentSerializers.put(alias, serializerClass);
      }
    } else {
      throw new IllegalArgumentException("Serializer class '" + serializerClass + "' does not identify as persistent or transient.");
    }
```

Personally, I feel that the second code block looks too complicated with too many nested if-else blocks, even though both does the same thing. I feel the the first one looks much cleaner. Now, if I get one more veto against my opinion I'll happily change it to the second one :)

ljacomet: I find myself in agreement with @albinsuresh here.

PR #619, opened 2015-09-30
Issue #369 OffHeapValueHolder with lazy deserialization

### LINKED ISSUE #369: Only minimal (de)serialization between tiers

ljacomet:
This PR pushes further the reduction of serialization / deserialisation required between tiers.

It reaches that goal by doing the following:
* Improve invalidation in the `CompoundCacheStore` by introducing a `HigherCachingTier` interface and a specific method for cross tiers invalidation.
* Makes `OffHeapValueHolder` have different modes - `BINARY`, `VALUE` and `BOTH` - that are used depending on context inside `AbstractOffHeapStore`, it also extends a new interface `BinaryValueHolder`
* Improves faulting on-heap when the faulted `ValueHolder` implements `BinaryValueHolder`

The second change is what enables the third one and probably others. However, I am looking for opinions on whether the changes done inside `AbstractOffHeapStore` and `OffHeapValueHolder` to reach this are considered worth it.

This PR can be merged but does not close #369.

### DIFF
--- a/impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
+++ b/impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
@@ -48,6 +48,8 @@ import org.ehcache.statistics.StoreOperationOutcomes;
 import org.terracotta.offheapstore.Segment;
 import org.terracotta.offheapstore.exceptions.OversizeMappingException;
 import static org.terracotta.statistics.StatisticBuilder.operation;
+
+import org.terracotta.offheapstore.util.FindbugsSuppressWarnings;
 import org.terracotta.statistics.observer.OperationObserver;
 
 public abstract class AbstractOffHeapStore<K, V> implements AuthoritativeTier<K, V>, LowerCachingTier<K, V> {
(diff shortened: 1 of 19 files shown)


impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -843,14 +857,37 @@ public boolean test(Map.Entry<K, OffHeapValueHolder<V>> argument) {
...
       next = null;
       while (next == null && mapIterator.hasNext()) {
-        Map.Entry<K, OffHeapValueHolder<V>> entry = mapIterator.next();
-        final long now = timeSource.getTimeMillis();
-        if (entry.getValue().isExpired(now, TimeUnit.MILLISECONDS)) {
-          containsKey(entry.getKey());  //This is done here to remove the expired entry without any side effects
-          continue;
-        }
+        final Map.Entry<K, OffHeapValueHolder<V>> entry = mapIterator.next();
+        final OffHeapValueHolder<V> value = (OffHeapValueHolder<V>) internalGet(entry.getKey(), false, true);
```
ljacomet: This is one of the ugly changes. There is no easy way to tell the iterator to eagerly deserialize and when we have the `OffHeapValueHolder` back, it is out of lock scope so we can't do anything with it. Hence the call to internalGet without updating access stats but forcing deserialization.

ljacomet:
After discussion with @lorban, @alexsnaps and @chrisdennis decision has been made to postpone the lazy off heap value holder changes. So closing this and will open another PR for the invalidation changes.

PR #623, opened 2015-10-02
Issue #369 Improve CompoundCachingTier.invalidate(K key)

ljacomet:
* Skip pushing down to lower tier and flush to authority directly instead
* Rename and move invalidate variant taking a function to HigherCachingTier

### DIFF
--- a/impl/src/main/java/org/ehcache/internal/store/tiering/CompoundCachingTier.java
+++ b/impl/src/main/java/org/ehcache/internal/store/tiering/CompoundCachingTier.java
@@ -22,6 +22,7 @@ import org.ehcache.function.NullaryFunction;
 import org.ehcache.spi.ServiceProvider;
 import org.ehcache.spi.cache.Store;
 import org.ehcache.spi.cache.tiering.CachingTier;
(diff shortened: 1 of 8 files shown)


impl/src/main/java/org/ehcache/internal/store/tiering/CompoundCachingTier.java
```diff
@@ -194,7 +180,7 @@ public void setInvalidationListener(InvalidationListener<K, V> invalidationListe
       }
 
       CachingTier.Provider higherProvider = serviceProvider.getService(compoundCachingTierServiceConfiguration.higherProvider());
-      CachingTier<K, V> higherCachingTier = higherProvider.createCachingTier(storeConfig, serviceConfigs);
+      HigherCachingTier<K, V> higherCachingTier = (HigherCachingTier<K, V>) higherProvider.createCachingTier(storeConfig, serviceConfigs);
```
chrisdennis: Should we not have a specialized provider interface for the higher tier too?

ljacomet: Indeed, will work on adding that

Issue #620, opened 2015-09-30
NPE for invalid Cache config & non existing XML


JSAhuja:
Cont.. from the chain mail, getting NPE exception for below case.
Something meaningful messages can help user understand the NPE which are cause by user errors.

i)       Pointed xml is not available.
````
public void identityCopierTest() throws ClassNotFoundException, SAXException, InstantiationException, IOException, IllegalAccessException {
  CopierCallCounter.resetCounter();
  String initialAlias="Hello", modifiedAlias="Verifying store by reference";
  Configuration xmlConfig = new XmlConfiguration(getClass().getResource("/XMLNotAvailable123.xml"));//NPE, #42
  CacheManager cacheManager = CacheManagerBuilder.newCacheManager(xmlConfig);
  cacheManager.init();
  Cache<Integer, Description> cache = cacheManager.getCache("IdentityCopier", Integer.class, Description.class);
  cacheManager.close();
}
````

Stack trace:
````
java.lang.NullPointerException
    at org.ehcache.config.xml.XmlConfiguration.parseConfiguration(XmlConfiguration.java:158)
    at org.ehcache.config.xml.XmlConfiguration.<init>(XmlConfiguration.java:152)
    at org.ehcache.config.xml.XmlConfiguration.<init>(XmlConfiguration.java:127)
    at org.ehcache.config.xml.XmlConfiguration.<init>(XmlConfiguration.java:107)
    at com.terracotta.qa.ehcache.CustomCopierTest.identityCopierTest(CustomCopierTest.java:42)
```` 
 
ii) Cache config is not available in the XML, i.e. the xml is present but have only <ehcache:config> </ehcache:config> tags
````` 
public void identityCopierTest() throws ClassNotFoundException, SAXException, InstantiationException, IOException, IllegalAccessException {
  CopierCallCounter.resetCounter();
  String initialAlias="Hello", modifiedAlias="Verifying store by reference";
  Configuration xmlConfig = new XmlConfiguration(getClass().getResource("/ehcache/CustomCopierTest.xml"));
  CacheManager cacheManager = CacheManagerBuilder.newCacheManager(xmlConfig);
  cacheManager.init();
  Cache<Integer, Description> cache = cacheManager.getCache("CacheNotDefined", Integer.class, Description.class);// CacheNotDefined is not defined in the XML

  Description desc = new Description(1234, initialAlias);
  cache.put(1, desc);//NPE #48

  cacheManager.close();
}
```` 
XML Something like this:
````
[...]

PR #624, opened 2015-10-05
Issue #620 Throw NPE with a meaningful message while instantiating an…

albinsuresh:
… XmlConfiguration with a null URL

### DIFF
--- a/xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
+++ b/xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
@@ -137,6 +137,7 @@ public class XmlConfiguration implements Configuration {
    *
    * @param url URL pointing to the XML file's location
    * @param classLoader ClassLoader to use to load user types.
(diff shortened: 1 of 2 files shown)


xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -146,6 +146,9 @@ public XmlConfiguration(URL url, final ClassLoader classLoader)
    */
   public XmlConfiguration(URL url, final ClassLoader classLoader, final Map<String, ClassLoader> cacheClassLoaders)
       throws ClassNotFoundException, SAXException, InstantiationException, IllegalAccessException, IOException {
+    if(url == null) {
+      throw new NullPointerException("The resource specified by the URL does not exist");
```
ljacomet: That's not the correct message. The message should be: "Url cannot be null".
And while you are at it, I believe null in the following two parameters is an error also.

PR #625, opened 2015-10-05
PR for #552: ManagementRegistry Refactoring

mathieucarbou:
The goals / use cases of this PR: See issues #552,

* A ManagementRegistry should be linked to one and only one cache manager
* ManagementRegistry should re-use ThreadPoolService - issue #350 
* Way to manage several cache managers from one "shared" service
* Refactored ManagementRegistry interface to facilitate statistic collection
* Decoupling of EhcacheManager class from all statistics and management classes.
* Management registry uses CacheManagerListener to listen for cache manager init / close and cache added / removed
* Supports configuration to alias a cache manager for the management module
* Simplified capability naming to not depend on a fully qualified name
* Updated JSR 107 classes to not depend on a static way to recover configured cache manager alias
* Updated tests and doc.


management/src/test/java/org/ehcache/docs/ManagementTest.java
```diff
@@ -51,25 +55,26 @@ public void usingManagementRegistry() throws Exception {
         .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).build())
         .buildConfig(Long.class, String.class);
 
-    ManagementRegistry managementRegistry = new DefaultManagementRegistry(); // <1>
+    DefaultManagementRegistryConfiguration registryConfiguration = new DefaultManagementRegistryConfiguration().setCacheManagerAlias("myCacheManager"); // <1>
+    ManagementRegistry managementRegistry = new DefaultManagementRegistry(registryConfiguration); // <2>
```
ljacomet: This is not needed as it would be handled automatically by the service locator. Our samples should promote recommended usage.

management/src/main/resources/META-INF/services/org.ehcache.spi.service.ServiceFactory
```diff
@@ -1 +1,2 @@
-org.ehcache.management.registry.DefaultManagementRegistryFactory
\ No newline at end of file
+org.ehcache.management.registry.DefaultManagementRegistryFactory
+org.ehcache.management.registry.DefaultSharedManagementServiceFactory
```
ljacomet: Given what I understand of the shared management registry, having it declared here makes no sense.

ljacomet:
General code style comment: can you make sure trailing blank chars are trimmed.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -462,6 +434,10 @@ public void deregisterListener(CacheManagerListener listener) {
   public void init() {
     final StatusTransitioner.Transition st = statusTransitioner.init();
 
+    if (serviceLocator.getService(CacheManagerProviderService.class) == null) {
+      this.serviceLocator.addService(new DefaultCacheManagerProviderService(this));
```
chrisdennis: What does it mean if the `getService(...)` call returns non-null?

ljacomet: That we did (init - close)+ - init cycles

management/src/test/java/org/ehcache/docs/ManagementTest.java
mathieucarbou: Added a better comment in management.adoc ;-)

mathieucarbou:
@ljacomet : PR updated!

Issue #214, opened 2014-12-09
Ehcache iterator throws RuntimeException in response to CacheAccessException in Store on Iterator.next() 


timeck:
See `org.ehcache.Ehcache.CacheEntryIterator.next()`

This iterator was reworked to make it interact with the Store in the way the 107 TCK expects. In particular it expected that the access time of each entry returned in the iterator is not touched until `next()` is called. In the previous implementation of `org.ehcache.Ehcache.CacheEntryIterator` the call to the store iterator's `next()` happened eagerly when the `org.ehcache.Ehcache.CacheEntryIterator` created

I don't think the 107 expectation is horrible and completely wrong, but it could maybe be relaxed to just expect the total number of access checks after iterating the cache.

albinsuresh:
Identified an issue with the `CacheAccessException` handling contract between the store and the cache. In case of a store failure, store is supposed wrap/handle the exceptions thrown from that level and only throw a `CacheAccessException` to the cache level. But this contract is not honoured in most of the store APIs. `AbsractOffheapStore.get()` being an example. Because of this all the runtime exceptions happening at the store level bleed through to the cache level bypassing the resilience strategy as expected.

ljacomet:
@albinsuresh Please isolate in its own commit the work to transform `RuntimeException` from the underlying data structures into `CacheAccessException` at the store level.
Then do the discussed work on making the iterator remember exception raised during `advance` and throw them when used.

albinsuresh:
A single resilience strategy is insufficient for both iterator creation and performing an iteration.

With a resilience strategy in place, the iterator creation need not fail even in the case of an underlying store failure. As our default implementation a non-iterable iterator can be created and returned in that case(by remembering the failure and returning false on _hasNext()_ and throwing on _next()_).
But in the case of a failure happening in the middle of an iteration, the safest default resilience strategy we could take is to throw a runtime exception.

So the default resilience strategy for iterator creation is not to throw and return an non-iterable iterator. The strategy for an actual iteration is to throw. This is the reason why I'm thinking of splitting the resilience strategy for iteration operations.

After the discussion that I had with @lorban, we have decided to remove the _void iteratorFailure(CacheAccessException e)_ method from the `ResilienceStrategy` interface and replace that with 2 more specific ones:

```java
  /**
   * Called when a cache iterator creation fails due to an underlying store
   * failure.
   *
   * @param e the triggered failure
   * @return the entry to return from the operation
   */
[...]

PR #629, opened 2015-10-06
Issue #214 Delegate the CacheAccessException thrown from the cache it…

albinsuresh:
…erator to the resilience strategy

### DIFF
--- a/core/src/main/java/org/ehcache/Ehcache.java
+++ b/core/src/main/java/org/ehcache/Ehcache.java
@@ -1395,19 +1395,7 @@ public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V> {
 
     public CacheEntryIterator(boolean quiet) {
       this.quiet = quiet;
-      
-      Store.Iterator<Entry<K, Store.ValueHolder<V>>> storeIterator = null;
-      try {
-        storeIterator = store.iterator();
-      } catch (CacheAccessException e) {
-        cacheAccessError(e);
-      }
-      this.iterator = storeIterator;
-    }
-
-    private void cacheAccessError(CacheAccessException e) {
-      resilienceStrategy.iteratorFailure(e);
-      cacheAccessError = true;
+      this.iterator = store.iterator();
     }
     
     @Override
@@ -1417,20 +1405,13 @@ public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V> {
       if (cacheAccessError) {
         return false;
       }
-      
-      try {
-        return iterator.hasNext();
-      } catch (CacheAccessException e) {
-        cacheAccessError(e);
-        return false;
-      }
+
+      return iterator.hasNext();
     }
 
     @Override
     public Entry<K, V> next() {
-      statusTransitioner.checkAvailable();
-      
-      if (cacheAccessError) {
+      if (!hasNext()) {
         throw new NoSuchElementException();
       }
       
@@ -1440,10 +1421,8 @@ public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V> {
         if (!quiet) getObserver.end(GetOutcome.HIT_NO_LOADER);
       } catch (CacheAccessException e) {
         if (!quiet) getObserver.end(GetOutcome.FAILURE);
-        cacheAccessError(e);
-        
-        // XXX: not what we want!
-        throw new RuntimeException(e);
+        cacheAccessError = true;
+        return resilienceStrategy.iteratorFailure(e);
       }
       
       return new ValueHolderBasedEntry<K, V>(current);
--- a/core/src/main/java/org/ehcache/Ehcache.java
+++ b/core/src/main/java/org/ehcache/Ehcache.java
@@ -26,6 +26,7 @@ import org.ehcache.exceptions.BulkCacheLoadingException;
 import org.ehcache.exceptions.BulkCacheWritingException;
 import org.ehcache.exceptions.CacheAccessException;
 import org.ehcache.exceptions.CacheLoadingException;
+import org.ehcache.exceptions.CachePassThroughException;
 import org.ehcache.exceptions.CacheWritingException;
 import org.ehcache.expiry.Duration;
 import org.ehcache.function.BiFunction;
@@ -188,7 +189,7 @@ public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V> {
               }
             } catch (Exception e) {
               cacheLoadingObserver.end(CacheLoadingOutcome.FAILURE);
-              throw newCacheLoadingException(e);
+              throw new CachePassThroughException(newCacheLoadingException(e));
             }
             
             return loaded;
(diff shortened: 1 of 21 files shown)


core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -1379,9 +1383,13 @@ public boolean remove(K key) {
...
-      for (Iterator<Cache.Entry<K, V>> iter = new CacheEntryIterator(true); iter.hasNext(); ) {
-        K key = iter.next().getKey();
-        remove(key);
+      try {
+        for (Iterator<Entry<K, V>> iter = new CacheEntryIterator(true); iter.hasNext(); ) {
+          K key = iter.next().getKey();
+          remove(key);
+        }
+      } catch (CacheAccessException e) {
+        throw new RuntimeException(e);
```
albinsuresh: Throwing RuntimeException to be consistent with the behaviour of the other methods. If we decide to handle these too with resilience strategy, that will be done as a separate commit.

albinsuresh:
Transformation of RuntimeExceptions from the store level to CacheAccessExceptions will be done as a separate commit. Raising the PR with just this commit to get some feedback on the resilience strategy split approach.

ljacomet:
Had a discussion with @albinsuresh and I recommend that the underlying Store iterators be adapted to remember when they encounter a problem and only throw on `Store.Iterator.next()` which will then simplify the resilience strategy interactions by having a single method that could return an entry if a user later chooses to do so but throw by default.

api/src/main/java/org/ehcache/exceptions/CachePassThroughException.java
```diff
@@ -0,0 +1,48 @@
...
+package org.ehcache.exceptions;
+
+/**
+ * A generic wrapper runtime exception that will not be caught and
+ * handled at the store level. This wrapper can be used to leak any
+ * runtime exceptions that you don't want to be caught at the store level.
+ * 
+ * @author Albin Suresh
+ */
+public class CachePassThroughException extends RuntimeException {
```
ljacomet: I do not think we want that type in the API module, it is more of an internal implementation concern. Can it live in `impl` (ideal) or `core`?

core-spi-test/src/main/java/org/ehcache/internal/store/StorePutIfAbsentTest.java
```diff
@@ -143,9 +144,8 @@ public void wrongKeyTypeThrowsException()
         kvStore2.putIfAbsent("key", value);
       }
       throw new AssertionError("Expected ClassCastException because the key is of the wrong type");
-    } catch (ClassCastException e) {
-      // expected
-    } catch (CacheAccessException e) {
+    } catch (CacheAccessException cae) {
+      assertThat(cae.getCause(), instanceOf(ClassCastException.class));
```
ljacomet: Wait, that changes the test semantics. You should do nothing if the cause is a `ClassCastException` and throw otherwise.

core-spi-test/src/main/java/org/ehcache/internal/store/StorePutIfAbsentTest.java
```diff
@@ -165,9 +165,8 @@ public void wrongValueTypeThrowsException()
         kvStore2.putIfAbsent(key, "value");
       }
       throw new AssertionError("Expected ClassCastException because the value is of the wrong type");
-    } catch (ClassCastException e) {
-      // expected
-    } catch (CacheAccessException e) {
+    } catch (CacheAccessException cae) {
+      assertThat(cae.getCause(), instanceOf(ClassCastException.class));
```
ljacomet: Wait, that changes the test semantics. You should do nothing if the cause is a `ClassCastException` and throw otherwise.

core/src/test/java/org/ehcache/EhcacheBasicGetTest.java
```diff
@@ -148,7 +150,8 @@ public void testGetNoStoreEntryCacheLoadingException() throws Exception {
     try {
       ehcache.get("key");
       fail();
-    } catch (CacheLoadingException e) {
+    } catch (CachePassThroughException e) {
```
ljacomet: Ehcache should not leak `CachePassThroughException`, this should still work by catching the `CacheLoadingException`

ljacomet: Looking a bit more at the code, that probably means doing some work on the `FakeStore` used here to match what happened in the real stores.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -698,64 +831,86 @@ public String toString() {
...
-        if (mappedValue != null) {
-          long expirationTime = mappedValue.expirationTime(OnHeapValueHolder.TIME_UNIT);
-          return newUpdateValueHolder(key, existingValue, computedValue, now, expirationTime);
-        } else {
-          return newCreateValueHolder(key, computedValue, now);
         }
+      });
+      return enforceCapacityIfValueNotNull(computeResult);
+    } catch (CachePassThroughException cpte) {
+      Throwable cause = cpte.getCause();
```
ljacomet: These few lines are repeated all over the place and in the `AbstractOffHeapStore` as well.
So they should be extracted. What is also interesting is that none of this requires the context of `this` class, so it would be a valid candidate for a method on `CachePassThroughException`.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -401,26 +491,34 @@ private void invalidate() {
...
-          next = entry;
+        } catch (RuntimeException re) {
+          prefetchFailure = new CacheAccessException(re);
         }
       }
       
       @Override
-      public boolean hasNext() throws CacheAccessException {
+      public boolean hasNext() {
         return next != null;
```
ljacomet: I believe this should become `return next != null || prefetchFailure != null`

albinsuresh: But that is wrong, no? _hasNext_ is supposed to return whether or not a next element is there. A prefetchFailure happens when a next element "is there" but we just failed to retrieve that. So _hasNext_ returning false in that case is like lying to the user, right? The next element is actually there. Even though there is no value in returning *true* either since that is not useful to the user in any way.

ljacomet: According to my reading of the code, if you have a `prefetchFailure`, then `next` is null. So in order to have `hasNext()` return true, it should check both `next` and `prefetchFailure` and either different than `null` indicates there is/was something ... a real one or an exception waiting to happen.

impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -842,25 +1013,35 @@ public boolean test(Map.Entry<K, OffHeapValueHolder<V>> argument) {
...
+        prefetchFailure = ce;
+      } catch (RuntimeException re) {
+        prefetchFailure = new CacheAccessException(re);
       }
     }
 
     @Override
-    public boolean hasNext() throws CacheAccessException {
+    public boolean hasNext() {
       return next != null;
```
ljacomet: See `OnHeapStore` comment

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
albinsuresh: Aah yes. My bad, It is an OR of both the conditions. I misread that as an AND.

albinsuresh:
Committed the exception handling code refactoring changes.

Issue #626, opened 2015-10-06
Write Behind reorders operations when assembling batches


chrisdennis:
If you run a write behind setup without coalescing but with batching write behind can end up reordering puts and removes to the same key.  When assembling batches write behind simply takes all writes amongst the selected operations and makes a write batch, and the same to form a delete batch.  This logic should be corrected to prevent reordering of writes and deletes to the same key.

PR #630, opened 2015-10-06
Issue #626 Ensure correct ordering of writes and deletes over each key

chrisdennis:


impl/src/test/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindTestBase.java
```diff
@@ -29,113 +25,223 @@
...
+      loaderWriter.setLatch(countDownLatch);
+      
+      testCache.remove("key");
+      testCache.put("key", "value1");
+      testCache.remove("key");
+      testCache.put("key", "value2");
+      testCache.remove("key");
+      testCache.put("key", "value3");
+      testCache.remove("key");
+      testCache.put("key", "value4");
```
ljacomet: Why do we do 8 operations when batch is configured at 10? This forces the whole system to wait for the max delay.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindQueue.java
```diff
@@ -395,40 +394,26 @@ private void processQuarantinedItems(List<SingleOperation<K, V>> quarantinedItem
 
   private void processBatchedOperations(List<SingleOperation<K, V>> quarantinedItems) throws Exception {
     final int batchSize = determineBatchSize(quarantinedItems);
-
+
```
ljacomet: Picky on code style: remove trailing blank chars

Issue #615, opened 2015-09-29
Create Store 'access' statistic


ljacomet:
We need to have a statistic indicating Store usage. As there is no mapping between cache methods and store methods, this statistic should not be a single operation but rather a view on what happens inside the store.

Note: This does not cover eviction and expiry which have their own stat already.

ljacomet:
Had a discussion with @chrisdennis and @alexsnaps on this topic.

* Each store will have an _access_ statistic with outcomes amongst: _hit, miss, write, remove_
* In addition there will be a second statistic for caching tiers that will be dedicated to faulting follow up with outcomes: _hit, miss_

ljacomet:
After some more discussion, the second statistic will be a _tier efficiency_ one, which will aim at indicating how a tier is working. So having to fault is unhelpful while returning the value is, etc ...
This stat will be focused on tier specific methods.

ljacomet:
After seeing the implementation result, changing approach:

* Record operation level statistics in the stores
* Logic to expose a tier usefulness statistic based on the operation ones will be part of the management module

PR #631, opened 2015-10-07
Issue #615: Introduce store level statistics

ljacomet:
* Access stat for store mutations
* Tier efficiency stat for all tier specific operations, trying to indicate how well the tier is working

### DIFF
--- a/core/src/main/java/org/ehcache/spi/cache/tiering/LowerCachingTier.java
+++ b/core/src/main/java/org/ehcache/spi/cache/tiering/LowerCachingTier.java
@@ -38,7 +38,7 @@ public interface LowerCachingTier<K, V> extends ConfigurationChangeSupport {
    * @return the value holder, or null.
    * @throws CacheAccessException
    */
-  Store.ValueHolder<V> getOrComputeIfAbsent(K key, Function<K, Store.ValueHolder<V>> source) throws CacheAccessException;
+  Store.ValueHolder<V> installMapping(K key, Function<K, Store.ValueHolder<V>> source) throws CacheAccessException;
 
   /**
    * Return the value holder currently in the caching tier and remove it.
--- a/core/src/main/java/org/ehcache/statistics/StoreOperationOutcomes.java
+++ b/core/src/main/java/org/ehcache/statistics/StoreOperationOutcomes.java
@@ -22,35 +22,6 @@ package org.ehcache.statistics;
  */
 public interface StoreOperationOutcomes {
   /**
-   * The Enum GetOutcome.
-   */
-  enum GetOutcome {
-
-    /** hit. */
-    HIT,
-    /** miss. */
-    MISS;
-  }
-
-  /**
-   * The Enum PutOutcome.
-   */
-  enum PutOutcome {
-    /** added. */
-    ADDED,
-    /** updated. */
-    UPDATED
-  };
-
-  /**
-   * The Enum RemoveOutcome.
-   */
-  enum RemoveOutcome {
-    /** success. */
-    SUCCESS
-  };
-
-  /**
    * The access outcomes
    */
   enum AccessOutcome implements StoreOperationOutcomes {
--- a/core/src/main/java/org/ehcache/statistics/StoreOperationOutcomes.java
+++ b/core/src/main/java/org/ehcache/statistics/StoreOperationOutcomes.java
@@ -51,9 +51,24 @@ public interface StoreOperationOutcomes {
   };
 
   /**
+   * The access outcomes
+   */
+  enum AccessOutcome implements StoreOperationOutcomes {
+    HIT,
+    MISS,
+    WRITE,
+    REMOVE
+  }
+
+  enum TierHelpfulnessOutcome implements StoreOperationOutcomes {
+    HELPFUL,
+    UNHELPFUL
+  }
+
+  /**
    * The eviction outcomes.
    */
-  enum EvictionOutcome implements CacheOperationOutcomes {
+  enum EvictionOutcome implements StoreOperationOutcomes {
     /** success. */
     SUCCESS,
     /** failure */
--- a/impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
+++ b/impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
@@ -124,6 +124,8 @@ public class OnHeapStore<K, V> implements Store<K,V>, HigherCachingTier<K, V> {
     }
   };
 
+  private final OperationObserver<StoreOperationOutcomes.AccessOutcome> accessObserver = operation(StoreOperationOutcomes.AccessOutcome.class).named("access").of(this).tag("onheap-store").build();
+  private final OperationObserver<StoreOperationOutcomes.TierHelpfulnessOutcome> tierHelpfulnessObserver = operation(StoreOperationOutcomes.TierHelpfulnessOutcome.class).named("helpfulness").of(this).tag("onheap-store").build();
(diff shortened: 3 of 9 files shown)


core/src/main/java/org/ehcache/spi/cache/tiering/LowerCachingTier.java
```diff
@@ -38,7 +38,7 @@
    * @return the value holder, or null.
    * @throws CacheAccessException
    */
-  Store.ValueHolder<V> getOrComputeIfAbsent(K key, Function<K, Store.ValueHolder<V>> source) throws CacheAccessException;
+  Store.ValueHolder<V> installMapping(K key, Function<K, Store.ValueHolder<V>> source) throws CacheAccessException;
```
chrisdennis: Why are we renaming this method in this PR?  And if it's necessary could we do the rename in it's own commit?

core/src/main/java/org/ehcache/statistics/StoreOperationOutcomes.java
```diff
@@ -22,38 +22,24 @@
...
-  };
-
-  /**
-   * The Enum RemoveOutcome.
-   */
-  enum RemoveOutcome {
-    /** success. */
-    SUCCESS
-  };
+  enum TierEfficiencyOutcome implements StoreOperationOutcomes {
```
chrisdennis: Efficiency seems like a weird name.  Efficiency would be something derived from this raw data right?  TierUtilityOutcome?  TierHelpfullnessOutcome?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -124,6 +124,8 @@ public void cacheConfigurationChange(CacheConfigurationChangeEvent event) {
     }
   };
 
+  private final OperationObserver<StoreOperationOutcomes.AccessOutcome> accessObserver = operation(StoreOperationOutcomes.AccessOutcome.class).named("access").of(this).tag("onheap-store").build();
+  private final OperationObserver<StoreOperationOutcomes.TierEfficiencyOutcome> tierEfficiencyObserver = operation(StoreOperationOutcomes.TierEfficiencyOutcome.class).named("faulting").of(this).tag("onheap-store").build();
```
chrisdennis: Why is the TierEfficiencyOutcome which we have tried to abstract from the implementation called "faulting"?

core/src/main/java/org/ehcache/spi/cache/tiering/LowerCachingTier.java
ljacomet: It is a `LowerCachingTier` method that has the exact same name as a `CachingTier` method but a completely different contract. Since I was touching code in the area I could not resist the rename that really should have been part of the PR introducing this interface. I will isolate in its own commit and reference that issue instead.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -174,7 +176,8 @@ public OnHeapStore(final Configuration<K, V> config, TimeSource timeSource, Copi
   }
   
   private OnHeapValueHolder<V> internalGet(final K key, final boolean updateAccess) throws CacheAccessException {
-    return map.computeIfPresent(key, new BiFunction<K, OnHeapValueHolder<V>, OnHeapValueHolder<V>>() {
+    accessObserver.begin();
```
chrisdennis: I'm ignorant of lots of the details here... but it seems weird that we're calling the `accessObserver` here, but not calling the `tierEfficiencyObserver`.

impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -149,15 +146,17 @@ public void put(final K key, final V value) throws CacheAccessException {
             }
           }
         }, false);
-        return;
+        break;
```
chrisdennis: Could we not just put the accessObserver call before the return?

impl/src/test/java/org/ehcache/internal/store/heap/BaseOnHeapStoreTest.java
```diff
@@ -581,14 +639,16 @@ public void testComputeIfAbsentExisting() throws Exception {
     OnHeapStore<String, String> store = newStore();
 
     store.put("key", "value");
-    ValueHolder<String> newValue = store.computeIfAbsent("key", new Function<String, String>() {
+    ValueHolder<String> newValue =  store.computeIfAbsent("key", new Function<String, String>() {
```
chrisdennis: no change here

impl/src/test/java/org/ehcache/internal/store/heap/BaseOnHeapStoreTest.java
```diff
@@ -618,7 +679,7 @@ public String apply(String mappedKey) {
           throw RUNTIME_EXCEPTION;
         }
       });
-      throw new AssertionError();
+      fail("Expected exception");
```
chrisdennis: why the change?

impl/src/test/java/org/ehcache/internal/store/heap/BaseOnHeapStoreTest.java
```diff
@@ -852,7 +904,7 @@ public void testGetOrComputeIfAbsentContention() throws InterruptedException {
 
     final OnHeapStore<Long, String> store = newStore();
 
-    int threads = 10;
+    int threads = 5;
```
chrisdennis: why?

core/src/main/java/org/ehcache/statistics/StoreOperationOutcomes.java
ljacomet: Indeed - will use Helpfulness as it relates to the outcomes.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
ljacomet: Good catch!

impl/src/test/java/org/ehcache/internal/store/heap/BaseOnHeapStoreTest.java
ljacomet: To add a message in case this ever fails, applying boy scout rule while navigating (and changing this code)

impl/src/test/java/org/ehcache/internal/store/heap/BaseOnHeapStoreTest.java
ljacomet: Will revert

impl/src/test/java/org/ehcache/internal/store/heap/BaseOnHeapStoreTest.java
ljacomet: Will revert

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
ljacomet: When the `OnHeapStore` acts as anything but an authority, this method is never called, so no need to involve the tier specific stat.

impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
ljacomet: Indeed, will move back in there and take care of the other `break` the same.

ljacomet:
Having second thoughts - will open an other PR with a different impl for this.

Issue #627, opened 2015-10-06
Remove extractable (and rarely used?) write-behind features


chrisdennis:
The write behind retry functionality can be safely made the concern of the CacheLoaderWriter implementor.  This simplifies the write behind implementation and configuration.  If we want to support the useless ourselves this can be accomplished by writing a CacheLoaderWriter wrapper implementation that handles the retry concerns.

ljacomet:
As commented on the PR that removes this (and the rate limiting logic), we should consider providing the wrapper implementation although at a later stage.

chrisdennis:
In the same vein as retry I'm also removing the rate limiting from write behind and pushing it to a potential future CacheLoaderWriter wrapper.

Issue #628, opened 2015-10-06
Write behind failure should clean up the cache?


chrisdennis:
Currently failures during write behind do not attempt to clean up the cache state, thus potentially leaving the cache and SOR inconsistent with each other.  I'm on the fence right now as to whether cleanup should be write-behinds concern, or the concern of the loader-writer implementation.

chrisdennis:
This need a decision on whose responsibility the cleanup is. If it's the loader-writer implementors then it should be documented as such. If it's the write behind implementation then we need to do the handling. My vote currently is for us handling it on the grounds that we do cleanup in write-through already.

PR #632, opened 2015-10-07
Issue #627 Stripping write-behind back to the bone

chrisdennis:
This commit removes two features of write behind (and cleans up a needlessly exposed concept) with the proposal that both these features (retry and rate limiting) can be replaced with `CacheLoaderWriter` wrappers dedicated to this task.  This simplifies the central logic of write behind.

In this initial PR I have not added implementations of these wrappers.  I would be interested in getting feedback on whether we think these should be provided by us now (or ever).


ljacomet:
Ok with the removal, we should consider providing the wrapper implementations although later.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindQueue.java
```diff
@@ -519,28 +467,10 @@ private void processSingleOperation(List<SingleOperation<K, V>> quarantinedItems
...
-            } catch (InterruptedException e1) {
-              Thread.currentThread().interrupt();
-              throw new Exception("Exception while processing key '" + item.getKey() + "' write behind queue", e);
-            }
-          }
-        }
+      try {
+        item.performSingleOperation(cacheLoaderWriter);
+      } catch (Exception e) {
+        LOGGER.warn("Exception while processing key '{}' write behind queue : {}", item.getKey(), e);
```
ljacomet: I believe we need to delegate to the resilience strategy here and by default remove the mapping from the cache.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindQueue.java
```diff
@@ -398,60 +381,25 @@ private void processBatchedOperations(List<SingleOperation<K, V>> quarantinedIte
...
+        for (SingleOperation<K, V> singleOperation : batch) {
+          if (successes.contains(singleOperation.getKey()))
+            batch.remove(singleOperation);
+        }
+        if (failures != null) {
+          for (Map.Entry<?, Exception> entry : failures.entrySet()) {
+            LOGGER.warn("Exception while processing key '{}' write behind queue", entry.getKey());
           }
         }
+      } catch (Exception e) {
```
ljacomet: I believe we need to delegate to the resilience strategy here and by default remove the mappings from the cache

impl/src/main/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindQueue.java
```diff
@@ -398,60 +381,25 @@ private void processBatchedOperations(List<SingleOperation<K, V>> quarantinedIte
...
+        Map<?, Exception> failures = bulkCacheWritingException.getFailures();
+        Set<?> successes = bulkCacheWritingException.getSuccesses();
+        // remove successful items
+        for (SingleOperation<K, V> singleOperation : batch) {
+          if (successes.contains(singleOperation.getKey()))
+            batch.remove(singleOperation);
+        }
+        if (failures != null) {
+          for (Map.Entry<?, Exception> entry : failures.entrySet()) {
+            LOGGER.warn("Exception while processing key '{}' write behind queue", entry.getKey());
```
ljacomet: We should delegate to the resilience strategy and remove the set of failed keys from the cache.

ljacomet:
The run time of the write behind tests is not acceptable. They add 30 seconds to the build. This was introduced in the PR for #626 and should be fixed here.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindQueue.java
chrisdennis: I differ slightly in the details of how this should be handled, but regardless I had intended to fix this under #628

Issue #636, opened 2015-10-12
Started services don't get stopped when an exception happens during initialization


squirrelala:
Incidentally found due to breakages in the serializer tests (we will try create a new standalone test for this specifically):

A test tries to create a cache manager from xml with persistent caches with different serializers.
It checks one cache/combination at a time, closing and reiniting the cache manager in between.
When one case failed due to an incorrectly implemented serializer:
````
Caused by: 
at org.ehcache.StatusTransitioner$Transition.failed(StatusTransitioner.java:223)
at org.ehcache.EhcacheManager.init(EhcacheManager.java:505)
at com.terracotta.qa.ehcache.configuration.XmlProvider.init(XmlProvider.java:47)
at com.terracotta.qa.framework.BaseTest.test(BaseTest.java:182)
... 25 more
Caused by: java.lang.IllegalStateException: Cache 'cacheLongCharSequence' creation in EhcacheManager failed.
at org.ehcache.EhcacheManager.createCache(EhcacheManager.java:239)
at org.ehcache.EhcacheManager.init(EhcacheManager.java:490)
... 27 more
Caused by: java.lang.RuntimeException: org.ehcache.spi.serialization.UnsupportedTypeException: No serializer found for type 'java.lang.CharSequence'
at org.ehcache.EhcacheManager.createNewEhcache(EhcacheManager.java:322)
at org.ehcache.EhcacheManager.createCache(EhcacheManager.java:215)
... 28 more
Caused by: org.ehcache.spi.serialization.UnsupportedTypeException: No serializer found for type 'java.lang.CharSequence'
at org.ehcache.spi.serialization.DefaultSerializationProvider$AbstractProvider.getClassFor(DefaultSerializationProvider.java:189)
at org.ehcache.spi.serialization.DefaultSerializationProvider$PersistentProvider.createSerializer(DefaultSerializationProvider.java:128)
at org.ehcache.spi.serialization.DefaultSerializationProvider$AbstractProvider.createValueSerializer(DefaultSerializationProvider.java:167)
at org.ehcache.spi.serialization.DefaultSerializationProvider.createValueSerializer(DefaultSerializationProvider.java:78)
at org.ehcache.EhcacheManager.createNewEhcache(EhcacheManager.java:318)
... 29 more
````

even though we call cacheManager.close(), it seems that subsequent attempts to init a cache manager using the same xml config / storage location gets an OverlappingFileLockException due to the previously failed init (and/or close) not cleaning up

````
[...]

PR #639, opened 2015-10-19
Issue #636 : Stop all services if failure in initializing CacheManager

rishabhmonga:


core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -471,6 +471,7 @@ public void init() {
               LOGGER.error("Cache '{}' could not be removed due to ", toBeClosed, exceptionClosingCache);
           }
         }
+        serviceLocator.stopAllServices();
```
ljacomet: If this throws, it will shadow the original exception. As a consequence it will make fixing the problem harder.

integration-test/src/test/java/org/ehcache/integration/SerializerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    CacheConfiguration<Integer, CharSequence> configuration = configurationBuilder.buildConfig(Integer.class, CharSequence.class);
+    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder().withCache("myCache", configuration)
+        .using(new TimeSourceConfiguration(testTimeSource))
+        .with(new CacheManagerPersistenceConfiguration(folder.newFolder("tempData")))
+        .build(false);
+    try {
+      cacheManager.init();
+    } catch (Exception e) {
+      // expected
+      assertNotNull(e);
```
ljacomet: What's the point here?

integration-test/src/test/java/org/ehcache/integration/SerializerTest.java
```diff
@@ -0,0 +1,84 @@
...
+            .disk(10, MemoryUnit.MB))
+        .add(new DefaultSerializerConfiguration<CharSequence>(TestSerializer.getBuggySerializer(true), SerializerConfiguration.Type.VALUE));
+    CacheConfiguration<Integer, CharSequence> configuration = configurationBuilder.buildConfig(Integer.class, CharSequence.class);
+    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder().withCache("myCache", configuration)
+        .using(new TimeSourceConfiguration(testTimeSource))
+        .with(new CacheManagerPersistenceConfiguration(folder.newFolder("tempData")))
+        .build(false);
+    try {
+      cacheManager.init();
+    } catch (Exception e) {
```
ljacomet: Can't you be more specific as to what you catch?
Also, if no exception is thrown, the test will proceed fine which is not what we want ...

integration-test/src/test/java/org/ehcache/integration/SerializerTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+/**
+ * @author rism
+ */
+public class SerializerTest {
```
ljacomet: How can this test be named like that? It is testing something completely different that has to do with `CacheManager` lifecycle.
Was there not already a test class where to put this code?

integration-test/src/test/java/org/ehcache/integration/SerializerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @SuppressWarnings("unchecked")
+  @Test
+  public void testSerializers() throws Exception {
+    TestTimeSource testTimeSource = new TestTimeSource();    
+    CacheConfigurationBuilder configurationBuilder = CacheConfigurationBuilder.newCacheConfigurationBuilder()
+        .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder()
+            .heap(10, EntryUnit.ENTRIES)
+            .disk(10, MemoryUnit.MB))
+        .add(new DefaultSerializerConfiguration<CharSequence>(TestSerializer.getBuggySerializer(true), SerializerConfiguration.Type.VALUE));
```
ljacomet: Is that really the simpler way to fail at cache initialisation?

rishabhmonga: This was closest to the reported issue

ljacomet:
@rishabhmonga Could you add a test? I know I made remarks on the previous one, but no tests is not good.

rishabhmonga:
@ljacomet just added it. Sorry about that. Just realised commit -a doesnt add new files. I was trying to rename and move the test to a more appropriate package but forgit to add it

ljacomet:
Had a chat with @rishabhmonga in order to trim down the test.

Once that is done, this is good to merge for me.

rishabhmonga:
updated the test.

rishabhmonga:
test this please

ljacomet:
test this please

Issue #611, opened 2015-09-24
Copier at Cache Manager for Jsr107, XML Cache Template Configuration


JSAhuja:
For Jsr107 using XML cache template configuration, with copiers defined in service tags(cache manager level), are not getting picked up.
Same thing for the serializers is working.

XML:
````
<ehcache:config
    xmlns:ehcache='http://www.ehcache.org/v3' xmlns:jsr107="http://www.ehcache.org/v3/jsr107">

  <ehcache:service>
    <ehcache:default-copiers>
      <ehcache:copier type="com.terracotta.qa.Type.Description">com.terracotta.qa.Copiers.DescriptionCopier</ehcache:copier>
      <ehcache:copier type="com.terracotta.qa.ehcache.generator.Person">com.terracotta.qa.Copiers.PersonCopier</ehcache:copier>
    </ehcache:default-copiers>
  </ehcache:service>

  <ehcache:service>
    <ehcache:default-serializers>
      <ehcache:serializer type="com.terracotta.qa.Type.Description">com.terracotta.qa.ehcache.serializers.CustomSerializerGeneral</ehcache:serializer>
      <ehcache:serializer type="com.terracotta.qa.ehcache.generator.Person">com.terracotta.qa.ehcache.serializers.CustomSerializerGeneral</ehcache:serializer>
    </ehcache:default-serializers>
  </ehcache:service>


  <ehcache:service>
    <jsr107:defaults>
      <jsr107:cache name="jcacheKeyValueCopier" template="cacheKeyValueCopierTemplate"/>
    </jsr107:defaults>
  </ehcache:service>

  <ehcache:cache-template name="cacheKeyValueCopierTemplate">
    <ehcache:key-type>com.terracotta.qa.Type.Description</ehcache:key-type>
    <ehcache:value-type>com.terracotta.qa.ehcache.generator.Person</ehcache:value-type>
    <ehcache:heap size="10"/>
  </ehcache:cache-template>
</ehcache:config>
````

Test
````
CachingProvider provider = Caching.getCachingProvider();
    CacheManager cacheManager = provider.getCacheManager(getClass().getResource("/jsr107/DefaultKeyValueCopierTest.xml").toURI(), getClass().getClassLoader());
    Cache<Description, Person> cache = cacheManager.createCache("jcacheKeyValueCopier", createMutableConfiguration(Description.class, Person.class));
cacheManager.close();
````

Console output:
````
[...]

PR #643, opened 2015-10-21
Issue #611 Fix JSR-107 + template + default copiers

ljacomet:
The ConfigurationMerger was not taking into account the CacheManager
level settings, which can contain default copier configurations.

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
+++ b/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
@@ -20,11 +20,16 @@ import org.ehcache.config.CacheConfiguration;
 import org.ehcache.config.CacheConfigurationBuilder;
 import org.ehcache.config.copy.CopierConfiguration;
 import org.ehcache.config.copy.DefaultCopierConfiguration;
+import org.ehcache.config.copy.DefaultCopyProviderConfiguration;
 import org.ehcache.config.loaderwriter.DefaultCacheLoaderWriterConfiguration;
 import org.ehcache.config.xml.XmlConfiguration;
+import org.ehcache.internal.classes.ClassInstanceConfiguration;
 import org.ehcache.internal.copy.IdentityCopier;
 import org.ehcache.internal.copy.SerializingCopier;
+import org.ehcache.spi.copy.Copier;
+import org.ehcache.spi.copy.CopyProvider;
 import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -39,6 +44,7 @@ import javax.cache.integration.CacheLoader;
 import javax.cache.integration.CacheWriter;
 
 import static org.ehcache.config.CacheConfigurationBuilder.newCacheConfigurationBuilder;
+import static org.ehcache.spi.ServiceLocator.findSingletonAmongst;
 
 /**
  * ConfigurationMerger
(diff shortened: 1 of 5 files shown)


107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -128,6 +133,21 @@
     LOG.info("No user configured copiers found. Falling back to JSR-107 defaults.");
```
albinsuresh: That log message needs to be moved

ljacomet: Indeed

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -123,15 +128,39 @@
...
   }
 
-  private <K, V> CacheConfigurationBuilder<K, V> handleStoreByValue(Eh107CompleteConfiguration<K, V> jsr107Configuration, CacheConfigurationBuilder<K, V> builder) {
+  private <K, V> CacheConfigurationBuilder<K, V> handleStoreByValue(Eh107CompleteConfiguration<K, V> jsr107Configuration, CacheConfigurationBuilder<K, V> builder, String cacheName) {
     DefaultCopierConfiguration copierConfig = builder.getExistingServiceConfiguration(DefaultCopierConfiguration.class);
-    LOG.info("No user configured copiers found. Falling back to JSR-107 defaults.");
     if(copierConfig == null) {
       if(jsr107Configuration.isStoreByValue()) {
+        if (xmlConfiguration != null) {
+          for (ServiceCreationConfiguration<?> serviceCreationConfiguration : xmlConfiguration.getServiceCreationConfigurations()) {
```
chrisdennis: findSingletonAmongst?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -123,15 +128,39 @@
...
+            if (serviceCreationConfiguration instanceof DefaultCopyProviderConfiguration) {
+              Map<Class<?>, ClassInstanceConfiguration<Copier<?>>> defaults = ((DefaultCopyProviderConfiguration) serviceCreationConfiguration)
+                  .getDefaults();
+              boolean matchingDefault = false;
+              if (!defaults.containsKey(jsr107Configuration.getKeyType())) {
```
chrisdennis: I would have inverted this if block.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -123,15 +128,39 @@
...
+              } else {
+                matchingDefault = true;
+              }
+              if (!defaults.containsKey(jsr107Configuration.getValueType())) {
+                builder = builder.add(new DefaultCopierConfiguration<K>((Class) SerializingCopier.class, CopierConfiguration.Type.VALUE));
+              } else {
+                matchingDefault = true;
+              }
+              if (matchingDefault) {
+                LOG.info("CacheManager level copier configuration overwriting JSR-107 by-value semantics for cache {}", cacheName);
```
chrisdennis: Do we care about the distinction between the key and value copier here?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Indeed

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: indeed

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: I do not think so

PR #651, opened 2015-10-23
Issue #350 Final Review

chrisdennis:


chrisdennis:
This currently requires a build of offheap-store from this branch: https://github.com/chrisdennis/offheap-store/tree/thread_abstraction  This is why the CI build failed.

core/src/main/java/org/ehcache/config/writebehind/DefaultWriteBehindConfiguration.java
```diff
@@ -25,94 +25,57 @@
...
 
-  public void setWriteBehindMaxQueueSize(int writeBehindMaxQueueSize) {
+  public void setMaxQueueSize(int writeBehindMaxQueueSize) {
     if(writeBehindMaxQueueSize < 1) {
       throw new IllegalArgumentException("WriteBehind queue size cannot be less than 1.");
     }
     this.writeBehindMaxQueueSize = writeBehindMaxQueueSize;
   }
 
+  public void setExecutorAlias(String executorAlias) {
```
ljacomet: Configuration classes should be immutable. While I know this was not the case before, I believe it is worth making it so.

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -334,21 +334,20 @@ public void close() {
...
+        decorator = loaderWriter;
+      } else {
+        final WriteBehindProvider factory = serviceLocator.getService(WriteBehindProvider.class);
+        decorator = factory.createWriteBehindLoaderWriter(loaderWriter, writeBehindConfiguration);
         if(decorator != null) {
           lifeCycledList.add(new LifeCycledAdapter() {
             @Override
             public void close() {
-              factory.releaseWriteBehindDecoratorCacheLoaderWriter(decorator);
+              factory.releaseWriteBehindLoaderWriter(decorator);
```
ljacomet: This cannot work with regards to lifecycle. The write behind stuff needs to be closed before the loader writer part, so the registration needs to be for loader writer first and write behind second.

ljacomet:
Looks good to me, I would go ahead and make the necessary release of offheap-store and if build passes, good to merge.

chrisdennis:
I have created a PR for Terracotta-OSS/offheap-store#21 that needs merging (and releasing) before this PR can be merged.

Issue #650, opened 2015-10-23
Reduce history size of collected stats based on a timestamp since last call


mathieucarbou:
Review ManagementRegistry's collectStatistics method to:

- add a parameter (since) to reduce history size of samples based on a last called timestamp passed by the client to the api
- change the parameters to a DSL or builder: because much easier, enables to add further parameters easily (such as since Timestamp above), and remove duplicated methods 
- fix generic returns of the the collectStatistics method to remove the <T> type

mathieucarbou:
merged!

PR #654, opened 2015-10-27
Issue #650 - Refactored way of collecting statistics

mathieucarbou:
- DSL and more type-safe way to query statistics
- updated to management-model 1.0.0 (types added and renamed)
- Added support to query statistic history since a date
- Fixed generic signatures in management module
- Removing unused “Setting” statistic for cache manager name and cache name
- Default window average set to 1 minute
- Removed Statistic Manager from EhcacheManager

Thanks your suggestions and help @lorban and @chrisdennis :+1:


management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
```diff
@@ -0,0 +1,302 @@
...
+  static class StatisticQueryBuilder implements StatisticQuery, StatisticQuery.Builder {
+
+    private final CapabilityManagement capabilityManagement;
+    private final String capabilityName;
+    private final Collection<String> statisticNames;
+    private final List<Map<String, String>> contexts;
+    private final long since;
+
+    StatisticQueryBuilder(CapabilityManagement capabilityManagement, String capabilityName, Collection<String> statisticNames) {
+      this(capabilityManagement, capabilityName, statisticNames, Collections.<Map<String, String>>emptyList(), 0);
```
lorban: While correct, 0 is a bit confusing here. I think Long.MIN_VALUE would be a better choice.

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
```diff
@@ -0,0 +1,302 @@
...
+  public CallQuery.Builder call(String methodName, String[] argClassNames, Object[] args) {
+    return new CallQueryBuilder(capabilityManagement, capabilityName, methodName, argClassNames, args);
+  }
+
+  @Override
+  public CallQuery.Builder call(String methodName) {
+    return call(methodName, new String[0], new Object[0]);
+  }
+
+  static class StatisticQueryBuilder implements StatisticQuery, StatisticQuery.Builder {
```
lorban: Weird construct: the query builder also is a query?

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    @Override
+    public StatisticQuery.Builder on(Map<String, String> context) {
+      if (!contexts.contains(context)) {
+        List<Map<String, String>> contexts = new ArrayList<Map<String, String>>(this.contexts);
+        contexts.add(Collections.unmodifiableMap(context));
+        return new StatisticQueryBuilder(capabilityManagement, capabilityName, statisticNames, contexts, since);
+      }
+      return this;
```
lorban: This doesn't follow the Ehcache builder pattern which states that builders are immutable.

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    @Override
+    public Builder on(List<? extends Map<String, String>> contexts) {
+      StatisticQuery.Builder newBuilder = this;
```
lorban: Same as above.

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
```diff
@@ -0,0 +1,302 @@
...
+      for (Map<String, String> context : contexts) {
+        newBuilder = newBuilder.on(context);
+      }
+      return newBuilder;
+    }
+
+    @Override
+    public StatisticQuery.Builder since(long unixTimestampMs) {
+      return new StatisticQueryBuilder(capabilityManagement, capabilityName, statisticNames, contexts, unixTimestampMs);
```
lorban: Not very consistent with the remarks on the builder methods above.

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
```diff
@@ -0,0 +1,302 @@
...
+        }
+        contextualStatistics.put(context, new ContextualStatistics(context, statistics));
+      }
+
+      return new DefaultResultSet<ContextualStatistics>(contextualStatistics);
+    }
+
+  }
+
+  static class CallQueryBuilder implements CallQuery, CallQuery.Builder {
```
lorban: Same remarks as for StatisticQueryBuilder.

management/src/test/java/org/ehcache/docs/ManagementTest.java
```diff
@@ -71,9 +72,16 @@ public void usingManagementRegistry() throws Exception {
 
     Map<String, String> context = createContext(managementRegistry); // <5>
 
-    Collection<Counter> counters = managementRegistry.collectStatistics(context, "StatisticsCapability", "GetCounter"); // <6>
+    Collection<Statistic<?, ?>> counters = managementRegistry.withCapability("StatisticsCapability") // <6>
```
lorban: I don't find the following DSL to be particularly readable. build().execute().get(0).getStatistics() gives me a felling that I have no idea what kind of objects are being returned in between calls.

management/src/test/java/org/ehcache/management/registry/DefaultManagementRegistryTest.java
```diff
@@ -85,7 +97,7 @@ public void testCanGetCapabilities() {
     assertThat(new ArrayList<Capability>(managementRegistry.getCapabilities()).get(1).getName(), equalTo("StatisticsCapability"));
 
     assertThat(new ArrayList<Capability>(managementRegistry.getCapabilities()).get(0).getDescriptions(), hasSize(4));
-    assertThat(new ArrayList<Capability>(managementRegistry.getCapabilities()).get(1).getDescriptions(), hasSize(16));
+    assertThat(new ArrayList<Capability>(managementRegistry.getCapabilities()).get(1).getDescriptions(), hasSize(15));
```
lorban: What made this change from 16 to 15?

management/src/test/java/org/ehcache/management/registry/DefaultManagementRegistryTest.java
```diff
@@ -121,18 +133,185 @@ public void testCanGetStats() {
...
+
+    cacheManager1.getCache("aCache1", Long.class, String.class).put(1L, "1");
+    cacheManager1.getCache("aCache1", Long.class, String.class).put(2L, "2");
+    cacheManager1.getCache("aCache1", Long.class, String.class).put(2L, "2");
+
+    do {
+      Thread.sleep(100);
+      statistics = builder.build().execute().get(context);
+      putCount = statistics.getStatistic(0, CounterHistory.class);
+    } while (putCount.getValue().size() < 1);
```
lorban: There should be a "safety exit" from this loop, in case a bug makes it loop forever. For readability, this should also be extracted into a "assert*" method.

management/src/main/java/org/ehcache/management/StatisticQuery.java
```diff
@@ -0,0 +1,53 @@
...
+  long getSince();
+
+  interface Builder extends QueryBuilder<Builder, StatisticQuery> {
+
+    /**
+     * The query will only collect statistics computed since this time.
+     * This allows for example to filter out the list of returned values within some {@link org.terracotta.statistics.Statistic}
+     * such as {@link org.terracotta.management.stats.sampled.SampledRate}
+     *
+     * @param unixTimestampMs The unix timestamp
```
lorban: Simply known as "timestamp".

management/src/main/java/org/ehcache/management/providers/ManagementProvider.java
```diff
@@ -81,9 +81,10 @@
    *
    * @param context the context.
    * @param statisticNames the statistic names to collect.
+   * @param since The unix time in ms from where to return the statistics for statistics based on samples.
    * @return the statistic values.
    */
-  <T extends Statistic<?>> Collection<T> collectStatistics(Map<String, String> context, String... statisticNames);
+  List<Statistic<?, ?>> collectStatistics(Map<String, String> context, Collection<String> statisticNames, long since);
```
lorban: This one is a bit ugly. Since you're accepting a Collection of statisticNames, I fail to see why you're returning a List of statistic instead of a collection: the ordering cannot be guaranteed unless you pass a List.

It would have been preferable IMHO to move "long since" as the 2nd parameter and keep the String... statisticNames intact.

management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatistics.java
```diff
@@ -101,7 +99,7 @@
   }
 
   @SuppressWarnings("unchecked")
-  public <T extends Statistic<?>> Collection<T> queryStatistic(String statisticName) {
+  public List<? extends Statistic<?, ?>> queryStatistic(String statisticName, long since) {
```
lorban: Why a List instead of a Collection?

lorban:
As a general note, there are too many logical changes conflated into a single commit here. It would be *much* better to split concerns into multiple commits. That would at least tremendously help reviewers and later code maintenance.

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
mathieucarbou: Since it is a timestamp, we can set System.currentTimeMillis() ? It would work because at the time of creation, the stat framework is not yet started so we won't miss some stats.

management/src/main/java/org/ehcache/management/registry/DefaultCapabilityBasedQuery.java
mathieucarbou: Just the implementation, not the interfaces. To avoid repeating code and classes. Do you see that it could cause an issue somewhere ? calls to the builder methods always create a new instance so the previous one is not change.

management/src/test/java/org/ehcache/management/registry/DefaultManagementRegistryTest.java
mathieucarbou: "Setting" statistic has been removed :-)

Issue #635, opened 2015-10-12
Allow configuring instances for services extending ClassInstanceProvider


lorban:
Currently, services extending ClassInstanceProvider like DefaultCacheEventListenerProvider and DefaultCacheLoaderWriterProvider only accept a class name in their configuration.

Shouldn't we allow configuring either a class name or an instance? That would make the configuration a bit more flexible.

BTW, DefaultCopyProvider already accepts an instance as that was required by the XAStore.

ljacomet:
I am in favour of generalising this.

Others?

albinsuresh:
Inorder to generalise this `ClassInstanceConfiguration` will be modified to hold the instance provided by the user. `ClassInstanceProvider.newInstance` also will have to be modified to return the user provided instance if one exists, else do the instantation as it is currently. With these changes the only requirement at the service configuration levels is to have an extra constructor that takes in an actual service instance.

This would work for copiers, loader-writers and event listeners but not for serialisers because the serialisation provider does not extend `ClassInstanceProvider` provider anymore. @chrisdennis did the refactoring of `DefaultSerializationProvider` recently and he must have had a good enough reason to do that. The goal can be achieved even with this exceptional case of serialisation provider by handling it separately. But I'm just attempting a refactoring of `DefaultSerializationProvider ` as well to make it extend `ClassInstanceProvider` so that everything is consistent. If anyone has any knowledge on why that refactoring was not done earlier and why doing it now wouldn't work either , please share.

albinsuresh:
With the changes done for this task, the user will be able to configure the services as follows:

```java
CacheConfigurationBuilder.newCacheConfigurationBuilder()
        .add(new DefaultCopierConfiguration<Description>(new DescriptionCopier(),
            CopierConfiguration.Type.KEY))
```

in addition to the current way of doing this:

```java
CacheConfigurationBuilder.newCacheConfigurationBuilder()
        .add(new DefaultCopierConfiguration<Description>(DescriptionCopier.class,
            CopierConfiguration.Type.KEY))
```

Even though the former is more intuitive for a regular java user, the amount of code that he has to write still remains somewhat same. Having to create a service configuration explicitly and sticking in an actual service instance in it is still bit of an overkill. How-about having specialised methods at the `CacheConfigurationBuilder` like _withKeyCopier_, _withValueSerializer_ etc so that it is much more intuitive to the user? Internally we can still get the job done with the service configurations. But the user facing API would look much cleaner this way. 

[...]

PR #655, opened 2015-10-29
Issue #635 Configure services at the cache level using actual service…

albinsuresh:
… instances

### DIFF
--- a/docs/src/docs/asciidoc/user/getting-started.adoc
+++ b/docs/src/docs/asciidoc/user/getting-started.adoc
@@ -293,7 +293,7 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
 <4> The `PersonCopier` configured at the cache manager level will be used for this cache.
 
 The samples above demonstrated the configuration of different services at the cache and the cache manager levels.
-In none of the samples did we really create instances of services such as copiers, serializers etc.
+In none of the samples did we really create instances of copiers, serializers etc.
 Instead we supplied the class types with the service configurations and the instantiation was done internally by Ehcache.
 Even though that is the recommended way of configuring services, it is still possible to configure a service with actual
 instances.
@@ -312,6 +312,12 @@ WARNING: The former way of configuring services using class types is still the r
 the service instances are not shared between caches. Sharing services can have disastrous effects in your application if
 the user is not careful enough. The latter way of configuring services using instances opens the door for sharing the
 same service instance across caches but that feature should be used with extra care.
+
+NOTE: By now you are familiar with configuring different services using actual instances or using class types of
+loader-writers, serializers etc. It is desirable to have some cleanup performed on these instances once you are done
+using the cache that uses these instances. You can achieve this by making your custom class types `Closeable` and
+keeping your resource release logic in the _close_ method. Ehcache will make sure that this _close_ will be invoked when
+the cache is removed from the cache manager or when the cache manager is closed. 
 
 === It wouldn't be Java without _some_ XML
 
--- a/docs/src/docs/asciidoc/user/getting-started.adoc
+++ b/docs/src/docs/asciidoc/user/getting-started.adoc
@@ -292,6 +292,27 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
 <3> The `DescriptionCopier` and `PersonCopier` configured at the cache manager level will be used for this cache.
 <4> The `PersonCopier` configured at the cache manager level will be used for this cache.
 
+The samples above demonstrated the configuration of different services at the cache and the cache manager levels.
+In none of the samples did we really create instances of services such as copiers, serializers etc.
+Instead we supplied the class types with the service configurations and the instantiation was done internally by Ehcache.
+Even though that is the recommended way of configuring services, it is still possible to configure a service with actual
+instances.
+
+The following sample demonstrates the same for copier service. Other services can also be configured in a similar way:
+
+[source,java,indent=0]
+----
+include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag=cacheServiceConfigurations]
+----
+
+<1> Configuring the `DescriptionCopier` copier using class type(the old way). 
+<2> Configuring the copier using an actual instance: The `PersonCopier` instance here.
+
+WARNING: The former way of configuring services using class types is still the recommended way as that ensures that
+the service instances are not shared between caches. Sharing services can have disastrous effects in your application if
+the user is not careful enough. The latter way of configuring services using instances opens the door for sharing the
+same service instance across caches but that feature should be used with extra care.
+
 === It wouldn't be Java without _some_ XML
 
 You can create a XML file to configure a `CacheManager`:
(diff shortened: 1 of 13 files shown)


impl/src/main/java/org/ehcache/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -158,12 +158,22 @@ private AbstractProvider(Map<Class<?>, Class<? extends Serializer<?>>> serialize
     @Override
     public <T> Serializer<T> createKeySerializer(Class<T> clazz, ClassLoader classLoader, ServiceConfiguration<?>... configs) throws UnsupportedTypeException {
       DefaultSerializerConfiguration<T> conf = find(SerializerConfiguration.Type.KEY, configs);
+      if(conf != null) {
```
albinsuresh: This had to be done since serialisation provider classes does not extend `ClassInstanceProvider` anymore. So this workaround was required. If I manage to refactor `DefaultSerializationProvider` to extend `ClassInstanceProvider` again, this special handling won't be required.

albinsuresh: Changes done in the `ClassInstanceProvider` class for issue 432 is preventing me from doing this refactoring. I have three choices here:

1. Leave this code as it is as an exceptional case
2. Revert some of the changes done in 432 (the changes removing `ConstructorArguments parameters from the _newInstance()_ methods in `ClassInstanceProvider` class)
3. Add a setter to `ClassInstanceConfiguration` that can be used to modify the *arguments* field. Using that setter I'll be able to inject the constructor args that I want for instantiation.

I personally prefer the third one. Please let me know if there are any objections/comments/feedback.

ljacomet: 3 is not an option, we want configuration to be immutable.

ljacomet:
Going down the path you are taking - and which was started at the CopyProvider - as an unfortunate consequence: it means the configuration builder cannot be reused, as it contains real state.
Could you investigate if we would have options to pass in these instances at a different place in the setup flow?

albinsuresh:
How about adding a new service that provides the service instances? The instance configuration can be extracted out to a new set of configurations. Like a `ServiceInstanceConfiguration`

```java
public interface ServiceInstanceConfiguration<T> {

  /**
   * Indicates the type of service this configuration holds.
   *
   * @return the service type
   */
  Class<T> getServiceType();
}
```

and specialised instance configurations for different services like the following for copiers:

```java
public interface CopierInstanceConfiguration extends ServiceInstanceConfiguration<Copier> {
  
  public Copier<?> getKeyCopier();

  public Copier<?> getValueCopier();
}
```

The `CacheConfigurationBuilder`'s _buildConfig_ needs to be altered to take in these service instance configurations as follows:

```java
public <CK extends K, CV extends V> CacheConfiguration<CK, CV> buildConfig(Class<CK> keyType,
  Class<CV> valueType,
  ServiceInstanceConfiguration<?>... instanceConfigurations) {
```

Users will have to use concrete instance configuration classes to register the instances  and pass those while building the cache configuration as follows:

```java
    DefaultCopierInstanceConfiguration instanceConfiguration = 
        new DefaultCopierInstanceConfiguration(new DescriptionCopier(), new PersonCopier());
    CacheConfiguration<Long, Person> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder()
        .add(new DefaultCopierConfiguration<Person>((Class)SerializingCopier.class,   //<1>
            CopierConfiguration.Type.VALUE))
        .buildConfig(Long.class, Person.class, instanceConfiguration);
```

Here is the concrete reference implementation of `ServiceInstanceConfiguration` for copiers:

```java
public class DefaultCopierInstanceConfiguration implements CopierInstanceConfiguration {
  
  public final Copier<?> keyCopier;
  public final Copier<?> valueCopier;
  
  public DefaultCopierInstanceConfiguration(Copier<?> keyCopier, Copier<?> valueCopier) {
    this.keyCopier = keyCopier;
    this.valueCopier = valueCopier;
  }
  
  @Override
  public Copier<?> getKeyCopier() {
    return null;
  }

  @Override
  public Copier<?> getValueCopier() {
    return null;
  }

  @Override
  public Class<Copier> getServiceType() {
    return Copier.class;
  }
}
```

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -293,6 +293,27 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
 <3> The `DescriptionCopier` and `PersonCopier` configured at the cache manager level will be used for this cache.
 <4> The `PersonCopier` configured at the cache manager level will be used for this cache.
 
+The samples above demonstrated the configuration of different services at the cache and the cache manager levels.
+In none of the samples did we really create instances of services such as copiers, serializers etc.
```
ljacomet: copiers and serializers are not _services_

impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
```diff
@@ -84,25 +84,44 @@ private T newInstance(K alias, ClassInstanceConfiguration<T> config) {
...
+      providedVsCount.put(instance, new AtomicInteger(1));
+    } else {
+      currentCount.incrementAndGet();
     }
+    return instance;
   }
 
   protected void releaseInstance(T instance) throws IOException {
-    if(!created.remove(instance)) {
+    if(providedVsCount.containsKey(instance)) {
```
ljacomet: What's the purpose of this `providedVsCount`?
How come you can close even when not removing?

docs/src/docs/asciidoc/user/getting-started.adoc
albinsuresh: Yes. Yes. I'll correct that. It'd have really helped if there was some common term for these provided things.

impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
albinsuresh: For the instances that are actually *created* by this class the count is never gonna be more than 1. And those instances when released will be removed in the if-block below. Now for the user provided instances the count can be more than 1 if the user reuses the same instance across caches. So here we are doing reference counting to handle that case and again the if block below would handle removing the last instance.

ljacomet: I believe it makes no sense to even try to do reference counting - for no added value as there is no conditional behaviour linked to it.
However, this discussion made me realise that we want the `created` type to be looking for instance equality not object equality, so you probably need to wrap a `IdentityHashMap` as a `Set` as your type there. And as your usage of an `AtomicInteger` indicates, the data structure must be thread safe and so needs even more wrapping to be synchronised - as this is good enough for lifecycle related methods.

docs/src/docs/asciidoc/user/getting-started.adoc
ljacomet: I would use something like _companion objects_  or _dependencies_ and define what these are.

impl/src/main/java/org/ehcache/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -55,9 +55,9 @@
 
   private final TransientProvider transientProvider;
   private final PersistentProvider persistentProvider;
-  
-  final Set<Serializer<?>> created = new HashSet<Serializer<?>>();
-  
+
+  protected final Map<Serializer<?>, AtomicInteger> providedVsCount = new HashMap<Serializer<?>, AtomicInteger>();
```
ljacomet: Like for the `CalssInstanceProvider` this needs to become safe for a concurrent environment.

Issue #659, opened 2015-11-01
Execute on build related feedback


ljacomet:
During JavaOne 2015, I spent some time on the Gradle booth discussing the Ehcache 3 build.

This led to interesting feedback and this issue is to track build improvements based on said feedback.

Issue #644, opened 2015-10-21
Clean up / complete pom metadata


ljacomet:
* Developers / contributors

ljacomet:
Regarding developers / contributors, let's replace what we have with a generic entry indicating the work is done by Terracotta developers.

PR #660, opened 2015-11-01
Issue #659 JavaOne Gradle build feedback

ljacomet:
* Bump to Gradle 2.8 + enable parallel build in daemon
* Apply maven plugin on modules subset limited to those needing an install / deploy
* Clean up and refactor scripts


ljacomet:
test this please

buildSrc/src/main/groovy/scripts/Utils.groovy
```diff
@@ -0,0 +1,129 @@
...
+        developerConnection = 'scm:git:git@github.com:ehcache/ehcache3.git'
+      }
+      licenses {
+        license {
+          name = 'The Apache Software License, Version 2.0'
+          url = 'http://www.apache.org/licenses/LICENSE-2.0.txt'
+          distribution = 'repo'
+        }
+      }
+      developers {
```
albinsuresh: Can't we do the issue #644 cleanup here itself?

ljacomet: Well, we may ... but since the changes contained here seem to trigger a test issue, I would rather have it happen in a separate commit that can be merged sooner.

ljacomet:
Added the fix for #644 in this PR

Issue #647, opened 2015-10-22
Core Ehcache Services


chrisdennis:
Would it be better if we didn't force the core ehcache service configurations to have to be nested inside service tags?  As an example would it better if we traded this:
```xml
<ehcache:config>
  <ehcache:service>
    <ehcache:default-copiers>
      <ehcache:copier type="com.pany.ehcache.copier.Description">com.pany.ehcache.copier.DescriptionCopier</ehcache:copier>
      <ehcache:copier type="com.pany.ehcache.copier.Person">com.pany.ehcache.copier.PersonCopier</ehcache:copier>
    </ehcache:default-copiers>
  </ehcache:service>

  <ehcache:cache alias="bar">
    <ehcache:key-type>com.pany.ehcache.copier.Description</ehcache:key-type>
    <ehcache:value-type>com.pany.ehcache.copier.Person</ehcache:value-type>
    <ehcache:heap size="10"/>
  </ehcache:cache>
</ehcache:config>
```
for this:
```xml
<ehcache:config>
  <ehcache:default-copiers>
    <ehcache:copier type="com.pany.ehcache.copier.Description">com.pany.ehcache.copier.DescriptionCopier</ehcache:copier>
    <ehcache:copier type="com.pany.ehcache.copier.Person">com.pany.ehcache.copier.PersonCopier</ehcache:copier>
  </ehcache:default-copiers>

  <ehcache:cache alias="bar">
    <ehcache:key-type>com.pany.ehcache.copier.Description</ehcache:key-type>
    <ehcache:value-type>com.pany.ehcache.copier.Person</ehcache:value-type>
    <ehcache:heap size="10"/>
  </ehcache:cache>
</ehcache:config>
```

ljacomet:
I am in favour of such a change.

Others?

PR #661, opened 2015-11-02
Issue #647 : Moving ehcache core services out of service tag

rishabhmonga:


xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -171,34 +171,34 @@ private void parseConfiguration()
...
-      if (serviceType.getDefaultSerializers() != null) {
-        DefaultSerializationProviderConfiguration configuration = new DefaultSerializationProviderConfiguration();
-
-        for (SerializerType.Serializer serializer : serviceType.getDefaultSerializers().getSerializer()) {
-          try {
-            configuration.addSerializerFor(getClassForName(serializer.getType(), classLoader), (Class) getClassForName(serializer.getValue(), classLoader));
-          } catch (ClassNotFoundException e) {
-            throw new RuntimeException(e);
-          }
+      final ServiceCreationConfiguration<?> serviceConfiguration1 = configurationParser.parseExtension((Element)serviceType.getAny());
```
ljacomet: Can this be inlined or the name of the var changed to not have that `1` in it?

docs/src/docs/asciidoc/user/xml.adoc
```diff
@@ -35,6 +35,21 @@ These `Service` instances can then be used by `Cache` instances managed by the `
 extension point of the XML configuration (and Ehcache 3's modular architecture), as explained in the
 link:107{outfilesuffix}#_supplement_jsr_107_s_configurations[JSR-107 configuration section].
 
+=== `<default-serializers>` element
+
+A `<default-serializers>` element represents `Serializers` configured at `CacheManager` level.
```
ljacomet: Update this to reflect it contains a collection of elements doing what you explain

docs/src/docs/asciidoc/user/xml.adoc
```diff
@@ -35,6 +35,21 @@ These `Service` instances can then be used by `Cache` instances managed by the `
...
+Requires serializer `type` and a fully qualified class name of the `Serializer`.   
+
+=== `<default-copiers>` element
+
+A `<default-copiers>` element represents `Copiers` configured at `CacheManager` level.
+Requires copier `type` and a fully qualified class name of the `Copier`.
+
+=== `<persistence>` element
+
+A `<persistence>` element represents `Persistence`, to be used in case of persistent `Cache` instances.
```
ljacomet: Double check and document this actually creates a `PersistentCacheManager`

docs/src/docs/asciidoc/user/xml.adoc
```diff
@@ -35,6 +35,21 @@ These `Service` instances can then be used by `Cache` instances managed by the `
 extension point of the XML configuration (and Ehcache 3's modular architecture), as explained in the
 link:107{outfilesuffix}#_supplement_jsr_107_s_configurations[JSR-107 configuration section].
 
+=== `<default-serializers>` element
+
+A `<default-serializers>` element represents a collection of `Serializers` configured at `CacheManager` level. 
+Requires serializer `type` and a fully qualified class name of the `Serializer`.
```
ljacomet: Prefix with "Eache serializer ..." to indicate that the requires applies to the elements of the collection. Applies to copiers next

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -29,6 +29,27 @@
           </xs:documentation>
         </xs:annotation>
       </xs:element>
+      <xs:element name="default-serializers" type="ehcache:serializer-type" minOccurs="0" maxOccurs="1">
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            Type and fully qualified class name of serializers that are to be registered
```
ljacomet: Indicate here too that "this is a collection of elements containing ..."

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -29,6 +29,27 @@
...
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            Type and fully qualified class name of serializers that are to be registered 
+          </xs:documentation>
+        </xs:annotation>
+      </xs:element>
+      <xs:element name="default-copiers" type="ehcache:copier-type" minOccurs="0" maxOccurs="1">
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            Type and fully qualified class name of copiers that are to be registered
```
ljacomet: Same as above

ljacomet:
Documentation still needs work here. My previous comments were about adding the information that the high level element (with an s) is a collection of elements that have specific requirements.
It contained a type (each spelled as _eache_) that has been interpreted by adding _ehcache_ in front of the description - but it still makes no sense.
So to be very specific:
> A `<default-serializers>` element represents a collection of `Serializers` configured at `CacheManager` level. 
> Ehcache `Serializer` requires serializer `type` and a fully qualified class name of the `Serializer`.

needs to become:
> A `<default-serializers>` element represents `Serializers` configured at `CacheManager` level. 
> It is a collection of `serializer` tags that require a `type` and the fully qualified class name of the `Serializer`.

And the same inside the XSD.

To be applied to the other documentation that relates to a collection of elements as well.

PR #664, opened 2015-11-04
Introduce a resource releaser method for SerializationProvider as a precursor to the real fix for #635

albinsuresh:
1. The responsibility of closing a persistent serializer has been moved to the SerializationProvider via the releaseSerializer method. This way we achieve consistency in handling the life cycle of the provided objects.
2. Serializers are not Closeable anymore. So caches or rather the serialization provider won't be closing the serializers when they are closed.

### DIFF
--- a/impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
+++ b/impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
@@ -20,10 +20,14 @@ import org.ehcache.config.CacheConfiguration;
 import org.ehcache.spi.ServiceProvider;
 import org.ehcache.spi.service.ServiceConfiguration;
 
+import java.io.Closeable;
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
 
 import static org.ehcache.internal.classes.commonslang.reflect.ConstructorUtils.invokeConstructor;
 
@@ -36,6 +40,11 @@ public class ClassInstanceProvider<K, T> {
    * The order in which entries are put in is kept.
    */
   protected final Map<K, ClassInstanceConfiguration<T>> preconfigured = Collections.synchronizedMap(new LinkedHashMap<K, ClassInstanceConfiguration<T>>());
+
+  /**
+   * Instances created by this provider
+   */
+  protected final Set<T> created = new HashSet<T>();
 
   private final Class<? extends ClassInstanceConfiguration<T>> cacheLevelConfig;
 
@@ -76,7 +85,9 @@ public class ClassInstanceProvider<K, T> {
       }
     }
     try {
-      return invokeConstructor(config.getClazz(), config.getArguments());
+      T instance = invokeConstructor(config.getClazz(), config.getArguments());
+      created.add(instance);
+      return instance;
     } catch (InstantiationException e) {
       throw new RuntimeException(e);
     } catch (IllegalAccessException e) {
@@ -88,6 +99,20 @@ public class ClassInstanceProvider<K, T> {
     }
   }
 
+  protected void releaseInstance(T instance) {
+    if(!created.contains(instance)) {
+      throw new IllegalArgumentException("Given instance of " + instance.getClass().getName() + " is not managed by this provider");
+    }
+    if(instance instanceof Closeable) {
+      try {
+        ((Closeable)instance).close();
+        created.remove(instance);
+      } catch (IOException e) {
+        throw new RuntimeException("Failed to close the instance of " + instance.getClass().getName(), e);
+      }
+    }
+  }
+  
   public void start(ServiceProvider serviceProvider) {
     // default no-op
   }
--- a/impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
+++ b/impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
@@ -99,17 +99,12 @@ public class ClassInstanceProvider<K, T> {
     }
   }
 
-  protected void releaseInstance(T instance) {
-    if(!created.contains(instance)) {
+  protected void releaseInstance(T instance) throws IOException {
+    if(!created.remove(instance)) {
       throw new IllegalArgumentException("Given instance of " + instance.getClass().getName() + " is not managed by this provider");
     }
     if(instance instanceof Closeable) {
-      try {
-        ((Closeable)instance).close();
-        created.remove(instance);
-      } catch (IOException e) {
-        throw new RuntimeException("Failed to close the instance of " + instance.getClass().getName(), e);
-      }
+      ((Closeable)instance).close();
     }
   }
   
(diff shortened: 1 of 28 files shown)


impl/src/main/java/org/ehcache/internal/serialization/CompactJavaSerializer.java
```diff
@@ -150,12 +150,6 @@ protected int getOrAddMapping(ObjectStreamClass desc) throws IOException {
     }
   }
 
-  @Override
-  public void close() {
```
ljacomet: I believe this logic should remain here, and be called when `SerializationProvider.releaseSerializer` is called.
This could be achieved by having our default serialisers implementations still implement `Closeable` or some other interface.

impl/src/main/java/org/ehcache/internal/serialization/CompactPersistentJavaSerializer.java
```diff
@@ -51,13 +48,8 @@ public CompactPersistentJavaSerializer(ClassLoader classLoader, FileBasedPersist
     }
   }
 
-  @Override
-  public final void close() throws IOException {
```
ljacomet: Same as other comment, the full logic should remain, just called from `SerializationProvider.releaseSerializer` now.

impl/src/main/java/org/ehcache/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -63,23 +69,43 @@ public DefaultSerializationProvider(DefaultSerializationProviderConfiguration co
...
+    created.add(serializer);
+    return serializer;
   }
 
   @Override
+  public void releaseSerializer(final Serializer<?> serializer) {
+    if (!created.remove(serializer)) {
+      throw new IllegalArgumentException("Given serializer: " + serializer.getClass().getName() + " is not managed by this provider");
+    }
+    if(serializer instanceof CompactPersistentJavaSerializer) {
```
ljacomet: As stated in my previous comments, we want to invoke the former `close` method on all our Serializer implies. Same `Closeable` interface or new custom one?

albinsuresh: Having one more instances check for CompactJavaSerializer isn't enough? Reusing `Closable` has the side effect of ehcache closing any custom Serializer implementation that is closable. The next option is to have a new interface just for the sake of these 2 implementations. Is it really worth compared to having an additional instanceof check?

ljacomet: Agreed with not using a known type for this, however I believe this is a useful customisation seam where serialisers implementers can hook up to have their serialiser be "closed" properly.

albinsuresh:
Check-in incorporating review feedback.

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
```diff
@@ -838,7 +842,12 @@ public void releaseStore(Store<?, ?> resource) {
         // release the underlying store first, as it may still need the serializer to flush down to lower tiers
         underlyingStoreProvider.releaseStore(xaStore.underlyingStore);
         try {
-          helper.softLockSerializerRef.getAndSet(null).close();
+          Serializer<?> serializer = helper.softLockSerializerRef.getAndSet(null);
```
ljacomet: There is some weird code around type here - the `Serializer` held by the `AtomicReference` is always of a  single type that does not need to be closed.
I recommend changing the reference to be of the more specific type `SoftLockSerializer` and thus dropping the if block after it because that other guy does not need to be `Closeable`.

impl/src/main/java/org/ehcache/internal/classes/ClassInstanceProvider.java
```diff
@@ -88,6 +99,20 @@ private T newInstance(K alias, ClassInstanceConfiguration<T> config) {
...
   }
 
+  protected void releaseInstance(T instance) {
+    if(!created.contains(instance)) {
+      throw new IllegalArgumentException("Given instance of " + instance.getClass().getName() + " is not managed by this provider");
+    }
+    if(instance instanceof Closeable) {
+      try {
+        ((Closeable)instance).close();
+        created.remove(instance);
```
ljacomet: I suggest inverting these two calls, so that the instance no longer is in the created collection.

impl/src/main/java/org/ehcache/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -63,23 +70,44 @@ public DefaultSerializationProvider(DefaultSerializationProviderConfiguration co
...
 
   @Override
+  public void releaseSerializer(final Serializer<?> serializer) {
+    if (!created.contains(serializer)) {
+      throw new IllegalArgumentException("Given serializer: " + serializer.getClass().getName() + " is not managed by this provider");
+    }
+    if(serializer instanceof Closeable) {
+      try {
+        ((Closeable)serializer).close();
+        created.remove(serializer);
```
ljacomet: Same inversion as for the `ClassInstanceProvider`

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
```diff
@@ -808,7 +812,7 @@ public Duration getExpiryForUpdate(K key, SoftLock oldSoftLock, SoftLock newSoft
         AtomicReference<Serializer<SoftLock<V>>> softLockSerializerRef = new AtomicReference<Serializer<SoftLock<V>>>();
         SoftLockValueCombinedSerializer softLockValueCombinedSerializer = new SoftLockValueCombinedSerializer<V>(softLockSerializerRef, storeConfig.getValueSerializer());
 
-        // create the underlying store
+        // create the underlying store
```
ljacomet: Remove the trailing white space

transactions/src/main/java/org/ehcache/transactions/xa/XAStore.java
```diff
@@ -838,7 +842,10 @@ public void releaseStore(Store<?, ?> resource) {
         // release the underlying store first, as it may still need the serializer to flush down to lower tiers
         underlyingStoreProvider.releaseStore(xaStore.underlyingStore);
         try {
-          helper.softLockSerializerRef.getAndSet(null).close();
+          Serializer<SoftLock> serializer = helper.softLockSerializerRef.getAndSet(null);
```
ljacomet: Like I said, the type here can be restricted to `SoftLockSerializer` by changing the type of the AtomicReference and once done, you know it is not `Closeable` so the code below is useless.

ljacomet:
test this please

impl/src/main/java/org/ehcache/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -63,23 +70,39 @@ public DefaultSerializationProvider(DefaultSerializationProviderConfiguration co
 
   @Override
   public <T> Serializer<T> createKeySerializer(Class<T> clazz, ClassLoader classLoader, ServiceConfiguration<?>... configs) throws UnsupportedTypeException {
+    Serializer<T> serializer;
     if (findSingletonAmongst(PersistenceSpaceIdentifier.class, (Object[]) configs) == null) {
-      return transientProvider.createKeySerializer(clazz, classLoader, configs);
+      serializer = transientProvider.createKeySerializer(clazz, classLoader, configs);
```
ljacomet: trailing white space

Issue #133, opened 2014-11-05
Handling exceptions from Expiry methods


timeck:
At the moment unchecked exceptions from the methods `org.ehcache.expiry.Expiry.getExpiryForAccess(K, V)` and `org.ehcache.expiry.Expiry.getExpiryForUpdate(K, V)` are thrown back to the caller and the cache operation in which the `Expiry` call was part of will not mutate the cache. 

There are at least two other options proposed:
(1) Catch exceptions and use a default (either Duration.ZERO or Duration.FOREVER). This is similar to 107 behavior for exceptions in `ExpiryPolicy`
(2) Give the exception to a `ResilienceStrategy` to deal with as it sees fit. It's not clear what to do with the cache operation or what Duration it should apply if it is allowed to succeed

rishabhmonga:
Reiterating the proposals above :
(1) We either conform to JSR107 and handle the exception we receive and in turn revert with a `Duration.ZERO` or `Duration.FOREVER`. This allows the cache operation to succeed.
(2) We fail the operation, destroying the mapping and throw the exception using `CachePassThroughException`.

ljacomet:
Discussion with @chrisdennis and @rishabhmonga: we go for option 2 and if doing so in the JSR-107 bridge still passes the TCK, then we apply the same logic there.

rishabhmonga:
As per the discussion, we will implement the option 2 by logging the exception received at error log level and returning a `Duration.ZERO`, essentially leading to a removal of the mapping.

PR #665, opened 2015-11-05
Issue #133 : Fixing exceptions from Expiry methods

rishabhmonga:


ljacomet:
This PR does not build on the `CachePassThroughException`, have a chat with @albinsuresh and if it cannot fit please explain why.
Also, the solution description includes destroying the mapping which will not happen here AFAICT.

core-spi-test/src/main/java/org/ehcache/internal/tier/CachingTierClear.java
```diff
@@ -103,7 +103,9 @@ public Object apply(final Object o) {
           }
         });
 
-        assertThat(newReturnedValueHolder.value(), is(equalTo(newValueHolder.value())));
+        if (newReturnedValueHolder != null) {
```
ljacomet: No, this makes no sense with regards to the changes done here, same applies to the two other changed tests. None of these changes belong here.
Find the root cause of the failure and fix that instead.

core/src/main/java/org/ehcache/Ehcache.java
```diff
@@ -285,9 +285,19 @@ private boolean newValueAlreadyExpired(K key, V oldValue, V newValue) {
     
     final Duration duration;
     if (oldValue == null) {
-      duration = runtimeConfiguration.getExpiry().getExpiryForCreation(key, newValue);
+      try {
+        duration = runtimeConfiguration.getExpiry().getExpiryForCreation(key, newValue);
+      } catch (RuntimeException re) {
+        logger.error("Expiry caused an exception ", re);
```
ljacomet: Clarify message: _Expiry computation caused an exception - Expiry duration will be 0_

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -336,7 +339,10 @@ public void remove(final K key) throws CacheAccessException {
           }
 
           returnValue.set(mappedValue);
-          setAccessTimeAndExpiry(key, mappedValue, now);
+          if (setAccessTimeAndExpiry(key, mappedValue, now) == null) {
```
ljacomet: Quite a number of the changes here are the same. Refactor into a new method like `setAccessTimeAndExpiryThenReturnMapping` which clarifies intent and includes expiration notification when needed.
Then you can have the few places where it happens differently still invoke a refactored `setAccessTimeAndExpiry`.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -546,8 +558,11 @@ public boolean hasNext() {
         
         final Map.Entry<K, OnHeapValueHolder<V>> thisEntry = next;
         advance();
-    
-        setAccessTimeAndExpiry(thisEntry.getKey(), thisEntry.getValue(), timeSource.getTimeMillis());
+
+        if (setAccessTimeAndExpiry(thisEntry.getKey(), thisEntry.getValue(), timeSource.getTimeMillis()) == null) {
+          onExpiration(thisEntry.getKey(), thisEntry.getValue());
+          advance();
```
ljacomet: No reason to advance here - please review the iterator logic to understand it

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -546,8 +558,11 @@ public boolean hasNext() {
         
         final Map.Entry<K, OnHeapValueHolder<V>> thisEntry = next;
         advance();
-    
-        setAccessTimeAndExpiry(thisEntry.getKey(), thisEntry.getValue(), timeSource.getTimeMillis());
+
+        if (setAccessTimeAndExpiry(thisEntry.getKey(), thisEntry.getValue(), timeSource.getTimeMillis()) == null) {
+          onExpiration(thisEntry.getKey(), thisEntry.getValue());
```
ljacomet: `onExpiration` cannot be called while not under lock, and nothing here will remove the mapping. Find another way of solving this.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -644,7 +666,13 @@ public float getHitRate(TimeUnit unit) {
           return null;
         }
         // TODO find a way to increment hit count on a fault
-        setAccessTimeAndExpiry(key, cachedValue, now);
+        if (setAccessTimeAndExpiry(key, cachedValue, now) == null) {
+          if (backEnd.remove(key, cachedValue)) {
+            onExpiration(key, cachedValue);
```
ljacomet: `onExpiration` must be called under key lock

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -603,7 +618,14 @@ public float getHitRate(TimeUnit unit) {
             ValueHolder<V> value = fault.get();
             final OnHeapValueHolder<V> newValue;
             if(value != null) {
-              newValue = importValueFromLowerTier(key, value, now);
+              try {
+                newValue = importValueFromLowerTier(key, value, now);
```
ljacomet: In case we fail here, we ideally should be able to remove the mapping from the authority. Can you try to find a way for doing this?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1119,8 +1156,16 @@ public void disableStoreEventNotifications() {
     this.eventListener = CacheEvents.nullStoreEventListener();
   }
   
-  private void setAccessTimeAndExpiry(K key, OnHeapValueHolder<V> valueHolder, long now) {
-    valueHolder.accessed(now, expiry.getExpiryForAccess(key, valueHolder.value()));
+  private OnHeapValueHolder<V> setAccessTimeAndExpiry(K key, OnHeapValueHolder<V> valueHolder, long now) {
```
ljacomet: I suggest a rename to indicate new contract

impl/src/main/java/org/ehcache/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -141,7 +144,10 @@ public AbstractOffHeapStore(String statisticsTag, Configuration<K, V> config, Ti
           }
 
           if (updateAccess) {
-            setAccessTimeAndExpiry(mappedKey, mappedValue, now);
+            if (setAccessTimeAndExpiry(mappedKey, mappedValue, now) == null) {
```
ljacomet: Same remark as `OnHeapStore`

impl/src/test/java/org/ehcache/internal/store/tiering/CompoundCachingTierSPITest.java
```diff
@@ -72,7 +76,7 @@ public void setUp() {
 
       private CachingTier<String, String> newCachingTier(Long capacity) {
         Store.Configuration<String, String> config = new StoreConfigurationImpl<String, String>(getKeyType(), getValueType(), null, null,
-                ClassLoader.getSystemClassLoader(), null, buildResourcePools(capacity), new JavaSerializer<String>(getSystemClassLoader()), new JavaSerializer<String>(getSystemClassLoader()));
+                ClassLoader.getSystemClassLoader(), expiry, buildResourcePools(capacity), new JavaSerializer<String>(getSystemClassLoader()), new JavaSerializer<String>(getSystemClassLoader()));
```
ljacomet: You can simply use `Expirations.noExpiration()` here

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -220,7 +220,10 @@ public OnHeapStore(final Configuration<K, V> config, TimeSource timeSource, Copi
           }
 
           if (updateAccess) {
-            setAccessTimeAndExpiry(key, mappedValue, now);
+            if (setAccessTimeAndExpiryThenReturnMapping(key, mappedValue, now) == null) {
```
ljacomet: Still believe this whole block can be hidden inside a method. Also your first version had many more locations where equivalent code was called, but you changed to a version that no longer notifies.
I believe the initial version was more _correct_  just needed some refactoring to extract duplication.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -644,7 +667,13 @@ public float getHitRate(TimeUnit unit) {
           return null;
         }
         // TODO find a way to increment hit count on a fault
-        setAccessTimeAndExpiry(key, cachedValue, now);
+        if (setAccessTimeAndExpiryThenReturnMapping(key, cachedValue, now) == null) {
+          if (backEnd.remove(key, cachedValue)) {
+            onExpiration(key, cachedValue);
```
ljacomet: Again, you need to have _exclusivity_ on the key to make that call.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -603,7 +618,15 @@ public float getHitRate(TimeUnit unit) {
             ValueHolder<V> value = fault.get();
             final OnHeapValueHolder<V> newValue;
             if(value != null) {
-              newValue = importValueFromLowerTier(key, value, now);
+              try {
+                newValue = importValueFromLowerTier(key, value, now);
+              } catch (RuntimeException re) {
+                LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
+                backEnd.remove(key, fault);
+                onExpiration(key, value);
```
ljacomet: This call can only be performed under _exclusive_ key access.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1118,9 +1126,36 @@ public void enableStoreEventNotifications(StoreEventListener<K, V> listener) {
...
+    }
+    valueHolder.accessed(now, duration);
+    return valueHolder;
+  }
+
+  private void expireMapping(final K key, final ValueHolder<V> value) {
+    map.computeIfPresent(key, new BiFunction<K, OnHeapValueHolder<V>, OnHeapValueHolder<V>>() {
+      @Override
+      public OnHeapValueHolder<V> apply(K mappedKey, final OnHeapValueHolder<V> mappedValue) {
+        if(mappedValue.equals(key)) {
```
ljacomet: Pretty sure that's not the equal check you want!

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -603,7 +602,15 @@ public float getHitRate(TimeUnit unit) {
             ValueHolder<V> value = fault.get();
             final OnHeapValueHolder<V> newValue;
             if(value != null) {
-              newValue = importValueFromLowerTier(key, value, now);
+              try {
+                newValue = importValueFromLowerTier(key, value, now);
+              } catch (RuntimeException re) {
+                LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
+                backEnd.remove(key, fault);
+                expireMapping(key, value);
```
ljacomet: Unless another thread installs something, this will have no effect as you just removed the mapping ...

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -644,7 +651,11 @@ public float getHitRate(TimeUnit unit) {
           return null;
         }
         // TODO find a way to increment hit count on a fault
-        setAccessTimeAndExpiry(key, cachedValue, now);
+        if (setAccessTimeAndExpiryThenReturnMapping(key, cachedValue, now, true) == null) {
+          backEnd.remove(key, cachedValue);
```
ljacomet: Why? is that not covered by the method you call one line above?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -546,8 +543,10 @@ public boolean hasNext() {
         
         final Map.Entry<K, OnHeapValueHolder<V>> thisEntry = next;
         advance();
-    
-        setAccessTimeAndExpiry(thisEntry.getKey(), thisEntry.getValue(), timeSource.getTimeMillis());
+
+        if (setAccessTimeAndExpiryThenReturnMapping(thisEntry.getKey(), thisEntry.getValue(), timeSource.getTimeMillis(), true) == null) {
+          return null;
```
ljacomet: You cannot return null, this breaks the iterator contract. We indeed clean the mapping from the store in case of failure, but we still return it.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -600,10 +598,28 @@ public float getHitRate(TimeUnit unit) {
...
-              newValue = importValueFromLowerTier(key, value, now);
+              try {
+                newValue = importValueFromLowerTier(key, value, now);
+              } catch (RuntimeException re) {
+                LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
+                map.computeIfPresent(key, new BiFunction<K, OnHeapValueHolder<V>, OnHeapValueHolder<V>>() {
+                  @Override
+                  public OnHeapValueHolder<V> apply(K mappedKey, final OnHeapValueHolder<V> mappedValue) {
+                    if(mappedValue.equals(fault)) {
+                      final AbstractValueHolder<V> valueHolder = (AbstractValueHolder<V>)value;
```
ljacomet: I do not like the blind cast. I suggest doing something like:
 - check if expired already - if yes, call into onExpiration directly
 - if not check if `instanceof AbstractValueHolder` and if yes update expiration time
 - in case it is not the type we expect, do nothing as we have no way of sending expiry info to authority

ljacomet:
Replaced by PR #678

Issue #667, opened 2015-11-05
Handle exceptions from EvictionVeto and EvictionPrioritizer properly


ljacomet:
Need to make sure we behave as decided:
* `EvictionPriotizer` exception means we evict in whatever order we sampled
* `EvictionVeto` exception means no veto is applied

PR #668, opened 2015-11-06
Issue #667 Handling exceptions from EvictionVeto and Prioritizer

ljacomet:


albinsuresh:
Just a suggestion not really related to this issue. Looks like we are ending up with a lot of anonymous `Store.Configuration` classes in different test classes(twice each in `OnHeapStoreByRefTest` and `OnHeapStoreByValueTest`). If we would extract a generic one out, then different tests can just override the config that is relevant for the test and reuse the rest. That would really reduce some amount of duplicated code.

ljacomet:
Both remarks have been addressed:
* Replaced the weird byte array creation with one using random
* I was lazy and introduced much duplication in store creation - changed that to be cleaner

PR #669, opened 2015-11-12
Extracted some management types to terracotta-management module

mathieucarbou:
@anthonydahanne, @lorban : here is the PR for extracting types


management/src/main/java/org/ehcache/management/QueryBuilder.java
```diff
@@ -15,9 +15,9 @@
  */
 package org.ehcache.management;
 
+import org.terracotta.management.context.Context;
+
 import java.util.Collection;
-import java.util.List;
-import java.util.Map;
```
anthonydahanne: hum... that was not used before ? (unused import ?)

management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatisticsProvider.java
```diff
@@ -24,13 +23,12 @@
...
 import org.terracotta.management.capabilities.descriptors.Descriptor;
+import org.terracotta.management.context.Context;
 import org.terracotta.management.stats.Statistic;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
```
anthonydahanne: those types were not used ?

anthonydahanne: apparently not (just checked)

management/src/main/java/org/ehcache/management/QueryBuilder.java
anthonydahanne: apparently not (just checked)

anthonydahanne:
LGTM , thanks for fixing the unused imports

cschanck:
At Anthony's request...

PR #687, opened 2015-11-27
Fixing the write-behind queue size calculation of batched write-behind

albinsuresh:


impl/src/test/java/org/ehcache/loaderwriter/writebehind/AbstractWriteBehindTestBase.java
```diff
@@ -592,4 +594,40 @@ public void testAgedBatchedIsWritten() throws Exception {
...
+    }
+
+    TestWriteBehindProvider writeBehindProvider = new TestWriteBehindProvider();
+    WriteBehindTestLoaderWriter<String, String> loaderWriter = new WriteBehindTestLoaderWriter<String, String>();
+
+    CacheManager cacheManager = managerBuilder().using(writeBehindProvider).build(true);
+    try {
+      Cache<String, String> testCache = cacheManager.createCache("testAgedBatchedIsWritten", configurationBuilder()
+          .add(new DefaultCacheLoaderWriterConfiguration(loaderWriter))
+          .add(newBatchedWriteBehindConfiguration(5, SECONDS, 2).build())
```
ljacomet: This test will run for 5 seconds+ if left like that, do complete the batch after your assertion to make it finish faster

impl/src/main/java/org/ehcache/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -200,7 +203,7 @@ public void run() {
             latest.remove(op.getKey(), op);
           }
         } finally {
-          LOGGER.info("Cancelling Batch " + this);
+          LOGGER.debug("Batch processing completed");
```
ljacomet: The statement here is to inform that we are cancelling the scheduled batch expiry task, while it needed a level change and clarification, your change no longer reflects this.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -152,9 +153,11 @@ private Batch newBatch() {
     return executor.submit(batch);
   }
   
+  @SuppressFBWarnings("IS2_INCONSISTENT_SYNC")
   @Override
   public long getQueueSize() {
-    return executorQueue.size() * batchSize;
+    return executorQueue.size() * batchSize   // This multiplication with batchSize is still not accurate as the batches are not always full
+           + (openBatch == null ? 0 : openBatch.size());
```
ljacomet: As discussed in chat, the openBatch reference could change between the two accesses to it, better create a local var for it.

albinsuresh:
Addressed the review feedback.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -152,9 +153,12 @@ private Batch newBatch() {
     return executor.submit(batch);
   }
   
+  @SuppressFBWarnings("IS2_INCONSISTENT_SYNC")
```
ljacomet: Making the `openBatch` volatile makes this warning goes away - I believe that's the solution rather than the ignore.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -154,7 +154,9 @@ private Batch newBatch() {
   
   @Override
   public long getQueueSize() {
-    return executorQueue.size() * batchSize;
+    Batch snapshot = openBatch;
+    return executorQueue.size() * batchSize   // This multiplication with batchSize is still not accurate as the batches are not always full
```
ljacomet: Move the comment to javadoc please

albinsuresh:
Addressed the review feedback.

impl/src/main/java/org/ehcache/loaderwriter/writebehind/WriteBehind.java
```diff
@@ -42,7 +42,9 @@
   
   /**
    * Gets the best estimate for items in the queue still awaiting processing.
-   * Not including elements currently processed
+   * Not including elements currently processed.
+   * Since the value returned is a rough estimate, it can sometimes be more than
```
ljacomet: So that's interesting - I was expecting you to put the javadoc on the specific implementation. But I realise this level also contains some wording that make the batching impl wrong.
So I suggest:
* Move the paragraph about the size being bigger down to the batching impl.
* Remove from here the bit about not including elements currently processed.

albinsuresh:
Updated the javadocs as suggested.

Issue #690, opened 2015-12-02
Remove eviction prioritize feature and compute eviction veto state on insert/update


ljacomet:
While it sounded like a good idea to have customisable eviction priorities, it turns out the be a performance problem that will be hard to address with the feature as it currently exists.

So in this context, we are dropping the feature to not block performance work later.

Regarding eviction veto, we are aligning the behaviour across tiers to a computation of the veto on insertion / update of the mapping.

lorban:
There's a nice optimization/rationalization we could do regarding the veto state: if it was available from the ValueHolder interface, we could simply copy it across when mappings are being faulted to a caching tier instead of calling the veto predicate during each fault.

This would also simplify the veto calling code a bit as the only place it would be called is during initial insertion/update, ie: only ever triggered by a user call and never by the internal plumbing.

ljacomet:
I can see usefulness in computing the veto again when a value moves between tiers. When it happens, internal stats of the Cache.Entry have changed and so the veto may end up with a different value.

PR #692, opened 2015-12-03
Issue 690

lorban:
- remove eviction prioritizer feature
 - make on-heap only execute veto predicate during insertion/update and not during eviction

Note: there's an extra optimization/rationalization we could do: with this patch each store has its own veto/no-veto state so when a mapping is faulted from the authority to the caching tier, the veto has to be reprocessed. Maybe we could make veto a common ValueHolder property so that its state is simply copied when mappings are copied to the caching tier.


impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1227,27 +1235,31 @@ private void expireMapping(final K key, final ValueHolder<V> value) {
...
-      return new CopiedOnHeapValueHolder<V>(valueHolder, realValue, valueCopier, now, expiration);
+      return new CopiedOnHeapValueHolder<V>(valueHolder, realValue, veto, valueCopier, now, expiration);
     }
   }
 
-  private OnHeapValueHolder<V> makeValue(V value, long creationTime, long expirationTime, Copier<V> valueCopier) {
-    if(valueCopier instanceof SerializingCopier) {
-      return makeSerializedValue(value, creationTime, expirationTime, ((SerializingCopier)valueCopier).getSerializer());
+  private OnHeapValueHolder<V> makeValue(K key, V value, long creationTime, long expirationTime, Copier<V> valueCopier) {
+    boolean veto = checkVeto(CacheStoreHelper.cacheEntry(key, value, creationTime, OnHeapValueHolder.TIME_UNIT));
```
ljacomet: This creates a cache entry that never has a valid hit rate, while there may be a valid one. Can't we find a way to compute the veto with ALL the information it can use internally?

ljacomet:
@lorban Can you ask your question on issue #690 ?

api/src/main/java/org/ehcache/config/Eviction.java
```diff
@@ -22,7 +22,7 @@
 import java.util.concurrent.TimeUnit;
 
 /**
- * Utility class for getting predefined {@link EvictionVeto} and {@link EvictionPrioritizer} instances.
+ * Utility class for getting predefined {@link EvictionVeto} instance.
```
chrisdennis: still needs a plural here.

impl/src/main/java/org/ehcache/internal/concurrent/ConcurrentHashMap.java
```diff
@@ -6300,12 +6302,12 @@ public final void compute() {
         }
     }
 
-    public Set<Entry<K, V>> getRandomValues(Random rndm, int size, Predicate<Entry<K, V>> veto) {
-        Set<Entry<K, V>> sampled = new HashSet<Entry<K, V>>(size);
+    public List<Entry<K, V>> getRandomValues(Random rndm, int size, Predicate<Entry<K, V>> veto) {
```
chrisdennis: It should be safe to eliminate the new entry creation in this method I think.

impl/src/main/java/org/ehcache/internal/concurrent/ConcurrentHashMap.java
```diff
@@ -6356,7 +6358,7 @@ public final void compute() {
         return sampled;
     }
     
-    private static <T> boolean add(Set<? super T> to, Predicate<? super T> veto, T value) {
+    private static <T> boolean add(List<? super T> to, Predicate<? super T> veto, T value) {
```
chrisdennis: Since a list won't ever refuse to add we can probably just eliminate this method, by inlining the veto call and assuming a true return for the add.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1282,11 +1294,16 @@ boolean evict() {
     final Random random = new Random();
 
     @SuppressWarnings("unchecked")
-    Set<Map.Entry<K, OnHeapValueHolder<V>>> values = map.getRandomValues(random, SAMPLE_SIZE, (Predicate<Map.Entry<K, OnHeapValueHolder<V>>>)evictionVeto);
+    Set<Map.Entry<K, OnHeapValueHolder<V>>> values = map.getRandomValues(random, SAMPLE_SIZE, new Predicate<OnHeapValueHolder<V>>() {
```
chrisdennis: Could we create this wrapper once only at initialization time?

chrisdennis: Also how does this compile?  Doesn't getRandomValues return a set now?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
lorban: With the currently discussed changes, there are only two reasons the veto predicate would be called:
 - during initial insertion
 - during faulting

In the 1st case, this is useless as the hit rate will always be invalid since the mapping is currently being inserted. Regarding the 2nd case, we could add the extra complexity but I'd rather follow the route of my note (making the veto flag common across all tiers) which would stop calling the veto predicate during faulting.

impl/src/main/java/org/ehcache/internal/concurrent/ConcurrentHashMap.java
lorban: Are you suggesting to directly return the Node instances? Can't these change under your feet, making you potentially evict a candidate whose value has changed since you chose it?

api/src/main/java/org/ehcache/config/Eviction.java
lorban: thank you

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
lorban: Sure thing.

impl/src/main/java/org/ehcache/internal/concurrent/ConcurrentHashMap.java
lorban: I'm not sure that would make the code any more readable.

ljacomet:
After discussions with @lorban and @chrisdennis we decided to merge this but keep working on the issue to improve things further.

PR #700, opened 2015-12-07
Issue 690 Internalize more of the eviction logic inside CHM

chrisdennis:


impl/src/main/java/org/ehcache/internal/concurrent/ConcurrentHashMap.java
```diff
@@ -6302,63 +6299,71 @@ public final void compute() {
...
+
+        return getEvictionCandidateWrap(tab, start, size, maxKey, maxValue, prioritizer, veto);
     }
-    
-    private static <T> boolean add(List<? super T> to, Predicate<? super T> veto, T value) {
-      return !veto.test(value) && to.add(value);
+
+    private Entry<K, V> getEvictionCandidateWrap(Node<K,V>[] tab, int start, int size, K maxKey, V maxVal, Comparator<? super V> prioritizer, EvictionVeto<? super K, ? super V> veto) {
+        Traverser<K, V> t = new Traverser<K, V>(tab, tab.length, 0, start);
+        for (Node<K, V> p; (p = t.advance()) != null;) {
```
ljacomet: This iteration logic is duplicated. Can't you extract it?

PR #725, opened 2016-01-06
Issue #165 - intermediate review

ljacomet:
This is partial work for issue #165 on which I want an initial design review.

I am mostly interested in the logic to handle events inside the store, so please have a look at commits 9ad1b89 and 944debc mostly.

This is not to be merged.


core/src/main/java/org/ehcache/CacheConfigurationProperty.java
```diff
@@ -16,12 +16,19 @@
...
 package org.ehcache;
 
-/**
- * @author rism
- */
 public enum CacheConfigurationProperty {
   /**
    * used to update size of ResourcePool in RuntimeConfig
    */
-  UPDATESIZE
```
lorban: Underscores would make those names more readable IMHO.

ljacomet:
Closing this as we had multiple discussions and the current implementation is actually evolving a lot, including changing what I wanted to be reviewed here.

Issue #718, opened 2016-01-06
Ehcache ServerEntityService Implementation (Active Service Only)


chrisdennis:
This includes using the voltron configuration area to store the “identity” of the CacheManager entity - I think this can just be a UUID for now.

PR #730, opened 2016-01-12
Issue #718 Add server side entity (and enough client side logic to integration test things)

chrisdennis:


chrisdennis:
Test this please

ljacomet:
I believe the build errors are caused because there is no reference to the terracotta snapshot repository in the gradle config.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheServerEntityService.java
```diff
@@ -0,0 +1,46 @@
...
+public class EhcacheServerEntityService implements ServerEntityService<EhcacheActiveEntity, PassiveServerEntity<EntityMessage, EntityResponse>>{
+
+  @Override
+  public long getVersion() {
+    return 0L;
+  }
+
+  @Override
+  public boolean handlesEntityType(String typeName) {
+    return true;
```
ljacomet: Really? Don't we need something concrete here?

PR #731, opened 2016-01-19
Issue 616

AbfrmBlr:
OSGI tests failing


api/src/main/java/org/ehcache/spi/sizeof/SizeOfEngine.java
```diff
@@ -0,0 +1,60 @@
...
+   * @return size of the value along with the offset
+   */
+  long sizeofValue(Object value);
+  
+  /**
+   * Bytes required on adding each key-value pair
+   * 
+   * @return size overhead of adding a key-value pair
+   */
+  long getCHMOffset();
```
lorban: There must be a better way to abstract that mechanism. We should meet around this.

impl/src/main/java/org/ehcache/internal/sizeof/DefaultSizeOfEngineProviderConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+package org.ehcache.internal.sizeof;
+
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+import org.ehcache.spi.sizeof.SizeOfEngineProvider;
+
+/**
+ * @author Abhilash
+ *
+ */
+public class DefaultSizeOfEngineProviderConfiguration implements ServiceCreationConfiguration<SizeOfEngineProvider> {
```
lorban: I wonder if introducing a SizeOfEngineProviderConfiguration interface wouldn't make things a bit more readable.

impl/src/main/java/org/ehcache/internal/sizeof/DefaultSizeOfEngineConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+    return SizeOfEngineProvider.class;
+  }
+
+  @Override
+  public Long getMaxDepth() {
+    return this.maxDepth;
+  }
+
+  @Override
+  public Long getMaxSize() {
```
lorban: Why Long instead of long?

impl/src/main/java/org/ehcache/internal/sizeof/DefaultSizeOfEngineConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+    this.maxSize = maxSize;
+  }
+  
+  @Override
+  public Class<SizeOfEngineProvider> getServiceType() {
+    return SizeOfEngineProvider.class;
+  }
+
+  @Override
+  public Long getMaxDepth() {
```
lorban: I think int would be enough for max depth. We may also want to rename that field since it's not about graph depth anymore.

impl/src/main/java/org/ehcache/internal/sizeof/DefaultSizeOfEngineProviderFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * @author Abhilash
+ *
+ */
+public class DefaultSizeOfEngineProviderFactory implements ServiceFactory<SizeOfEngineProvider> {
+
+  @Override
+  public SizeOfEngineProvider create(ServiceCreationConfiguration<SizeOfEngineProvider> configuration) {
+    long maxDepth = Long.MAX_VALUE;
```
lorban: I wonder if that's the best default value. IIRC ehcache 2 used to use 100 as the default, but I'm open to any suggestion over this.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -137,7 +147,12 @@ public void cacheConfigurationChange(CacheConfigurationChangeEvent event) {
         if(updatedPools.getPoolForResource(ResourceType.Core.HEAP).getSize() !=
            configuredPools.getPoolForResource(ResourceType.Core.HEAP).getSize()) {
           LOG.info("Setting size: " + updatedPools.getPoolForResource(ResourceType.Core.HEAP).getSize());
-          capacity = updatedPools.getPoolForResource(ResourceType.Core.HEAP).getSize();
+          ResourcePool pool = updatedPools.getPoolForResource(ResourceType.Core.HEAP);
+          if (pool.getUnit() instanceof MemoryUnit) {
+            capacity = ((MemoryUnit)pool.getUnit()).toBytes(pool.getSize());
```
lorban: I think the size has to be converted to bytes.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -243,6 +265,9 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
       });
       if (result == null) {
         getObserver.end(StoreOperationOutcomes.GetOutcome.MISS);
+        if (byteSized && expiredValue.get() != null) {
+          decrementCurrentUsageInBytesIfRequired(getSizeOfKeyWithOffset(key) + getSizeOfValueWithOffset(expiredValue.get().value()) + sizeOfEngine.getCHMOffset());
```
lorban: I don't think the size should be recalculated when the mapping gets evicted/removed/expired. This should be stored somewhere like in the ValueHolder.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -120,6 +124,12 @@ public int compare(ValueHolder<?> t, ValueHolder<?> u) {
   private final Class<V> valueType;
   private final Copier<K> keyCopier;
   private final Copier<V> valueCopier;
+  
+  private final SizeOfEngine sizeOfEngine;
+  private final long faultOffset;
+  
+  private final boolean byteSized;
```
lorban: I'm not fond of that variable and all those if (byteSized) {} blocks. Couldn't we inject a no-op sizeof engine that always return 0 as the size when a store is configured with max entries?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -292,9 +329,13 @@ public void put(final K key, final V value) throws CacheAccessException {
 
       if (entryActuallyAdded.get()) {
         putObserver.end(StoreOperationOutcomes.PutOutcome.PUT);
-        enforceCapacity(1);
+        enforceCapacity(delta);
```
lorban: That delta estimate isn't a bad idea, but it shouldn't be recalculated each time and you should loop around enforceCapacity() until enough bytes have been freed to make sure the store's consumption goes below the configured max bytes.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -492,6 +551,8 @@ public boolean replace(final K key, final V oldValue, final V newValue) throws C
   public void clear() throws CacheAccessException {
     try {
       map.clear();
+      //TODO: Is this thread safe ?
+      currentUsageinBytes.set(0l);
```
lorban: It's not, but I wonder if the alternatives wouldn't be worse? We could iterate over all mappings and remove them one by one but I'm slightly worried about the performance impact.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -120,6 +124,12 @@ public int compare(ValueHolder<?> t, ValueHolder<?> u) {
   private final Class<V> valueType;
   private final Copier<K> keyCopier;
   private final Copier<V> valueCopier;
+  
+  private final SizeOfEngine sizeOfEngine;
+  private final long faultOffset;
```
lorban: What do you mean with "offset"? Shouldn't that be called "faultSizeInBytes"?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1283,14 +1413,70 @@ private static long safeExpireTime(long now, Duration duration) {
...
+  private void decrementCurrentUsageInBytesIfRequired(long delta) {
+    if(byteSized) {
+      while(true) {
+        long expect =  currentUsageinBytes.get();
+        long update = expect - delta;
+        if(update < 0l) {
+          //Either this happened due to race 
+          //or the size estimation screwed up somewhere
+          //come out silently or throw ?
+          throw new AssertionError("It's screwed");
```
lorban: I don't think we'll want to keep this assertion.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
AbfrmBlr: toBytes should convert to bytes I guess

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
AbfrmBlr: Cool idea, storing in valueholder will help in reducing the sizeof overhead, but add a little bit to the storage which should be fine.

AbfrmBlr:
test this please

ljacomet:
test this please

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -120,6 +126,14 @@ public int compare(ValueHolder<?> t, ValueHolder<?> u) {
   private final Class<V> valueType;
   private final Copier<K> keyCopier;
   private final Copier<V> valueCopier;
+  
+  private final SizeOfEngine sizeOfEngine;
+  private final long faultSizeInBytes;
```
ljacomet: Do we really need all these fields here? Can't they be encapsulated somewhere else?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -271,30 +298,40 @@ public void put(final K key, final V value) throws CacheAccessException {
 
     final AtomicBoolean entryActuallyAdded = new AtomicBoolean();
     final long now = timeSource.getTimeMillis();
-
+    
+    final AtomicLong replacedValueSize = new AtomicLong(0l);
```
ljacomet: Please no lower case L for longs

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -271,30 +298,40 @@ public void put(final K key, final V value) throws CacheAccessException {
...
+          OnHeapValueHolder<V> toReturn = null;
           if (mappedValue == null) {
-            return newCreateValueHolder(key, value, now);
+            toReturn = newCreateValueHolder(key, value, now);
           } else {
-            return newUpdateValueHolder(key, mappedValue, value, now);
+            replacedValueSize.set(sizeOfEngine.sizeof(mappedValue.value()));// get it only for mapped value
+            toReturn = newUpdateValueHolder(key, mappedValue, value, now);
           }
+          entryActuallyAdded.set(toReturn != null);
```
ljacomet: This completely breaks the put statistics as _replaced_ will no longer be triggered. And thus I believe it breaks size accounting as well since it depends on the same test.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -271,30 +298,40 @@ public void put(final K key, final V value) throws CacheAccessException {
...
         putObserver.end(StoreOperationOutcomes.PutOutcome.PUT);
-        enforceCapacity(1);
+        long delta = 1;
+        if (byteSized) { 
+          delta = valuePut.size();
+        }
+        enforceCapacity(delta);
       } else {
         putObserver.end(StoreOperationOutcomes.PutOutcome.REPLACED);
+        long replacedDelta = sizeOfEngine.sizeof(value) - replacedValueSize.get();
```
ljacomet: Between the call to sizeof here and the one above on line 316, sizing of a single value happens multiple times. That does not seem like a good thing to do.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1283,14 +1386,70 @@ private static long safeExpireTime(long now, Duration duration) {
...
+        if(currentUsageinBytes.compareAndSet(expect, update)) {
+          break;
+        }
+      }
+    }
+  }
+  
+  private void decrementCurrentUsageInBytesIfRequired(long delta) {
+    if(byteSized) {
+      while(true) {
```
ljacomet: Replacing with a do ... while would clarify the loop

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1283,14 +1386,70 @@ private static long safeExpireTime(long now, Duration duration) {
...
+  
+  protected long getCurrentUsageInBytes() {
+    if (byteSized) {
+      return currentUsageinBytes.get();
+    }
+    return 0l;
+  }
+  
+  protected void enforceCapacity(long delta) {
+    if(byteSized) {
```
ljacomet: This is wrong. We need to keep the logic of eviction doing bounded work, even if it means that the cache stays over size for a while. Here you have no bounds on the number of elements to evict. Also, once this logic is applied, then capacity enforcement might need to happen not only after the size grows.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1283,14 +1386,70 @@ private static long safeExpireTime(long now, Duration duration) {
...
-            && capacity < map.size(); attempts++) {
+  private void enforceByteCapacity() {
+    while (capacity < currentUsageinBytes.get()) {
+      evict();
+    }
+  }
+ 
+  private void incrementCurrentUsageInBytesIfRequired(long delta) {
+    if(byteSized) {
+      while(true) {
```
ljacomet: Why the loop? Why not use `AtomicLong.addAndGet`?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1388,7 +1552,7 @@ public void releaseStore(Store<?, ?> resource) {
       close(onHeapStore);
     }
 
-    static void close(final OnHeapStore onHeapStore) {
+    public static void close(final OnHeapStore onHeapStore) {
```
ljacomet: Why does this need to become public?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1374,7 +1534,11 @@ private static boolean eq(Object o1, Object o2) {
       CopyProvider copyProvider = serviceProvider.getService(CopyProvider.class);
       Copier<K> keyCopier  = copyProvider.createKeyCopier(storeConfig.getKeyType(), storeConfig.getKeySerializer(), serviceConfigs);
       Copier<V> valueCopier = copyProvider.createValueCopier(storeConfig.getValueType(), storeConfig.getValueSerializer(), serviceConfigs);
-      OnHeapStore<K, V> onHeapStore = new OnHeapStore<K, V>(storeConfig, timeSource, keyCopier, valueCopier);
+      
+      SizeOfEngineProvider sizeOfEngineProvider = serviceProvider.getService(SizeOfEngineProvider.class);
+      boolean isByteSized = storeConfig.getResourcePools().getPoolForResource(ResourceType.Core.HEAP).getUnit() instanceof MemoryUnit;
+      SizeOfEngine sizeOfEngine = sizeOfEngineProvider.createSizeOfEngine(isByteSized, serviceConfigs);
```
ljacomet: Can't we let the provider handle the above logic?

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -103,6 +103,17 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
 <4> You define a resource pool for the disk.
 
+==== Byte Sized Tiering
+
+[source,java,indent=0]
+----
+include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag=byteSizedTieredCache]
+----
+
+<1> You can also size a tiered cache based on bytes only.You can limit the amount of java heap used by cache for storing key-value pairs.
```
ljacomet: Would rephrase that:
You can also size the heap tier in bytes. This will limit the amount of heap used by that tier for storing key-value pairs. Note that there is a cost associated to sizing objects.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -103,6 +103,17 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
 
+==== Byte Sized Tiering
+
+[source,java,indent=0]
+----
+include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag=byteSizedTieredCache]
+----
+
+<1> You can also size a tiered cache based on bytes only.You can limit the amount of java heap used by cache for storing key-value pairs. 
+<2> A SizeOfEngine can also be provided which actually ascertains the size of objects on java heap. It uses sizeof library to do so. The configuration for the same takes two arguments. First one specifies the maximum
```
ljacomet: Would rephrase that:
The sizing mechanism can be configured along two axis: The first one specifies the maximum number of objects to visit while walking the object graph, the second defines the maximum size of a single object. If the sizing goes above any of these two limits, the mutative operation on the cache will be ignored.

impl/src/main/java/org/ehcache/internal/sizeof/DefaultSizeOfEngine.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.internal.sizeof;
+
+import org.ehcache.internal.copy.IdentityCopier;
```
ljacomet: Unused imports

impl/src/test/java/org/ehcache/internal/store/heap/bytesized/ByteAccountingTest.java
```diff
@@ -0,0 +1,429 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.internal.store.heap.bytesized;
+
+import static org.ehcache.config.ResourcePoolsBuilder.newResourcePoolsBuilder;
```
ljacomet: Most of the existing code places static imports at the bottom

impl/src/main/java/org/ehcache/internal/sizeof/DefaultSizeOfEngineProviderConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.internal.sizeof;
+
+import org.ehcache.sizeof.SizeOfEngineProviderConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
```
ljacomet: Unused import

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -40,6 +40,9 @@
 import org.ehcache.event.EventType;
 import org.ehcache.internal.copy.ReadWriteCopier;
 import org.ehcache.spi.copy.Copier;
+import org.ehcache.internal.copy.SerializingCopier;
```
ljacomet: Unused imports

AbfrmBlr:
Rebased

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -679,6 +722,7 @@ public float getHitRate(TimeUnit unit) {
       getOrComputeIfAbsentObserver.end(CachingTierOperationOutcomes.GetOrComputeIfAbsentOutcome.HIT);
       return getValue(cachedValue);
     } catch (RuntimeException re) {
+      LOG.error(re.getMessage(), re);
```
ljacomet: Why logging here and at such a high level?

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -689,23 +733,27 @@ public void invalidate(final K key) throws CacheAccessException {
...
-      if (removed.get()) {
+      OnHeapValueHolder<V> invalidated = null;
+      if ((invalidated = invalidatedValue.get()) != null) {
         invalidateObserver.end(CachingTierOperationOutcomes.InvalidateOutcome.REMOVED);
+        decrementCurrentUsageInBytesIfRequired(invalidated.size());
       } else {
         invalidateObserver.end(CachingTierOperationOutcomes.InvalidateOutcome.MISS);
       }
     } catch (RuntimeException re) {
+      LOG.error(re.getMessage(), re);
```
ljacomet: See other log related comment

AbfrmBlr: those remanants of debuging effort. WIll remove all those

lorban:
test this please

PR #733, opened 2016-01-27
#734 - Added gradle task (gradlew createPom) to generate module pom.xml files.

mathieucarbou:
This allows to import directly ehcache snapshot modules in our IDE by using the standard maven dependency mechanism to handle dependencies across all opened projects and modules.

Note: the task is the same. But if I put it in the parent graddle build, the generated poms to not contain the project dependencies, but only external dependencies. I didn't find a way to have it in the parent file, but the task configuration resolved only at execution in sub modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ build/
 .nb-gradle-properties
 
 /classes
+
+pom.xml
--- a/build.gradle
+++ b/build.gradle
@@ -41,6 +41,7 @@ subprojects {
   apply plugin: 'checkstyle'
   apply plugin: 'findbugs'
   apply plugin: 'jacoco'
+  apply plugin: 'maven'
 
   group = 'org.ehcache.modules'
   version = baseVersion
@@ -143,6 +144,28 @@ subprojects {
       csv.enabled false
     }
   }
+
+  afterEvaluate {
+
+    task createPom {
+      pom {
+        project {
+          groupId project.group
+          artifactId "ehcache-${project.name}"
+          version project.version
+          licenses {
+            license {
+              name 'The Apache Software License, Version 2.0'
+              url 'http://www.apache.org/licenses/LICENSE-2.0.txt'
+              distribution 'repo'
+            }
+          }
+        }
+      }.writeTo("pom.xml")
+    }
+
+  }
+
 }
 
 allprojects {


mathieucarbou:
See issue #734

mathieucarbou:
You cannot within the same IDEA window have several Maven projects opened with a graddle one.

I need to have:

1 - terracotta-management project opened 
2 - ehcache project opened (specifically the management module), which depends on 1
3 - another project, which depends on 1 and 2.

1 and 3 are maven projects, so when you import pom files within the same IDE window, the IDE is smart enough to detect inter-project dependencies. And thus you do not need to generate each time your snapshots from the command line. It's fast development!

2 is not maven. So you cannot import any maven module. and 2 needs modifications depending on 1 and 3 depends on 2. I do not want to build each time I need to see the changes some snapshots jars for 1 and then back to 3.

There is a feature in Intellij that enables the importing of several pom files, event if there are in different projects, and when pointing to snapshots, the IDE automatically create the module dependencies. 

Probably you did not used this feature before, hence your question ;-) I would need to show you because it makes development and modifications across projects (which I need to do) quite fast and less error-prone.

ljacomet:
Seriously not too fond of adding this in the project, especially when duplicated across all modules, including the ones we have no plan on publishing to a maven repository.

mathieucarbou:
I agree for the duplication, it's really dirty. But I strongly think we still need a way for maven users (which is pretty a large user base) to be able to use the ehcache project (snapshot) with their project in the same IDE.

A concrete use case would be to facilitate debugging and patching. Once a user found an issue he would be able to directly unit test it and patch it by making his maven project depending on the opened ehcache project by using the snapshot dep.

Otherwise, it's all back and forth between several windows and we are forced to build and install deps, which is not fun and long ;-) I guess if we increase the difficulty of integration, I fear users would be less prone to do the effort in sending us patches or PR. Because the project setup is much more difficult.

Concretely at the moment I generated these poms in ehcache because I have a pretty big inter-project refactor to achieve so it's less risky to have all working modules withing the same IDE window. But I can see plenty of other use cases where we could benefits of having some maven poms :-)

mathieucarbou:
@ljacomet @anthonydahanne @chrisdennis : I've found a solution to avoid the duplication of the task. The trick is to use afterEvaluate() so that the task is register with its configuration for each sub-module.

=> Now there is only one task, inherited, and configuration is resolved per sub-module. Thus, each sub-module pom contains the right dependencies to other ehcache modules.

`./gradlew createPom`

chrisdennis:
:stuck_out_tongue:

mathieucarbou:
@anthonydahanne : If you need the poms (as I will for the next days), you can generate them with:

`curl -Ls https://goo.gl/qIapwy | sh`

See: https://gist.github.com/mathieucarbou/fb791e084995bb5c12ffd817608c753d

PR #735, opened 2016-01-29
Issue #165

ljacomet:
As decided a while ago, all eventing has been pushed down at the `Store` level.

Note that this will most likely not pass on Cloudbees right now, investigating that. And I also need to do a final pass as I believe some dead code and TODOs are still hiding in there.

But the design is final and that can be reviewed.
Please have a look at `Store.getStoreEventSource()` and the types related to its return type.


api/src/main/java/org/ehcache/config/CacheRuntimeConfiguration.java
```diff
@@ -45,7 +45,7 @@
    * @throws java.lang.IllegalStateException if the listener is already registered
    */
   void registerCacheEventListener(CacheEventListener<? super K, ? super V> listener,
-                                  EventOrdering ordering, EventFiring firing, Set<EventType> forEventTypes);
+                                  EventOrdering ordering, EventFiring firing, EnumSet<EventType> forEventTypes);
```
chrisdennis: A specific Implementation in an API interface?

ljacomet: Hehe, got carried away with the `EnumSet` indeed, let me revert that back to a `Set`

core/src/main/java/org/ehcache/CacheConfigurationProperty.java
```diff
@@ -16,12 +16,19 @@
...
 
-/**
- * @author rism
- */
 public enum CacheConfigurationProperty {
   /**
    * used to update size of ResourcePool in RuntimeConfig
    */
-  UPDATESIZE
+  UPDATESIZE,
```
chrisdennis: Underscores?

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -324,7 +324,8 @@ public void close() throws Exception {
         }
       }
     }
-    Store.Configuration<K, V> storeConfiguration = new StoreConfigurationImpl<K, V>(config, keySerializer, valueSerializer);
+    Store.Configuration<K, V> storeConfiguration = new StoreConfigurationImpl<K, V>(config, keySerializer,
+        valueSerializer);
```
chrisdennis: whitespace change

core/src/main/java/org/ehcache/config/StoreConfigurationImpl.java
```diff
@@ -34,17 +34,18 @@
   private final ResourcePools resourcePools;
   private final Serializer<K> keySerializer;
   private final Serializer<V> valueSerializer;
-  
-  public StoreConfigurationImpl(CacheConfiguration<K, V> cacheConfig, Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+
+  public StoreConfigurationImpl(CacheConfiguration<K, V> cacheConfig,
+                                Serializer<K> keySerializer, Serializer<V> valueSerializer) {
```
chrisdennis: whitespace

core/src/main/java/org/ehcache/events/EventListenerWrapper.java
```diff
@@ -0,0 +1,56 @@
...
+package org.ehcache.events;
+
+import org.ehcache.event.CacheEventListener;
+import org.ehcache.event.EventFiring;
+import org.ehcache.event.EventOrdering;
+import org.ehcache.event.EventType;
+
+import java.util.EnumSet;
+
+public final class EventListenerWrapper {
```
chrisdennis: This wrapper seems to be leaking quite a bit...

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -77,12 +89,30 @@ public CacheEventDispatcherImpl(ExecutorService orderedDelivery, ExecutorService
...
+  }
+
+  private void registerCacheEventListener(EventListenerWrapper wrapper) {
+    if(aSyncListenersSet.contains(wrapper) || syncListenersSet.contains(wrapper)) {
+      throw new IllegalStateException("Cache Event Listener already registered: " + wrapper.listener);
     }
-    if (doRegister) {
-      store.enableStoreEventNotifications(storeListener);
+
+    if(wrapper.ordering == EventOrdering.ORDERED) {
```
chrisdennis: Switches?

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -77,12 +89,30 @@ public CacheEventDispatcherImpl(ExecutorService orderedDelivery, ExecutorService
...
+      if (orderedListenerCount.incrementAndGet() == 1) {
```
chrisdennis: Would getAndIncrement == 0 be more explicit?

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -95,157 +125,107 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
+      if (!removeWrapperFromSet(wrapper, syncListenersSet)) {
+        throw new IllegalStateException("Unknown cache event listener: " + listener);
+      }
     }
-    if (!hasListeners()) {
-      store.disableStoreEventNotifications();
+  }
+
+  private boolean removeWrapperFromSet(EventListenerWrapper wrapper, Set<EventListenerWrapper> listenersSet) {
+    for (EventListenerWrapper listenerWrapper : listenersSet) {
```
chrisdennis: This is either more complicated than it needs to be, or I'm missing something... why do we look checking equality and only then remove?

core/src/main/java/org/ehcache/events/OrderedEventDispatcher.java
```diff
@@ -0,0 +1,38 @@
...
+  public OrderedEventDispatcher(ExecutorService executorService) {
+    this.executorService = executorService;
+  }
+
+  @Override
+  public Future<?> dispatch(CacheEvent<K, V> cacheEvent, Iterable<EventListenerWrapper> listenerWrappers) {
+    return executorService.submit(new EventDispatchTask<K, V>(cacheEvent, listenerWrappers));
+  }
+
+  public ExecutorService getExecutorService() {
```
chrisdennis: This getter seems ugly... could we replace it with a shutdown method on this class itself?

impl/src/main/java/org/ehcache/internal/events/ScopedStoreEventDispatcher.java
```diff
@@ -0,0 +1,313 @@
...
+    if (listeners.isEmpty()) {
+      return NO_OP_EVENT_SINK;
+    } else {
+      return new InvocationScopedEventSink<K, V>(filters, ordered, orderedQueues);
+    }
+  }
+
+  @Override
+  public void releaseEventSink(StoreEventSink<K, V> eventSink) {
+    if (eventSink != NO_OP_EVENT_SINK) {
```
chrisdennis: Isn't this spilling the guts of InvocationScopedEventSink on to the table?  The if check here is basically a type check right?

impl/src/main/java/org/ehcache/internal/events/ScopedStoreEventDispatcher.java
```diff
@@ -0,0 +1,313 @@
...
+          events.removeLast();
+          Thread.currentThread().interrupt();
+        }
+      }
+    }
+  }
+
+  private static class FireableStoreEventWrapper<K, V> {
+    private final Lock lock = new ReentrantLock();
+    private final AtomicBoolean fireable = new AtomicBoolean(false);
```
chrisdennis: A single AtomicReference and an enum based state machine?

impl/src/main/java/org/ehcache/internal/events/ScopedStoreEventDispatcher.java
```diff
@@ -0,0 +1,313 @@
...
+      fireable.set(true);
+    }
+
+    void waitTillFired() {
+      int interruptCount = 0;
+      while (!fired.get()) {
+        lock.lock();
+        try {
+          if (!fired.get()) {
+            condition.await();
```
chrisdennis: awaitUninterruptibly?

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
ljacomet: Indeed

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
ljacomet: Not sure about that, only using a single value. Seems an `if` is more compact in that case.

ljacomet: However, having an if/else further down for a two value enum - using switch there

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
ljacomet: The wrapper I have to remove only contains the _public_ listener, not its configurations (ordered or not, async / sync), so I need to access the one in the set to do proper accounting.

core/src/main/java/org/ehcache/events/OrderedEventDispatcher.java
ljacomet: Getting rid of that whole abstraction. It is the business of the `CacheEventDispatcher` and its associated Factory, no need to have one layer there.

107/src/main/java/org/ehcache/jsr107/Eh107CacheEntryEvent.java
```diff
@@ -27,11 +27,11 @@
 
   private static final long serialVersionUID = 8460535666272347345L;
 
-  private final org.ehcache.event.CacheEvent<K, V> ehEvent;
+  private final org.ehcache.event.CacheEvent<? extends K, ? extends V> ehEvent;
```
chrisdennis: Looks like we can eliminate these fully-qualified class references now.

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -48,20 +51,25 @@
...
-
-  public CacheEventDispatcherImpl(ExecutorService orderedDelivery, ExecutorService unorderedDelivery, Store<K, V> store) {
-    this.orderedDelivery = orderedDelivery;
-    this.unorderedDelivery = unorderedDelivery;
-    this.store = store;
-    storeListener.setEventNotificationService(this);
+  private final ExecutorService unOrderedExectuor;
+  private final ExecutorService orderedExecutor;
+  private final AtomicInteger listenersCount = new AtomicInteger(0);
+  private final AtomicInteger orderedListenerCount = new AtomicInteger(0);
```
chrisdennis: These are now only accessed under lock, and so can now be replaced with regular ints

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -95,157 +130,115 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
-  }
+    EventListenerWrapper wrapper = new EventListenerWrapper(listener);
 
-  // TODO this should be really the shutdown method for the service
-  @Override
-  public void releaseAllListeners() {
-    for (EventListenerWrapper wrapper: registeredListeners) {
-      registeredListeners.remove(wrapper);
+    if (!removeWrapperFromList(wrapper, aSyncListenersList)) {
+      if (!removeWrapperFromList(wrapper, syncListenersList)) {
```
chrisdennis: &&

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -95,157 +130,115 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
-        @Override
-        public void run() {
-          CacheEventListener<K, V> listener = wrapper.getListener();
-          listener.onEvent(event);
-        }
-      };
-      
-      ExecutorService eventDelivery = wrapper.config.orderingMode().equals(EventOrdering.UNORDERED) ? unorderedDelivery : orderedDelivery;
-      notificationResults.put(wrapper, eventDelivery.submit(notificationTask));
+  public void shutdown() {
```
chrisdennis: This could just be made synchronized... then we can just call "clear" etc.

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
```diff
@@ -95,157 +130,115 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
-      return listener.hashCode();
+    if (!aSyncListenersList.isEmpty()) {
+      executor.submit(new EventDispatchTask<K, V>(event, aSyncListenersList));
     }
-    
-    @Override
-    public boolean equals(Object other) {
-      if (!(other instanceof EventListenerWrapper)) {
-        return false;
+    if (future != null) {
```
chrisdennis: Can we avoid the null check here by pushing the future de-ref inside the if, and make associated scope change too.

core/src/main/java/org/ehcache/events/ComplexEventThreadLocalList.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.events;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexEventThreadLocalList<T> {
```
chrisdennis: Dead

core/src/main/java/org/ehcache/events/CacheEvents.java
```diff
@@ -17,10 +17,10 @@
...
 
 import org.ehcache.Cache;
-import org.ehcache.Cache.Entry;
 import org.ehcache.event.CacheEvent;
 import org.ehcache.event.EventType;
-import org.ehcache.spi.cache.Store;
+import org.ehcache.spi.cache.events.StoreEvent;
+import org.ehcache.spi.cache.events.StoreEventListener;
 
 public final class CacheEvents {
```
chrisdennis: nullStoreEventListener method is dead.

impl/src/main/java/org/ehcache/internal/events/FireableStoreEventHolder.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+class FireableStoreEventHolder<K, V> {
+
+  enum Status {
+    CREATED, FIREABLE, FIRED
+  }
+
+  private final Lock lock = new ReentrantLock();
+  private final AtomicReference<Status> status = new AtomicReference<Status>(Status.CREATED);
+  private final AtomicBoolean failed = new AtomicBoolean(false);
```
chrisdennis: This could just be a volatile boolean.

chrisdennis:
Changes in the xml module are just whitespace... might be best to leave cleaning those files until later.

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
ljacomet: Actually I find this indicates intent better

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
ljacomet: Fixing that indeed

core/src/main/java/org/ehcache/events/CacheEventDispatcherImpl.java
ljacomet: Correct

core/src/main/java/org/ehcache/events/ComplexEventThreadLocalList.java
ljacomet: Removing

core/src/main/java/org/ehcache/events/CacheEvents.java
ljacomet: Removing

ljacomet:
test this please

ljacomet:
Working on fixing that racy test

ljacomet:
While investigating the racy test, I found a major issue with regards to eviction in `AbstractOffHeapStore`. As soon as you are ordered and the eviction kicks out a key sharing the same ordering queue as the value being written, there is a deadlock. Trying to figure out how to get around that.

ljacomet:
This is the updated version that fixes the issue mentioned above.
I would like to see it merged while not closing the associated issue as the XAStore related testing still needs to happen, but I believe that can happen in a follow up PR.

ljacomet:
Failing build is the current JDK 1.8 / cloudbees issue. All tests are green.

Issue #719, opened 2016-01-06
Ehcache Voltron Entity Internal Configuration Type


chrisdennis:
This is the configuration data that is stored internal to the entity (as of this point it will not be persistent - but it will be at a later date).  This will include the 'matching' logic for validating incoming client configuration information against the entity’s configuration.

PR #736, opened 2016-02-02
Issue #719 Add configure and validate implementations, both server an…

chrisdennis:
…d client side, and the related messages.


AbfrmBlr:
Is there a reason to not use proxy for method dispatch in slice1 ?

PR #737, opened 2016-02-03
Issue 704

lorban:
Parameters previously given to *Builder.build(...) are now given to *Buider.new*Builder(...). This has two nice side effects:

 - All builders now implement the Builder interface that mandates a build() method.
 - The ugly casts in GettingStarted are gone as the compiler can now infer the types.

### DIFF
--- a/impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
+++ b/impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
@@ -36,22 +36,28 @@ import static org.ehcache.config.ResourcePoolsBuilder.newResourcePoolsBuilder;
 /**
  * @author Alex Snaps
  */
-public class CacheConfigurationBuilder<K, V> {
+public class CacheConfigurationBuilder<K, V> implements Builder<CacheConfiguration<K, V>> {
 
   private final Collection<ServiceConfiguration<?>> serviceConfigurations = new HashSet<ServiceConfiguration<?>>();
   private Expiry<? super K, ? super V> expiry;
   private ClassLoader classLoader = null;
   private EvictionVeto<? super K, ? super V> evictionVeto;
   private ResourcePools resourcePools = newResourcePoolsBuilder().heap(Long.MAX_VALUE, EntryUnit.ENTRIES).build();
+  private Class<? super K> keyType;
+  private Class<? super V> valueType;
 
-  private CacheConfigurationBuilder() {
+  public static <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilder(Class<K> keyType, Class<V> valueType) {
+    return new CacheConfigurationBuilder<K, V>(keyType, valueType);
   }
 
-  public static <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilder() {
-    return new CacheConfigurationBuilder<K, V>();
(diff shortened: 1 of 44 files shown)


xml/src/main/java/org/ehcache/config/xml/XmlConfiguration.java
```diff
@@ -273,6 +278,7 @@ private void parseConfiguration()
         builder = builder.add(new DefaultCopierConfiguration(valueCopier, DefaultCopierConfiguration.Type.VALUE));
       }
       EvictionVeto evictionVeto = getInstanceOfName(cacheDefinition.evictionVeto(), cacheClassLoader, EvictionVeto.class);
+      builder = builder.evictionVeto(evictionVeto);
```
ljacomet: Pretty sure this causes compilation error

impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -174,16 +167,16 @@ public boolean hasDefaultExpiry() {
     return otherBuilder;
   }
 
-  public CacheConfigurationBuilder<K, V> withKeyCopier(Copier<K> keyCopier) {
+  public CacheConfigurationBuilder<K, V> withKeyCopier(Copier<? super K> keyCopier) {
```
ljacomet: Pretty sure that's incorrect as shown by the cast introduced below

impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -192,16 +185,16 @@ public boolean hasDefaultExpiry() {
     return otherBuilder;
   }
 
-  public CacheConfigurationBuilder<K, V> withValueCopier(Copier<V> valueCopier) {
+  public CacheConfigurationBuilder<K, V> withValueCopier(Copier<? super V> valueCopier) {
```
ljacomet: same as for `withKeyCopier`

impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -210,16 +203,16 @@ public boolean hasDefaultExpiry() {
     return otherBuilder;
   }
 
-  public CacheConfigurationBuilder<K, V> withKeySerializer(Serializer<K> keySerializer) {
+  public CacheConfigurationBuilder<K, V> withKeySerializer(Serializer<? super K> keySerializer) {
```
ljacomet: same as for `withXXXCopier`

impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
```diff
@@ -228,16 +221,16 @@ public boolean hasDefaultExpiry() {
     return otherBuilder;
   }
 
-  public CacheConfigurationBuilder<K, V> withValueSerializer(Serializer<V> valueSerializer) {
+  public CacheConfigurationBuilder<K, V> withValueSerializer(Serializer<? super V> valueSerializer) {
```
ljacomet: Same as before

impl/src/main/java/org/ehcache/config/CacheConfigurationBuilder.java
lorban: In my understanding, ? super K is needed as you can define Long as the key type but configure a Number key copier.

What did I miss?

Issue #739, opened 2016-02-03
Problem ehcache 3 with WAS7


fabriciofconde:
Hi,

I have been with problem to use Ehcache 3 with WAS7 (Websphere Application Server 7).
I did simple application web, and run test some application servers different:
- Tomcat 7: works
- Jetty 8: works
- WAS7: Doesn't work

I did same test with other versions (3.0.0.m4, 3.0.0.m2 and 3.0.0.m1) into WAS7:
- 3.0.0.m4 doesn't work
- 3.0.0.m2 doesn't work
- 3.0.0.m1 works

**Follow the logs**
_[03/02/16 15:09:45:661 BRST] 00000011 SystemErr     R 	at com.ibm.ws.http.channel.inbound.impl.HttpInboundLink.ready(HttpInboundLink.java:272)
[03/02/16 15:09:45:661 BRST] 00000011 SystemErr     R 	at com.ibm.ws.tcp.channel.impl.NewConnectionInitialReadCallback.sendToDiscriminators(NewConnectionInitialReadCallback.java:214)
[03/02/16 15:09:45:662 BRST] 00000011 SystemErr     R 	at com.ibm.ws.tcp.channel.impl.NewConnectionInitialReadCallback.complete(NewConnectionInitialReadCallback.java:113)
[03/02/16 15:09:45:662 BRST] 00000011 SystemErr     R 	at com.ibm.ws.tcp.channel.impl.AioReadCompletionListener.futureCompleted(AioReadCompletionListener.java:165)
[03/02/16 15:09:45:662 BRST] 00000011 SystemErr     R 	at com.ibm.io.async.AbstractAsyncFuture.invokeCallback(AbstractAsyncFuture.java:217)
[03/02/16 15:09:45:662 BRST] 00000011 SystemErr     R 	at com.ibm.io.async.AsyncChannelFuture.fireCompletionActions(AsyncChannelFuture.java:161)
[03/02/16 15:09:45:662 BRST] 00000011 SystemErr     R 	at com.ibm.io.async.AsyncFuture.completed(AsyncFuture.java:138)
[03/02/16 15:09:45:663 BRST] 00000011 SystemErr     R 	at com.ibm.io.async.ResultHandler.complete(ResultHandler.java:202)
[03/02/16 15:09:45:663 BRST] 00000011 SystemErr     R 	at com.ibm.io.async.ResultHandler.runEventProcessingLoop(ResultHandler.java:766)
[...]

PR #741, opened 2016-02-08
#739: fix ConcurrentWeakIdentityHashMap keySet() and entrySet()

lorban:
Fix the implementations so that the returned sets do reflect map changes. The IBM JDK 1.6's Collections.newSetFromMap() implementation directly relies on this contract which made the problem obviously apparent, but we could have run into the same kind of problem with other JDKs too.

### DIFF
--- a/core/src/main/java/org/ehcache/util/ConcurrentWeakIdentityHashMap.java
+++ b/core/src/main/java/org/ehcache/util/ConcurrentWeakIdentityHashMap.java
@@ -18,9 +18,10 @@ package org.ehcache.util;
 
 import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
-import java.util.AbstractMap;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.AbstractSet;
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/ehcache/util/ConcurrentWeakIdentityHashMap.java
```diff
@@ -117,15 +119,56 @@ public void clear() {
...
-    for (WeakReference<K> kWeakReference : map.keySet()) {
-      final K k = kWeakReference.get();
-      if (k != null) {
-        ks.add(k);
+    return new AbstractSet<K>() {
+      @Override
+      public Iterator<K> iterator() {
+        purgeKeys();
+        final Iterator<WeakReference<K>> iterator = map.keySet().iterator();
+        final AtomicReference<K> ref = new AtomicReference<K>();
```
chrisdennis: Why an AtomicReference?

core/src/main/java/org/ehcache/util/ConcurrentWeakIdentityHashMap.java
```diff
@@ -117,15 +119,56 @@ public void clear() {
...
+        advance(iterator, ref);
+
+        return new Iterator<K>() {
```
chrisdennis: Could we create a common abstract iterator class that could be shared between the tow derived collections?

core/src/main/java/org/ehcache/util/ConcurrentWeakIdentityHashMap.java
ljacomet: Because of the required final to use it inside the anonymous iterator. Or do you suggest that a simple array would be better?

core/src/main/java/org/ehcache/util/ConcurrentWeakIdentityHashMap.java
ljacomet: +1

lorban: Yes, I could and should have done that.

Issue #724, opened 2016-01-06
Create Ehcache Kit


chrisdennis:
Building on the Voltron skeleton kit, we will need to create an Ehcache specific kit that bundles the Voltron kit with the Ehcache specific code that needs to reside on the server, along with the client side Ehcache libraries.

chrisdennis:
Waiting on Jeff to complete the integration/kit testing framework so that we can add some automated kit testing before this issue is closed.

PR #744, opened 2016-02-10
Issue #724 Initial Ehcache Kit Skeleton

chrisdennis:
What I'm really asking for review of here is the gradle changes.  This isn't a final kit version, it will need tweaking, but doing that will be a lot simpler once the other tasks are complete.  So I'd like this to be considered for merging, but I'll hold off on closing the issue.


ljacomet:
Not too happy with the resulting `MANIFEST.MF` having `Export-Package` for `LICENSE` and all XSDs.

Adding the following: `instruction 'Export-Package', '!LICENSE', '!ehcache-*.xsd'` fixes the issue but makes the XSDs appear in the `Private-Package` section.

@anthonydahanne and others: any idea what are the consequences - if any - of these two options?

PR #745, opened 2016-02-10
Management API layer refactoring and CollectorService

mathieucarbou:
Goals:

- Extracted management registry api to [terracotta-management](https://github.com/Terracotta-OSS/terracotta-management/) so that other products could re-use the same mechanism to expose their stats and capabilities
- Added Ehcache service that uses the management registry and its configuration to frequently send stats and also sends ehcache notifications to a management message consumer.

Some details:

- Frequency to send stats is by default based on the TimeToDisable (75% of this value). Default value for ttd is 30 seconds, so by default, stats would be sent each 22 seconds.
- Depends on new version 2.1.0 of [terracotta-management](https://github.com/Terracotta-OSS/terracotta-management/) (not yet released)
- New module dependency added: ___terracotta-management-registry___ (besides current existing _terracotta-management-model_)
- Test / example added: `DefaultCollectorServiceTest `


management/src/main/java/org/ehcache/management/providers/CacheBinding.java
```diff
@@ -38,7 +38,7 @@ public String getAlias() {
     return alias;
   }
 
-  public Cache<?, ?> getCache() {
+  public Cache getCache() {
```
chrisdennis: What was wrong with the wildcards here?

mathieucarbou: If I leave them here, `EhcacheActionWrapper` won't compile at lines 64, 71 and 79.

I don't know what would be the best way to solve that... It's caused by the reflection invocation of the management calls (Object type).

```
  @Exposed
  public Object get(@Named("key") Object key) {
    Object convertedKey = convert(key, managedObject.getCache().getRuntimeConfiguration().getKeyType());
    return managedObject.getCache().get(convertedKey); // 64
  }

  @Exposed
  public void remove(@Named("key") Object key) {
    Object convertedKey = convert(key, managedObject.getCache().getRuntimeConfiguration().getKeyType());
    managedObject.getCache().remove(convertedKey); // 71
  }

  @Exposed
  public void put(@Named("key") Object key, @Named("value") Object value) {
    Object convertedKey = convert(key, managedObject.getCache().getRuntimeConfiguration().getKeyType());
    Object convertedValue = convert(value, managedObject.getCache().getRuntimeConfiguration().getValueType());
    managedObject.getCache().put(convertedKey, convertedValue); // 74
  }
```

mathieucarbou:
test this please

mathieucarbou:
@chrisdennis @lorban @ljacomet @anthonydahanne : terracotta-management 2.1.0 has been release, so this PR is complete now.

Also, after discussing with Anthony, we removed the cluster topology classes from terracotta-management, to only keep monitoring and management metadata inside this lib.

management/src/main/java/org/ehcache/management/providers/CacheBinding.java
chrisdennis: Yeah - this makes sense.  The pedant in me wants to add the generics to `CacheBinding` and as far out as we can make them work... but that is ultimately futile.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -0,0 +1,215 @@
...
+    this.messageConsumer = messageConsumer;
+  }
+
+  @Override
+  public void start(ServiceProvider serviceProvider) {
+    timeSource = serviceProvider.getService(TimeSourceService.class);
+    managementRegistry = serviceProvider.getService(ManagementRegistryService.class);
+    configuration = managementRegistry.getConfiguration();
+    cacheManagerProviderService = serviceProvider.getService(CacheManagerProviderService.class);
+    scheduledExecutorService = serviceProvider.getService(ExecutionService.class).getScheduledExecutor(DefaultCollectorService.class.getSimpleName());
```
chrisdennis: This isn't the correct way to use the ExecutionService... you need to be pulling the alias from user provided config.  This will work in the default case where the service ignores the aliases... but if the user tries to constrain threads in to named pools they need to be able to make sure the set of aliases used within the system matches their pool names.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -0,0 +1,215 @@
...
+  public void cacheRemoved(String alias, Cache<?, ?> cache) {
+    messageConsumer.accept(new DefaultMessage(
+        timeSource.getTimeSource().getTimeMillis(),
+        new ContextualNotification(
+            configuration.getContext().with("cacheName", alias),
+            EhcacheNotification.CACHE_REMOVED.name())));
+  }
+
+  @Override
+  @SuppressFBWarnings("SF_SWITCH_NO_DEFAULT")
```
chrisdennis: We should add no-op cases for the 'ignored' transitions, and an AssertionError throw on default.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -0,0 +1,215 @@
...
+        messageConsumer.accept(new DefaultMessage(
+            timeSource.getTimeSource().getTimeMillis(),
+            new ContextualNotification(
+                configuration.getContext(),
+                EhcacheNotification.CACHE_MANAGER_AVAILABLE.name())));
+        break;
+
+      case MAINTENANCE:
+        messageConsumer.accept(new DefaultMessage(
+            timeSource.getTimeSource().getTimeMillis(),
```
chrisdennis: Why are we always going back to the services to get stuff?  We can probably store the `TimeSource` and `CacheManager` directly, no?

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+      }, pollingIntervalMs, pollingIntervalMs, TimeUnit.MILLISECONDS);
+    }
+  }
+
+  @Override
+  public void stopStatisticCollector() {
+    if (collectingStats.compareAndSet(true, false)) {
+      task.cancel(false);
+      task = null;
```
chrisdennis: This is racy... you might accidentally cancel, or null out an incoming task reference (unless their is synchronization outside this somewhere).  You could potentially switch to an AtomicReference here to avoid this issue.

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryService.java
```diff
@@ -0,0 +1,146 @@
...
+  private volatile ScheduledExecutorService statisticsExecutor;
+  private volatile EhcacheManager cacheManager;
+
+  public DefaultManagementRegistryService() {
+    this(new DefaultManagementRegistryConfiguration());
+  }
+
+  public DefaultManagementRegistryService(ManagementRegistryServiceConfiguration configuration) {
+    if (configuration == null) {
+      throw new NullPointerException();
```
chrisdennis: Worth reporting what was null?

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryService.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+    this.configuration = configuration;
+  }
+
+  @Override
+  public void start(final ServiceProvider serviceProvider) {
+    this.statisticsExecutor = serviceProvider.getService(ExecutionService.class).getScheduledExecutor(EhcacheStatisticsProvider.class.getSimpleName());
```
chrisdennis: Execution service usage wrong here too.

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryService.java
```diff
@@ -0,0 +1,146 @@
...
+  @Override
+  public void cacheRemoved(String alias, Cache<?, ?> cache) {
+    unregister(cache);
+    unregister(new CacheBinding(alias, cache));
+
+    StatisticsManager.dissociate(cache).fromParent(cacheManager);
+  }
+
+  @Override
+  @SuppressFBWarnings("SF_SWITCH_NO_DEFAULT")
```
chrisdennis: That same switch default issue.

management/src/main/java/org/ehcache/management/registry/DefaultSharedManagementService.java
```diff
@@ -63,14 +66,18 @@ public void cacheRemoved(String alias, Cache<?, ?> cache) {
       }
 
       @Override
+      @SuppressFBWarnings("SF_SWITCH_NO_DEFAULT")
```
chrisdennis: Default again.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
mathieucarbou: ok! I've changed what was there before because of these 2 reasons:

1. because I ended to have some threads with null in names since the alias is null by default.

    `return new Thread(r, "Ehcache [" + alias + "]-" + threadCount.getAndIncrement());`

2. also because I wanted to separate the thread pool responsible of computing statistics (in the  `EhcacheStatisticsProvider` class and the thread pool responsible for collecting and pushing messages (`DefaultCollectorService`)

=> What would be the best way to fix that @chrisdennis ? I need to redo and thus have 2 configurable aliases in the `ManagementRegistryServiceConfiguration` ? Thanks :-)

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
mathieucarbou: ok!

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
mathieucarbou: yes!

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
mathieucarbou: Thanks! Will do!

Note: would synchronized work also ? because I am wondering if I also need to make sure that start / stop / stopCollector and startCollector are exclusive ?

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryService.java
mathieucarbou: yes, I'll fix!

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryService.java
mathieucarbou: No, we could just create a default config. I'll change.

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryService.java
mathieucarbou: yep!

management/src/main/java/org/ehcache/management/registry/DefaultSharedManagementService.java
mathieucarbou: ... and yes ;-)

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
chrisdennis: Yes you could synchronize if you wanted instead.  If you used an AtomicReference you would then have to deal with cancelling a task if you failed to install it in the AtomicReference.  Synchronization might be the better solution if it's not going to be a bottleneck, which I doubt it will.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
chrisdennis: Yes... you'll need to have those aliases be configurable in your service config.

mathieucarbou:
@chrisdennis : PR updated. Could you please check again (specifically for the racy code, if it is fixed correctly). Thank you!

mathieucarbou:
@ljacomet @chrisdennis @lorban : just rebased.
Please tell us (@anthonydahanne and I) if anything else is required for this to be merge :-) 
Thank you!

mathieucarbou:
test this please

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * @author Mathieu Carbou
+ */
+@ServiceDependencies({CacheManagerProviderService.class, ManagementRegistryService.class, ExecutionService.class, TimeSourceService.class})
+public class DefaultCollectorService implements CollectorService, CacheManagerListener {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultCollectorService.class);
+
+  private volatile ScheduledFuture<?> task;
```
chrisdennis: Doesn't need to be volatile now does it?

chrisdennis:
From my point of view this looks okay now... might want to get @anthonydahanne to review the management aspects of it, but other than that it's fine.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
mathieucarbou: Right! I'll update! now it doesn't :-)

mathieucarbou:
test this please

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryConfiguration.java
```diff
@@ -29,22 +30,29 @@
...
     addConfiguration(new EhcacheStatisticsProviderConfiguration(1, TimeUnit.MINUTES, 100, 1, TimeUnit.SECONDS, 30, TimeUnit.SECONDS));
   }
 
   public DefaultManagementRegistryConfiguration setCacheManagerAlias(String alias) {
-    this.cacheManagerAlias = alias;
+    return setContext(Context.create("cacheManagerName", alias));
+  }
+
+  public DefaultManagementRegistryConfiguration setContext(Context context) {
+    if (!this.context.contains("cacheManagerName") && !context.contains("cacheManagerName")) {
```
anthonydahanne: this.context.contains("cacheManagerName") || context.contains("cacheManagerName")

anthonydahanne:
@chrisdennis  could you please merge ? LGTM

Issue #740, opened 2016-02-04
Inconsistent duplicate service configuration handling


lorban:
Some services are meant to have multiple configs, while some other should just have one. There are checks here and there but not everywhere there should be.

For instance:

```
        CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class)
            .withLoaderWriter(new SampleLoaderWriter<Long, String>(singletonMap(41L, "zero"))) // <1>
            .withLoaderWriter(new CacheLoaderWriter<Long, String>() {...}
...
```

does not fail while it should.

```
    DefaultSerializationProviderConfiguration service = new DefaultSerializationProviderConfiguration();
    service.addSerializerFor(Long.class, LongSerializer.class);
    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
        .withCache("cache", cacheConfiguration)
        .withSerializer(String.class, StringSerializer.class)
        .using(service)
        .build(true);
```

also doesn't fail, while

```
    CacheConfiguration<Long, String> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class)
        .withResourcePools(ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).offheap(10, MemoryUnit.MB))
        .withKeySerializer(new LongSerializer()) // <1>
        .withValueSerializer(new CharSequenceSerializer()) // <2>
        .add(new DefaultSerializerConfiguration<Long>(LongSerializer.class, SerializerConfiguration.Type.KEY))
        .build();
```

properly fails with a readable error message.

There probably are other instances of the problem, which probably means we should revise all our uses of ServiceLocator.findAmongst() and ServiceLocator.findSingletonAmongst().

PR #750, opened 2016-02-12
Issue #740 Prevent duplicate service configuration in builder

ljacomet:
When creating the cache configuration, most service configuration are
 expected to exist as a single instance. The builder now takes care of
 achieving this. At the same time, service consuming these configuration
 now properly check that no duplicates are found when this is not valid.

This pull requests takes care of the `CacheConfigurationBuilder` level, `CacheManagerBuilder` level to come later.

### DIFF
--- a/core/src/main/java/org/ehcache/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/EhcacheManager.java
@@ -60,6 +60,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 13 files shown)


ljacomet:
Added the `CacheManagerBuilder` handling of duplicate `ServiceCreationConfiguration`

ljacomet:
test this please

impl/src/test/java/org/ehcache/internal/store/tiering/CacheStoreWith3TiersSPITest.java
```diff
@@ -78,6 +79,7 @@
  *
  * @author Ludovic Orban
  */
+@Ignore
```
lorban: Why the @Ignore ?

lorban:
test this please

core/src/main/java/org/ehcache/EhcacheManager.java
```diff
@@ -113,6 +114,17 @@ public EhcacheManager(Configuration config, Collection<Service> services, boolea
     this.configuration = new DefaultConfiguration(config);
   }
 
+
+  private void validateServicesConfigs() {
```
lorban: Where is that called?

core/src/main/java/org/ehcache/config/DefaultConfiguration.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
```
lorban: Did you forget to cleanup the imports?

core/src/main/java/org/ehcache/EhcacheManager.java
ljacomet: Well ... should be in the constructor ... fixing that, thanks

core/src/main/java/org/ehcache/config/DefaultConfiguration.java
ljacomet: Might have, will check

core/src/main/java/org/ehcache/EhcacheManager.java
lorban: If the build did not fail despite this, this means there is a test lacking somewhere. :)

ljacomet: Yep, took care of that as well.

Issue #616, opened 2015-09-29
Byte sized support for on heap sizing


ljacomet:
Use and integrate the [sizeof library](https://github.com/ehcache/sizeof).

ljacomet:
This task should also improve the documentation by documenting supported units.

ljacomet:
Before releasing this feature, we will need to decide if we want the currently [open issues of sizeof](https://github.com/ehcache/sizeof/issues) to be resolved and released first.

ljacomet:
Following the fix for issue ehcache/sizeof#28, we want to revise the max sizing depth that existed in Ehcache 2.x.

Proposition is:
* Provide a max depth option to limit CPU consumption on sizing
* Provide a max size option to limit memory consumption of a single mapping

By default, there will be no defaults for these two options - object graphs as deep and as big as given will be sized then inserted.
Hitting any of these limits will cause the mapping to not be cached and we will log the fact that we dropped it for such a reason.

ljacomet:
First major part of the work made it in.

Known missing things:
* XML sizeof limitations configuration
* Integration with reworked builders (see #705)
* Handling sizeof refusing to size due to imposed limits

PR #753, opened 2016-02-16
Sizeof exception handling #616

AbfrmBlr:


AbfrmBlr:
This should close #616

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -567,7 +570,7 @@ public boolean replace(final K key, final V oldValue, final V newValue) throws C
         }
       });
       if (returnValueHolder.get() != null) {
-        long replacedDelta = replacedValue.size() - returnValueHolder.get().size();
+        long replacedDelta = replacedValue == null ? 0 : replacedValue.size() - returnValueHolder.get().size();
```
ljacomet: Please add the parenthesis as done in the change above this one

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1376,18 +1401,23 @@ private void expireMapping(final K key, final ValueHolder<V> value) {
...
-  private OnHeapValueHolder<V> importValueFromLowerTier(K key, ValueHolder<V> valueHolder, long now) {
+  private OnHeapValueHolder<V> importValueFromLowerTier(K key, ValueHolder<V> valueHolder, long now) throws LimitExceededException {
     V realValue = valueHolder.value();
     Duration expiration = expiry.getExpiryForAccess(key, realValue);
-    return cloneValueHolder(key, valueHolder, now, expiration);
+    return cloneValueHolder(key, valueHolder, now, expiration, false);
   }
 
-  private OnHeapValueHolder<V> cloneValueHolder(K key, ValueHolder<V> valueHolder, long now, Duration expiration) {
+  private OnHeapValueHolder<V> cloneValueHolder(K key, ValueHolder<V> valueHolder, long now, Duration expiration, boolean dontSize) throws LimitExceededException {
```
ljacomet: Can you invert the added boolean parameter - so that `true` means sizing happens, `false` means it does not.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -779,6 +779,26 @@ public K getKey() {
...
+  private OnHeapValueHolder<V> invalidateInGetorComputeIfAbsent(MapWrapper<K, V> map, final K key, final ValueHolder<V> value, final Fault<V> fault, final long now) {
+    final AtomicReference<OnHeapValueHolder<V>> toInvalidate = new AtomicReference<OnHeapValueHolder<V>>();
+    map.computeIfPresent(key, new BiFunction<K, OnHeapValueHolder<V>, OnHeapValueHolder<V>>() {
+      @Override
+      public OnHeapValueHolder<V> apply(K mappedKey, final OnHeapValueHolder<V> mappedValue) {
+        if(mappedValue.equals(fault)) {
+          try {
+            toInvalidate.set(cloneValueHolder(key, value, now, Duration.ZERO, true));
+          } catch (LimitExceededException ex) {
+            throw new AssertionError("Should have already been thrown.");
```
ljacomet: That message is confusing. Just say that it should not be thrown here as sizing is not happening.

impl/src/main/java/org/ehcache/internal/store/heap/OnHeapStore.java
```diff
@@ -1355,9 +1373,16 @@ private void expireMapping(final K key, final ValueHolder<V> value) {
       }
     }
 
-    return makeValue(key, newValue, now, expirationTime, this.valueCopier);
+    OnHeapValueHolder<V> holder = null;
+    try {
+      holder = makeValue(key, newValue, now, expirationTime, this.valueCopier);
+      eventSink.updated(key, oldValue, newValue);
+    } catch (LimitExceededException e) {
+      LOG.warn(e.getMessage());
```
ljacomet: There is a missing removed event here. See what happens when expiry result is a duration of zero.

Issue #683, opened 2015-11-24
Modularize APIs to account for JDK9


alexsnaps:
May want to use this to figure out where we screwed up:
http://jqassistant.org/get-started/

ljacomet:
* API module is master
 * a package present here cannot be repeated anywhere else.
* core module: 
 * `org.ehcache.core.*`
* impl module:
 * `org.ehcache.impl.*`
 * `org.ehcache.config.builders`
* xml module:
 * `org.ehcache.xml.*`
* transactions module:
 * `org.ehcache.transactions.*`
* 107 module:
 * `org.ehcache.jsr107.*`
* clustered modules:
 * `org.ehcache.clustered.*`

PR #756, opened 2016-02-16
Issue #683 Moving packages around

ljacomet:
Goal si to not have duplication in packages across modules.


lorban:
I like this change because it has a very nice side-effect: by just looking at your import list, you can know what modules you're depending on.

And this makes me wonder if moving the builders from core to impl was such a smart move as IIRC impl was supposed to only contain classes that end-users would never have to play with. Or maybe that was a too extreme approach?

Food for thoughts...

PR #757, opened 2016-02-17
Issue 710

AbfrmBlr:


impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -268,7 +293,43 @@ public void close() throws Exception {
     }
 
   }
-  
+
+  private void validateListenerConfig() {
+    if (!eventListenerConfigurations.isEmpty() && eventNotifier instanceof DisabledCacheEventNotificationService) {
+      if (orderedExecutor == null && unOrderedExecutor == null) {
+        throw new IllegalArgumentException("Listeners will not work unless exceutors or EventDispatcher is configured.");
```
ljacomet: typo in exception message

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -89,7 +97,7 @@ private ServiceDeps() {
   private ClassLoader classLoader = ClassLoading.getDefaultClassLoader();
   private EvictionVeto<? super K, ? super V> evictionVeto;
   private CacheLoaderWriter<? super K, V> cacheLoaderWriter;
-  private CacheEventDispatcher<K, V> cacheEventNotificationService = new DisabledCacheEventNotificationService<K, V>();
+  private CacheEventDispatcher<K, V> eventNotifier = new DisabledCacheEventNotificationService<K, V>();
```
ljacomet: Name is weird - why not `eventDispatcher`?

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -241,6 +259,11 @@ public void close() throws Exception {
         storeProvider.releaseStore(store);
       }
     });
+
+    if (this.eventNotifier instanceof DisabledCacheEventNotificationService && (orderedExecutor != null & unOrderedExecutor != null)) {
+      this.eventNotifier = new CacheEventDispatcherImpl<K, V>(store, unOrderedExecutor, orderedExecutor);
```
ljacomet: So this works indeed. But if the user tries to provide a different implementation of `CacheEventDispatcher`, it will never be connected to the `Store`.
To have a working solution, you need to be able to pass the `CacheEventDispatcherImpl` from the outside and get a working setup. And with the current contract, this does not work, which indicates that we are hiding an implicit contract. We need to make it explicit. That's why I mentioned that the `CacheEventDispatcher` probably needs a setter for the `StoreEventSource`.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -268,7 +293,43 @@ public void close() throws Exception {
...
+        throw new IllegalArgumentException("Listeners will not work unless exceutors or EventDispatcher is configured.");
+      }
+    }
+  }
+
+  private void registerListeners(Cache<K, V> cache, ServiceProvider serviceLocator, List<LifeCycled> lifeCycledList) {
+    if (!eventListenerConfigurations.isEmpty()) {
+      final CacheEventListenerProvider listenerProvider = serviceLocator.getService(CacheEventListenerProvider.class);
+      if (listenerProvider == null) {
+        throw new IllegalStateException("No CacheEventListenerProvider could be found - did you configure one?");
```
ljacomet: How can we get here?

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -315,9 +376,31 @@ public T build() {
...
-  public final UserManagedCacheBuilder<K, V, T> withCacheEvents(CacheEventDispatcher<K, V> cacheEventNotificationService) {
+  public final UserManagedCacheBuilder<K, V, T> withEventDispatcher(CacheEventDispatcher<K, V> eventNotifier) {
+    UserManagedCacheBuilder<K, V, T> otherBuilder = new UserManagedCacheBuilder<K, V, T>(this);
+    otherBuilder.eventNotifier = eventNotifier;
+    return otherBuilder;
+  }
+
+  public final UserManagedCacheBuilder<K, V, T> withOrderedEventExecutor(ExecutorService orderedExecutor) {
+    UserManagedCacheBuilder<K, V, T> otherBuilder = new UserManagedCacheBuilder<K, V, T>(this);
+    otherBuilder.orderedExecutor = orderedExecutor;
```
ljacomet: The builder methods should take care of not allowing to get in an invalid state. Either by throwing early or by performing the necessary _overwrite_ when user specifies conflicting configuration.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
AbfrmBlr: I thought of the same name, but Ehcache was using this, so for the sake of it

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
AbfrmBlr: If the user does not configure a listenerprovider, servicelocator fails to find this as this not one of the required services in dependency

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
AbfrmBlr: cool

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
AbfrmBlr: its taken care in build method as of now. Will move it.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
ljacomet: That makes little sense then. Looking at the code around all this, I believe there is something not working that well ... While I do not want to revise all of that here, I still believe that if the user provides a `CacheEventListenerConfiguration` we should make that service available.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
ljacomet: So for the sake of it, you do the rename in Ehcache as well :smile:

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
AbfrmBlr: We can do that, just use the default provider to create listeners from their config.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
AbfrmBlr: By overwrite you mean that is user configured the `eventNotifier` then it is given preference over the configured executors, right ? or do we want to throw when the user configures both `eventNotifier` and executors ? 

As of now if both `eventNotifier` & executors are configured the preference is given to the provided `eventNotifier`.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -268,7 +296,46 @@ public void close() throws Exception {
...
+  }
+
+  private void registerListeners(Cache<K, V> cache, Store<K, V> store, ServiceProvider serviceLocator, List<LifeCycled> lifeCycledList) {
+    if (!eventListenerConfigurations.isEmpty()) {
+      final CacheEventListenerProvider listenerProvider;
+      CacheEventListenerProvider provider;
+      if ((provider = serviceLocator.getService(CacheEventListenerProvider.class)) != null) {
+        listenerProvider = provider;
+      } else {
+        listenerProvider = new DefaultCacheEventListenerProvider(new DefaultCacheEventListenerProviderConfiguration());
```
ljacomet: Please get rid of the configuration class. It is useless. The `ClassInstanceProvider` does not need it, it accepts a null as the first parameter in its constructor.

impl/src/main/java/org/ehcache/UserManagedCacheBuilder.java
```diff
@@ -268,7 +295,46 @@ public void close() throws Exception {
...
+
+  private void validateListenerConfig() {
+    if (!eventListenerConfigurations.isEmpty() && eventDispatcher instanceof DisabledCacheEventNotificationService) {
+      if (orderedExecutor == null && unOrderedExecutor == null) {
+        throw new IllegalArgumentException("Listeners will not work unless Executors or EventDispatcher is configured.");
+      }
+    }
+  }
+
+  private void registerListeners(Cache<K, V> cache, Store<K, V> store, ServiceProvider serviceLocator, List<LifeCycled> lifeCycledList) {
```
ljacomet: store param is unused

ljacomet:
test this please

Issue #20, opened 2014-08-21
Polish config framework


alexsnaps:
i.e. everything around config parsing from XML, creating Builders and building actual `Cache` and `CacheManager` instances. But also wiring the `RuntimeConfiguration` of some kind, that let someone read the configuration currently in use by a given `Cache` or `CacheManager`, also enabling mutating the bits that are mutable

ljacomet:
Improvement areas:
* Documenting dynamic updatable properties

chrisdennis:
What about Hungarian?

alexsnaps:
Revisit: 
 - attribute vs. element in XML config. 
    Some "logical" consistency has to be applied here
 - `newBuilder()` method params vs. `Builder.build()` params
   where the former feels like a nice way to force required attributes to have some values, but which can be overridden with the setters and makes the builder more reusable.

albinsuresh:
With #635 done, users will be able to configure instances of loader-writers, serializers etc directly using the cache configuration builder as follows:
```java
CacheConfigurationBuilder.newCacheConfigurationBuilder()
        .add(new DefaultCopierConfiguration<Description>(new DescriptionCopier(),
            CopierConfiguration.Type.KEY))
```

This way the user still has to use a service configuration explicitly to stick in an actual instance. As an improvement ver this, specialized methods like _withLoaderWriter_, _withValueSerializer_ etc can be added to the `CacheConfigurationBuilder` so that the configuration of instances would look much more intuitive. Internally we can still get the job done with the overloaded service configuration constructors that take in instances. But at least the user facing API would look much cleaner this way.

So configuring instances could be done this way:
```
CacheConfigurationBuilder.newCacheConfigurationBuilder()
        .withLoaderWriter(new SimpleLoaderWriter())
        .withKeyCopier(new DescriptionCopier())
        .withValueSerializer(new PersonSerializer())
```

ljacomet:
Created #702, #703, #704 and #705 to cover the 4 specific issues.

This issue remains open to do a pass on the whole configuration.

lorban:
Event listeners config still need some form of polish too, but we may want to wait until #710 gets resolved first.

PR #766, opened 2016-02-19
Issue #20

ljacomet:
* Add default serializers for `String`, `Integer`, `Float`, `Long`, `Double` and `Character`
* Move one lost builder to proper package
* Fix null handling in `SoftLock`


docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -282,6 +281,8 @@ Serializer instances are created per cache for keys and values separately.
 By default, serializers are automatically used by tiers that require them (off-heap and disk) but the on-heap tier, while it can
 use them or not, will not by default.
 
+NOTE: By default, there are serializers already configured for types `java.io.Serializable`, `java.lang.String` and `java.lang.Long`.
+
```
lorban: and Float and Character and Integer and... ?

Issue #722, opened 2016-01-06
Terracotta Clustering Service Configuration


chrisdennis:
Things under this task:
* ServiceCreationConfiguration Implementation (also a CacheManagerConfiguration<PersistentCacheManager> - c.f. CacheManagerPersistenceConfiguration/PersistenceConfiguration
* Programmatic builder for the above.
* XML Configuration schema and parsing code for the above.

Issue #747, opened 2016-02-11
Decoupling jsr 107 module from management module


mathieucarbou:
So that ehcache "core" features do no depend on management module and its transitive dependencies. Since jsr 107 is not using any history data it should probably be possible to use directly the `SatsisticManager` class.

PR #769, opened 2016-02-19
Issue #722 Terracotta clustering service configuration support

cljohnso:
This PR is in two parts:

1. Refactors EhcacheManager to split CachePersistenceManager methods out.  This creates a BaseCacheManager abstract class holding the non-persistent methods leaving EhcacheManager as a subclass maintaining the current functionality. This is a positioning move for supporting clustering.

  * Adds requiresSerialization method to ResourceType
  * Splits EhcacheManager.newStore into BaseCacheManager (non-DISK)
  * Generalizes CacheManagerBuilder to
    * return builder-designated CacheManager type (subclass)
    * return CacheManagerConfiguration-driven CacheManagerBuilder subclass
    * adds newSpecializedCacheManager method to obtain type-specific CacheManager from configuration

1. Adds interfaces and concrete classes in support of configuring and
building a clustered CacheManager.

  * ehcache-clustered-ext.xsd defining the <service> element extension
  * ClusteringServiceConfigurationParser for parsing XML <service> extension
  * ClusteringServiceConfiguration providing both ServiceCreationConfiguration and CacheManagerConfiguration support
  * ClusteringService, DefaultClusteringService and ClusteringServiceFactory providing the anchor for the clustering service configured with the ClusteringServiceConfiguration
  * ClusteredCacheManager, ClusteredCacheManagerImpl, and ClusteredCacheManagerBuilder providing support for a clustered CacheManager

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
@@ -63,7 +63,7 @@ class ClusteredCacheManagerImpl extends BaseCacheManager implements ClusteredCac
 
   @Override
   public Maintainable toMaintenance() {
-    return null;
+    throw new UnsupportedOperationException("not yet implemented");
   }
 
   @Override
--- a/clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered;
+
+import org.ehcache.Maintainable;
+import org.ehcache.clustered.exceptions.CacheDurabilityException;
+import org.ehcache.config.Configuration;
+import org.ehcache.core.BaseCacheManager;
+import org.ehcache.core.Ehcache;
+import org.ehcache.core.events.CacheEventDispatcherFactory;
+import org.ehcache.core.spi.cache.Store;
+import org.ehcache.event.CacheEventListenerProvider;
+import org.ehcache.spi.loaderwriter.CacheLoaderWriterProvider;
+import org.ehcache.spi.loaderwriter.WriteBehindProvider;
+import org.ehcache.spi.service.Service;
+import org.ehcache.spi.service.ServiceDependencies;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Provides an implementation for a clustered {@link org.ehcache.CacheManager CacheManager}.
+ *
+ * @author Clifford W. Johnson
+ */
+class ClusteredCacheManagerImpl extends BaseCacheManager implements ClusteredCacheManager {
+
+  // TODO: Determine a way to identify the dependencies of BaseCacheManager ...
+  @ServiceDependencies({
+      Store.Provider.class,
+      CacheLoaderWriterProvider.class,
+      WriteBehindProvider.class,
+      CacheEventDispatcherFactory.class,
+      CacheEventListenerProvider.class
+  })
+  private static class ServiceDeps {
+    private ServiceDeps() {
+      throw new UnsupportedOperationException("This is an annotation place holder, not to be instantiated");
+    }
+  }
+
+  public ClusteredCacheManagerImpl(final Configuration configuration) {
+    this(configuration, Collections.<Service>emptyList());
+  }
+
+  public ClusteredCacheManagerImpl(final Configuration configuration, final Collection<Service> services) {
+    super(configuration, services, true);
+  }
+
+  @Override
+  public Maintainable toMaintenance() {
+    return null;
+  }
+
+  @Override
+  public void destroyCache(final String alias) throws CacheDurabilityException {
+  }
+
+  @Override
+  protected Class<?> getServiceDependencyMarkerClass() {
+    return ServiceDeps.class;
+  }
+
+  @Override
+  protected void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
+  }
+}
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.config;
+
+import org.ehcache.CacheManager;
+import org.ehcache.clustered.ClusteredCacheManager;
+import org.ehcache.clustered.ClusteredCacheManagerBuilder;
+import org.ehcache.clustered.service.ClusteringService;
+import org.ehcache.config.builders.CacheManagerBuilder;
+import org.ehcache.config.builders.CacheManagerConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+
+import java.net.URI;
+
+/**
+ * Specifies the configuration for a {@link ClusteringService}.
+ *
+ * @author Clifford W. Johnson
+ */
+// TODO: Should this accept/hold a *list* of URIs?
+// TODO: Is 'alias' *required*?
+public class ClusteringServiceConfiguration
+    implements ServiceCreationConfiguration<ClusteringService>,
+    CacheManagerConfiguration<ClusteredCacheManager> {
+
+  private final String alias;
+  private final URI connectionUrl;
+
+  public ClusteringServiceConfiguration(final String alias, final URI connectionUrl) {
+    if (connectionUrl == null) {
+      throw new NullPointerException("connectionUrl");
+    }
+    this.alias = (alias == null ? "" : alias);
+    this.connectionUrl = connectionUrl;
+  }
+
+  public String getAlias() {
+    return alias;
+  }
+
+  public URI getConnectionUrl() {
+    return connectionUrl;
+  }
+
+  @Override
+  public Class<ClusteringService> getServiceType() {
+    return ClusteringService.class;
+  }
+
+  @Override
+  public CacheManagerBuilder<ClusteredCacheManager> builder(final CacheManagerBuilder<? extends CacheManager> other) {
+    return ClusteredCacheManagerBuilder.newCacheManagerBuilder(other).using(this);
+  }
+}
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
@@ -33,6 +33,7 @@ import java.net.URI;
  */
 // TODO: Should this accept/hold a *list* of URIs?
 // TODO: Is 'alias' *required*?
+// TODO: Add validation for connection URI(s)
 public class ClusteringServiceConfiguration
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<ClusteredCacheManager> {
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
@@ -32,25 +32,18 @@ import java.net.URI;
  * @author Clifford W. Johnson
  */
 // TODO: Should this accept/hold a *list* of URIs?
-// TODO: Is 'alias' *required*?
 // TODO: Add validation for connection URI(s)
 public class ClusteringServiceConfiguration
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<ClusteredCacheManager> {
 
-  private final String alias;
   private final URI connectionUrl;
 
-  public ClusteringServiceConfiguration(final String alias, final URI connectionUrl) {
+  public ClusteringServiceConfiguration(final URI connectionUrl) {
     if (connectionUrl == null) {
       throw new NullPointerException("connectionUrl");
     }
-    this.alias = (alias == null ? "" : alias);
     this.connectionUrl = connectionUrl;
-  }
-
-  public String getAlias() {
-    return alias;
   }
 
   public URI getConnectionUrl() {
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteredCacheConstants.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteredCacheConstants.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.config.xml;
+
+import java.net.URI;
+import java.net.URL;
+
+/**
+ * Constants shared between XML parsers used for clustered cache support.
+ *
+ * @author Clifford W. Johnson
+ */
+final class ClusteredCacheConstants {
+  /**
+   * Name of schema resource file.
+   */
+  static final String XSD = "ehcache-clustered-ext.xsd";
+  static final URL XML_SCHEMA = ClusteredCacheConstants.class.getResource("/" + XSD);
+  /**
+   * Namespace for cluster configuration elements.  Must match {@code targetNamespace} in <code>{@value #XSD}</code>.
+   */
+  static final URI NAMESPACE = URI.create("http://www.ehcache.org/v3/clustered");
+}
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
@@ -20,6 +20,7 @@ import org.ehcache.clustered.config.ClusteringServiceConfiguration;
 import org.ehcache.clustered.service.ClusteringService;
 import org.ehcache.spi.service.ServiceCreationConfiguration;
 import org.ehcache.xml.CacheManagerServiceConfigurationParser;
+import org.ehcache.xml.exceptions.XmlConfigurationException;
 import org.w3c.dom.Attr;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
@@ -27,6 +28,7 @@ import org.w3c.dom.NodeList;
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
 
 import javax.xml.transform.Source;
 import javax.xml.transform.stream.StreamSource;
@@ -84,13 +86,13 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
             final Attr urlAttribute = ((Element)item).getAttributeNode("url");
             final String urlValue = urlAttribute.getValue();
             try {
-              connectionUri = URI.create(urlValue);
-            } catch (IllegalArgumentException e) {
-              throw new RuntimeException(
+              connectionUri = new URI(urlValue);
+            } catch (URISyntaxException e) {
+              throw new XmlConfigurationException(
                   String.format("Value of %s attribute on XML configuration element <%s> in <%s> is not a valid URI - '%s'",
                       urlAttribute.getName(), fragment.getTagName(),
                       (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName()),
-                      connectionUri));
+                      connectionUri), e);
             }
           }
         }
@@ -98,7 +100,7 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
       // TODO: Validate connectionUri is valid URL with proper content
       return new ClusteringServiceConfiguration(alias, connectionUri);
     }
-    throw new RuntimeException(String.format("XML configuration element <%s> in <%s> is not supported",
+    throw new XmlConfigurationException(String.format("XML configuration element <%s> in <%s> is not supported",
         fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
   }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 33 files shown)


cljohnso:
The _max memory_ commit is expected to be replaced by something more finely-tuned.

clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
```diff
@@ -0,0 +1,81 @@
...
+    this(configuration, Collections.<Service>emptyList());
+  }
+
+  public ClusteredCacheManagerImpl(final Configuration configuration, final Collection<Service> services) {
+    super(configuration, services, true);
+  }
+
+  @Override
+  public Maintainable toMaintenance() {
+    return null;
```
lorban: I'd rather throw UnsupportedOperationException for now.

clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public void destroyCache(final String alias) throws CacheDurabilityException {
```
lorban: Same as above.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+
+import java.net.URI;
+
+/**
+ * Specifies the configuration for a {@link ClusteringService}.
+ *
+ * @author Clifford W. Johnson
+ */
+// TODO: Should this accept/hold a *list* of URIs?
```
lorban: This should be a Set of URIs IMHO.

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -0,0 +1,104 @@
...
+      for (int i = 0; i < childNodes.getLength(); i++) {
+        final Node item = childNodes.item(i);
+        if (Node.ELEMENT_NODE == item.getNodeType()) {
+          if ("connection".equals(item.getLocalName())) {
+            final Attr urlAttribute = ((Element)item).getAttributeNode("url");
+            final String urlValue = urlAttribute.getValue();
+            try {
+              connectionUri = URI.create(urlValue);
+            } catch (IllegalArgumentException e) {
+              throw new RuntimeException(
```
lorban: Throwing RuntimeException never is a brilliant idea especially since it is also done in other locations and there's no specialized exception. IMHO this is an indicator that a little bit of refactoring should happen here, but outside the context of this task.

core/src/main/java/org/ehcache/core/BaseCacheManager.java
```diff
@@ -0,0 +1,634 @@
...
+ *
+ * @author Clifford W. Johnson
+ */
+public abstract class BaseCacheManager implements CacheManager, ObservableCacheManager {
+
+  private final DefaultConfiguration configuration;
+  private final ClassLoader cacheManagerClassLoader;
+  private final boolean useLoaderInAtomics;
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
lorban: I think BaseCacheManager.class would be better than getClass()

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -111,76 +61,26 @@ public EhcacheManager(Configuration config) {
...
-  void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
+  protected void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
     boolean diskTransient = isDiskTransient(ehcache);
-    ehcache.close();
     if (diskTransient) {
       try {
         destroyPersistenceSpace(alias);
       } catch (CachePersistenceException e) {
-        LOGGER.debug("Unable to clear persistence space for cache {}", alias, e);
+        this.getLogger().debug("Unable to clear persistence space for cache {}", alias, e);
```
lorban: Why this.getLogger()? I also wonder if this shouldn't be a warning instead of a debug.

core/src/main/java/org/ehcache/spi/cache/ObservableCacheManager.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+import org.ehcache.core.events.CacheManagerListener;
+
+/**
+ * Specifies methods for a cache manager observable by the management framework.
+ *
+ * @author Clifford W. Johnson
+ */
+public interface ObservableCacheManager {
```
lorban: This should extend CacheManager IMHO.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -84,7 +85,9 @@ public synchronized void start(ServiceProvider serviceProvider) {
     cacheManager = serviceProvider.getService(CacheManagerProviderService.class).getCacheManager();
     scheduledExecutorService = serviceProvider.getService(ExecutionService.class).getScheduledExecutor(configuration.getCollectorExecutorAlias());
 
-    cacheManager.registerListener(this);
+    if (cacheManager instanceof ObservableCacheManager) {
```
lorban: If this test evaluates to false, this is going to make the cache manager silently un-monitorable. There should at least be a warning about that.

management/src/main/java/org/ehcache/management/registry/DefaultSharedManagementService.java
```diff
@@ -53,39 +54,42 @@
...
     final ManagementRegistryService managementRegistry = serviceProvider.getService(ManagementRegistryService.class);
     final Context cmContext = managementRegistry.getConfiguration().getContext();
-    final EhcacheManager ehcacheManager = serviceProvider.getService(CacheManagerProviderService.class).getCacheManager();
+    final CacheManager cacheManager = serviceProvider.getService(CacheManagerProviderService.class).getCacheManager();
 
-    ehcacheManager.registerListener(new CacheManagerListener() {
-      @Override
-      public void cacheAdded(String alias, Cache<?, ?> cache) {
-      }
+    if (cacheManager instanceof ObservableCacheManager) {
```
lorban: Same as above: if this evaluates to false you silently have no monitoring.

clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
cljohnso: `destroyCache` has nothing to do at this point .. should it really throw an `UnsupportedOperationException` now?

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
cljohnso: This was modeled after code in `org.ehcache.transactions.xml.TxCacheManagerServiceConfigurationParser` -- which throws `RuntimeException`.  Should we settle on something else?

core/src/main/java/org/ehcache/core/BaseCacheManager.java
cljohnso: A discussion with @chrisdennis led to this.  I'll follow any agreed upon model.

core/src/main/java/org/ehcache/core/EhcacheManager.java
cljohnso: The use of `debug` was pre-existing.  As for `Logger` access, I'll follow any agreed upon model.

ljacomet:
Can you get rid of the max memory commit? Something similar is now part of trunk

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
cljohnso: Hmmm ... looks like I dropped the `TODO` in one of my rebases.

This affects `DefaultCollectorService`, `DefaultSharedManagementService`, and `DefaultManagementRegistryService`.  Do we need the warning in each place?

cljohnso:
I'll remove the max memory commit at my next rebase.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
cljohnso: For a `Set`,  `ClusteringServiceConfigurationParser#parseServiceCreationConfiguration` will need to do a check for duplicates ... or not.  Should we fail a configuration with duplicate entries or just accept what fits in the set, first come/first served.  Should there be any attempt at resolving hosts during validation?  Should the set constructed from an XML configuration be _ordered_ (e.g. a `LinkedHashSet`) so the iteration order over the set will reflect the ordering in the XML?

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: It is a debug because there is not a lot to do here. If the file cannot be destroyed, it will be an issue only if you try re-creating in the same location.

core/src/main/java/org/ehcache/core/BaseCacheManager.java
ljacomet: If I understand the change here, the goal is to have the logging appear as coming from the sub type. Not sure I like it that much. Unless we decide that logs should be "functional" in nature, that is abstract the code getting executed ... but that seems weird.

build.gradle
```diff
@@ -79,6 +79,10 @@ subprojects {
     utils.fillManifest(manifest,"ehcache-${project.name}")
   }
 
+  test {
+    jvmArgs "-Xmx256m"
+  }
+
```
chrisdennis: Louis handled this in another commit so it should disappear on rebase I think.

clustered/client/src/main/java/org/ehcache/clustered/ClusteredCacheManagerImpl.java
chrisdennis: I'm happy with either... I imagine this going to see some significant refactoring when you and Louis start looking at the tighter integrations of store factories with resource and the various persistence mechanisms.

cljohnso: I think I'll leave `destroyCache` as-is (no action) for the time being -- there's presently nothing to destroy so the empty method isn't, technically, incorrect.  `toMaintenance`, on the other hand, will throw.

build.gradle
cljohnso: It will disappear.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
chrisdennis: I would stick with one URL here for the moment.  We know that AA and AP might change this, but for simplicity I'd stick with one until forced to change.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+    CacheManagerConfiguration<ClusteredCacheManager> {
+
+  private final String alias;
+  private final URI connectionUrl;
+
+  public ClusteringServiceConfiguration(final String alias, final URI connectionUrl) {
+    if (connectionUrl == null) {
+      throw new NullPointerException("connectionUrl");
+    }
+    this.alias = (alias == null ? "" : alias);
```
chrisdennis: Why would we support a null alias?

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+    this.connectionUrl = connectionUrl;
+  }
+
+  public String getAlias() {
+    return alias;
+  }
+
+  public URI getConnectionUrl() {
```
chrisdennis: Which is it?  URI or URL?

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
cljohnso: Don't know that we should.  Do we want to enforce a non-empty alias?

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteredCacheConstants.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Constants shared between XML parsers used for clustered cache support.
+ *
+ * @author Clifford W. Johnson
+ */
+final class ClusteredCacheConstants {
+  /**
+   * Name of schema resource file.
+   */
+  static final String XSD = "ehcache-clustered-ext.xsd";
```
chrisdennis: Do we ever use the string constant directly?

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -0,0 +1,104 @@
...
+      URI connectionUri = null;
+      final NodeList childNodes = fragment.getChildNodes();
+      for (int i = 0; i < childNodes.getLength(); i++) {
+        final Node item = childNodes.item(i);
+        if (Node.ELEMENT_NODE == item.getNodeType()) {
+          if ("connection".equals(item.getLocalName())) {
+            final Attr urlAttribute = ((Element)item).getAttributeNode("url");
+            final String urlValue = urlAttribute.getValue();
+            try {
+              connectionUri = URI.create(urlValue);
```
chrisdennis: You should be calling new URI here as urlValue isn't a compile time constant.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
cljohnso: From an XML point of view, it comes across as a URI.  From the programmatic specification, we could limit it to a URL ... we _are_ supposed to be able to connect to the thing, right?  Not only that, we could limit it to `http`/`https` URLs -- unless we want to designate a scheme of our own.

I have altered the extension schema to ensure the connection URL is a reasonably formed `http`/`https URL.  That will be included in the next commit.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
cljohnso: We can certainly do that and, when the time comes, handle the `Set` version as an extension to this class.  The extension schema (as it currently stands) will need to be changed to increase the `maxOccurs` for the `connection` element -- also upward compatible.

impl/src/main/java/org/ehcache/config/builders/CacheManagerBuilder.java
```diff
@@ -68,37 +70,46 @@ public T build() {
...
+
   @SuppressWarnings("unchecked")
-  T cast(EhcacheManager ehcacheManager) {
+  private T cast(EhcacheManager ehcacheManager) {
     return (T) ehcacheManager;
   }
 
   public <K, V> CacheManagerBuilder<T> withCache(String alias, CacheConfiguration<K, V> configuration) {
-    return new CacheManagerBuilder<T>(this, configBuilder.addCache(alias, configuration));
+    return this.newBuilder(configBuilder.addCache(alias, configuration));
```
chrisdennis: Why all the 'this' references?

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
cljohnso: Doesn't change much but I see your point ... will change.

impl/src/main/java/org/ehcache/config/builders/CacheManagerBuilder.java
```diff
@@ -212,4 +230,52 @@ T cast(EhcacheManager ehcacheManager) {
...
+   * @param <T> the type of the desired {@code CacheManager}
+   *
+   * @return a new {@code CacheManager} instance of the type designated by {@code specialization};
+   *      the returned {@code CacheManager} is <b>not</b> initialized
+   *
+   * @throws ClassCastException if the {@link CacheManagerBuilder} associated with the
+   *        {@link CacheManagerConfiguration} instances appearing in the configuration does
+   *        not build a {@code CacheManager} of the {@code specialization} type
+   */
+  public static <T extends CacheManager> T newSpecializedCacheManager(Class<T> specialization, Configuration config)
```
chrisdennis: Does this need to be public?

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
cljohnso: BTW ... right now, the call path to this method makes no provisions for directly handling faulty input.  Should it?  I'm throwing a `RuntimeException` because that was how it was handled elsewhere.  As Ludovic mentioned elsewhere, `RuntimeException` isn't very "expressive".

impl/src/main/java/org/ehcache/config/builders/CacheManagerBuilder.java
cljohnso: It's presently used from the `org.ehcache.clustered.ClusteredCacheManagerBuilder` class ...

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteredCacheConstants.java
cljohnso: It was, at one time, used in a test -- that is no longer the case.  It is, however, referred to in javadoc in `ClusteringServiceConfigurationParser` and, assuming we continue the transaction model and use the same extension schema for `<cache>` extensions, will be referred to in the javadoc `CacheServiceConfigurationParser` implementation handling the `<cache>` extensions.

impl/src/main/java/org/ehcache/config/builders/CacheManagerBuilder.java
cljohnso: For of habit from many years spent programming Java where `this` was a requirement ...

core/src/main/java/org/ehcache/core/BaseCacheManager.java
cljohnso: @chrisdennis ... would you like to weigh in on this?

I came from a shop where we didn't want to expose structural/compositional details in our logging ... we went as far as defining loggers at the **interface** level -- logical vs "physical" logging, if you will.  Like I said, I'll follow any agreed upon model.

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
chrisdennis: I don't think a RuntimeException is terrible here since this is startup time code... but maybe we should standardize on something more specific?  @ljacomet, thoughts?

core/src/main/java/org/ehcache/core/BaseCacheManager.java
chrisdennis: To my mind using the concrete type for the logger makes sense... but it does make the logging code itself more convoluted.  For that reason maybe either a private logger per type in hierarchy makes more sense... but either way we should be consistent.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
chrisdennis: We are supposed to connect to it... but generally I prefer sticking to URI until the last minute, since URLs have a nasty tendency to trigger DNS calls at the wrong moments.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
chrisdennis: That depends... what does the alias mean?  Does it tie a cache to the clustering service that backs it, or does it identify the name of the entity to connect to?

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
mathieucarbou: @cljohnso @lorban @chrisdennis @ljacomet : I have an alternative solution:

When I created the `CacheManagerProviderService`, the goal was to be able within the services to access a cache manager. Firstly because if the m&m needs (to listen on some events).

So perhaps instead of checking whether the CM is an instance of Observable, and silently discard a feature the user explicitly installed, why don't we change the `CacheManagerProviderService ` so that we have something like this:

```
CacheManager cacheManager = serviceProvider.getService(CacheManagerProviderService.class).getCacheManager();

ObservableCacheManager cacheManager = serviceProvider.getService(CacheManagerProviderService.class).getObservableCacheManager();
```
`getObservableCacheManager` would do a sanity check and throw an exception if the CM is not observable.

or...

`ObservableCacheManager cacheManager = serviceProvider.getService(CacheManagerProviderService.class).getCacheManager(ObservaleCacheManager.class);`

and the method `getCacheManager(Class<T extends CacheManager>)` would "unwrap" to the wanted class or throw.

The idea behind this is that a service gets what it needs, and if he can't then it crashes simply because it won't work anyway. I think it's not the responsibility of the service to check / know all the different potential flavors of a CM (because at the end there could be more than just the Observable flavor). The service just need to ask what it needs to work properly, and the init just need to fail if what is needed cannot be provided. It would be the responsibility of the developer to fix such error.

Management services are optional (theoretically because jsr107 depends on it at the moment, at least until #747 is fixed). So if somebody register such services within a CM that is not observable, it should fail at init (so when the cache manager provider would throw an exception), because the required feature won't work.

Let me know what you decided: I am interested in following the changes in the management module :-)

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
lorban: FIY I was thinking about something simple like XMLConfigurationException that extends from RuntimeException.

core/src/main/java/org/ehcache/core/EhcacheManager.java
lorban: Regarding debug vs warn, I generally am in favor of warning the user when something not fatal but worth looking into happened like here.

core/src/main/java/org/ehcache/core/BaseCacheManager.java
cljohnso: For my two cents, I think using `BaseCacheManager` (the "parent" class name) hides information potentially useful in problem source identification -- what we really need to know is the _actual_ class responsible for getting is here (`EhcacheManager` or `ClusteredCacheManager` in this case) ... simply using `BaseCacheManager` hides that so we have to rely on clues obtained from elsewhere to know which of the cache manager implementations is involved.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
cljohnso: Yes ... for some reason, `equals` wants to resolve the host ... would have been better if that wasn't part of the `URL.equals` contract.

core/src/main/java/org/ehcache/core/EhcacheManager.java
cljohnso: I agree with @lorban on this one ... something non-fatal happened that might deserve attention by the user.  We're logging at an info level cache creation, cache closure, and (normal) cache removal -- so this can't be a concern about noise.

This is pre-existing code but I'm going to change this to warn unless someone has a strong objection.

core/src/main/java/org/ehcache/spi/cache/ObservableCacheManager.java
cljohnso: Anyone else want to weight in on this?  @chrisdennis @ljacomet ?  I have no objection to making this extend `CacheManager` ...

chrisdennis: No objections from me.

clustered/client/src/main/java/org/ehcache/clustered/config/ClusteringServiceConfiguration.java
cljohnso: On the other hand, should I change the field holding the connection URI to a `Set` or `List` and changing the associated getter?  I believe the ultimate target is to consume more than one connection URL (for AA & AP) -- have we any concern/desire to position the consumers of this bit of configuration for that eventuality?

clustered/client/src/main/java/org/ehcache/clustered/config/xml/ClusteringServiceConfigurationParser.java
cljohnso: I've added an `XmlConfigurationException` to the **xml** module ... looks like there are a number of places where it might be brought to bear.  Should I create an issue to cover that work?

chrisdennis: Yes.

cljohnso:
Rebased, made some review adjustments, and dropped the max memory commit.

In this update, I did **not** address `ClusteringServiceConfiguration` _alias_ handling or the `CacheManagerProvider` service changes suggested by @mathieucarbou -- which @chrisdennis has recommended handling in this PR.

cljohnso:
The latest commit addresses `ClusteringServiceConfiguration` _alias_ and the `CacheManagerProviderService` changes.

After discussion with @chrisdennis, `ObservableCacheManager` is renamed to `InternalCacheManager`.  The intent is that `InternalCacheManager` be the interface consumed internally (by services) and includes methods needed internally that we do not wish to expose on `CacheManager` -- such as those needed by the management layer.  `CacheManagerProviderService`  is changed to return an `InternalCacheManager` instead of `CacheManager` or one of the concrete implementations.  This was considered a better approach than introducing a casting-based approach.

The `ClusteringServiceConfiguration` _alias_ field was dropped -- it will be added in the future as needed (though, given other identification mechanisms, it's not likely to be necessary).

`ObservableCacheManager`/`InternalCacheManager` was relocated -- a package change (adding 'core' was previously missed).

cljohnso:
@chrisdennis @lorban @ljacomet 
We have yet to resolve the logging discussion anchored against `BaseCacheManager` ...

chrisdennis:
I'm happy with this.  @lorban @ljacomet I'd prefer if one of you guys gave this another pass and merged if you are happy too.

mathieucarbou:
:+1:

PR #773, opened 2016-02-21
Issue #724 Fix empty source and javadoc jars

ljacomet:
Contains a ugly hack. No issue for me replacing it with something better ... but I would rather have a hacked working build for now than a broken one.


ljacomet:
Updated PR uses what I believe is a clean design ... even though it remains rough on edges.

Instead of applying configuration from the root gradle build file, deploy and distribution configurations have been moved to be project plugins that can be applied per module.

The rough bit is that the distribute plugin must be applied after dependencies declaration.

ljacomet:
Created #775 for the occasionaly failing test

Issue #782, opened 2016-02-24
higher tier stores stats of 3-tier caches not connected in the ContextManager


lorban:
In a 3-tier cache (e.g.: heap / offheap / disk) the two highest stores' statistics do not appear in the stats framework's ContextManager.

Issue #775, opened 2016-02-22
Investigate and fix sometimes failing test


ljacomet:
The following test `CountSizedOnHeapStoreByRefTest.testIteratorExpiryHappensUnderExpiredKeyLockScope` fails at times. As it is a test shared between different configurations, it may appear under a different class.

It should be investigated and fixed so that it no longer has occasional failures.

ljacomet:
updates to the test have been posted which will do a thread dump in case of failure. I will leave this issue open for a few days to see if we can get it to fail again. Otherwise I will close it assuming the increased timeout was the fix.

ljacomet:
Closing as issue no longer occurred.

ljacomet:
Failed again: https://ehcache.ci.cloudbees.com/job/ehcache3-pull-requests/1780/jdk=Oracle%20JDK%201.8(latest),label_exp=standard/testReport/junit/org.ehcache.impl.internal.store.heap/CountSizedOnHeapStoreByValueTest/testIteratorExpiryHappensUnderExpiredKeyLockScope/

ljacomet:
Actually that's a different test, most likely failing for the same reason - timeout too short. Will adapt in the same way.

PR #783, opened 2016-02-24
#782 associate higher and lower caching tiers' stats with ContextManager

lorban:


ljacomet:
That build failure is a Java 1.6 compile failure

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CompoundCachingTier.java
```diff
@@ -67,6 +68,9 @@ public void onInvalidation(final K key, final Store.ValueHolder<V> valueHolder)
         }
       }
     });
+
+    StatisticsManager.associate(higher).withParent(this);
+    StatisticsManager.associate(lower).withParent(this);
```
chrisdennis: Calling static method through a subclass reference... might be better to call ContextManager.associate(...)

ljacomet:
Failure is racy test, already filed as #775

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CompoundCachingTier.java
lorban: StatisticsManager.associate() is used everywhere. Should I change all refs while I'm at it?

chrisdennis: I guess not then... the cat's out of the bag... go with convention.

PR #788, opened 2016-02-24
Changed to coderay for source-highlighting, including bundling the CS…

GaryWKeim:
…S so things work when offline.

Set stylesdir to ./css and moved all stylesheets there.
Bundled font-awesome styles and fonts.
Changed generate script to remove html5 from $siteDocDir.
Altered asciidoc files to properly place property declarations and header content.
Ignore emacs crumbs.


terracotta-org:

GaryWKeim:
Replaced by #795

Issue #791, opened 2016-02-25
OnHeapStore.MapWrapper should store the cache's key as such when identity copier is configured


lorban:
Currently, `MapWrapper` uses `OnHeapKey<K>` as key which forces the creation of a new object on each `get()`. Such behavior is required because of potential key copies.

In case store-by-ref is configured (which is the default) `IdentityCopier` gets wired in and does not perform key copies. We could directly use the cache's key directly as the `MapWrapper` key in that case, saving an object allocation per `get()`.

PR #792, opened 2016-02-25
Issue #791 Extract MapWrapper from OnHeapStore

ljacomet:
This enables two different implementation:
* Existing one which store OnHeapKey<K> to handle copy semantics
for the key
* New one which acts as a passthrough when the key copier is an
IdentityCopier

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/heap/Backend.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/heap/Backend.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


ljacomet:
@lorban asked about renaming `MapWrapper` but I could not find something so far. Ideas welcome!

Also, this does not have a lot of tests but my feeling is that the code is simple enough and type constrained so that it works. And also all the `OnHeapStore` testing is effectively testing it.

ljacomet:
Again build failure to blame on #775

impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleMapWrapper.java
```diff
@@ -0,0 +1,99 @@
...
+import org.ehcache.impl.internal.store.heap.holders.OnHeapValueHolder;
+
+import java.util.Comparator;
+import java.util.Map;
+import java.util.Random;
+
+/**
+ * Simple passthrough map wrapper, no key translation
+ */
+class SimpleMapWrapper<K, V> implements MapWrapper<K, V> {
```
chrisdennis: Would it really be so bad to eliminate this class and have the map implement map wrapper directly?

ljacomet: I thought about it but I am not too fond of the idea mostly because it pushes that code further away from its original source.
And also because the current decoupling is a nice place where to fix racy clear when using byte size heap

chrisdennis: Fair enough.

lorban:
Do we plan on keeping that ugly `MapWrapper` name?

PR #795, opened 2016-02-25
Changed to coderay for source-highlighting, including bundling the CS…

GaryWKeim:
…S so things work when offline.

Set stylesdir to ./css and moved all stylesheets there.
Bundled font-awesome styles and fonts.
Changed generate script to remove html5 from $siteDocDir.
Altered asciidoc files to properly place property declarations and header content.
Ignore emacs crumbs.


ljacomet:
I need to investigate more but in the current status this breaks website generation.

GaryWKeim:
Can you please show me how. I ran the ./docs/generate script, copied the generated content into place, and ran a static web server and all seemed fine.

ljacomet:
Have a look at the README in https://github.com/ehcache/ehcache.org-site

Issue #760, opened 2016-02-18
Creation of 'clustered' store implementations within the current store provider service mechanism


chrisdennis:
We need to figure out how a cache configured as clustered ends up with a clustered store without leaking details of clustered in to the unclustered code.  As part of this we also need to tackle the desire to support at a minimum clustered and unclustered caches in the same manager.  We may also want to consider supporting both clustered and disk persistent caches in the same manager (although this seems at first glance to be significantly more complicated).

PR #796, opened 2016-02-26
Back to single CacheManager impl

ljacomet:
The goal is to move back to a single `CacheManager` implementation. This is achieved by introducing a specific `Service` subtype for _persistable_ resources.
It enables the `CacheManager` implementation to simply deal with these without ever knowing what the resource is and the specific service implementation.
It is rough on the edges but should be a good indication of what can be achieved.

Comments and remarks expected!


ljacomet:
Sorry for that compile error - fixed now

api/src/main/java/org/ehcache/spi/service/PersistableResourceService.java
```diff
@@ -0,0 +1,54 @@
...
+import org.ehcache.exceptions.CachePersistenceException;
+
+import java.util.Collection;
+
+/**
+ * PersistableResourceService
+ */
+public interface PersistableResourceService extends Service {
+
+  boolean handlesResourceType(ResourceType resourceType);
```
chrisdennis: This looks similar to what Clifford and I were discussing today... except that we had started extending these `PersistableResourceService` instances to have two additional properties:

1. They create stores (which seemed to indicate that this method would consume Set<ResourceType> rather than just a single resource)
2. They have state machines of their own... thereby decoupling the maintainable state of the local cache manager from the maintainable state of any one of a number of potential maintainable states of associated persistence services.

ljacomet:
Added bits to make `PersistableResourceService` _maintenance_ aware

api/src/main/java/org/ehcache/spi/service/PersistableResourceService.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Specific interface for services that are dedicated to handling a {@link ResourceType} which is
+ * {@link ResourceType#isPersistable() persistable}.
+ */
+public interface PersistableResourceService extends Service {
+
+  /**
+   * Start this service for maintenance, based on its default configuration.
+   */
+  void startForMaintenance();
```
chrisdennis: Do we want to decouple the "three-state-lifecycle"-ness of a service from it's "PersistableResource"-ness?

chrisdennis: Shouldn't we be passing the ServiceProvider in here as well?  In fact do we want to have a pseudo-three phase startup.

1. Create services: new FooService()
2. Wire services: wire(ServiceProvider services)
3. Lifecycle transition services: start(), maintenance(), stop()

lorban: I'm in favor of decoupling the "persistability" from the "maintainability". But just renaming the interface to something like `MaintainableService` should suffice IMHO.

ljacomet: Agreed with the `MaintainableService` abstraction.

Still thinking about the proportion on the startup phases.

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -157,22 +675,95 @@ public void close() {
...
+    for (MaintainableService service : services) {
+      service.startForMaintenance(provider);
+    }
   }
 
-  private LocalPersistenceService startPersistenceService() {
-    LocalPersistenceService persistenceService = serviceLocator.getService(LocalPersistenceService.class);
-    persistenceService.start(serviceLocator);
-    return persistenceService;
+  private ServiceProvider<MaintainableService> getMaintenableServiceProvider() {
```
cljohnso: Sneaking in some French? (_Maintenable_)  ;-)

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -16,32 +16,67 @@
```
cljohnso: The differences in this file are easier to observe using 

```
git difftool 3ad87cafc20586140675ce179e9ce90ba993a0ed:core/src/main/java/org/ehcache/core/BaseCacheManager.java HEAD:core/src/main/java/org/ehcache/core/EhcacheManager.java
```

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: Indeed :-)

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -63,76 +110,548 @@ public EhcacheManager(Configuration config, Collection<Service> services) {
...
+      @Override
+      public void close() {
+        storeProvider.releaseStore(store);
+      }
+    });
+
+    return store;
+  }
+
+  private PersistableResourceService getPersistableResourceService(ResourceType resourceType) {
```
cljohnso: Confirming a presumption -- the current implementation presumes a single service supporting a given `ResourceType` ... support for multiple services for a given `ResourceType` is for #760.

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -63,76 +110,548 @@ public EhcacheManager(Configuration config, Collection<Service> services) {
...
+      throw new NullPointerException("Alias cannot be null");
+    }
+    this.getLogger().info("Destroying Cache '{}' in {}.", alias, simpleName);
+    removeAndCloseWithoutNotice(alias);
+    destroyPersistenceSpace(alias);
+    this.getLogger().info("Cache '{}' is successfully destroyed in {}.", alias, simpleName);
+  }
+
+  private void destroyPersistenceSpace(String alias) throws CachePersistenceException {
+    Collection<PersistableResourceService> services = serviceLocator.getServicesOfType(PersistableResourceService.class);
```
cljohnso: Are there any concerns about ordering of destruction of multiple persistent services?

api/src/main/java/org/ehcache/spi/ServiceProvider.java
```diff
@@ -23,17 +23,18 @@
  * This acts as a repository for {@link Service} instances, that can be used to
  * look them up by type.
  *
+ * @param <T> A bound on service types this provider can return
  * @author Alex Snaps
  */
-public interface ServiceProvider {
+public interface ServiceProvider<T extends Service> {
```
cljohnso: There are a number of `ServiceProvider` uses to which the generic type parameter was not added.  Oversight?  Intentional?

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: That's a good point. The alt option would be to pass in the `ResourcePool`

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: I certainly hope not. What are you hinting at?

api/src/main/java/org/ehcache/spi/ServiceProvider.java
ljacomet: Oversight ... I was relying on the compiler errors only

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: After discussion with @cljohnso the `ResourceType` is enough to handle that future.

Issue #798, opened 2016-02-29
XAStore.Provider and TransactionManagerProvider dependencies are in the wrong direction


ljacomet:
While working on proper handling of resource types and the relation with `Store.Provider`,  I realised that the current transaction support has its dependencies backwards.

Currently the `XAStore.Provider` declares a dependency on the `TransactionManagerProvider` while only the second has a mandatory `ServiceCreationConfiguration`.

If we were to reverse the dependency, it enables to get rid of the useless `XAStoreProviderConfiguration` and matching XML entry.

ljacomet:
Added to this change, proper ranking needs to be done as introduced in #811 and discussed in PR #838.

ljacomet:
Requires work done in #811

PR #800, opened 2016-03-01
Issue #683 Move transactions classes to internal package

ljacomet:
* Reduces code exposure as internal package are private in OSGi and
javadoc will not be created for them
* Fixed export package declaration in XML module

This was a missing set of changes from the original work on #683


chrisdennis:
Assuming tests pass this looks reasonable to me.

cljohnso:
There is a little _leakage_ around `XAStore.Provider` from the `configuration` package ... is that of any concern?

ljacomet:
I do not think that's an issue, especially as it will be transient. Once #798 is resolved, it will go away.

PR #801, opened 2016-03-01
Suggestion: use standard logger name for logging

mathieucarbou:
Logging systems use package separator to filter

`<logger name="org.ehcache.spi" level="WARN"/>`

So a suggestion would be to use that so that we can filter the different cache names in logs, if you are using cache names in logger names.

I saw it in `UserManagedCacheBuilder `. I don't know if other classes needs to be changed though...


lorban:
Why not? I like the idea of keeping the dot ('.') as the separator.

ljacomet:
I agree that this makes sense. For completeness, `org.ehcache.core.EhcacheManager#createNewEhcache` should be adapted as well then.

mathieucarbou:
closing PR - created branch at wrong place. Will reopen and add the other modification :-)

Issue #776, opened 2016-02-22
Improve javadoc


ljacomet:
While releasing m5 javadoc, I noticed that outside of the api module, there are a number of classes that are user facing that lack proper documentation.
This includes package level documentation.

ljacomet:
First PR (#802) pushed out for the builders javadoc

ljacomet:
From @lorban, `Store.replace(K, V)` javadoc needs to be looked at.

lorban:
`Store.computeIfPresent()` and `Store.bulkComputeIfAbsent()` should be looked at too, especially their returned value contract.

ljacomet:
Create #844 to track further javadoc improvements.

PR #802, opened 2016-03-02
Issue #776 Builders javadoc

ljacomet:
* Remove javadoc for `internal` packages
* Javadoc for builders
* Misc clean ups

This is a first part of the work needed for #776, more to come.

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
+++ b/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
@@ -84,12 +84,12 @@ class ConfigurationMerger {
 
       builder = handleStoreByValue(jsr107Configuration, builder, cacheName);
 
-      final boolean useJsr107Expiry = builder.hasDefaultExpiry();
-      if (useJsr107Expiry) {
+      final boolean hasConfiguredExpiry = builder.hasConfiguredExpiry();
+      if (hasConfiguredExpiry) {
+        LOG.info("Cache {} will use expiry configuration from template {}", cacheName, templateName);
+      } else {
         expiryPolicy = initExpiryPolicy(jsr107Configuration);
         builder = builder.withExpiry(expiryPolicy);
-      } else {
-        LOG.info("Cache {} will use expiry configuration from template {}", cacheName, templateName);
       }
 
       boolean useEhcacheLoaderWriter;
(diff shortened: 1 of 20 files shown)


107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -111,13 +111,13 @@
 
       CacheConfiguration<K, V> cacheConfiguration = builder.build();
 
-      if (!useJsr107Expiry) {
+      if (hasConfiguredExpiry) {
```
lorban: Why changing the test from `(!useJsr107Expiry)` to `(hasConfiguredExpiry)` instead of `(!hasConfiguredExpiry)`? An overlook maybe?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -111,13 +111,13 @@
...
         expiryPolicy = new EhcacheExpiryWrapper<K, V>(cacheConfiguration.getExpiry());
       }
 
       return new ConfigHolder<K, V>(
           new CacheResources<K, V>(cacheName, loaderWriter, expiryPolicy, initCacheEventListeners(jsr107Configuration)),
-          new Eh107CompleteConfiguration<K, V>(jsr107Configuration, cacheConfiguration, !useJsr107Expiry, useEhcacheLoaderWriter),
+          new Eh107CompleteConfiguration<K, V>(jsr107Configuration, cacheConfiguration, hasConfiguredExpiry, useEhcacheLoaderWriter),
```
lorban: Same as above, or have you spotted a bug?

xml/src/test/java/org/ehcache/xml/FromTemplateCacheConfigurationBuilderDefaultTest.java
```diff
@@ -38,7 +38,7 @@ public void setUp() throws Exception {
   }
 
   @Test
-  public void testDefaultExpiry() throws Exception {
-    assertThat(minimalTemplateBuilder.hasDefaultExpiry(), is(true));
+  public void testNoConfiguredExpiry() throws Exception {
+    assertThat(minimalTemplateBuilder.hasConfiguredExpiry(), is(false));
```
lorban: What's the story behind this test change? Is that related to the inverted condition above?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Because a configured expiry means we should not use the JSR-107 provided one. This stems from changing method name (and contract) on the builder: `CacheConfigurationBuilder.hasConfiguredExpiry()` vs the former `CacheConfigurationBuilder.hasDefaultExpiry`
When I was writing the javadoc, expressing the contract with default name was really cumbersome and felt way more natural with changing the name and meaning of `true` and `false`.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Same as above indeed - so not a bug but a desired change

xml/src/test/java/org/ehcache/xml/FromTemplateCacheConfigurationBuilderDefaultTest.java
ljacomet: Correct

Issue #799, opened 2016-03-01
EhcacheOffHeapBackingMap.computeIfPinned()  too eagerly deserializes the OffHeapValueHolder's value


lorban:
`AbstractOffHeapStore.flush()` calls `EhcacheOffHeapBackingMap.computeIfPinned()` which fully deserializes the `OffHeapValueHolder`'s value even if the latter only is necessary when the flushed value expires *and* there is a registered cache event listener. The expiration check could be done solely with the `OffHeapValueHolder`'s metadata, saving the enormous cost of object deserialization.

lorban:
This relates to #797 but is more important as more perf is lost here than in the other compute methods.

ljacomet:
Requires offheap-store release (most likely 2.2.1) to fix a bug

Issue #790, opened 2016-02-25
OnHeapStore.setAccessTimeAndExpiryThenReturnMapping() calls valueHolder.value()


lorban:
This is required per the `Expiry` contract to pass the value to the expiry instance.

The nasty drawback is that this makes deserialization happen twice: once to pass to the expiry and once in `Ehcache.get()` to return to the user call.

Deserializing twice has obvious disastrous performance consequence, especially since our default `Expiry` implementation has no use for the deserialized value.

One thing that could be done is change the `Expiry` contract to pass `Holder<V>` instead of `V` to its methods so that only the expiries making use of the value would pay the extra deserialization price.

ljacomet:
So this is not as easy as it sounds.

* To be really efficient, not only the expiry contract has to be adapted, the events one too.
 * to be fair the event one could be an internal adaptation only, not exposed to `CacheEventListener`
* Doing this change actually adds overhead on create events which usually have the value in _clear_ and it would now have to be wrapped
* Generics are a mess :cry:

ljacomet:
I will soon push a PR for these changes. I really do not like the results and what we have to do to enable this.

So I will be looking for conversation on this topic and most likely wait for all others perf issues to be resolved and see how much we are talking about exactly here.

ljacomet:
Awaiting more performance results after #777, #787 and #799 have been applied.

ljacomet:
Realised that the changes in the initial version of the pull request were badly broken for offheap where value holders are now lazy. So the change is that in all methods where the value needs to be returned, it has to be forced to deserialize or detached because expiry computation no longer triggers it. And remember that off heap will only deserialize a value once since the reference has a life limited to the invoking thread. So while there will be differences, I am afraid they will be minimal.

On heap the discussion is a bit different because values are deserialized at each invocation indeed. But that is limited to use cases where the heap tier is configured with a `SerializingCopier` which is clearly NOT the recommended way if you care for maximum performance.

So all this really makes me doubt this PR even more.

lorban:
Except that JSR107 caches enable store-by-value by default, I do agree with the above.

We are close to our last opportunity to change the API to make this possible, even if we decide to always deserialize for now and postpone the optimization.

Food for thoughts...

chrisdennis:
[...]

PR #803, opened 2016-03-02
Issue #799 Lazy OffHeapValueHolder deserialization

ljacomet:
This is a proposition for a solution to issue #799.
It was refused back when in the context of #369 as the number of deserialisation that we gained were not that high.
However in the face of performance test results, I believe it is worth having the discussion again.

Note that the real effectiveness of this change depends on the outcome of #790

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/offheap/EhcacheConcurrentOffHeapClockCache.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/offheap/EhcacheConcurrentOffHeapClockCache.java
@@ -21,6 +21,7 @@ import org.ehcache.config.EvictionVeto;
 import org.ehcache.function.BiFunction;
 import org.ehcache.function.Function;
 
+import org.ehcache.impl.internal.store.offheap.factories.EhcacheSegmentFactory;
 import org.terracotta.offheapstore.MetadataTuple;
 import org.terracotta.offheapstore.concurrent.AbstractConcurrentOffHeapCache;
 import org.terracotta.offheapstore.pinning.PinnableSegment;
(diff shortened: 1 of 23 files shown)


ljacomet:
Latest update of this PR is not ready for merge. It even triggers a potential off heap bug.

ljacomet:
This is now up to date and ready for review prior to merge.

impl/src/main/java/org/ehcache/impl/internal/store/offheap/EhcacheConcurrentOffHeapClockCache.java
```diff
@@ -21,6 +21,7 @@
 import org.ehcache.function.BiFunction;
 import org.ehcache.function.Function;
 
+import org.ehcache.impl.internal.store.offheap.factories.EhcacheSegmentFactory;
```
chrisdennis: unused import

impl/src/main/java/org/ehcache/impl/internal/store/offheap/LazyOffHeapValueHolder.java
```diff
@@ -0,0 +1,133 @@
...
+
+  @Override
+  public V value() {
+    forceDeserialization();
+    return value;
+  }
+
+  @Override
+  public ByteBuffer getBinaryValue() throws IllegalStateException {
+    if (mode == Mode.DETACHED) {
```
chrisdennis: Reads nicely if this is written as `if (isBinaryValueAvailable())`.

chrisdennis:
Line 884 of AbstractOffHeapStore should use `TimeUnit.MILLISECONDS` not `OffHeapValueHolder.TIME_UNIT`.

ljacomet:
Indeed, I will fix that issue as well.

Issue #787, opened 2016-02-24
Make OnheapStore.get() call into MapWrapper.get() instead of  MapWrapper.computeIfPresent()


lorban:
According to the investigation done in #507, this is a big deal for performance.

PR #804, opened 2016-03-02
Issue 777

AbfrmBlr:
WIth XA tests failing.
Added two interfaces JSRIntegrableCache and Hookable.
Store apis return StoreOpStatus , which does not look good.


api/src/main/java/org/ehcache/spi/Hookable.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.spi;
+
+/**
+ * @author Abhilash
+ *
+ */
+public interface Hookable {
```
ljacomet: I would rename that to `InternalCache` and fold in the `JSRIntegrableCache` and for sure move it out of API module.

ljacomet: And it probably should extend `UserManagedCache` as well. This will enable to get rid of a large number of the casts introduced in various places.

ljacomet:
This PR needs a rebase which might be a bit messy given some of the changes that came and went around `CacheManager` implementations.

core/src/main/java/org/ehcache/core/spi/cache/Store.java
```diff
@@ -85,11 +88,12 @@
    *
    * @param key   key with which the specified value is to be associated
    * @param value value to be associated with the specified key
+   * @return <tt>true</tt> if the mapping is actually added
```
ljacomet: Javadoc update needed

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -217,44 +162,29 @@ public V apply(final K k) {
...
 
     try {
-      ValueHolder<V> computed = store.compute(key, remappingFunction);
-      if (computed != null) {
-        putObserver.end(PutOutcome.ADDED);
-      } else {
+      PutStatus status = store.put(key, value);
+      switch (status) {
+      case PUT:
+        putObserver.end(PutOutcome.PUT);
```
ljacomet: I would use the `PutStatus` enum as the statistics value and get rid of `PutOutcome`, which would simplify a lot this code by getting rid of the switch.

Applies to other methods similarly updated.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -375,8 +383,16 @@ public void remove(final K key) throws CacheAccessException {
       map.computeIfPresent(key, new BiFunction<K, OnHeapValueHolder<V>, OnHeapValueHolder<V>>() {
         @Override
         public OnHeapValueHolder<V> apply(K mappedKey, OnHeapValueHolder<V> mappedValue) {
-          removedValue.set(mappedValue);
-          eventSink.removed(mappedKey, mappedValue.value());
+
+          if (mappedValue != null && mappedValue.isExpired(now, TimeUnit.MILLISECONDS)) {
```
ljacomet: `computeIfPresent` means `mappedValue` will always be non null

impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -294,8 +309,16 @@ public void remove(K key) throws CacheAccessException {
       backingMap().computeIfPresent(key, new BiFunction<K, OffHeapValueHolder<V>, OffHeapValueHolder<V>>() {
         @Override
         public OffHeapValueHolder<V> apply(K mappedKey, OffHeapValueHolder<V> mappedValue) {
-          removed.set(true);
-          eventSink.removed(mappedKey, mappedValue.value());
+
+          if (mappedValue != null && mappedValue.isExpired(now, TimeUnit.MILLISECONDS)) {
```
ljacomet: `computeIfPresent` means `mappedValue` will always be non null

impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -294,8 +309,16 @@ public void remove(K key) throws CacheAccessException {
...
+            onExpiration(key, mappedValue, eventSink);
```
ljacomet: use `mappedKey` here

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -375,8 +383,16 @@ public void remove(final K key) throws CacheAccessException {
...
+            onExpiration(key, mappedValue, eventSink);
```
ljacomet: use `mappedKey` here

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -482,17 +502,21 @@ public boolean remove(final K key, final V value) throws CacheAccessException {
       if (removedValue.get() != null) {
         conditionalRemoveObserver.end(StoreOperationOutcomes.ConditionalRemoveOutcome.REMOVED);
         decrementCurrentUsageInBytesIfRequired(removedValue.get().size());
-        return true;
+        return RemoveStatus.REMOVED;
       } else {
         conditionalRemoveObserver.end(StoreOperationOutcomes.ConditionalRemoveOutcome.MISS);
```
ljacomet: Since we are anyway tracking the different between key present and key missing, I would also update the `conditionalRemoveObserver` to take a `RemoveStatus`, simplifying the code here.

This may apply to others methods modified in a similar way.

core/src/main/java/org/ehcache/core/Ehcache.java
lorban: I wouldn't do that as this would bastardize the statistics types and would make M&M work slightly harder.

If you like this idea, think about doing it the other way: returning a `PutStatus` from put but IMHO this would couple the `Store` contract with the statistics one. I personally prefer this switch statement.

impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
AbfrmBlr: Ah, that was there to make findbugs happy

AbfrmBlr: it failed even for jdk6

AbfrmBlr:
Rebased, all test are passing.

AbfrmBlr:
Addresses #787 as well

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -172,7 +173,7 @@ private void removeCache(final String alias, final boolean removeFromConfig) {
     statusTransitioner.checkAvailable();
     final CacheHolder cacheHolder = caches.remove(alias);
     if(cacheHolder != null) {
-      final Ehcache<?, ?> ehcache = cacheHolder.retrieve(cacheHolder.keyType, cacheHolder.valueType);
+      final Cache<?, ?> ehcache = cacheHolder.retrieve(cacheHolder.keyType, cacheHolder.valueType);
```
ljacomet: Using `InternalCache` here and elsewhere will clean up a lot of casts

core/src/main/java/org/ehcache/core/PersistentUserManagedEhcache.java
```diff
@@ -46,14 +48,18 @@
 
   private final StatusTransitioner statusTransitioner;
   private final Logger logger;
-  private final Ehcache<K,V> ehcache;
+  private final Cache<K,V> ehcache;
```
ljacomet: Again, using `InternalCache` should remove the need for casts

core/src/main/java/org/ehcache/core/spi/cache/StoreOpStatus.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @author Abhilash
+ *
+ */
+public interface StoreOpStatus {
+
+  /**
+   * Put operation status
+   */
+  public enum PutStatus implements StoreOpStatus {
```
ljacomet: I would move all these enums inside the `Store` interface and drop their common super type which is not used anywhere.

ljacomet:
Two general things:
* Please break out the commits for #787, they make their review more complicated for no added value.
* Please plan to add variants of the `EhcacheBasic*Test` that test an `Ehcache`. I am fine with these appearing once this specific PR is merged though.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -224,14 +227,14 @@ public boolean containsKey(K key) throws CacheAccessException {
...
-  public void put(K key, V value) throws CacheAccessException {
+  public PutStatus put(K key, V value) throws CacheAccessException {
     checkKey(key);
     checkValue(value);
     XATransactionContext<K, V> currentContext = getCurrentContext();
     if (currentContext.touched(key)) {
       V oldValue = currentContext.oldValueOf(key);
       currentContext.addCommand(key, new StorePutCommand<V>(oldValue, new XAValueHolder<V>(value, timeSource.getTimeMillis())));
-      return;
+      return PutStatus.PUT;
```
lorban: That's a PUT only if oldValue is null, otherwise it's an UPDATE.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -245,6 +248,7 @@ public void put(K key, V value) throws CacheAccessException {
     } else {
       currentContext.addCommand(key, new StorePutCommand<V>(null, new XAValueHolder<V>(value, timeSource.getTimeMillis())));
     }
+    return PutStatus.PUT;
```
lorban: If the current context already enlisted an Evict command, then this should be a NOOP. The same is true above BTW. `addCommand()` probably should return a boolean to indicate such case.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -253,13 +257,13 @@ public void put(K key, V value) throws CacheAccessException {
...
   @Override
-  public void remove(K key) throws CacheAccessException {
+  public boolean remove(K key) throws CacheAccessException {
     checkKey(key);
     XATransactionContext<K, V> currentContext = getCurrentContext();
     if (currentContext.touched(key)) {
       V oldValue = currentContext.oldValueOf(key);
       currentContext.addCommand(key, new StoreRemoveCommand<V>(oldValue));
-      return;
+      return true;
```
lorban: except if `oldValue` is null, or if there is an eviction command.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -271,10 +275,11 @@ public void remove(K key) throws CacheAccessException {
         currentContext.addCommand(key, new StoreRemoveCommand<V>(softLock.getOldValue()));
       }
     }
+    return true;
```
lorban: except if there is an eviction command.

lorban:
The return codes of `XAStore.put()` and `XAStore.remove()` are rather approximative. 

for put:
 - when in doubt, it's a NOOP.
 - when oldValue is null, it's a PUT.
 - otherwise it's an UPDATE.

for remove:
 - when in doubt, it's false.
 - when oldValue is null, it's false.
 - otherwise it's true.

AbfrmBlr:
Rebased & review comments taken care of

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -224,27 +224,34 @@ public boolean containsKey(K key) throws CacheAccessException {
...
     if (softLockValueHolder != null) {
       SoftLock<V> softLock = softLockValueHolder.value();
       if (isInDoubt(softLock)) {
         currentContext.addCommand(key, new StoreEvictCommand<V>(softLock.getOldValue()));
+        status = PutStatus.NOOP;
       } else {
         currentContext.addCommand(key, new StorePutCommand<V>(softLock.getOldValue(), new XAValueHolder<V>(value, timeSource.getTimeMillis())));
       }
     } else {
       currentContext.addCommand(key, new StorePutCommand<V>(null, new XAValueHolder<V>(value, timeSource.getTimeMillis())));
```
lorban: Status should be PutStatus.PUT here.

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -194,12 +194,12 @@ private void closeCache(final String alias, final Ehcache<?, ?> ehcache) {
 
   /**
    * Perform cache closure actions specific to a cache manager implementation.
-   * This method is called <i>after</i> the {@code Ehcache} instance is closed.
+   * This method is called <i>after</i> the {@code EhcacheWithLoaderWriter} instance is closed.
```
ljacomet: This should become `InternalCache` I assume

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -194,12 +194,12 @@ private void closeCache(final String alias, final Ehcache<?, ?> ehcache) {
...
    *
    * @param alias the cache alias
-   * @param ehcache the {@code Ehcache} instance for the cache to close
+   * @param ehcache the {@code EhcacheWithLoaderWriter} instance for the cache to close
    */
-  protected void closeEhcache(final String alias, final Ehcache<?, ?> ehcache) {
+  protected void closeEhcache(final String alias, final Cache<?, ?> ehcache) {
```
ljacomet: I would use `InternalCache` here

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -720,16 +725,16 @@ void destroy() {
   private static final class CacheHolder {
     private final Class<?> keyType;
     private final Class<?> valueType;
-    private volatile Ehcache<?, ?> cache;
+    private volatile Cache<?, ?> cache;
```
ljacomet: Should be `InternalCache` as well

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -16,80 +16,74 @@
...
-import static org.ehcache.core.exceptions.ExceptionFactory.newCacheWritingException;
 import static org.terracotta.statistics.StatisticBuilder.operation;
 
 /**
- * @author Alex Snaps
+ * @author Abhilash
+ *
  */
-public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V> {
+public class Ehcache<K, V> implements Cache<K, V>, UserManagedCache<K, V>, InternalCache<K, V> {
```
ljacomet: `implements InternalCache<K, V>` is enough

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -253,28 +260,35 @@ public void put(K key, V value) throws CacheAccessException {
...
     if (softLockValueHolder != null) {
       SoftLock<V> softLock = softLockValueHolder.value();
       if (isInDoubt(softLock)) {
         currentContext.addCommand(key, new StoreEvictCommand<V>(softLock.getOldValue()));
+        status = false;
       } else {
         currentContext.addCommand(key, new StoreRemoveCommand<V>(softLock.getOldValue()));
       }
     }
+    return status;
```
lorban: should be false here

107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
```diff
@@ -79,7 +79,7 @@
       registerEhcacheListeners(entry.getKey(), entry.getValue());
     }
 
-    this.jsr107Cache = ((Ehcache<K, V>) ehCache).getJsr107Cache();
+    this.jsr107Cache = ((InternalCache<K, V>) ehCache).getJsr107Cache();
```
ljacomet: Can't we change the type to `InternalCache` on the field?

107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -99,7 +99,7 @@ private void refreshAllCaches() {
   }
 
   private <K, V> Eh107Cache<K, V> wrapEhcacheCache(String alias, org.ehcache.Cache<K, V> cache) {
-    CacheLoaderWriter<? super K, V> cacheLoaderWriter = ((Ehcache<K, V>)cache).getCacheLoaderWriter();
+    CacheLoaderWriter<? super K, V> cacheLoaderWriter = ((InternalCache<K, V>)cache).getCacheLoaderWriter();
```
ljacomet: What if we use `InternalCache` as the type param?

107/src/main/java/org/ehcache/jsr107/Eh107CacheStatisticsMXBean.java
```diff
@@ -65,7 +65,7 @@
 
   Eh107CacheStatisticsMXBean(String cacheName, Eh107CacheManager cacheManager, Cache<?, ?> cache, ManagementRegistryService managementRegistry) {
     super(cacheName, cacheManager, "CacheStatistics");
-    this.bulkMethodEntries = ((Ehcache<?, ?>) cache).getBulkMethodEntries();
+    this.bulkMethodEntries = ((InternalCache<?, ?>) cache).getBulkMethodEntries();
```
ljacomet: Can't we change the type param to be `InternalCache`?

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
AbfrmBlr: But it is in doubt right ? and it's adding a  EvictCommand

PR #807, opened 2016-03-03
Issue 723

chrisdennis:
Wiring the clustered cache API to the voltron client code.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/service/DefaultClusteringService.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/service/DefaultClusteringService.java
@@ -16,6 +16,13 @@
 
 package org.ehcache.clustered.service;
 
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Collection;
+import java.util.Properties;
+
+import org.ehcache.clustered.client.EhcacheClientEntityFactory;
 import org.ehcache.clustered.config.ClusteringServiceConfiguration;
 import org.ehcache.config.ResourcePool;
 import org.ehcache.config.ResourceType;
@@ -25,7 +32,11 @@ import org.ehcache.spi.service.MaintainableService;
 import org.ehcache.spi.service.Service;
 import org.ehcache.spi.service.ServiceConfiguration;
 
-import java.util.Collection;
+import org.terracotta.connection.Connection;
+import org.terracotta.connection.ConnectionException;
+import org.terracotta.connection.ConnectionFactory;
+import org.terracotta.exception.EntityAlreadyExistsException;
+import org.terracotta.exception.EntityNotFoundException;
 
 import static java.util.Collections.emptyList;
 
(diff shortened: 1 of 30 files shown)


ljacomet:
Pretty sure that the empty file `clustered/client/private/cache/retriever/catalog.xml` should not be part of this

clustered/client/src/main/java/org/ehcache/clustered/service/DefaultClusteringService.java
```diff
@@ -34,46 +48,113 @@
...
+    entityFactory = new EhcacheClientEntityFactory(clusterConnection);
+    if (!entityFactory.acquireLeadership(entityIdentifier)) {
+      throw new IllegalStateException("Couldn't acquire cluster-wide maintenance lease");
+    }
   }
 
   @Override
   public void stop() {
-    // TODO: Implement stop
+    entityFactory.abandonLeadership(entityIdentifier);
```
ljacomet: Is that safe to call when not holding it?

clustered/client/src/main/java/org/ehcache/clustered/service/DefaultClusteringService.java
```diff
@@ -34,46 +48,113 @@
...
+    try {
+      clusterConnection.close();
+    } catch (IOException ex) {
```
ljacomet: No need to clear the entityFactory?

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -117,6 +117,7 @@ public EhcacheManager(Configuration config, Collection<Service> services, boolea
     this.serviceLocator = new ServiceLocator(services.toArray(new Service[services.size()]));
     this.useLoaderInAtomics = useLoaderInAtomics;
     validateServicesConfigs();
+    resolveServices();
```
ljacomet: Why is this needed at this time?

chrisdennis: So that the full set of services is loaded in to the ServiceLocator before maintenance is called.  Before this change it wasn't being done until init, which means maintenance won't see the whole set.

clustered/client/src/main/java/org/ehcache/clustered/service/DefaultClusteringService.java
chrisdennis: Yes... I believe so but I will verify this again.

- [ ] Verify it's safe to abandon leadership when not owning it (with a test maybe?)

clustered/client/src/main/java/org/ehcache/clustered/service/DefaultClusteringService.java
chrisdennis: That would probably be a good move, will make the change.

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -523,24 +537,8 @@ public void deregisterListener(CacheManagerListener listener) {
...
-          throw new IllegalArgumentException("Couldn't resolve Service " + serviceConfig.getServiceType().getName());
-        }
-      }
-      serviceLocator.loadDependenciesOf(ServiceDeps.class);
-      try {
-        serviceLocator.startAllServices();
-      } catch (Exception e) {
-        throw st.failed(e);
-      }
+      serviceLocator.startAllServices();
```
ljacomet: Removing the surrounding catch block will cause an exception to be triggered when trying to stop in the lower catch block which will end up shadowing the original exception.
So AFAICT the try/catch block is required here.

core/src/main/java/org/ehcache/core/StatusTransitioner.java
```diff
@@ -30,7 +30,7 @@
 /**
  * @author Alex Snaps
  */
-final class StatusTransitioner {
+public final class StatusTransitioner {
```
ljacomet: None of these visibility changes are required. Please revert.

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: Indeed - makes sense

core/src/main/java/org/ehcache/core/StatusTransitioner.java
chrisdennis: Yeah... thats gunk pulled across from earlier hackery... will revert.

core/src/main/java/org/ehcache/core/EhcacheManager.java
chrisdennis: Confused... let's chat about this.

ljacomet: Sorry, reviewer stupid

PR #809, opened 2016-03-04
heap store.get directly calls map.get #787

AbfrmBlr:


core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -109,7 +124,15 @@ public long lastAccessTime(TimeUnit unit) {
   }
 
   public void setLastAccessTime(long lastAccessTime, TimeUnit unit) {
-    this.lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    while (true) {
```
lorban: Why a while loop here and a do...while loop above?

AbfrmBlr:
I think for `lastaccesstime` loop is not required. removing that.

AbfrmBlr:
That was stupid. Loop is required for both.

AbfrmBlr:
`setLastAccessTime` had a bug - resolved now

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -109,7 +124,16 @@ public long lastAccessTime(TimeUnit unit) {
...
 
   public void setLastAccessTime(long lastAccessTime, TimeUnit unit) {
-    this.lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    do {
+      long read = this.lastAccessTime;
+      if (read >= lastAccessTime) {
+        break;
+      }
+      if (ACCESSTIME_UPDATER.compareAndSet(this, read, unit.convert(lastAccessTime, nativeTimeUnit()))) {
```
lorban: `lastAccessTime` has already been converted, don't do it again.

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -109,7 +124,16 @@ public long lastAccessTime(TimeUnit unit) {
   }
 
   public void setLastAccessTime(long lastAccessTime, TimeUnit unit) {
-    this.lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
```
ljacomet: Removing the `this.` does not magically create a local var. You are still setting it blindly.

ljacomet: hehe scratch that, you are assigning to the param. Well, could we have a local var with a name instead? This is confusing with the name shadowing in place already.

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -54,14 +57,26 @@ public long creationTime(TimeUnit unit) {
...
+  private void updateExpirationTime(long update) {
+    do {
+      long current = this.expirationTime;
+      if (current >= update) {
+        break;
+      }
+      if (EXPIRATIONTIME_UPDATER.compareAndSet(this, current, update)) {
+        break;
+      }
+    } while (true);
```
ljacomet: Given the condition, I am in favour of a `while(true)` loop that has the benefit of clearly indicating you are doing something special.

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -109,7 +124,16 @@ public long lastAccessTime(TimeUnit unit) {
...
+    lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    do {
+      long read = this.lastAccessTime;
+      if (read >= lastAccessTime) {
+        break;
+      }
+      if (ACCESSTIME_UPDATER.compareAndSet(this, read, unit.convert(lastAccessTime, nativeTimeUnit()))) {
+        break;
+      }
+    } while (true);
```
ljacomet: Same as for the other `do ... while`

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -109,7 +124,16 @@ public long lastAccessTime(TimeUnit unit) {
   }
 
   public void setLastAccessTime(long lastAccessTime, TimeUnit unit) {
-    this.lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    lastAccessTime = unit.convert(lastAccessTime, nativeTimeUnit());
+    while (true) {
+      long read = this.lastAccessTime;
```
lorban: Renaming the `read` variable `current` and the local `lastAccessTime` one `update` would make this method trivial to mentally match it with the `updateExpirationTime()` one. That would help future maintenance.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -263,36 +263,29 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
...
-        if (expiredValue.get() != null) {
-          decrementCurrentUsageInBytesIfRequired(expiredValue.get().size());
-        }
-      } else {
+        storeEventDispatcher.releaseEventSink(eventSink);
+        return null;
+      }
+
+      if (updateAccess) {
+        mapping = setAccessTimeAndExpiryThenReturnMapping(key, mapping, timeSource.getTimeMillis(), eventSink);
```
ljacomet: This is the only place we need the event sink, which means it should be created and released around this call.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -263,36 +263,29 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
...
+      }
+      if (mapping != null) {
```
ljacomet: You are missing the else branch here

PR #812, opened 2016-03-04
Issue #724

chrisdennis:
Initial cut of the Ehcache kitting code... requires a fix to Terracotta-OSS/terracotta-platform but should be otherwise okay.


ljacomet:
For some reason, the kit produced by this is `.tar` and not `.tar.gz` and is indeed not compressed.
I believe it would make sense to output a compressed tarball.

PR #819, opened 2016-03-08
Issue 716: RFC

lorban:
This isn't the final PR, a lot of cleanup is due before that

### DIFF
--- a/core/src/main/java/org/ehcache/core/Ehcache.java
+++ b/core/src/main/java/org/ehcache/core/Ehcache.java
@@ -133,6 +133,11 @@ public class Ehcache<K, V> implements InternalCache<K, V> {
   }
 
   @Override
+  public V getNoLoader(K key) {
+    return get(key);
+  }
+
+  @Override
   public V get(final K key) {
     getObserver.begin();
     statusTransitioner.checkAvailable();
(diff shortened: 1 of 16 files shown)


core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -648,6 +648,69 @@ public void loadAll(Set<? extends K> keys, boolean replaceExistingValues, Functi
...
+        public Entry<K, V> next() {
+          try {
+            final Entry<K, ValueHolder<V>> next = iterator.next();
+            // call Cache.get() here to check for expiry *and* account for a get in the stats
+            if (get(next.getKey()) == null) {
+              current = null;
+              return null;
+            }
+            // call Store.get() here to check check for hasNext()' expiry *and do not* account for an extra get in the stats
+            store.get(next.getKey());
```
ljacomet: Confused, even with the comment above, as to what this achieves since we ignore the result of the operation.

ljacomet:
Compilation error because other `InternalCache` implementation is not covered.

core/src/main/java/org/ehcache/core/Ehcache.java
lorban: without the call to Store.get(), CacheExpiryTest.expire_whenAccessed() assertion at line 371 fails because iterator.hasNext() returns true while the store only contains expired mappings.

ljacomet: After discussions, we should remove mappings when the new expiry computed is a duration of zero, while still return the value to the caller because it was not expired when accessed.

This probably means changes to `Store.get` logic and potential adaptations to the `Store` method called by the `EhcacheWithLoaderWriter`

ljacomet:
Given the TODO introduced in `Crappyterator`, I would suggest to not merge this before RC1 and then make sure the TODO is addressed before RC2.
@lorban sounds good to you?

ljacomet:
Or we even create an issue for that immediate expiration optimisation and make sure it gets fixed before this is merged.

lorban:
I prefer the 2nd option, especially since I noticed some corner-cases in the expiration handling that could make byte-sized OnHeapStore leak.

My eyes tell me that expiration cries for a rework.

Issue #820, opened 2016-03-08
Clean up configuration interfaces


ljacomet:
Stemming from #776 I believe a number of interfaces that relate to service configuration can be removed.

For example `CopierConfiguration` adds a `getType` method and declares an enum for the return of the said method. But it does not explicit any of the contracts that the `DefaultCopyProvider` relies upon in the `DefaultCopierConfiguration`. And even if it did, what would be the point of imposing that API to custom implementations of `CopyProvider`.

As such I want to remove such useless interfaces that relate to configuration in order to clean up the code.

PR #821, opened 2016-03-08
Issue #820

ljacomet:
Removal of a number of interfaces related to configuration that had no real purpose.
Includes some visibility and imports clean up in package `org.ehcache.core`.


chrisdennis:
This all seems logical to me... do you want to wait for another reviewer?

Issue #811, opened 2016-03-04
Support multiple instances of Service subtype via ServiceLocator


cljohnso:
Change ServiceLocator to support registration and discovery of Service subtypes having multiple implementations.  Initially targeted at Store.Provider and PersistableResourceService, the goal is to permit registration and discovery of multiple concrete implementations of these interfaces.

As part of this work, the current use of the SupplementaryService annotation will be replaced by a new annotation that will be used to mark Service subtypes for which registration of multiple implementations is permitted.

This is a spin-off of Issue #760.

cljohnso:
Split into multiple PRs:

* PR #838 to handle multiple Service instance registration
* PR #848 to add underlying Provider discovery to XAStore.Provider
* PR TBD to merge DefaultStoreProvider into CacheStore.Provider

PR #838, opened 2016-03-10
Issue #811 Support registering multiple instances of a single Service subtype

cljohnso:
The commits underlying this PR replace the `SupplementaryService` annotation with a `PluralService` annotation and generalize `ServiceLocator` support for registering multiple instances of a single `Service` subtype.  

Each `Service` supporting multiple instance registration is marked with the `PluralService` annotation.  The following existing `Service` interfaces are marked with `PluralService`:

* `Store.Provider`
* `AuthoritativeTier`
* `CachingTier`

`EhcacheManager` and `UserManagedCacheBuilder` are changed to select a `Store.Provider` implementation based on the results returned by a the new `Store.Provider.rank` method.  The ranking is based on the configured resources and service configurations.

In addition to the changes in `ServiceLocator` and the introduction of `PluralService`, this PR also includes updates to some unit tests needed to accommodate the multiple service support.


core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
```diff
@@ -0,0 +1,74 @@
...
+   * @param resourceTypes the set of {@code ResourceType}s that must be supported by the provider
+   * @param serviceConfigs the collection of {@code ServiceConfiguration}s used to influence the
+   *                       selection
+   *
+   * @return the non-{@code null} {@code Store.Provider} implementation chosen
+   *
+   * @throws IllegalStateException if no suitable {@code Store.Provider} is available
+   */
+  public static Store.Provider selectStoreProvider(
+      final ServiceLocator serviceLocator, final Set<ResourceType> resourceTypes, final Collection<ServiceConfiguration<?>> serviceConfigs) {
```
ljacomet: Can't the type here be a `ServiceProvider`?

cljohnso: Yes, it could be -- but `ServiceLocator` is in the same module under 'spi' so I didn't see the point.  If you'd prefer it be generalized to `ServiceProvider<Service>` I can certainly do that.

chrisdennis: That would seem to make sense to me if it's the minimal necessary interface.

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
```diff
@@ -0,0 +1,74 @@
...
+    final Collection<Store.Provider> storeProviders = serviceLocator.getServicesOfType(Store.Provider.class);
+    if (storeProviders.isEmpty()) {
```
chrisdennis: This is a little redundant, but I guess you think the distinct message is helpful?  Maybe we should eliminate this, and include the set of available providers in the later ISE?

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    int highRank = 0;
+    Store.Provider rankingProvider = null;
+    for (final Store.Provider provider : storeProviders) {
+      int rank = provider.rank(resourceTypes, serviceConfigs);
+      if (rank > highRank) {
+        highRank = rank;
+        rankingProvider = provider;
+      }
+    }
```
chrisdennis: Could we do this as a Collections.max(storeProviders, RANKING_COMPARATOR) call, followed by a conditional on the return that the rank is above 0?

core/src/main/java/org/ehcache/core/spi/cache/Store.java
```diff
@@ -446,6 +450,19 @@
...
 
+    /**
+     * Gets the internal ranking for the {@link Store} instances provided by this {@code Provider} of the store's
+     * ability to handle the specified resources.  A higher rank value indicates a more capable {@code Store}.
+     *
+     * @param resourceTypes the set of {@code ResourceType}s for the store to handle
+     * @param serviceConfigs the collection of {@code ServiceConfiguration} instances that may contribute
+     *                       to the ranking
+     *
+     * @return a non-negative rank indicating the ability of a {@code Store} created by this {@code Provider}
```
chrisdennis: Is there any real reason to disallow negative ranks?

core/src/test/java/org/ehcache/core/UserManagedCacheTest.java
```diff
@@ -100,7 +94,6 @@ private void assertFailingTransitionGoesToLowestStatus(InternalCache cache) thro
     } catch (StateTransitionException e) {
       assertThat(cache.getStatus(), is(Status.UNINITIALIZED));
     }
-
```
chrisdennis: Could we avoid this cleanup? (or do it in a separate commit)

impl/src/main/java/org/ehcache/impl/internal/store/DefaultStoreProvider.java
```diff
@@ -46,10 +49,22 @@
 public class DefaultStoreProvider implements Store.Provider {
   private static final Logger LOG = LoggerFactory.getLogger(DefaultStoreProvider.class);
 
+  private static final Set<ResourceType> SUPPORTED_RESOURCE_TYPES =
```
chrisdennis: This could be an enum set right?

osgi-test/build.gradle
```diff
@@ -47,6 +47,7 @@ test {
   systemProperty 'ehcache.osgi.jar', project(':dist').jar.archivePath.getPath()
   systemProperty 'ehcache.osgi.jcache.version', parent.jcacheVersion
   systemProperty 'ehcache.osgi.slf4j.version', parent.slf4jVersion
+  systemProperty 'jta.osgi.jar', "$projectDir/lib/com.springsource.javax.transaction-1.1.0.jar"
```
chrisdennis: This is the ugliness that Louis is fixing right?  Could we stick a comment in here to that effect?

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -656,6 +657,20 @@ public V apply(final K k) {
     private final Map<Store<?, ?>, SoftLockValueCombinedSerializerLifecycleHelper> createdStores = new ConcurrentWeakIdentityHashMap<Store<?, ?>, SoftLockValueCombinedSerializerLifecycleHelper>();
 
     @Override
+    public int rank(final Set<ResourceType> resourceTypes, final Collection<ServiceConfiguration<?>> serviceConfigs) {
```
chrisdennis: This seems wrong... XAStore should be doing a recursive call on StoreSupport.selectStoreProvider that is passed a StoreProvider that filters out this provider from the set of returned store providers.

core/src/main/java/org/ehcache/core/spi/cache/Store.java
cljohnso: No real reason ... I just wanted a value of 0 to mean the provider didn't provide support.

core/src/test/java/org/ehcache/core/UserManagedCacheTest.java
cljohnso: Hmmm ... not my code ... I just removed stuff from this class.

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
chrisdennis: It would be called once for each provider as it is currently... then once again for the actual max (to check the max rank against 0).

core/src/test/java/org/ehcache/core/UserManagedCacheTest.java
chrisdennis: Sorry, I meant could we leave the imports in place since nothing else was modified in the class.

impl/src/main/java/org/ehcache/impl/internal/store/DefaultStoreProvider.java
cljohnso: If we can assume that anything supported by `DefaultStoreProvider` (or its successor) will be in `ResourceType.Core`, then an enum set will work.  I'll tweak this.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
cljohnso: Discussed with @chrisdennis -- unless we need to bundle it with this work, the associated updates will be another PR.

impl/src/main/java/org/ehcache/impl/internal/store/DefaultStoreProvider.java
```diff
@@ -46,10 +49,22 @@
 public class DefaultStoreProvider implements Store.Provider {
```
cljohnso: Discussed with @chrisdennis -- the functionality currently performed by `DefaultStoreProvider` can be subsumed into `CacheStore.Provider` along with changes to `CacheStore.Provider` to have it be specific about what resource/config combinations it supports.  Likely needs additional discussion but this work will be in another PR ... possibly combined with the `XAStore.Provider` changes.

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
cljohnso: I suppose the likelihood of having a situation where **no** `Store.Provider` is available is exceedingly rare.  I can ditch this.

Do you want to itemize the available providers in the "No Store.Provider found ..." message?  If so, should this just be a `storeProviders.toString()` or something better formatted for readability?

core/src/test/java/org/ehcache/core/UserManagedCacheTest.java
cljohnso: Yes, this could be reverted.  I had originally touched the file because of some test that was subsequently relocated (I'd put it in the wrong place).

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
ljacomet: `Store.Providers.toString()` will be hard to digest, I believe the ideal is a list of `Store.Provider` class names.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
ljacomet: Agree with the change, propose to cover it in #798

impl/src/main/java/org/ehcache/impl/internal/store/DefaultStoreProvider.java
```diff
@@ -46,10 +49,22 @@
...
+  private static final Set<ResourceType> SUPPORTED_RESOURCE_TYPES =
+      new HashSet<ResourceType>(Arrays.asList(ResourceType.Core.HEAP, ResourceType.Core.OFFHEAP, ResourceType.Core.DISK));
+
   private volatile ServiceProvider<Service> serviceProvider;
   private final ConcurrentMap<Store<?, ?>, Store.Provider> providersMap = new ConcurrentWeakIdentityHashMap<Store<?, ?>, Store.Provider>();
 
   @Override
+  public int rank(final Set<ResourceType> resourceTypes, final Collection<ServiceConfiguration<?>> serviceConfigs) {
+    if (SUPPORTED_RESOURCE_TYPES.containsAll(resourceTypes)) {
+      return resourceTypes.size();
```
ljacomet: This is valid only if bigger than `1`, otherwise it may get selected and later fail.

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
```diff
@@ -0,0 +1,74 @@
...
+      final ServiceLocator serviceLocator, final Set<ResourceType> resourceTypes, final Collection<ServiceConfiguration<?>> serviceConfigs) {
+    final Collection<Store.Provider> storeProviders = serviceLocator.getServicesOfType(Store.Provider.class);
+    if (storeProviders.isEmpty()) {
+      throw new IllegalStateException("No Store.Provider available");
+    }
+    int highRank = 0;
+    Store.Provider rankingProvider = null;
+    for (final Store.Provider provider : storeProviders) {
+      int rank = provider.rank(resourceTypes, serviceConfigs);
+      if (rank > highRank) {
```
ljacomet: What should be the behaviour in case multiple providers answer with the same rank?
Initial thinking indicates this is a bad idea.

impl/src/test/java/org/ehcache/impl/internal/store/tiering/CacheStoreTest.java
```diff
@@ -680,6 +683,36 @@ public void testReleaseStoreFlushes () throws CacheAccessException {
     verify(cachingTierProvider, times(1)).releaseCachingTier(any(CachingTier.class));
   }
 
+  @Test
+  public void testRank() throws Exception {
+    CacheStore.Provider provider = new CacheStore.Provider();
+
+    Assert.assertThat(provider.rank(
+        Collections.<ResourceType>singleton(ResourceType.Core.DISK), Collections.<ServiceConfiguration<?>>emptyList()),
+        Matchers.is(0));
```
ljacomet: More style than anything, but using static imports here helps readability for me.

impl/src/test/java/org/ehcache/impl/internal/store/tiering/CacheStoreTest.java
```diff
@@ -680,6 +683,36 @@ public void testReleaseStoreFlushes () throws CacheAccessException {
     verify(cachingTierProvider, times(1)).releaseCachingTier(any(CachingTier.class));
   }
 
+  @Test
+  public void testRank() throws Exception {
+    CacheStore.Provider provider = new CacheStore.Provider();
+
+    Assert.assertThat(provider.rank(
+        Collections.<ResourceType>singleton(ResourceType.Core.DISK), Collections.<ServiceConfiguration<?>>emptyList()),
```
ljacomet: Extracting the collections creation could help a lot test readability.

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
cljohnso: K ... `Store.Provider` class names it is.

core/src/test/java/org/ehcache/core/UserManagedCacheTest.java
cljohnso: This will be _removed_ in my next push of the PR.

impl/src/main/java/org/ehcache/impl/internal/store/DefaultStoreProvider.java
cljohnso: Addressing ...

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
cljohnso: Will throw instead ... though, in practice, after the `DefaultStoreProvider` change to respond more accurately, the chance of this happening should be close to nil.

impl/src/test/java/org/ehcache/impl/internal/store/tiering/CacheStoreTest.java
cljohnso: Most existing references to `ResourceType.Core` aren't statically imported.  I've introduced a method that cleans this up considerably -- have a look at the newer stuff when it gets pushed and let me know if you still think a static import will help.

impl/src/test/java/org/ehcache/impl/internal/store/tiering/CacheStoreTest.java
cljohnso: Introduced another method to clean this up.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -771,6 +773,28 @@ public V apply(final K k) {
...
+      if (findSingletonAmongst(XAStoreConfiguration.class, serviceConfigs) == null) {
+        // An XAStore must be configured for use
+        return 0;
+      }
+
+      // TODO: Introduce proper Provider discovery & tracking and remove this *HACK*
+      final int underlyingRank;
+      if (resourceTypes.equals(EnumSet.of(ResourceType.Core.HEAP))) {
```
cljohnso: I used this hack instead of simply presuming the underlying provider (`DefaultStoreProvider`) supported the configuration.  This will need changing once `XSStore.Provider` is capable of discovering (and tracking) the proper provider.

(As a note, `DefaultStoreProvider` continues to support heap-only -- just doesn't "advertise" that fact.)

cljohnso:
Rebased

api/src/main/java/org/ehcache/spi/service/PluralService.java
```diff
@@ -22,9 +22,12 @@
 import java.lang.annotation.Target;
 
 /**
- * @author Alex Snaps
+ * Used to indicate that a {@link Service} subtype is permitted to have more than
+ * one concrete implementation registered with a {@link org.ehcache.spi.ServiceProvider}.
+ *
+ * @author Clifford W. Johnson
```
chrisdennis: @ljacomet you just purged all the author tags right?  Do we care if they start coming back?

ljacomet: That's a good point. @cljohnso Could you remove it? I believe this is the kind of information better served by SCM annotate and that it quickens documentation rot.

cljohnso: Will do

core/src/main/java/org/ehcache/core/spi/cache/StoreSupport.java
```diff
@@ -0,0 +1,105 @@
...
+   * @param serviceConfigs the collection of {@code ServiceConfiguration}s used to influence the
+   *                       selection
+   *
+   * @return the non-{@code null} {@code Store.Provider} implementation chosen
+   *
+   * @throws IllegalStateException if no suitable {@code Store.Provider} is available or if
+   *        multiple {@code Store.Provider} implementations return the same top ranking
+   */
+  public static Store.Provider selectStoreProvider(
+      final ServiceProvider<Service> serviceLocator, final Set<ResourceType> resourceTypes, final Collection<ServiceConfiguration<?>> serviceConfigs) {
```
ljacomet: first parameter name should be `serviceProvider`

cljohnso: Changed

PR #842, opened 2016-03-11
Issue 829: replace 2 and 3 args implementation for the XAStore

lorban:


transactions/src/main/java/org/ehcache/transactions/xa/internal/XAValueHolder.java
```diff
@@ -30,7 +30,7 @@
  */
 public class XAValueHolder<V> extends AbstractValueHolder<V> {
 
-  private static final TimeUnit NATIVE_TIME_UNIT = TimeUnit.MILLISECONDS;
+  public static final TimeUnit NATIVE_TIME_UNIT = TimeUnit.MILLISECONDS;
```
ljacomet: Can be package private

Issue #762, opened 2016-02-18
Core Storage Service Configuration


chrisdennis:
The core storage service needs configuration to control space partitioning within a server instance, and a mechanism by which server entities, and ultimately accessing clients can designate which partition a specific cache is consuming space from.  Although by strict definitions this isn't an Ehcache issue, I'm filing it here since I consider core storage to be an extension of the server side of Ehcache, and keeping the issues in one place helps with my mental model.  We can file a linked or duplicate issue in the relevant repository when work begins on this issue.

chrisdennis:
Closed as it will be replaced with a set of tasks in Voltron/Platform.

PR #846, opened 2016-03-12
Issue #762

chrisdennis:
This is a first cut at XML based resource configuration for clustered caches.  What's here isn't necessarily ready for merging, but it's indicative of where I am headed.  Also included in here is a slight rework of how we do extensibility in core xsd.  It's a little more strictly typed, and I think things are a little cleaner as a result.


chrisdennis:
Badly named branch and PR... closing.

Issue #825, opened 2016-03-09
Change stores to inline-evict mappings for which the updated expiration timestamp's makes them expire


lorban:
Currently, when a `Store` implementation figures a mapping hasn't yet expired but after updating its timestamp the mapping has expired, it is kept into the store.

We could optimize that by checking for expiration after updating the mapping's timestamp to avoid keeping the latter.

lorban:
This is especially annoying in the context of JSR-107 where creation and access can have different expiration policies.

Configuring such expiration policy on a JSR-107 cache:
```
new ParameterizedExpiryPolicy(Duration.ETERNAL, Duration.ZERO, null)
```
makes this problem quite apparent as with our current implementation, a mapping will not expire before its first access (as it should be) but we only expire it on the 2nd access. This isn't incorrect, but can create some friction with the JSR-107 TCK's `Cache.iterator().hasNext()` tests.

ljacomet:
Requires #866 before it can be finished

Issue #823, opened 2016-03-09
byte-sized OnHeapStore.compute(3args) leaks if the 3rd arg returns FALSE and the expiration policy throws


lorban:
Assume that `OnHeapStore.compute(3args)` is called on an existing, not expired mapping. Also assume that the 2nd arg does not return null and the 3rd one returns `FALSE`.

This branch gets executed:
```
          } else if ((eq(existingValue, computedValue)) && (!replaceEqual.apply())) {
            if (mappedValue != null) {
              return setAccessTimeAndExpiryThenReturnMapping(key, mappedValue, now, eventSink);
            }
```

leaving `AtomicBoolean write` set to its default `false`. `setAccessTimeAndExpiryThenReturnMapping()` calls the expiration policy, and if the latter throws a `RuntimeException`, `null` is returned - which ends up being the computed method's return value which makes the backend remove the mapping.

`decrementCurrentUsageInBytesIfRequired` is only called when  `AtomicBoolean write` is set to true which won't happen here despite the fact that the mapping was dropped.

ljacomet:
Most likely be resolved through #825

ljacomet:
`computeIfPresent` suffered from the same issue

PR #847, opened 2016-03-14
Issue #825: inline expiry

ljacomet:
This PR takes care of the inline expiry in `OnHeapStore`. It also fixes a number of issues with byte size accounting (see #823).

`AbstractOffHeapStore` will be handled in a future PR, as well as more general `Store` tests on this.


ljacomet:
Raised a number of questions that will most likely be voided by #866 so closing until that one is resolved.

PR #848, opened 2016-03-14
Issue #798 Add underlying Provider discovery to XAStore.Provider

cljohnso:
This commit changes the 'rank' and 'createStore' methods in XAStore.Provider
to use discover the proper Store.Provider to use for the underlying provider
instead of being tied to DefaultStoreProvider.

This commit **does not** close this issue.


cljohnso:
There are whitespace changes that obscures what is actually changed in `XAStore` -- an if/the/else was removed and indentation changed accordingly.  This update is best viewed ignoring whitespace changes.

PR #849, opened 2016-03-14
Updated terracotta-management-model dependency: return type to calls …

mathieucarbou:
…should be serializable

(management module)

Depends on PR: https://github.com/Terracotta-OSS/terracotta-management/pull/21

 I just sent the PR now so that you can review @anthonydahanne

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ import scripts.*
 ext {
   baseVersion = '3.0.0-SNAPSHOT'
   offheapVersion = '2.2.1'
-  managementVersion = '2.2.0'
+  managementVersion = '2.3.0'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -18,7 +18,7 @@ import scripts.*
 ext {
   baseVersion = '3.0.0-SNAPSHOT'
   offheapVersion = '2.2.1'
-  managementVersion = '2.2.0'
+  managementVersion = '2.2.1'
```
anthonydahanne: hum... that's an api change that is pending for terracotta-management ; I'd rather assume 2.3.0 will be the next version...

mathieucarbou:
test this please

PR #854, opened 2016-03-14
Issue #811 Merge DefaultStoreProvider into CacheStore.Provider

cljohnso:
This commit merges the tiering logic from DefaultStoreProvider into
CacheStore.Provider and discards DefaultStoreProvider.


impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
```diff
@@ -103,6 +110,8 @@ public void onInvalidation(K key, ValueHolder<V> valueHolder) {
   }
 
   static class ComputationException extends RuntimeException {
+    private static final long serialVersionUID = -7117636543507991092L;
```
chrisdennis: Unrelated... if we want to 'fix' this we should do it in one fell swoop.

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
```diff
@@ -336,48 +345,126 @@ public void clear() throws CacheAccessException {
...
 
+  @ServiceDependencies({CompoundCachingTier.Provider.class,
+      OnHeapStore.Provider.class, OffHeapStore.Provider.class, OffHeapDiskStore.Provider.class})
   public static class Provider implements Store.Provider {
 
+    private static final Set<Set<ResourceType.Core>> SUPPORTED_RESOURCE_COMBINATIONS;
+    static {
+      // Logic in setTierConfigurations must mirror this set
+      final Set<Set<ResourceType.Core>> supported = new HashSet<Set<ResourceType.Core>>();
+      supported.add(EnumSet.of(ResourceType.Core.HEAP, ResourceType.Core.DISK));
```
chrisdennis: static imports would make this a lot cleaner.

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
```diff
@@ -336,48 +345,126 @@ public void clear() throws CacheAccessException {
...
+      supported.add(EnumSet.of(ResourceType.Core.HEAP, ResourceType.Core.OFFHEAP));
+      supported.add(EnumSet.of(ResourceType.Core.HEAP, ResourceType.Core.OFFHEAP, ResourceType.Core.DISK));
+      SUPPORTED_RESOURCE_COMBINATIONS = supported;
```
chrisdennis: unmodifiableSet(...) would be a sensible defensive move here.

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
```diff
@@ -336,48 +345,126 @@ public void clear() throws CacheAccessException {
...
+        return 0;
+      }
     }
 
     @Override
     public <K, V> Store<K, V> createStore(Configuration<K, V> storeConfig, ServiceConfiguration<?>... serviceConfigs) {
       CacheStoreServiceConfiguration cacheStoreServiceConfig = findSingletonAmongst(CacheStoreServiceConfiguration.class, (Object[])serviceConfigs);
-      if (cacheStoreServiceConfig == null) {
-        throw new IllegalArgumentException("Cache store cannot be configured without explicit config");
+      if (cacheStoreServiceConfig != null) {
```
chrisdennis: If I'm reading whats going on here right, would it maybe be simpler to make CacheStoreServiceConfiguration not implement ServiceConfiguration, and then rename it to CacheStoreConfiguration?

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
cljohnso: You'd do that when the field value is used only internally?

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
cljohnso: I really have no clue what you're talking about here; care to elaborate?

impl/src/test/java/org/ehcache/impl/internal/store/tiering/CacheStoreTest.java
```diff
@@ -63,28 +71,36 @@
...
 public class CacheStoreTest {
 
+  @Mock
+  private CachingTier<Number, CharSequence> numberCachingTier;
+  @Mock
+  private AuthoritativeTier<Number, CharSequence> numberAuthoritativeTier;
+  @Mock
+  private CachingTier<String, String> stringCachingTier;
+  @Mock
+  private AuthoritativeTier<String, String> stringAuthoritativeTier;
```
chrisdennis: If this is just eliminating doing the mocking in the first two lines of each test, then consider my vote a -1 against this specific change... just a vote however.  Would like others to weigh in.

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
cljohnso: You have two levels of static import that could be used here -- at what level would you recommend?

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
chrisdennis: If you are against I can live without it.

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
chrisdennis: `supported.add(of(HEAP, DISK));`

impl/src/test/java/org/ehcache/impl/internal/store/tiering/CacheStoreTest.java
cljohnso: This is actually the recommended way of mocking classes with generic parameters ...

impl/src/main/java/org/ehcache/impl/internal/store/tiering/CacheStore.java
ljacomet: Agreed with Chris + making config class package private

cljohnso: Just got out of a hangout with Chris -- we're going to reduce the `CacheStoreServiceConfiguration` to it's minimal form and make it an internal class.  There are a couple of tests that would need adjustment but that configuration is really only used internally to `CacheStore.Provider`.

cljohnso:
Updated and rebased ...

PR #857, opened 2016-03-15
Issue #790 Update expiry contract with some value indirection

ljacomet:
* Enables to indicate that recovering a value in some cases may be
expensive, i.e. deserialization
* Internal update of StoreEventSink contract to not have to do eager
deserialization when no listeners are registered.

I want the review to be about the API changes this introduces in addition to the technical aspect. I am still unsure if it should go in or not.
@lorban Might be interesting to get perf numbers with and without this


107/src/main/java/org/ehcache/jsr107/EhcacheExpiryWrapper.java
```diff
@@ -35,12 +36,12 @@ public Duration getExpiryForCreation(K key, V value) {
   }
 
   @Override
-  public Duration getExpiryForAccess(K key, V value) {
-    return wrappedExpiry.getExpiryForAccess(key, value);
+  public Duration getExpiryForAccess(K key, Invoker<V> value) {
+    return wrappedExpiry.getExpiryForAccess(key, (Invoker) value);
```
chrisdennis: Raw cast seems bad... no other way of avoiding it?

core/src/main/java/org/ehcache/core/util/Invokers.java
```diff
@@ -0,0 +1,61 @@
...
+    };
+  }
+
+  /**
+   * Returns a raw {@link Invoker} because generics ...
+   *
+   * @param value the value to hold
+   * @return an invoker with the given value
+   */
+  public static Invoker rawInvokerOf(final Object value) {
```
chrisdennis: wat?!

ljacomet:
Well, in its current state, this is badly broken ... my bad. Will push an updated version soon - principle and ugliness can still be discussed ;-)

ljacomet:
PR updated with a more correct handling of lazy value holders in off heap. This raises even more doubt for this PR as I had to explicitly request value deserialization in places because it no longer happens through expiry computation ... but the result really needs to be deserialized before being returned.
And in fact, while `OnHeap` deserialization happens on each invoke of the value, this is not relevant for `OffHeap`. So discussion to be continued on this topic ...

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -181,6 +181,6 @@ public long getId() {
 
   @Override
   public String toString() {
-    return "" + value();
+    return "" + invoke();
```
chrisdennis: Oh Hell No!

core/src/test/java/org/ehcache/core/util/Matchers.java
```diff
@@ -0,0 +1,56 @@
...
+      }
+
+      @Override
+      public void describeTo(Description description) {
+        description.appendText("holder containing value '").appendValue(value).appendText("'");
+      }
+    };
+  }
+
+  public static <V> Matcher<Invoker<V>> holdingAnyOf(final Class<V> clazz) {
```
chrisdennis: The 'hamcrest' way of doing this would be `holding(instanceOf(Foo.class))`

api/src/main/java/org/ehcache/Invoker.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Invoker creates an indirection to access a value.
+ * <P>
+ *   This can indicate that the value needs computation before it can be retrieved, such as deserialization.
+ * </P>
+ *
+ * @param <V> the value type
+ */
+public interface Invoker<V> {
```
chrisdennis: This feels like the name is wrong... it implies it has side affects (which it doesn't, right?)

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
cschanck: you know, "" + invoke().toString() would be clearer.

107/src/main/java/org/ehcache/jsr107/EhcacheExpiryWrapper.java
cljohnso: Alternatively, you could change the type of the `wrappedExpiry` field to `Expiry<K, V>` and cast the `wrappedExpiry` on the way in using an unchecked cast -- what you're concerned with in this case is solely that the `Expiry` implementation _accept_ `<K,V>` which you know because one handling _super_ was passed in.  This will eliminate the need to cast each value on its way in.

api/src/main/java/org/ehcache/Invoker.java
cljohnso: In Java 8 parlance, this is a _Supplier_, is it not?  Can we reuse that name?

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
cljohnso: Would we always want `toString` to materialize the value?  Wouldn't we want some way of reflecting that the value is, as yet, not materialized?

core/src/main/java/org/ehcache/core/events/StoreEventSink.java
```diff
@@ -27,7 +29,7 @@
    * @param key removed key
    * @param value removed value
    */
-  void removed(K key, V value);
+  void removed(K key, Invoker<V> value);
```
cljohnso: Is there any circumstance where the value presented to/via an event should be subject to delayed materialization?  Could it then be subject to being unavailable when the value is later fetched?

api/src/main/java/org/ehcache/Invoker.java
chrisdennis: Supplier seems better to me, yes.

api/src/main/java/org/ehcache/expiry/Expiry.java
```diff
@@ -56,6 +58,6 @@
    * @param newValue the new value of the entry
    * @return the updated expiration value for the given entry. A {@code null} return value indicates "no change" to the expiration time
    */
-  Duration getExpiryForUpdate(K key, V oldValue, V newValue);
+  Duration getExpiryForUpdate(K key, Invoker<V> oldValue, V newValue);
```
chrisdennis: I like these changes, they seem to make the API more honest.  Retrieving the values is something being done that may otherwise not be done... so it has a cost.  Associating a cost with an action is a nice way of communicating to the user that what they are doing might have a performance impact.

api/src/main/java/org/ehcache/Invoker.java
ljacomet: `ValueSupplier` to be specific?

chrisdennis: thinking about it more: ValueSupplier<V> seems a little redundant.  Of course Supplier collides with a JDK 8 type, so it's a bit ugly.  Maybe ValueSupplier<V> has a slight edge then.

107/src/main/java/org/ehcache/jsr107/EhcacheExpiryWrapper.java
ljacomet: Changing expiry to take `ValueSupplier<? extends V>` to resolve this

core/src/main/java/org/ehcache/core/util/Invokers.java
ljacomet: That's going away ... but can't replace it with the duck sadly :wink:

core/src/main/java/org/ehcache/core/events/StoreEventSink.java
ljacomet: This is necessary to not force deserialisation if no store event listeners are registered. Cache event listeners will anyway get the raw value, not wrapped.

api/src/main/java/org/ehcache/Invoker.java
ljacomet: Using `ValueSupplier.value()` in upcoming update

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
ljacomet: I am not going down the rabbit hole of making the `AbstractValueHolder` aware of values being materialised or not.

core/src/main/java/org/ehcache/core/util/Invokers.java
chrisdennis: Good news: http://unicode.org/emoji/charts/emoji-candidates.html#1f986

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
ljacomet: will use `format("%s", value());`

core/src/test/java/org/ehcache/core/util/Matchers.java
ljacomet: Well, matcher improved and I learned things about hamcrest :+1:

ljacomet:
So as discussed in issue #790 we are going through with this. I have pushed an up to date version that should include all review comments.

api/src/main/java/org/ehcache/config/CacheRuntimeConfiguration.java
```diff
@@ -38,8 +38,8 @@
    * Allows for registering {@link org.ehcache.event.CacheEventListener} on the cache
    *
    * @param listener the listener instance to register
-   * @param ordering the {@link org.ehcache.event.EventOrdering} to invoke this listener
-   * @param firing the {@link org.ehcache.event.EventFiring} to invoke this listener
+   * @param ordering the {@link org.ehcache.event.EventOrdering} to value this listener
```
chrisdennis: IntelliJ cannot grammar.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -218,14 +221,16 @@ private boolean newValueAlreadyExpired(K key, V oldValue, V newValue) {
     final Duration duration;
     if (oldValue == null) {
       try {
-        duration = runtimeConfiguration.getExpiry().getExpiryForCreation(key, newValue);
+        Expiry<? super K, ? super V> expiry = runtimeConfiguration.getExpiry();
+        duration = expiry.getExpiryForCreation(key, newValue);
```
chrisdennis: any reason for this change?

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -218,14 +221,16 @@ private boolean newValueAlreadyExpired(K key, V oldValue, V newValue) {
...
       } catch (RuntimeException re) {
         logger.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
         return true;
       }
     } else {
       try {
-        duration = runtimeConfiguration.getExpiry().getExpiryForUpdate(key, oldValue, newValue);
+        ValueSupplier<V> oldSupplier = supplierOf(oldValue);
+        duration = runtimeConfiguration.getExpiry().getExpiryForUpdate(key, oldSupplier, newValue);
```
chrisdennis: We've inlined this everywhere else, might as well do so here.

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
```diff
@@ -181,6 +183,6 @@ public long getId() {
 
   @Override
   public String toString() {
-    return "" + value();
+    return format("%s", value());
```
chrisdennis: This is sad

core/src/main/java/org/ehcache/core/util/ValueSuppliers.java
```diff
@@ -0,0 +1,45 @@
...
+ * Utility for creating basic {@link ValueSupplier} instances
+ */
+public class ValueSuppliers {
+
+  /**
+   * Returns a basic {@link ValueSupplier} that serves the value passed in
+   *
+   * @param value the value to hold
+   * @param <V> the value type
+   * @return an invoker with the given value
```
chrisdennis: not an invoker

core/src/main/java/org/ehcache/core/util/ValueSuppliers.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public static <U, V extends U> ValueSupplier<U> supplierOf(final V value) {
```
chrisdennis: No need for both of these generic types here, not with the `? extends V` usage.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -48,7 +48,7 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
     `Cache`, but will also close it. The `Cache` releases all locally held transient resources (such as memory).
     References to this `Cache` become unusable.
 <10> In order to release all transient resources (memory, threads, ...) a `CacheManager` provides to `Cache` instances
-     it manages, you have to invoke `CacheManager.close()`, which in turns closes all `Cache` instances known at the time.
+     it manages, you have to value `CacheManager.close()`, which in turns closes all `Cache` instances known at the time.
```
chrisdennis: More IntelliJ grammar loveliness

impl/src/main/java/org/ehcache/impl/internal/classes/commonslang/reflect/MemberUtils.java
```diff
@@ -47,7 +47,7 @@
      *
      * When a {@code public} class has a default access superclass with {@code public} members,
      * these members are accessible. Calling them from compiled code works fine.
-     * Unfortunately, on some JVMs, using reflection to invoke these members
+     * Unfortunately, on some JVMs, using reflection to value these members
```
chrisdennis: These files should really be locked in someway.

impl/src/main/java/org/ehcache/impl/internal/events/AbstractStoreEventDispatcher.java
```diff
@@ -43,17 +44,17 @@ public void closeOnFailure() {
     }
 
     @Override
-    public void removed(Object key, Object value) {
+    public void removed(Object key, ValueSupplier value) {
```
chrisdennis: generics?

impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -233,6 +235,7 @@ public Number call() throws Exception {
           }
 
           if (updateAccess) {
+            mappedValue.forceDeserialization();
```
chrisdennis: I wonder if this method should be named `detach`... or something similar... thats what this does right?

impl/src/test/java/org/ehcache/impl/internal/events/FudgingInvocationScopedEventSinkTest.java
```diff
@@ -131,4 +132,13 @@ public void testEvictedKeyDoesNotFudgeOlderEvents() {
     inOrder.verify(listener).onEvent(argThat(eventType(EventType.EVICTED)));
     verifyNoMoreInteractions(listener);
   }
+
+  private ValueSupplier<String> invokerOf(final String value) {
```
chrisdennis: Can't this use the real supplierOf method?

impl/src/test/java/org/ehcache/impl/serialization/CompactJavaSerializerClassUnloadingTest.java
```diff
@@ -116,7 +116,7 @@ private static void packHeap() {
 //  private static void dumpHeap(String fileName, boolean live) {
 //    try {
 //      new File(fileName).delete();
-//      ManagementFactory.getPlatformMBeanServer().invoke(ObjectName.getInstance("com.sun.management:type=HotSpotDiagnostic"), "dumpHeap", new Object[]{fileName, Boolean.valueOf(live)}, new String[]{"java.lang.String", "boolean"});
+//      ManagementFactory.getPlatformMBeanServer().value(ObjectName.getInstance("com.sun.management:type=HotSpotDiagnostic"), "dumpHeap", new Object[]{fileName, Boolean.valueOf(live)}, new String[]{"java.lang.String", "boolean"});
```
chrisdennis: over zealous ide

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -792,15 +794,16 @@ public Duration getExpiryForCreation(K key, SoftLock softLock) {
...
+        public Duration getExpiryForAccess(K key, final ValueSupplier<? extends SoftLock<V>> softLock) {
+          if (softLock.value().getTransactionId() != null) {
             // phase 1 prepare, access -> forever
             return Duration.FOREVER;
           } else {
             // phase 2 commit, or during a TX's lifetime, access -> some time
             Duration duration;
             try {
-              duration = configuredExpiry.getExpiryForAccess(key, (V) softLock.getOldValue());
+              ValueSupplier<V> supplierExpired = supplierOf(softLock.value().getOldValue());
```
chrisdennis: Did we want to delay the softLock.value() resolution too?  Is there any advantage to that @lorban ?

core/src/main/java/org/ehcache/core/Ehcache.java
ljacomet: Probably a left over of generics fighting, can go away.

api/src/main/java/org/ehcache/config/CacheRuntimeConfiguration.java
ljacomet: Refactor too much :wink:

core/src/main/java/org/ehcache/core/Ehcache.java
ljacomet: Java 1.7 compiler fails if inlined

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
ljacomet: Will need something else to understand what you would expect? Last option from me is just to delete

core/src/main/java/org/ehcache/core/util/ValueSuppliers.java
ljacomet: And that simplification enables the inlining on 1.7 it seems.

core/src/main/java/org/ehcache/core/Ehcache.java
ljacomet: But no longer does after the generics simplification done on `ValueSuppliers.supplierOf(V)`.

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
chrisdennis: Trolls!

chrisdennis:
```
V value = value();
return value == null ? "null" : value.toString();
```

impl/src/main/java/org/ehcache/impl/internal/events/AbstractStoreEventDispatcher.java
ljacomet: The whole point of that static field is to use it raw as it does nothing.

impl/src/test/java/org/ehcache/impl/internal/events/FudgingInvocationScopedEventSinkTest.java
ljacomet: Now it can

core/src/main/java/org/ehcache/core/spi/cache/AbstractValueHolder.java
ljacomet: Well, format takes care of that for me ...

ljacomet:
Doing a complete pass over the whole changeset to catch all these stupid rename in javadoc.

core/src/main/java/org/ehcache/core/events/StoreEventSink.java
```diff
@@ -25,26 +27,26 @@
    * Indicates the mapping was removed.
    *
    * @param key removed key
-   * @param value removed value
+   * @param value value supplier to removed value
```
chrisdennis: supplier of?

core/src/test/java/org/ehcache/core/util/Matchers.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Matchers
+ */
+public class Matchers {
+
+  public static <V> Matcher<ValueSupplier<V>> holding(final V value) {
+    return holding(equalTo(value));
+  }
+
+  public static <V> Matcher<ValueSupplier<V>> holding(final Matcher<V> matcher) {
```
chrisdennis: If this was `holding(Matcher<? super V> matcher)` it might solve the generics problem you are seeing in EhcacheWithLoaderWriterBasicReplaceTest?

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
chrisdennis: This got lost in the mix... fine for this to stay as is, if there is no advantage to delaying the soft lock call.

Issue #860, opened 2016-03-15
Optimized serializers aren't available to caches with a disk tier


lorban:
If a cache is configured with a disk tier, the optmized serializers for Integer, Long, String... don't get used.

See the following log:

```
INFO  o.e.i.i.s.s.DefaultSerializationProvider - Serializer for <java.lang.Long> : org.ehcache.impl.serialization.CompactPersistentJavaSerializer@67bd2497
INFO  o.e.i.p.DefaultLocalPersistenceService - Created /work/terracotta/github/lorban/ehcache3perf/target/rainfall/disk2tiers/ehcache3-persistence/cache1_379faab9b6fb27f87b177b067266d70b66ac64f2/DefaultSerializationProvider-Value
[main] INFO  o.e.i.i.s.s.DefaultSerializationProvider - Serializer for <java.lang.String> : org.ehcache.impl.serialization.CompactPersistentJavaSerializer@71282b42
```

PR #862, opened 2016-03-15
Issue #860 Default persistent serializers

ljacomet:
The default serializers that were added, while compatible with
persistent caches, were not properly configured. This is now resolved
and there are tests covering that.


impl/src/main/java/org/ehcache/impl/internal/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -208,6 +208,24 @@ public void start(ServiceProvider<Service> serviceProvider) {
       if (!serializers.containsKey(Serializable.class)) {
         serializers.put(Serializable.class, (Class) CompactPersistentJavaSerializer.class);
       }
+      if (!serializers.containsKey(Long.class)) {
```
chrisdennis: Time for a helper method?

Issue #822, opened 2016-03-08
EhcacheManager.close hangs after error creating cache


cljohnso:
If an `Error` is thrown after the `CacheHolder` is added to 'caches' but before `value.setCache` is called, then a `EhcacheManager.close` hangs (in `EhcacheManager.CacheHolder.retrieve`) waiting for the cache reference to be set in the `CacheHolder` instance.

This problem was discovered during testing when a place-holder update in `EhcacheManager.getStore` threw an `AssertionError`.  The fault could be caused by an `Error` thrown during `Service` resolution or `Store` creation.

ljacomet:
I am not sure this should be fixed at all. When an `Error` is thrown, we should not even pretend to clean up.
And our own code throwing `AssertionError` is such a case ... as it should not happen under any _normal_ circumstances.

cljohnso:
The way the code is now, an Error leaves us in a hard wait with no clue as to why.

ljacomet:
After discussions, while we do not want to catch `Error` or `Throwable` during cache construction, we need to make sure this does not cause `CacheManager.close()` to block.

AbfrmBlr:
````java
  @Test
  public void testCloseWhenCacheCreationFailsDuringInitialization() throws Exception {
    Store.Provider storeProvider = mock(Store.Provider.class);
    doThrow(new Error("Test EhcacheManager close.")).when(storeProvider).createStore(any(Store.Configuration.class), Matchers.<ServiceConfiguration>anyVararg());

    CacheConfiguration<Long, String> cacheConfiguration = new BaseCacheConfiguration<Long, String>(Long.class, String.class, null, null, null, ResourcePoolsHelper.createHeapOnlyPools());
    Map<String, CacheConfiguration<?, ?>> caches = new HashMap<String, CacheConfiguration<?, ?>>();
    caches.put("cache1", cacheConfiguration);
    DefaultConfiguration config = new DefaultConfiguration(caches, null);
    final CacheManager cacheManager = new EhcacheManager(config, Arrays.asList(
        storeProvider,
        mock(CacheLoaderWriterProvider.class),
        mock(WriteBehindProvider.class),
        mock(CacheEventDispatcherFactory.class),
        mock(CacheEventListenerProvider.class),
        mock(LocalPersistenceService.class)
    ));

    final CountDownLatch countDownLatch = new CountDownLatch(1);

    Executors.newSingleThreadExecutor().submit(new Runnable() {
      @Override
      public void run() {
[...]

PR #865, opened 2016-03-16
Failure of cache creation should clean up #822

AbfrmBlr:
This solves cacheholder retrieve problem.
Still making progress on StatusTransitioner stuff

### DIFF
--- a/core/src/main/java/org/ehcache/core/StatusTransitioner.java
+++ b/core/src/main/java/org/ehcache/core/StatusTransitioner.java
@@ -218,7 +218,16 @@ final class StatusTransitioner {
     }
 
     public StateTransitionException failed(Throwable t) {
+      if (st.done()) {
+        if (t != null) {
+          throw new AssertionError("Throwable cannot be null if Transition is done.");
+        }
+        return null;
+      }
       st.failed();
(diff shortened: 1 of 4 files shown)


ljacomet:
I believe a change around `org.ehcache.core.EhcacheManager#removeCache(java.lang.String, boolean)` needs to be done as well to make sure we do not fail just because we try to remove a `CacheHolder` where the cache ends up being `null`.

core/src/main/java/org/ehcache/core/StatusTransitioner.java
```diff
@@ -218,7 +218,13 @@ public void succeeded() {
     }
 
     public StateTransitionException failed(Throwable t) {
+      if (st.done()) {
```
ljacomet: This condition feels wrong - it probably has to be linked with the `Throwable` being null or not.

AbfrmBlr: The point is to repair the status transition only when it failed to.
So if transition was done even due to an `RuntimeException` the `finally` call should not do anything, should act like a noop.
And if the transition is not done, which creates problem of hang, the transition is done and then returned.

Issue #851, opened 2016-03-14
Byte sized heap: split accounting and eviction


ljacomet:
Byte sized heap store should do the accounting under lock and have a capacity check after mutations are done, similar to what exists for count sized heap store.

This could enable simplifying / reducing the `Atomic*` types that are used to track internal state in the lambdas passed down.

ljacomet:
Need to wait for #825 to be finished, too close to each other

ljacomet:
Adding more work here: let's try to find a way to handle the sizing race in `OnHeapStore.clear()`

PR #868, opened 2016-03-16
Issue #825: inline expiry

ljacomet:
* implement inline expiry
* proper return contracts for different methods in the face of inline expiry
* Added SPITests where relevant

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
@@ -278,11 +278,7 @@ public class OnHeapStore<K, V> implements Store<K,V>, HigherCachingTier<K, V> {
       if (updateAccess) {
         setAccessTimeAndExpiryThenReturnMappingOutsideLock(key, mapping, now);
       }
-      if (mapping != null) {
-        getObserver.end(StoreOperationOutcomes.GetOutcome.HIT);
-      } else {
-        getObserver.end(StoreOperationOutcomes.GetOutcome.MISS);
-      }
+      getObserver.end(StoreOperationOutcomes.GetOutcome.HIT);
       return mapping;
     } catch (RuntimeException re) {
       handleRuntimeException(re);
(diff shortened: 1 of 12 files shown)


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -340,7 +327,7 @@ private PutStatus putReturnHolder(final K key, final V value) throws CacheAccess
       if (entryActuallyAdded.get()) {
         enforceCapacity(valuePut.size(), eventSink);
       } else {
-        long replacedDelta = valuePut == null ? 0 : valuePut.size() - (replacedValue.get() == null ? 0 : replacedValue.get().size());
+        long replacedDelta = (valuePut == null ? 0 : valuePut.size()) - (replacedValue.get() == null ? 0 : replacedValue.get().size());
```
lorban: Two ternary operators on the same line is a bit much for my taste. I'd prefer to see this split onto two lines.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1233,30 +1234,35 @@ public V apply(final K k, final V oldValue) {
...
+    if (Duration.ZERO.equals(duration)) {
       expireMapping(key, valueHolder);
       return null;
     }
-    valueHolder.accessed(now, duration);
     return valueHolder;
   }
 
-  private OnHeapValueHolder<V> setAccessTimeAndExpiryThenReturnMapping(K key, OnHeapValueHolder<V> valueHolder, long now,
+  private OnHeapValueHolder<V> setAccessTimeAndExpiryThenReturnMappingUnderLock(K key, OnHeapValueHolder<V> valueHolder, long now,
```
lorban: The UnderLock and OutsideLock variants of this method are identical (practically so). Why having two variants then?

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
ljacomet: Agreed, but would like to leave it like that and address it in #851

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
ljacomet: While the difference is small, there is a huge impact. Remember that events cannot be fired outside of lock scope. So one variant will just fire an expiry event and return null while the other one will perform a full expiration (`computeIfPresent` with equality check that fires event + returns null).

lorban: Okay, then I would suggest to add a comment around line `expireMapping(key, valueHolder);` / `onExpiration(key, valueHolder, eventSink);` as it is very easy to miss the subtle difference.

ljacomet: Done

Issue #813, opened 2016-03-07
Store.containsKey should not trigger getObserver


ljacomet:
Currently main store implementations do trigger a `getObserver` when invoking `Store.containsKey` which is wrong.
The `getObserver` should only be triggered on the `Store.get` path.

PR #869, opened 2016-03-17
store contains does not trigger getobserver #813

AbfrmBlr:


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -262,19 +262,13 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
     getObserver.begin();
     StoreEventSink<K, V> eventSink = null;
     try {
-      OnHeapValueHolder<V> mapping = map.get(key);
+      OnHeapValueHolder<V> mapping = getQuiet(key);
```
ljacomet: Given this change, `get` and `getInternal` can be merged

AbfrmBlr: But `internalGet` is getting used in iterator with updateaccess as false.

ljacomet: Right, my bad

ljacomet:
Good after rebase

AbfrmBlr:
rebased

Issue #871, opened 2016-03-17
Add extra store statistics


lorban:
During perf tests, some internals of the stores are handy to monitor, hence a few more stats should be added.

PR #872, opened 2016-03-17
#871: add store-specific counters as pass-through statistics

lorban:


impl/src/main/java/org/ehcache/impl/internal/store/offheap/OffHeapMap.java
```diff
@@ -0,0 +1,23 @@
+package org.ehcache.impl.internal.store.offheap;
+
+/**
+ * Off-heap-backed maps counters.
+ *
+ * @author Ludovic Orban
+ */
+public interface OffHeapMap {
```
ljacomet: Why the extra interface instead of sticking these on `EhcacheOffHeapBackingMap<K, V>`?

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -250,6 +251,12 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
     getOrComputeIfAbsentObserver = operation(CachingTierOperationOutcomes.GetOrComputeIfAbsentOutcome.class).named("getOrComputeIfAbsent").of(this).tag("onheap-store").build();
     invalidateObserver = operation(CachingTierOperationOutcomes.InvalidateOutcome.class).named("invalidate").of(this).tag("onheap-store").build();
     silentInvalidateObserver = operation(HigherCachingTierOperationOutcomes.SilentInvalidateOutcome.class).named("silentInvalidate").of(this).tag("onheap-store").build();
+    StatisticsManager.createPassThroughStatistic(this, "size", Collections.singleton("onheap-store"), new Callable<Number>() {
+      @Override
+      public Number call() throws Exception {
+        return map.size();
```
ljacomet: This is incorrect for byte sized

lorban: Is it? It might still be useful to be able to figure out how many mappings you can store in a byte-sized heap tier, don't you think?

impl/src/main/java/org/ehcache/impl/internal/store/offheap/OffHeapMap.java
lorban: Just to avoid polluting the `EhcacheOffHeapBackingMap` interface with stats concerns. But I agree, at the very least this new interface is poorly named.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
ljacomet: Indeed, but then I would name it differently, like `mappingsCount` or something else.

impl/src/main/java/org/ehcache/impl/internal/store/offheap/OffHeapMap.java
ljacomet: `OffHeapMapStatistics`?

ljacomet:
Good for me once rebased

Issue #793, opened 2016-02-25
Convert RuntimeExceptions issued by XML parsing to XmlConfigurationException


cljohnso:
A `org.ehcache.xml.exceptions.XmlConfigurationException` was added to the **xml** module by PR #769 for use in reflecting a fault in an XML configuration file.  There are a number of places in the code, in the **xml** module in particular, where `RuntimeException` is thrown to reflect an error in XML configuration.  This issue is opened to track the conversion of that use of `RuntimeException` to `XmlConfigurationException`.

PR #876, opened 2016-03-18
#793: cleanup XML parsing exception handling

lorban:


xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -115,24 +115,14 @@ public ConfigurationParser(String xml, URL... sources) throws IOException, SAXEx
     factory.setXIncludeAware(true);
     factory.setSchema(XSD_SCHEMA_FACTORY.newSchema(schemaSources.toArray(new Source[schemaSources.size()])));
 
-    final DocumentBuilder domBuilder;
-    try {
-      domBuilder = factory.newDocumentBuilder();
-    } catch (ParserConfigurationException e) {
-      throw new AssertionError(e);
-    }
+    DocumentBuilder domBuilder = factory.newDocumentBuilder();
```
chrisdennis: I would still vote that a ParserConfigurationException should be converted to an AssertionError here.

chrisdennis:
Looks fine other than my vote on that ParserConfigurationException... another vote for your position would allow me to merge this.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
GaryWKeim: I thought assertions were for things that normally shouldn't occur. Messing up you XML is normal.

ljacomet: Changes looks good to me, not that I agree with Gary :wink:

Issue #716, opened 2016-01-06
Potential issue in iteration with multiple tiers


ljacomet:
While working on #165, I realised that we may have an issue when iterating on multiple tiers.
Currently, the iteration happens at the `AuthoritativeTier` level and will remove expired entries. But I believe this happens without looking at the pinning state, which means we may end up in weird state.

Did not write a test to confirm yet, so this may be bogus ... though I doubt it.

lorban:
I confirm this bug: iterating expired entires of a n-tier caches won't make the caching tier immediately flush their mappings. On the other hand, once expiration happened in the authority, it will eventually happen from the caching tier too which will flush the value that will instantly be dropped to the floor by the authority, and the cache will return null as it should. Not ideal, but not broken from a user's perspective.

On the other hand, there is a nasty side-effect: iterating the authority means iterating a store that doesn't have its metadata up-to-date as those happen only get updated in the caching tier as long as the mapping doesn't get flushed down.
This means that iterating an authority can very well prematurely expire perfectly valid mappings (think TTI).

lorban:
Resolving #825 first would make fixing this bug much simpler in the light of keeping the implementation JSR-107 compliant.

lorban:
Related to https://github.com/jsr107/jsr107spec/issues/337 as well.

PR #878, opened 2016-03-21
#716: rework Store.iterator() contract …

lorban:
…to move the expiration responsibility to the Cache layer

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
+++ b/107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
@@ -539,7 +539,8 @@ class Eh107Cache<K, V> implements Cache<K, V> {
       @Override
       public Entry<K, V> next() {
         checkClosed();
-        return new WrappedEhcacheEntry<K, V>(specIterator.next());
+        org.ehcache.Cache.Entry<K, V> next = specIterator.next();
+        return next == null ? null : new WrappedEhcacheEntry<K, V>(next);
       }
 
       @Override
--- a/107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
+++ b/107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
@@ -31,6 +31,7 @@ import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
 
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -524,8 +525,29 @@ class Eh107Cache<K, V> implements Cache<K, V> {
   }
 
   @Override
-  public java.util.Iterator<Cache.Entry<K, V>> iterator() {
-    return new Iterator<K, V>(ehCache);
+  public Iterator<Entry<K, V>> iterator() {
+    checkClosed();
+
+    final Iterator<org.ehcache.Cache.Entry<K, V>> specIterator = jsr107Cache.specIterator();
+    return new Iterator<Entry<K, V>>() {
+      @Override
+      public boolean hasNext() {
+        checkClosed();
+        return specIterator.hasNext();
+      }
+
+      @Override
+      public Entry<K, V> next() {
+        checkClosed();
+        return new WrappedEhcacheEntry<K, V>(specIterator.next());
+      }
+
+      @Override
+      public void remove() {
+        checkClosed();
+        specIterator.remove();
+      }
+    };
   }
 
   private void checkClosed() {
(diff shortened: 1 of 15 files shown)


107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
```diff
@@ -525,7 +525,28 @@ public void deregisterCacheEntryListener(CacheEntryListenerConfiguration<K, V> c
...
+      @Override
+      public boolean hasNext() {
+        checkClosed();
+        return specIterator.hasNext();
+      }
+
+      @Override
+      public Entry<K, V> next() {
+        checkClosed();
+        return new WrappedEhcacheEntry<K, V>(specIterator.next());
```
ljacomet: so we can no longer get a null here?

ljacomet: From the `SpecIterator` class, we can. So it seems weird to wrap null then.

ljacomet:
Needs a rebase

107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
lorban: Yes, good spot. I'll fix that and make sure there is test coverage for this particular case.

ljacomet: Don't get me wrong, if we can pass the JSR-107 TCK without ever returning null here, fine by me.

107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
```diff
@@ -525,7 +525,28 @@ public void deregisterCacheEntryListener(CacheEntryListenerConfiguration<K, V> c
 
   @Override
   public java.util.Iterator<Cache.Entry<K, V>> iterator() {
```
chrisdennis: We could import this now, right?

lorban: Right. Let me do it.

ljacomet:
That PR adds extra configuration options it seems. Care to comment on their objective?
Also, could you split that in a separate commit?

lorban:
I added a special JSR107 property to be able to pass pre-configured service creation configs to the 107 cache manager.

The sole purpose of that change is to be able to write tests with special services, like I did in IteratorTest that requires a special time source. My intent was not to have this documented in any way nor to let end-users make use of that.

lorban:
Reworked the way one can create a 107 cache manager using a Configuration object

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -126,6 +111,30 @@ public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties
     return cacheManager;
   }
 
+  public Eh107CacheManager createCacheManager(Configuration config) {
+    return createCacheManager(config, null, new Properties());
+  }
+
+  public Eh107CacheManager createCacheManager(Configuration config, URI uri, Properties properties) {
```
ljacomet: I would invert `URI` and `Configuration` parameters and make this one use the proper cache manager registration.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -126,6 +111,30 @@ public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties
     return cacheManager;
   }
 
+  public Eh107CacheManager createCacheManager(Configuration config) {
```
ljacomet: I would drop this method and force the user to give us a URI that we check is unique.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -78,6 +78,17 @@ public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties
...
+      if (URI_DEFAULT.equals(uri)) {
+        config = new DefaultConfiguration(classLoader);
+      } else {
+        config = new XmlConfiguration(uri.toURL(), classLoader);
+      }
+    } catch (Exception e) {
+      throw new javax.cache.CacheException(e);
+    }
+
     synchronized (cacheManagers) {
```
ljacomet: The combined proposed changes would allow this whole logic to move to the method `createCacheManager(URI, Configuration, Properties)`

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
ljacomet: And there could be a variant that does not take properties.

ljacomet:
Good to go once rebased

Issue #880, opened 2016-03-22
Merge back Cache implementation classes


ljacomet:
From various discussions and testing, it seems there was an alternate way to resolve the `Cache.get(K key)` performance issue without splitting implementation classes and which would end up speeding the call even when having a loader writer configured.

The idea is to have `Cache.get(K key)` invoke `Store.get(K key)` first and only do the lambda based invocation if it returns `null` and a `CacheLoaderWriter` is configured.

ljacomet:
Pending progress on clustered.

ljacomet:
From the clustered discussion currently happening, the different implementation will enable to make progress faster. While we may revisit a merge at a later stage, now is not the time to do so.

PR #882, opened 2016-03-22
Issue #880 Merge Cache implementations

ljacomet:
* Merging back `Cache` implementations in a single `Ehcache` class
* Implement `Ehcache.get` short circuit that invokes `Store.get`


core-spi-test/src/main/java/org/ehcache/internal/store/StoreReplaceKeyValueValueTest.java
```diff
@@ -151,7 +151,7 @@ public void wrongKeyTypeThrowsException()
 
     try {
       if (this.factory.getKeyType() == String.class) {
```
chrisdennis: Not your code, but this type check is wrong, it should be: `this.factory.getKeyType().isAssignableFrom(String.class)`

impl/src/test/java/org/ehcache/impl/config/serializer/SerializerCountingTest.java
```diff
@@ -92,7 +91,7 @@ public void testOnHeapPutGet() {
     printSerializationCounters("Get OnHeap");
 
     cache.put(42L, "Wrong ...");
-    assertCounters(2, 2, 0, 1, 0, 0);
+    assertCounters(2, 2, 0, 1, 1, 0);
```
chrisdennis: The numbers went up... do we care about that?

ljacomet: Put that underneath uses compute does indeed add extra deserialization of value. Adding a general comment to attract attention.

ljacomet:
This change introduces extra deserialization on put - see `SerializerCountingTest`. @lorban what was the perf status for put?

ljacomet:
Not doing for now given clustering discussion with @chrisdennis, @lorban and @cljohnso

Issue #834, opened 2016-03-10
Clustered Caching Configuration XML Parser


chrisdennis:
Implement an XML parser for the XML configuration defined in #833 that converts an XML document in to a series of calls against the builder API.

Issue #833, opened 2016-03-10
Clustered Caching Client Side Configuration


chrisdennis:
Design XSD and programmatic API for configuring clustered caches.  The current plan here is to support caches that are either explicitly sized or share an explicitly sized pool.  We will need 'resource' configuration for explicit sizing and sharing at the cache level, and 'service' level configuration for the shared pools.  Both pools and explicitly sized caches will allocate storage out of named server side resources that will have been configured in the server configuration.

cljohnso:
Need programmatic creation of ClusteredResourceType

PR #886, opened 2016-03-23
Issue #833 Create clustered store in support of clustered cache

cljohnso:
This commit addresses Issues #833 and #834 and partially addresses
Issue #760.

This commit lays the foundation for a ClusteredStore along with many
underlying support constructs.

* The byte-count 'memory-unit' and 'resource-unit' enumeration values
  in ehcache-core.xsd are changed to upper-case (with the exception of
  'kB').
* ResourcePool is split into ResourcePool and SizedResourcePool along
  with associated implementations.
* The XAStore.provider 'rank' method is changed to use 1000 as its base
  value.  This makes room for other 'grouping' providers.
* MaintainableService and PersistableResourceService are marked as
  plural services.
* Use of org.terracotta.offheapstore.util.MemoryUnit in the clustering
  code is changed to org.ehcache.config.units.MemoryUnit.
* A 'with(ResourcePool)' method is added to ResourcePoolsBuilder to
  accept non-core resources.
* The behavior of the ResourcePoolsBuilder 'with', 'heap', 'offheap',
  and 'disk' methods are changed to throw an IllegalArgumentException
  when a resource pool of a type already existing in the pool collection
  instead of silently replacing the existing pool.  The
  'withReplacing(ResourcePool)' method is added for preserve the old
  add/replace behavior.
* Removes the private ConfigurationParser.ResourcePoolImpl class in
  deference to the "regular" implementation.

ClusteredStore support specifics include:

* Introduction of the CacheResourceConfigurationParser and
  ClusteredResourceConfigurationParser to handle parsing of the
  <cluster-fixed> and <cluster-shared> cache resource elements.
  ConfigurationParser is updated to handle parsing resource extension
  elements.
* ClusteredResourceType, ClusteredResourcePool, and
  ClusteredResourcePoolBuilder are introduced to represent clustered
  cache resources.

-----

The `ClusteredStore` implementation in this commit is a rudimentary shell -- it relies on having a heap (or other store) configured to be functional.  This commit also lacks code which links clustered resources to the `ClusteringService` configuration.

### DIFF
--- a/107/src/test/java/org/ehcache/ParsesConfigurationExtensionTest.java
+++ b/107/src/test/java/org/ehcache/ParsesConfigurationExtensionTest.java
@@ -19,6 +19,8 @@ package org.ehcache;
 import com.pany.domain.Customer;
 import com.pany.domain.Product;
 import com.pany.ehcache.integration.ProductCacheLoaderWriter;
+
+import org.ehcache.config.SizedResourcePool;
 import org.ehcache.config.builders.CacheConfigurationBuilder;
 import org.ehcache.config.CacheRuntimeConfiguration;
 import org.ehcache.core.EhcacheManager;
@@ -80,7 +82,7 @@ public class ParsesConfigurationExtensionTest {
       // Test the config
       {
         final CacheRuntimeConfiguration<Long, Product> runtimeConfiguration = productCache.getRuntimeConfiguration();
-        assertThat(runtimeConfiguration.getResourcePools().getPoolForResource(ResourceType.Core.HEAP).getSize(), equalTo(200L));
+        assertThat(((SizedResourcePool)runtimeConfiguration.getResourcePools().getPoolForResource(ResourceType.Core.HEAP)).getSize(), equalTo(200L));
 
         final Expiry<? super Long, ? super Product> expiry = runtimeConfiguration.getExpiry();
         assertThat(expiry.getClass().getName(), equalTo("org.ehcache.expiry.Expirations$TimeToIdleExpiry"));
@@ -126,7 +128,7 @@ public class ParsesConfigurationExtensionTest {
     {
       final Cache<Long, Customer> customerCache = cacheManager.getCache("customerCache", Long.class, Customer.class);
       final CacheRuntimeConfiguration<Long, Customer> runtimeConfiguration = customerCache.getRuntimeConfiguration();
-      assertThat(runtimeConfiguration.getResourcePools().getPoolForResource(ResourceType.Core.HEAP).getSize(), equalTo(200L));
+      assertThat(((SizedResourcePool)runtimeConfiguration.getResourcePools().getPoolForResource(ResourceType.Core.HEAP)).getSize(), equalTo(200L));
     }
   }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
@@ -34,7 +34,8 @@ public interface FixedClusteredResourcePool extends ClusteredResourcePool, Sized
 
   /**
    * Identifies the server-side clustered resource from which space for this pool is reserved.
-   * If no server resource is identified, TODO: Identify the space source ...
+   * If no server resource is identified, the resource identified in the server configuration as
+   * the default is used.
    *
    * @return the server-side clustered resource id; may be {@code null}
    */
--- a/clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.config;
+
+import org.ehcache.config.SizedResourcePool;
+import org.ehcache.config.units.MemoryUnit;
+
+/**
+ * Specifies a {@link ClusteredResourcePool} reserving space from a server-based clustered resource.
+ * A {@code FixedClusteredResourcePool} uses dedicated space on the server.
+ */
+public interface FixedClusteredResourcePool extends ClusteredResourcePool, SizedResourcePool {
+
+  /**
+   * {@inheritDoc}
+   * @return {@inheritDoc}
+   */
+  @Override
+  MemoryUnit getUnit();
+
+  /**
+   * Identifies the server-side clustered resource from which space for this pool is reserved.
+   * If no server resource is identified, TODO: Identify the space source ...
+   *
+   * @return the server-side clustered resource id; may be {@code null}
+   */
+  String getFromResource();
+}
--- a/core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
+++ b/core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
@@ -67,12 +67,10 @@ public class ResourcePoolsImpl implements ResourcePools {
       throw new IllegalArgumentException("Pools to be updated cannot contain previously undefined resources pools");
     }
     // Can not update OFFHEAP
-    // TODO: Move to OFFHEAP ResourcePool
     if(toBeUpdated.getResourceTypeSet().contains(ResourceType.Core.OFFHEAP)) {
       throw new UnsupportedOperationException("Updating OFFHEAP resource is not supported");
     }
     // Can not update DISK
-    // TODO: Move to DISK ResourcePool
     if(toBeUpdated.getResourceTypeSet().contains(ResourceType.Core.DISK)) {
       throw new UnsupportedOperationException("Updating DISK resource is not supported");
     }
@@ -95,8 +93,6 @@ public class ResourcePoolsImpl implements ResourcePools {
    *
    * @param pools the resource pools to validate
    */
-  // TODO: Move to org.ehcache.impl.internal.store.tiering.CacheStore.Provider
-  // TODO: Generalize for all SizedResourcePool implementations (needs addition of defined ordering)
   public static void validateResourcePools(Collection<? extends ResourcePool> pools) {
     EnumMap<ResourceType.Core, SizedResourcePool> coreResources = new EnumMap<ResourceType.Core, SizedResourcePool>(ResourceType.Core.class);
     for (ResourcePool pool : pools) {
--- a/core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
+++ b/core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
@@ -34,9 +34,9 @@ import java.util.Set;
  */
 public class ResourcePoolsImpl implements ResourcePools {
 
-  private final Map<ResourceType, ResourcePool> pools;
+  private final Map<ResourceType<?>, ResourcePool> pools;
 
-  public ResourcePoolsImpl(Map<ResourceType, ResourcePool> pools) {
+  public ResourcePoolsImpl(Map<ResourceType<?>, ResourcePool> pools) {
     validateResourcePools(pools.values());
     this.pools = pools;
   }
@@ -45,15 +45,15 @@ public class ResourcePoolsImpl implements ResourcePools {
    * {@inheritDoc}
    */
   @Override
-  public ResourcePool getPoolForResource(ResourceType resourceType) {
-    return pools.get(resourceType);
+  public <P extends ResourcePool> P getPoolForResource(ResourceType<P> resourceType) {
+    return resourceType.getResourcePoolClass().cast(pools.get(resourceType));
   }
 
   /**
    * {@inheritDoc}
    */
   @Override
-  public Set<ResourceType> getResourceTypeSet() {
+  public Set<ResourceType<?>> getResourceTypeSet() {
     return pools.keySet();
   }
 
@@ -74,13 +74,13 @@ public class ResourcePoolsImpl implements ResourcePools {
     if(toBeUpdated.getResourceTypeSet().contains(ResourceType.Core.DISK)) {
       throw new UnsupportedOperationException("Updating DISK resource is not supported");
     }
-    for(ResourceType currentResourceType : toBeUpdated.getResourceTypeSet()) {
+    for(ResourceType<?> currentResourceType : toBeUpdated.getResourceTypeSet()) {
       getPoolForResource(currentResourceType).validateUpdate(toBeUpdated.getPoolForResource(currentResourceType));
     }
 
-    Map<ResourceType, ResourcePool> poolsMap = new HashMap<ResourceType, ResourcePool>();
+    Map<ResourceType<?>, ResourcePool> poolsMap = new HashMap<ResourceType<?>, ResourcePool>();
     poolsMap.putAll(pools);
-    for(ResourceType currentResourceType : toBeUpdated.getResourceTypeSet()) {
+    for(ResourceType<?> currentResourceType : toBeUpdated.getResourceTypeSet()) {
       ResourcePool poolForResource = toBeUpdated.getPoolForResource(currentResourceType);
       poolsMap.put(currentResourceType, poolForResource);
     }
--- a/core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
+++ b/core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
@@ -19,6 +19,7 @@ package org.ehcache.core.config;
 import org.ehcache.config.ResourcePool;
 import org.ehcache.config.ResourcePools;
 import org.ehcache.config.ResourceType;
+import org.ehcache.config.SizedResourcePool;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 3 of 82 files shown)


clustered/client/notes.adoc
```diff
@@ -0,0 +1,189 @@
+== Discussions with Chris Dennis
```
cljohnso: Please **IGNORE** this file ... it will be pulled.

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
```diff
@@ -69,25 +69,56 @@ public static ResourcePoolsBuilder newResourcePoolsBuilder() {
...
+  }
+
+  /**
+   * Add or replace the {@link ResourcePool} of {@link ResourceType} in the returned builder.
+   *
+   * @param resourcePool the non-{@code null} resource pool to add/replace
+   * @return a new builder with the added pool
+   */
+  // TODO: Determine the use case for this method
+  public ResourcePoolsBuilder withReplacing(ResourcePool resourcePool) {
```
cljohnso: I was unable to locate a place where this method could be effectively used -- it may, at this point, be a candidate for removal.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -745,7 +745,7 @@ public int rank(final Set<ResourceType> resourceTypes, final Collection<ServiceC
       }
 
       final Store.Provider candidateUnderlyingProvider = selectProvider(resourceTypes, serviceConfigs, xaServiceConfiguration);
-      return 10 + candidateUnderlyingProvider.rank(resourceTypes, serviceConfigs);
+      return 1000 + candidateUnderlyingProvider.rank(resourceTypes, serviceConfigs);
```
cljohnso: This was pushed up to make room for `ClusteredStore` on the presumption that an `XAStore` would _wrap_ a `ClusteredStore`.

107/src/test/java/org/ehcache/ParsesConfigurationExtensionTest.java
```diff
@@ -80,7 +82,7 @@ public void testXmlExample() throws ClassNotFoundException, SAXException, Instan
       // Test the config
       {
         final CacheRuntimeConfiguration<Long, Product> runtimeConfiguration = productCache.getRuntimeConfiguration();
-        assertThat(runtimeConfiguration.getResourcePools().getPoolForResource(ResourceType.Core.HEAP).getSize(), equalTo(200L));
+        assertThat(((SizedResourcePool)runtimeConfiguration.getResourcePools().getPoolForResource(ResourceType.Core.HEAP)).getSize(), equalTo(200L));
```
chrisdennis: This feels weird... I wonder if we can add generics to the ResourceType so that necessary cast be internalized and tied to the type of the resource being requested?

core/src/main/java/org/ehcache/core/config/SizedResourcePoolImpl.java
```diff
@@ -74,10 +68,25 @@ public ResourceUnit getUnit() {
...
    */
   @Override
-  public boolean isPersistent() {
-    return persistent;
+  public void validateUpdate(final ResourcePool newPool) {
+    super.validateUpdate(newPool);
+
+    SizedResourcePool sizedPool = (SizedResourcePool)newPool;
+    // Ensure unit type has not changed
+    if (!this.getUnit().getClass().equals(sizedPool.getUnit().getClass())) {
```
cljohnso: The old code in `org.ehcache.core.config.ResourcePoolsImpl#validateAndMerge` prevented any unit change -- even among `MemoryUnit` constants.  After a brief discussion with @chrisdennis, I switched to this method to prevent switching between `MemoryUnit` and `EntryUnit`.

core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
```diff
@@ -61,34 +62,29 @@ public ResourcePool getPoolForResource(ResourceType resourceType) {
    */
   @Override
   public ResourcePools validateAndMerge(ResourcePools toBeUpdated) {
+    // Ensure update pool types already exist in existing pools
     if(!getResourceTypeSet().containsAll(toBeUpdated.getResourceTypeSet())) {
       throw new IllegalArgumentException("Pools to be updated cannot contain previously undefined resources pools");
     }
+    // Can not update OFFHEAP
+    // TODO: Move to OFFHEAP ResourcePool
```
chrisdennis: I think it's better that we file issues for the todos in the non clustered code and remove the todos, this way we have a better chance of tracking and resolving all the todos before unclustered GA.

core/src/main/java/org/ehcache/core/config/SizedResourcePoolImpl.java
```diff
@@ -38,22 +37,17 @@
    * @param unit the unit for the size
    * @param persistent whether the pool is to be persistent
    */
-  public ResourcePoolImpl(ResourceType type, long size, ResourceUnit unit, boolean persistent) {
+  public SizedResourcePoolImpl(ResourceType type, long size, ResourceUnit unit, boolean persistent) {
+    super(type, persistent);
+    if (unit == null) {
+      throw new NullPointerException("ResourceUnit can not be null");
+    }
+    // TODO: Ensure 'type' can be sized?
```
chrisdennis: Could be covered by my generics proposal.

core/src/test/java/org/ehcache/core/config/ResourcePoolsImplTest.java
```diff
@@ -148,8 +149,10 @@ public void testMemoryResourceUnequalUnitInversion() {
 
   @Test
   public void testAddingNewTierWhileUpdating() {
-    ResourcePools existing = new ResourcePoolsImpl(Collections.singletonMap((ResourceType) ResourceType.Core.HEAP, (ResourcePool) new ResourcePoolImpl(ResourceType.Core.HEAP, 10L, EntryUnit.ENTRIES, false)));
-    ResourcePools toBeUpdated = new ResourcePoolsImpl(Collections.singletonMap((ResourceType) ResourceType.Core.DISK, (ResourcePool) new ResourcePoolImpl(ResourceType.Core.DISK, 10L, MemoryUnit.MB, false)));
+    ResourcePools existing = new ResourcePoolsImpl(Collections.singletonMap((ResourceType) ResourceType.Core.HEAP,
+        (ResourcePool) new SizedResourcePoolImpl(ResourceType.Core.HEAP, 10L, EntryUnit.ENTRIES, false)));
```
chrisdennis: These are weird casts... could we maybe use an explicit type parameter on the singletonMap method call.

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
```diff
@@ -69,25 +69,56 @@ public static ResourcePoolsBuilder newResourcePoolsBuilder() {
...
+    return new ResourcePoolsBuilder(newPools);
+  }
+
+  /**
+   * Add or replace the {@link ResourcePool} of {@link ResourceType} in the returned builder.
+   *
+   * @param resourcePool the non-{@code null} resource pool to add/replace
+   * @return a new builder with the added pool
+   */
+  // TODO: Determine the use case for this method
```
chrisdennis: The use case I'm aware of is reusing the builder for a different cache, and wanting to overwrite one of the resource definitions.  In that case though maybe we want the shortcut versions too?

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
```diff
@@ -96,6 +127,8 @@ public ResourcePoolsBuilder with(ResourceType type, long size, ResourceUnit unit
    * @param size the pool size
    * @param unit the pool size unit
    * @return a new builder with the added pool
+   *
+   * @throws IllegalArgumentException if the set of resource pools already contains a pool for {@code type}
```
chrisdennis: "...contains a heap resource"

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
chrisdennis: Did this get changed accidentally?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -89,16 +91,21 @@
 
   private static final SchemaFactory XSD_SCHEMA_FACTORY = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
 
+  private static final URL CORE_SCHEMA_URL = XmlConfiguration.class.getResource("/ehcache-core.xsd");
+  private static final String CORE_SCHEMA_NAMESPACE = "http://www.ehcache.org/v3";
+  private static final String CORE_SCHEMA_ROOT_ELEMENT = "config";
+  private static final String CORE_SCHEMA_JAXB_MODEL_PACKAGE = "org.ehcache.xml.model";
```
chrisdennis: Could we just stick ConfigType.class in a static, and then derive the model package from there?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -89,16 +91,21 @@
...
 
-  public ConfigurationParser(String xml, URL... sources) throws IOException, SAXException, JAXBException, ParserConfigurationException {
+  public ConfigurationParser(String xml) throws IOException, SAXException, JAXBException, ParserConfigurationException {
     Collection<Source> schemaSources = new ArrayList<Source>();
-    for (URL source : sources) {
-      schemaSources.add(new StreamSource(source.openStream()));
-    }
+    schemaSources.add(new StreamSource(CORE_SCHEMA_URL.openStream()));
+
     for (CacheManagerServiceConfigurationParser<?> parser : ClassLoading.libraryServiceLoaderFor(CacheManagerServiceConfigurationParser.class)) {
```
chrisdennis: Looking at this logic, I wonder if we want to file an issue for turning these in to a tree.  Top level service that provides the schema, followed by three accessor to get at the (up to) three parsers?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -541,27 +526,31 @@ public SizeOfEngineLimits heapStoreSettings() {
...
     ResourceType heapResource = resource.getValue();
-    return new ResourcePoolImpl(org.ehcache.config.ResourceType.Core.HEAP,
+    return new SizedResourcePoolImpl(org.ehcache.config.ResourceType.Core.HEAP,
             heapResource.getValue().longValue(), parseUnit(heapResource), false);
   }
 
   private ResourcePool parseResource(Element element) {
+    if (!CORE_SCHEMA_NAMESPACE.equals(element.getNamespaceURI())) {
+      return parseResourceExtension(element);
+    }
```
chrisdennis: I'm terrified of drop-through errors... could we add an else here?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -598,6 +587,15 @@ private static MemoryUnit parseMemory(MemoryType memoryType) {
     return xmlConfigurationParser.parseServiceConfiguration(element);
   }
 
+  ResourcePool parseResourceExtension(final Element element) {
+    URI namespace = URI.create(element.getNamespaceURI());
+    final CacheResourceConfigurationParser xmlConfigurationParser = resourceXmlParsers.get(namespace);
+    if (xmlConfigurationParser == null) {
+      throw new XmlConfigurationException("Can't find parser for namespace: " + namespace);
+    }
```
chrisdennis: again, else urges.

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -83,16 +83,12 @@
 
 /**
  * Exposes {@link org.ehcache.config.Configuration} and {@link CacheConfigurationBuilder} expressed
- * in a XML file that obeys the ehcache-core.xsd (todo link this to proper location, wherever this ends up being)
+ * in a XML file that obeys the ehcache-core.xsd. (todo link this to proper location, wherever this ends up being)
```
chrisdennis: @ljacomet we should probably create an issue to track doing this when we go GA.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
cljohnso: No, this was on purpose -- to make room for `ClusteredStore`.  If my presumption about the implementation orderings needs to be changed, let me know.

core/src/test/java/org/ehcache/core/config/ResourcePoolsImplTest.java
cljohnso: Old code ... but I agree that a type parameter is better than a cast in this case.  Want me to change it?

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
cljohnso: If we have unit tests that cover that scenario then the code works differently that you expect -- the original `with` methods are never called for a replacement.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
cljohnso: If you want to do it that way -- sure.  The constant that appears here is derived directly from the original code.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
cljohnso: Not sure I follow ... accessors on what?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
cljohnso: The else is the rest of the code in the method ... not sure I see the point.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
cljohnso: Again, I don't see the point ... the `if` in this case is checking for an exceptional condition.  You'd apply this _everywhere_ an if is used?

clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * {@inheritDoc}
+   * @return {@inheritDoc}
+   */
+  @Override
+  MemoryUnit getUnit();
+
+  /**
+   * Identifies the server-side clustered resource from which space for this pool is reserved.
+   * If no server resource is identified, TODO: Identify the space source ...
```
cljohnso: @chrisdennis What is the best way to describe the source of the space consumed when the "from resource" is not identified?

core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
cljohnso: I'm converting the `TODO`s to issues for all except the `ClusteredStore` class since this class still needs major work.

core/src/test/java/org/ehcache/core/config/ResourcePoolsImplTest.java
chrisdennis: Already touching the line so might as well.

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
chrisdennis: Wouldn't be surprised if there was no test coverage of that use case: @ljacomet care to weigh in?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
chrisdennis: Oh, I know... I didn't like the way Alex did it before... now that someone is touching this I can get the 'fixed'!

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
chrisdennis: I'll file an issue for this.

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
chrisdennis: Okay - I'll do it then.

clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
chrisdennis: It will use the ClusteringServiceConfiguration's default server resource... how you want to word that here I don't know.

transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
chrisdennis: This is fine then.

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
cljohnso: @chrisdennis is creating an issue.  Should this TODO be removed?  Replaced with a link _presuming_ the XSD is located at http://www.ehcache.org.v3?

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
cljohnso: I opened issue #893 for this.

core/src/test/java/org/ehcache/core/config/ResourcePoolsImplTest.java
cljohnso: Done

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
cljohnso: I added issue #893 for evaluating whether or not this method is needed.  At a minimum, one addressing the issue could add a unit test or two reinforcing it's use case.

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
cljohnso: Changed the ones for disk and offheap as well.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
cljohnso: Done.

```
  private static final String CORE_SCHEMA_JAXB_MODEL_PACKAGE = ConfigType.class.getPackage().getName();
```

clustered/client/src/main/java/org/ehcache/clustered/config/FixedClusteredResourcePool.java
cljohnso: "If no server resource is identified, the resource identified in the server configuration as the default is used."

Would be good to have some "name" reference ...

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
cljohnso: TODO removed; comment added to Issue #894.

cljohnso:
The 75ebb18 commit does not address the parameterization of `ResourceType`.

cljohnso:
This should close both #833 and #834.

chrisdennis:
Subsumed by PR #904

PR #889, opened 2016-03-24
Issue #851 Resolve race in clear when byte sizing

ljacomet:
Made the backend responsible for holding the bye size filed and
implement clear through simple backed re-creation.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
@@ -142,15 +142,13 @@ public class OnHeapStore<K, V> implements Store<K,V>, HigherCachingTier<K, V> {
   };
 
   static final int SAMPLE_SIZE = 8;
+  private volatile Backend<K, V> map;
 
-  private final Backend<K, V> map;
   private final Class<K> keyType;
   private final Class<V> valueType;
   private final Copier<V> valueCopier;
 
   private final SizeOfEngine sizeOfEngine;
-  private final boolean byteSized;
-  private final AtomicLong currentUsageinBytes = new AtomicLong(0L);
 
   private volatile long capacity;
   private final EvictionVeto<? super K, ? super V> evictionVeto;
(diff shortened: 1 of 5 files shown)


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -227,13 +227,10 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
     this.keyType = config.getKeyType();
     this.valueType = config.getValueType();
     this.expiry = config.getExpiry();
+    this.keyCopier = keyCopier;
```
chrisdennis: Could we eliminate the keyCopier here by having the backend return another backend on clear?  Is that cleaner?

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -642,17 +647,7 @@ public ReplaceStatus replace(final K key, final V oldValue, final V newValue) th
...
-      long current = currentUsageinBytes.get();
-      map.clear();
-      current = currentUsageinBytes.addAndGet(-current);
-      if(current < 0L) {
-        currentUsageinBytes.addAndGet(current);
-      }
-    } catch (RuntimeException re) {
-      handleRuntimeException(re);
-    }
+    createBackend();
```
chrisdennis: this would become `map = map.clear();`

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1458,27 +1453,11 @@ private static long safeExpireTime(long now, Duration duration) {
...
   }
 
-  protected long getCurrentUsageInBytes() {
-    if (byteSized) {
-      return currentUsageinBytes.get();
-    }
-    return 0L;
-  }
-
-  private long size() {
```
chrisdennis: Wat the wat?!

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
ljacomet: Yes, I considered doing it so your suggestion reinforces that.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
ljacomet: Doing that

ljacomet:
Remarks addressed

Issue #897, opened 2016-03-24
Current OnHeapStore implementation will not enforce capacities > Integer.MAX_VALUE correctly


chrisdennis:
The capacity enforcement in OnHeapStore will currently not work for capacities greater than `Integer.MAX_VALUE`, caches configured as such will have unbounded capacities.  This can be fixed by having OnHeapStore call `mappingCount()` instead of `size()` on the underlying CHM since it has a long return.

PR #898, opened 2016-03-25
Issue #897 Support for long mapping count

ljacomet:


impl/src/main/java/org/ehcache/impl/internal/store/heap/Backend.java
```diff
@@ -56,7 +56,7 @@
    *
    * @return the number of mappings
    */
-  int size();
```
lorban: This method should probably be renamed to something like `mappingCount()` as I find `size()` vs `naturalSize()` rather confusing.

ljacomet: Ok

Issue #899, opened 2016-03-25
Release RC2


ljacomet:
- [x] Tag release
- [x] Build and verify release
 - includes checking javadoc and source jars
- [x] Update and publish release on GitHub
- [x] Update and publish website
- [x] Publish jars to Maven Central
- [x] Email announcement

PR #901, opened 2016-03-25
Issue #899 Minor documentation improvements

ljacomet:


docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -67,6 +65,8 @@ NOTE: See link:usermanaged{outfilesuffix}[the user managed cache documentation]
 
 === Tiering
 
+Ehcache 3, as in previous version, offers a tiering model to allow to store growing amount of data on slower tiers.
```
chrisdennis: Ehcache 3, as in previous versions, offers a tiering model to allow storing increasing amounts of data on slower tiers.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -76,7 +76,10 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
 
 <1> If you wish to use off-heap, you'll have to define resource pool, giving the memory size you want to allocate.
 
-Do not forget to define in the java options the -XX:MaxDirectMemorySize option, according to the off-heap size you intend to use.
+The example above allocates a very small amount of off-heap.
+Remember that data stored off-heap will have to be serialized and deserialized.
+You should thus favor off-heap for large amounts of data where storing them on heap would have too severe impacts on garbage collection.
```
chrisdennis: You should thus favor off-heap for large amounts of data where on heap would have too severe an impact on garbage collection.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -76,7 +76,10 @@ include::{sourcedir}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag
...
+Do not forget to define in the java options the `-XX:MaxDirectMemorySize` option, according to the off-heap size you intend to use.
```
chrisdennis: This seems to imply setting these to the same value... thats not really what we want... but I'm not sure what we should say instead.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -207,6 +216,9 @@ In this particular example, the `OddKeysEvictionVeto` class will veto eviction o
 By the time the cache manager gets closed, only mappings with odd keys should be left in the cache as there isn't
 enough room for mapping with even keys.
 
+NOTE: Veto status is computed when a mapping is written to the cache.
+This means that proper veto implementation are expected to be constant for a key-value pair.
```
chrisdennis: implementations

docs/src/docs/asciidoc/user/getting-started.adoc
ljacomet: Let's not overthinking it right now, we can work on something better when we do a full section on tiering configuration, pros and cons, etc...

chrisdennis: ok

Issue #806, opened 2016-03-03
Create a set of default optimized copiers


lorban:
Much like we have a default set of optimized serializers (for `Long`, `String`...) we should create a comparable set of copiers.

This would make store-by-value perform as fast as store-by-ref in some cases, like when immutable key/value classes are configured and the copiers can simply return the object without making any copy.

This would mostly impact the out-of-the-box performance of JSR107 as it requires store-by-value to be the default mode.

lorban:
Or maybe more simply: register the `IdentityCopier` as the copier for all common immutable types like `Integer`, `Long`, `String`...

PR #908, opened 2016-03-29
Adding Default Copiers for better perf in jsr #806

AbfrmBlr:
* Creating a copy of IdentityCopier in JSR i.e. 'Eh107IdentityCopier' since thye logic of deciding whether as cache is storebyval or not depends on the kind of Copier configured for val type(in case of template overriding).

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
+++ b/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
@@ -30,7 +30,9 @@ import org.ehcache.xml.XmlConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import javax.cache.configuration.CacheEntryListenerConfiguration;
@@ -135,6 +137,7 @@ class ConfigurationMerger {
               xmlConfiguration.getServiceCreationConfigurations().toArray());
           if (defaultCopyProviderConfiguration != null) {
             Map<Class<?>, ClassInstanceConfiguration<Copier<?>>> defaults = defaultCopyProviderConfiguration.getDefaults();
+            handleCopierDefaultsforImmutableTypes(defaults);
             boolean matchingDefault = false;
             if (defaults.containsKey(jsr107Configuration.getKeyType())) {
               matchingDefault = true;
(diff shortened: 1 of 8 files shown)


107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -162,6 +164,43 @@
     return builder;
   }
 
+  private static <K, V> CacheConfigurationBuilder<K, V> addDefaultCopiers(CacheConfigurationBuilder<K, V> builder, Class keyType, Class valueType ) {
```
lorban: Why not having this logic in `DefaultCopyProvider` so that ehcache API users can also benefit from those copiers?

impl/src/main/java/org/ehcache/impl/config/copy/DefaultCopierConfiguration.java
```diff
@@ -55,7 +55,7 @@ public DefaultCopierConfiguration(Copier<T> instance, Type type) {
     this.type = type;
   }
 
-  DefaultCopierConfiguration(Class<? extends Copier<T>> copierClass) {
+  public DefaultCopierConfiguration(Class<? extends Copier<T>> copierClass) {
```
lorban: This constructor was not set as public for a good reason. Better not change that visibility and use a different constructor instead.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
AbfrmBlr: But in ehcache default copier is always 'IdentityCopier', no matter what is the type. So if no copier is provided IdentityCopier will be used. But in JSR, default copier is SerializingCopier.

lorban: Yeah, you're damn right. Ignore what I said, I just remembered that we changed the way to configure store-by-value in ehcache to configure copiers.

ljacomet:
Changes to `ConfigurationMerger` should have matching unit tests, accounting for the different use cases.

107/src/test/resources/ehcache-107-default-copiers.xml
```diff
@@ -35,4 +40,29 @@
     <ehcache:heap>10</ehcache:heap>
   </ehcache:cache-template>
 
+  <ehcache:cache-template name="stringtemplate">
```
ljacomet: Why reuse an existing XML instead of creating a new one? Separation of concerns.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
```diff
@@ -273,6 +278,53 @@ public void jsr107LoaderInitAlways() {
     assertThat(configHolder.useEhcacheLoaderWriter, is(false));
   }
 
+  @Test
+  public void jsr107DefaultEh107IdentityCopierForImmutableTypes() {
+    XmlConfiguration xmlConfiguration = new XmlConfiguration(getClass().getResource("/ehcache-107-default-copiers.xml"));
+    final DefaultJsr107Service jsr107Service = new DefaultJsr107Service(ServiceLocator.findSingletonAmongst(Jsr107Configuration.class, xmlConfiguration.getServiceCreationConfigurations().toArray()));
+    merger = new ConfigurationMerger(xmlConfiguration, jsr107Service, mock(Eh107CacheLoaderWriterProvider.class));
```
ljacomet: This big tests is a single code path for different types. What about testing the other code path?
No XML configuration, XML config without any default copiers, etc ...

107/src/test/resources/ehcache-107-default-copiers.xml
AbfrmBlr: Its reused because it already talks about default copiers, and that's the change here.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
AbfrmBlr: This test is about XML config without any default copiers provided for the immutable types, so what else is missing.
No XML config:- all the JSR caches are usable through XML config only, or you want mock based tests when you say No XML config.

107/src/test/resources/ehcache-107-default-copiers.xml
ljacomet: Well that XML is about testing the configuration of a default copier while your test is about getting smarter defaults when none configured.Not the same to me.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
ljacomet: Not true, with the code changes you have done, unless I missed something, there will be a smarter copier installed if you create a cache with immutable types purely in code using JSR-107 API. That is not tested for example.
And the code change you did in the configuration merger takes different path depending on the existence of default copier configurations at the CM level.

Issue #896, opened 2016-03-24
Bypass serialization of byte array as value while putting them in offheap/disk tier.


skbansal:
As of now we dont have byte array in white list of data types wherein we bypass the serialization. It would be advantageous if we bypass it while accessing it in offheap and disk tier.

skbansal:
I have tried same tests with custom serializer, but haven't noticed considerable improvement. 

TPS without custom serializer : 550k
TPS with custom serializer : 570k

Java : 1.7.0.72
Element (Long, ByteArray(1kb))
Usecase : R:W = 90:10

Custom Serializer Class:
public class ByteArraySerializer implements Serializer {
  public ByteArraySerializer() {  }
  public ByteArraySerializer(ClassLoader classLoader) {  }

  @Override
  public ByteBuffer serialize(Object o) throws SerializerException {
    byte[] array = (byte[]) o;
    ByteBuffer buffer = ByteBuffer.wrap(array);
    return buffer;
  }

  @Override
  public Object read(ByteBuffer byteBuffer) throws ClassNotFoundException, SerializerException {
    byte[] array = new byte[byteBuffer.remaining()];
    byteBuffer.get(array);
    return array;
  }

  @Override
  public boolean equals(Object o, ByteBuffer byteBuffer) throws ClassNotFoundException, SerializerException {
    byte[] array = (byte[]) o;
    return Arrays.equals(array, (byte[]) read(byteBuffer));
  }
}

skbansal:
Executed same tests and perf stats has not shown much gain. 

Before fix, it was 550k and with fix, it is 580k.

PR #910, opened 2016-03-30
#896 add optimized byte array serializer in default config

lorban:
Comments are welcome but I'd like to wait for the perf results before this gets merged in.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/serialization/ByteArraySerializer.java
+++ b/impl/src/main/java/org/ehcache/impl/serialization/ByteArraySerializer.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.impl.serialization;
+
+import org.ehcache.core.spi.service.FileBasedPersistenceContext;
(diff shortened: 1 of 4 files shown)


impl/src/main/java/org/ehcache/impl/serialization/ByteArraySerializer.java
```diff
@@ -0,0 +1,91 @@
...
+    binary.get(bytes);
+    return bytes;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public boolean equals(byte[] object, ByteBuffer binary) throws ClassNotFoundException, SerializerException {
+    return Arrays.equals(read(binary), object);
```
chrisdennis: This isn't the equals contract of arrays. (also wouldn't it be faster to do `binary.equals(serialize(object))`?)

lorban: Two remarks:

1) replacing `Arrays.equals(read(binary), object)` with `binary.equals(serialize(object))` would indeed be faster but the latter doesn't 'consume' the bytes of the byte buffer (position remains unchanged). All the other serializers do but that isn't documented. Should this impl also consume the bytes or not (by updating the position in your proposed impl)? And should this be documented or not?

2) I know that this isn't the real `byte[].equals()` contract, but without this 'improvement', atomic operations and the XAStore won't work. And somewhat unrelated, Hibernate stores `Object[]` values IIRC, which means we might need to add more array serializers to make it work with XA caches.

impl/src/main/java/org/ehcache/impl/serialization/ByteArraySerializer.java
```diff
@@ -0,0 +1,92 @@
...
+    return bytes;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public boolean equals(byte[] object, ByteBuffer binary) throws ClassNotFoundException, SerializerException {
+    boolean equals = binary.equals(serialize(object));
+    binary.position(binary.remaining());
```
chrisdennis: I think you want `binary.position(binary.limit())` (and I still think this is wrong)

lorban:
test this please

AbfrmBlr:
test this pleasea

ljacomet:
Any news on perf results? @lorban @skbansal

Issue #884, opened 2016-03-23
Persist type information for disk persistent caches (and use for type enforcement)


chrisdennis:
My current understanding is that disk persistent caches do not save any key/value type information.  This means there is no actual type safety for users when accessing persistent caches in an apparently type safe caches.  We should add test coverage and implement this feature.

PR #911, opened 2016-03-30
#884 Persist and use for type enforcement type information for disk …

lorban:
…persistent caches

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
@@ -70,6 +70,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
@@ -83,9 +84,15 @@ public class OffHeapDiskStore<K, V> extends AbstractOffHeapStore<K, V> implement
 
   private static final Logger LOGGER = LoggerFactory.getLogger(OffHeapDiskStore.class);
 
+  private static final String KEY_TYPE_PROPERTY_NAME = "keyType";
+  private static final String VALUE_TYPE_PROPERTY_NAME = "valueType";
+
   protected final AtomicReference<Status> status = new AtomicReference<Status>(Status.UNINITIALIZED);
 
   private final EvictionVeto<K, OffHeapValueHolder<V>> evictionVeto;
+  private final Class<K> keyType;
+  private final Class<V> valueType;
+  private final ClassLoader classLoader;
   private final Serializer<K> keySerializer;
   private final Serializer<V> valueSerializer;
   private final long sizeInBytes;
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -128,21 +135,51 @@ public OffHeapDiskStore(FileBasedPersistenceContext fileBasedPersistenceContext,
...
+    FileInputStream fis = new FileInputStream(metadataFile);
+    Properties properties = new Properties();
+    try {
+      properties.load(fis);
+    } finally {
+      fis.close();
+    }
+    try {
+      Class<?> persistedKeyType = classLoader.loadClass(properties.getProperty("keyType"));
+      if (!keyType.isAssignableFrom(persistedKeyType)) {
```
chrisdennis: Don't these need to be equality checks?  Otherwise anyone can retrieve this cache as say <Object, Object> and then pollute the cache with bad entries?

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -128,21 +135,51 @@ public OffHeapDiskStore(FileBasedPersistenceContext fileBasedPersistenceContext,
...
+
+    FileInputStream fis = new FileInputStream(metadataFile);
+    Properties properties = new Properties();
+    try {
+      properties.load(fis);
+    } finally {
+      fis.close();
+    }
+    try {
+      Class<?> persistedKeyType = classLoader.loadClass(properties.getProperty("keyType"));
```
chrisdennis: Could we put "keyType" and "valueType" in to static constants?

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
lorban: That's a question I had in mind: do we allow compatible type definitions in the cache config or not? I went this route because it allows using a Cache<Object, Object> with a Store<String, String> for instance.

But thinking twice about this, I believe you're right: configured types should be exactly the same as the store types otherwise polluting the store's contents become possible.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
lorban: Sure, will do.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -128,21 +138,51 @@ public OffHeapDiskStore(FileBasedPersistenceContext fileBasedPersistenceContext,
...
+    Properties properties = new Properties();
+    try {
+      properties.load(fis);
+    } finally {
+      fis.close();
+    }
+    try {
+      Class<?> persistedKeyType = classLoader.loadClass(properties.getProperty(KEY_TYPE_PROPERTY_NAME));
+      if (!keyType.equals(persistedKeyType)) {
+        throw new IllegalArgumentException("Persisted key type '" + persistedKeyType.getName() + "' is not assignable from configured key type '" + keyType.getName() + "'");
```
chrisdennis: Your exception messages should probably be updated to match the behavior change.

chrisdennis:
Test this please

PR #914, opened 2016-03-31
Initial proposal on clustered cache operations

chrisdennis:
Please place your review requirements here.  If you'd prefer a rendered copy then feel free to fetch and render at your leisure.


docs/src/docs/asciidoc/developer/clustered.adoc
```diff
@@ -0,0 +1,175 @@
...
+
+ * `List<binary-blob> get(hash)`: returns the list for the given hash
+ * `void append(hash, binary-blob)`: adds the given binary blob to the end of
+   the list for the given hash
+ * `list<binary-blob> getAndAppend(hash, binary-blob)`: the same as append but
+   also returns the list immediately prior to mutation (atomically).
+ * `void replace(hash, list<binary-blob>, list<binary-blob>)`: replaces the
+   first occurrence of the given subsequence with the supplied sequence
+
+This is the entirety of the functionality and data structure exposed by the
```
jhouserizer: 'exposed by the client'  - I know this is literally true, but "client" can be a matter of perspective, so it may be worth noting to readers that it is also the api exposed *to* the client side to work with / upon.

chrisdennis: It's a typo... I meant 'exposed to the client'

docs/src/docs/asciidoc/developer/clustered.adoc
```diff
@@ -0,0 +1,174 @@
...
+                +----------------------------------------------------------+
+                | +------+ /----------\          /----------\              |
+                | |k1, v1| |k1, f(k,v)|          |k1, g(k,v)|              |
+                | +------+ \----------/          \----------/              |
+                +----------------------------------------------------------+
+--------------------------------------------------------------------------------
+
+The structure above is interpreted by the client as meaning the current value of
+the k1 mapping is `g(f({k1, v1}))`, and that of the k2 mappings is
+`f({k2, v1}`.  More generally the current value of a mapping is always equal
```
anthonydahanne: missing closing parenthesis :-P

docs/src/docs/asciidoc/developer/clustered.adoc
```diff
@@ -0,0 +1,174 @@
...
+}
+
+value get(key) {
+  return resolve(server.get(key.hash), key);
+}
+
+value putIfAbsent(key, value) {
+  return resolve(server.getAndAppend(key.hash, operation(putIfAbsent(key, value))), key);
+}
+----
```
lorban: It's an implementation detail, but you don't say anything about it so I wonder when / how often `resolve()` should be called? With the naive approach above, doing millions of `put()` on the same key with no `get()` in-between would make the list grow to an absurd size which we probably want to avoid in practice.

jhouserizer: I had the same comment/question when talking this scheme through with Chris the other day.  He had an answer to it, that was a bit hand-wavey :)  but which sounded ok.   The server could 'notice' that the list related to a hash is getting very long and send a message to a client requesting it to do some resolve()ing for the given hash, which the client would then do.  Presumably it could do this noticing just as a client is doing an append() on that hash - so the server would then know a good candidate of a client to choose for doing the resolve().

ljacomet:
Maybe I am missing something, but the proposed scheme seems to favour writes over reads, since resolve happens on a read - implying a subsequent write - while a write _simply_ appends. And that is counter intuitive for a cache in my mind.

docs/src/docs/asciidoc/developer/clustered.adoc
lorban: As a reminder to myself: depending on how costly it would be to calculate the list size on the server, it could be the return value of the `append()` method then it would be up to the client to have the logic to proactively call `resolve()` or not. Otherwise, the server could hint the client by returning a boolean flag instead.

ljacomet:
Marking blocked as this depends on the design doc to be moved from wiki to asciidoc inside the project.

docs/src/docs/asciidoc/developer/clustered.adoc
henri-tremblay: and too many above

henri-tremblay:
Oh... And the build.gradle

docs/build.gradle
```diff
@@ -15,8 +15,13 @@
  */
 
 plugins {
-  id 'org.asciidoctor.convert' version '1.5.7'
+  id 'org.asciidoctor.convert' version '1.5.3'
```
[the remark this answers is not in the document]
chrisdennis: Hehe... there is a bug in asciidoctorj when used with jruby that prevents it from loading the asciidoctor-diagram gem properly.  This is the only way I managed to get everything working:

https://github.com/asciidoctor/asciidoctor-diagram/issues/159 (closed because it's not an asciidoctor-diagram bug)
https://github.com/asciidoctor/asciidoctorj/issues/562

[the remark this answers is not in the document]
henri-tremblay: Sigh...

PR #927, opened 2016-04-03
Issue #747 Replace limited management dependency in 107 code with direct statistics API usage

chrisdennis:
All tests pass, and I even managed to pull the management project out of the settings.gradle without breaking the build (modulo a change in :clustered:dist) - so management is effectively a leaf project now.

### DIFF
--- a/dist/build.gradle
+++ b/dist/build.gradle
@@ -21,7 +21,6 @@ dependencies {
   compile project(':api')
   compile project(':core')
   compile project(':impl')
-  compile project(':management')
   compile project(':107')
   compile project(':xml')
   compile project(':transactions')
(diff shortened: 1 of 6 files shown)


dist/build.gradle
```diff
@@ -21,7 +21,6 @@ dependencies {
   compile project(':api')
   compile project(':core')
   compile project(':impl')
-  compile project(':management')
```
ljacomet: Pretty sure that should not be pushed :wink:

chrisdennis: Why not?  I thought the plan was to cut management out of the OSS distribution for GA once 107 didn't need it.

ljacomet: Correct and indeed it is no longer needed in the current scope of unclustered GA.

Issue #836, opened 2016-03-10
Deadlock with ordered event listeners


ljacomet:
Needs analysis of the following thread dump:

    [testng] Full thread dump Java HotSpot(TM) 64-Bit Server VM (24.75-b04 mixed mode):
    [testng] 
    [testng] "pool-44-thread-1" prio=10 tid=0x00007f11c8016000 nid=0x2258 waiting on condition [0x00007f12b5438000]
    [testng]    java.lang.Thread.State: WAITING (parking)
    [testng] 	at sun.misc.Unsafe.park(Native Method)
    [testng] 	- parking to wait for  <0x00000007166fb448> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
    [testng] 	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
    [testng] 	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2043)
    [testng] 	at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:1079)
    [testng] 	at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:807)
    [testng] 	at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1068)
    [testng] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
    [testng] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    [testng] 	at java.lang.Thread.run(Thread.java:745)
    [testng] 
    [testng] "Rainfall-core Operations Thread" prio=10 tid=0x00007f11c8014800 nid=0x2257 waiting on condition [0x00007f12b4f33000]
    [testng]    java.lang.Thread.State: WAITING (parking)
    [testng] 	at sun.misc.Unsafe.park(Native Method)
    [testng] 	- parking to wait for  <0x00000007172fd678> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
    [testng] 	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
[...]

PR #928, opened 2016-04-04
#836: fix ordered event handling when OversizeMappingException happens

lorban:
`compute()` (and friends) might be called many times in case a OversizeMappingException happens. In that case, all pre-recorded events should be dropped as the repeated call recreates those events.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/events/InvocationScopedEventSink.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/events/InvocationScopedEventSink.java
@@ -23,6 +23,7 @@ import org.ehcache.core.spi.store.events.StoreEventListener;
 
 import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.Iterator;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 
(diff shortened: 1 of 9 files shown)


ljacomet:
test this please

impl/src/main/java/org/ehcache/impl/internal/events/InvocationScopedEventSink.java
```diff
@@ -150,26 +166,43 @@ private void fireOrdered(Set<StoreEventListener<K, V>> listeners, Deque<Fireable
...
+          fireableEvent.waitTillFired();
+        }
+        fireWaiters(listeners, orderedQueue);
       } else {
         // Waiting for another thread to fire - once that happens, done for this event
-        fireableEvent.waitTillFired();
+        try {
+          fireableEvent.waitTillFired();
+        } catch (RuntimeException e) {
+          throw new RuntimeException("can't wait anymore\nqueue = " + orderedQueue + "\n events = " + events, e);
```
ljacomet: These changes should not make it in - they were for debug purpose.

lorban: That's an overlook from my side, I'll revert that ASAP. Thanks!

Issue #902, opened 2016-03-25
OnHeapStore invalidate() ignores the return value of map.clear()


chrisdennis:
The OnHeapStore invalidate() method ignores the return of map.clear().  I think this might be "mostly harmless", but it should probably be fixed.  Simplest fix would seem to be calling the clear() method directly.

PR #931, opened 2016-04-04
onheapstore invalidate clears the backing map #902

AbfrmBlr:


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -650,7 +650,7 @@ private void invalidate() {
         LOG.warn("Failed to invalidate mapping for key {}", key, cae);
       }
     }
-    map.clear();
+    this.clear();
```
ljacomet: Code style: call `clear()` without the `this` prefix.

Issue #929, opened 2016-04-04
Get rid of PersistentCacheManager.toMaintenance()


ljacomet:
Following discussions with @chrisdennis, @cljohnso and @lorban in the context of upcoming clustered, we realise that the access to explicit maintenance mode may not be required for clustered in the face of the design of `auto-create` mode and as it is not mandatory for cache managers offering disk storage to their caches as they are auto creating from day one, we should drop this method.

However, we still need an option for destroying all persistent content for a cache manager, which could be achieved through `PersistentCacheManager.destroy()` which would internally have a similar behaviour as to what is achieved with going in maintenance mode and then destroy. 

The only scenario broken by this removal would be atomic _destroy then create_ at cache manager level but we are not sure this is a scenario we would want to promote in the clustering context in the face of the CLI as the recommended way to handle creation and destruction in production.

Issue #934, opened 2016-04-06
Lifecycle of UserManagedCache is broken


ljacomet:
There is currently no way of doing:
```java
cache.close();
cache.init();
```

The second init call will fail as the `Store.Provider` will no longer care about the `Store` instance.

ljacomet:
Once fixed, `org.ehcache.core.PersistentUserManagedEhcache#destroy` needs to properly exit maintenance mode.

PR #935, opened 2016-04-06
Issue #929 Get rid of toMaintenance

ljacomet:
* Add `Persistent(CacheManager|UserManagedCache).destroy()`
* Removed `Persistent(CacheManager|UserManagedCache).toMaintenance()`


core/src/main/java/org/ehcache/core/PersistentUserManagedEhcache.java
```diff
@@ -81,30 +80,16 @@ public PersistentUserManagedEhcache(CacheConfiguration<K, V> configuration, Stor
...
-      };
       st.succeeded();
-      return maintainable;
-    } catch (RuntimeException e) {
-      throw st.failed(e);
+    } catch (Throwable t) {
+      throw st.failed(t);
     }
+    destroyInternal();
+    // TODO Exit maintenance mode once #934 is solved
```
lorban: Another todo?

ljacomet: Yeah, let me get rid of it and instead add comments into #934

ljacomet:
Dropped TODO and replaced by comment in #934

lorban:
LGTM

ljacomet:
@chrisdennis @cljohnso @AbfrmBlr Anything to add?

chrisdennis:
Shouldn't Maintainable be moved out of the API... is it being exposed anywhere now?

ljacomet:
Not only it is no longer exposed, it is not even used at all. So I deleted it ... PR updated

PR #938, opened 2016-04-06
Issue 742: NOT A REAL PR

lorban:
but a request for comments.

### DIFF
--- a/api/src/main/java/org/ehcache/spi/copy/Copier.java
+++ b/api/src/main/java/org/ehcache/spi/copy/Copier.java
@@ -18,10 +18,14 @@ package org.ehcache.spi.copy;
 
 /**
  * Interface defining the contract used to make a copy of types.
- *
+ * <p>
+ *   The copied object's class must be preserved in such a way that copying for read that object will return
+ *   an object of the exact same class as before copy for write, i.e.: the following contract must always be true:
+ *   <p>
+ *   <code>object.getClass().equals( myCopier.copyForRead(myCopier.copyForWrite(object)).getClass() )</code>
+ *   </p>
+ * </p>
  * @param <T> the type of the instance to copy
- *
- * @author Albin Suresh
  */
 public interface Copier<T> {
 
--- a/docs/src/docs/asciidoc/user/serializers-copiers.adoc
+++ b/docs/src/docs/asciidoc/user/serializers-copiers.adoc
@@ -29,19 +29,7 @@ cannot store by reference is going to use a pair of Serializers: one for the key
 Serializers are scoped at the cache level: a cache can have a serializer for its keys and another one for its values.
 All stores of a cache will be using and sharing the same key and value serializer pair.
 
-=== Copiers
-
-As the on-heap store is capable of storing plain java objects as such,
-it is not necessary to use a costly serialization and deserialization mechanism to copy keys and values when a
-on-heap store is configured to store by value. For instance, some form of "copy constructor" mechanism is a lot
-more performant but requires custom code to be able to copy objects.
-
-Copiers are specific to the on-heap store: no other store makes use of copiers.
-
-By default, the on-heap mappings are stored by reference. The way to store them by value is to configure some copier(s)
-on it, either key, value or both.
-
-=== How is a serializer configured?
+==== How is a serializer configured?
 
 There are two places where serializers can be configured:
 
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/ehcache/spi/copy/Copier.java
```diff
@@ -18,7 +18,13 @@
...
+ * <p>
+ *   The copied object's class must be preserved in such a way that copying for read that object will return
+ *   an object of the exact same class as before copy for write, i.e.: the following contract must always be true:
+ *   <p>
+ *   <code>object.getClass().equals( myCopier.copyForRead(myCopier.copyForWrite(object)).getClass() )</code>
+ *   </p>
+ * </p>
  * @param <T> the type of the instance to copy
  *
  * @author Albin Suresh
```
ljacomet: Let's drop all author tags as soon as a class is touched

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+endif::sourcedir[]
+
+= Serializers and Copiers
+
+
+ifdef::notBuildingForSite[]
+include::menu.adoc[]
+endif::notBuildingForSite[]
+
+== Intro
```
ljacomet: I would use the full word here

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+The on-heap store is capable of storing cached objects either by reference (where the given key and value references
+are stored) or by value (where a copy of the given key and value are made and those copies are then stored) while all
+other stores are only capable of storing by value.
+
+Serializers and Copiers are the components that perform the required copies.
```
ljacomet: .. the required operations.

Because copy has meaning in this context.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+=== Serializers
+
+All stores but the on-heap one need some form of serialization/deserialization of
+objects to be able to store and retrieve them as they cannot internally store plain java objects but only bunch-of-bytes representations
+of them. Serializers are the Ehcache mechanism that revolves this logic: every cache that has at least one store that
+cannot store by reference is going to use a pair of Serializers: one for the key and another one for the value.
+
+Serializers are scoped at the cache level: a cache can have a serializer for its keys and another one for its values.
+All stores of a cache using serializers will share a common pair of key/value ones.
```
ljacomet: All stores of a cache will be using and sharing the same key and value serializer pair.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+By default, cache managers are pre-configured with specially optimized serializers that can handle the following types, in the following order:
+
+- `java.io.Serializable`
+- `java.lang.Long`
+- `java.lang.Integer`
+- `java.lang.Float`
+- `java.lang.Double`
+- `java.lang.Character`
+- `java.lang.String`
```
GaryWKeim: StringBuilder, StringBuffer?

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+=== Copiers
+
+As the on-heap store is capable of storing plain java objects as such,
+it is not necessary to use a costly serialization and deserialization mechanism to copy keys and values when a
+on-heap store is configured to store by value. For instance, some form of "copy constructor" mechanism is a lot
+more performant but requires custom code to be able to copy objects.
+
+Copiers are specific to the on-heap store: no other store makes use of copiers.
+
+By default, the on-heap store stores by reference. The way to make it to store by value is to configure some copier(s)
```
ljacomet: > ... store stores ... 

is not very nice.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+on it, either key, value or both.
+
+=== How is a serializer configured?
+
+There are two places where serializers can be configured:
+
+- at the cache level where one can use `CacheConfigurationBuilder.withKeySerializer()` and `CacheConfigurationBuilder.withValueSerializer()`.
+Both instances and class names can be configured.
```
ljacomet: I would be explicit in listing the method signatures.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+- at the cache manager level where one can use `CacheManagerBuilder.withSerializer(Class clazz, Class serializer)`
+
+If a serializer is configured directly at the cache level, it will be used.
```
ljacomet: ... it will be used, ignoring any cache manager level configuration.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+```
+CacheManagerBuilder.newCacheManagerBuilder().withSerializer(Employee.class, EmployeeSerializer.class).withSerializer(Person.class, PersonSerializer.class)
+```
+then configuring a `Cache<Long, Employee>` would make it use the `EmployeeSerializer` while a `Cache<Long, Customer>` would make it use the `PersonSerializer`.
+
+Each cache for which Ehcache instantiated a serializer (i.e.: the serializer was configured by its class name) will have its own instance.
+
+=== How is a copier configured?
+
+There is only one place where copiers can be configured: at the cache level where one can use `CacheConfigurationBuilder.withKeyCopier()`
```
ljacomet: Not true, copiers can be configured at the CM level as well.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
ljacomet: These are definitely not immutable types.

And I hope no one ever uses some as keys :wink:

impl/src/test/java/org/ehcache/docs/plugs/CharSequenceSerializer.java
```diff
@@ -39,18 +43,53 @@ public CharSequenceSerializer(ClassLoader classLoader) {
   @Override
   public ByteBuffer serialize(CharSequence object) {
     LOG.info("serializing {}", object);
-    ByteBuffer byteBuffer = ByteBuffer.allocate(object.length());
+
+    byte type;
+    if (object.getClass().equals(String.class)) {
```
ljacomet: Please, can we _not_ demonstrate this? I would consider is fishy code if found in my code base, so why advertise it as an example.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+For instance, let's say you have a `Person` interface and two subclasses: `Employee` and `Customer`. If you configure your cache manager as follows:
+```
+CacheManagerBuilder.newCacheManagerBuilder().withSerializer(Employee.class, EmployeeSerializer.class).withSerializer(Person.class, PersonSerializer.class)
+```
+then configuring a `Cache<Long, Employee>` would make it use the `EmployeeSerializer` while a `Cache<Long, Customer>` would make it use the `PersonSerializer`.
+
+Each cache for which Ehcache instantiated a serializer (i.e.: the serializer was configured by its class name) will have its own instance.
+
+=== How is a copier configured?
```
ljacomet: After going through the whole document, I believe outside of an introduction, topics on copiers should be grouped under the `Copiers` section, probably one level lower.

Same remark for serializers focused sections.

Issue #923, opened 2016-04-03
Transactional Caches currently ignore user defined vetoes


chrisdennis:
The transaction code currently ignore user defined eviction veto functions.  I think this might be as simple as the used veto function being a logical or of the users veto with the xa veto.

ljacomet:
Created some high level [design docs](https://github.com/ehcache/ehcache3/wiki/module.xa) around XA caches.
I believe there is no reason to not support user specified vetoes.

PR #943, opened 2016-04-06
Issue #923 Support eviction veto in XA caches

ljacomet:


transactions/src/main/java/org/ehcache/transactions/xa/internal/XAStore.java
```diff
@@ -979,4 +974,22 @@ public void stop() {
...
+
+    private final EvictionVeto<? super K, ? super V> wrappedVeto;
+
+    private XAEvictionVeto(EvictionVeto<? super K, ? super V> wrappedVeto) {
+      this.wrappedVeto = wrappedVeto;
+    }
+
+    @Override
+    public boolean vetoes(K key, SoftLock<V> softLock) {
+      if (isInDoubt(softLock)) {
```
chrisdennis:
```
return isInDoubt(softLock) || wrappedVeto.vetoes(key, softLock.getOldValue());
```

ljacomet: My IDE agrees with you ... still find the expanded version more readable.
Others to chime in?

GaryWKeim: I prefer conciseness.

ljacomet: Changing.

transactions/src/test/java/org/ehcache/transactions/xa/internal/XAStoreTest.java
```diff
@@ -1488,6 +1490,52 @@ public void testBulkComputeIfAbsent() throws Exception {
   }
 
   @Test
+  public void testCustomEvictionVeto() throws Exception {
```
chrisdennis: Are we ruling this test sufficient?  It obviously verifies that the custom veto is called... it doesn't actually verify that there is any effect on the eviction process however (of course whether that could be verified is another question).

ljacomet: The only thing that needs verification at this level is that the provided veto is called. Testing the proper handling of veto exists elsewhere.

PR #944, opened 2016-04-06
Issue 837

AbfrmBlr:


impl/src/main/java/org/ehcache/impl/internal/spi/copy/DefaultCopyProvider.java
```diff
@@ -54,6 +54,13 @@ public DefaultCopyProvider(DefaultCopyProviderConfiguration configuration) {
     return createCopier(DefaultCopierConfiguration.Type.VALUE, clazz, serializer, configs);
   }
 
+  @Override
+  public void releaseCopier(Copier<?> copier) throws Exception {
+    if (!(copier instanceof IdentityCopier) & !(copier instanceof SerializingCopier)) {
```
AbfrmBlr: Because these are the only ones that gets provided as default and its does not get instantiated by the instance provider

api/src/main/java/org/ehcache/spi/copy/CopyProvider.java
```diff
@@ -46,4 +48,14 @@
    * @return a {@link Copier} instance
    */
   <T> Copier<T> createValueCopier(Class<T> clazz, Serializer<T> serializer, ServiceConfiguration<?>... configs);
+
+  /**
+   * Releases the provided {@link Copier} instance
+   * If the copier instance is provided by the user, {@link Closeable#close()}
```
ljacomet: Use fqcn instead of adding imports for Javadoc

impl/src/main/java/org/ehcache/impl/internal/classes/ClassInstanceProvider.java
```diff
@@ -138,9 +146,10 @@ protected void releaseInstance(T instance) throws IOException {
       throw new IllegalArgumentException("Given instance of " + instance.getClass().getName() + " is not managed by this provider");
     }
 
-    if(instance instanceof Closeable) {
+    if(instantiated.contains(instance) && instance instanceof Closeable) {
```
ljacomet: This can be simplified by doing `if (instantiated.remove(instance)) && instance instance Closeable)` and thus dropping the remove after the block.

impl/src/main/java/org/ehcache/impl/internal/classes/ClassInstanceProvider.java
```diff
@@ -49,6 +51,7 @@
    * Instances provided by this provider vs their counts.
    */
   protected final ConcurrentWeakIdentityHashMap<T, AtomicInteger> providedVsCount = new ConcurrentWeakIdentityHashMap<T, AtomicInteger>();
+  protected final Set<T> instantiated = new HashSet<T>();
```
ljacomet: You probably want to use a `Set` with an reference identity contract and not an object identity one. See our `ConcurrentWeakIdentityHashMap` for example, which can easily be wrapped to act as a `Set` like we do in some `Store.Provider`.

impl/src/main/java/org/ehcache/impl/internal/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -66,6 +68,7 @@
   private final PersistentProvider persistentProvider;
 
   protected final ConcurrentWeakIdentityHashMap<Serializer<?>, AtomicInteger> providedVsCount = new ConcurrentWeakIdentityHashMap<Serializer<?>, AtomicInteger>();
+  protected final Set<Serializer<?>> instantiated = new HashSet<Serializer<?>>();
```
ljacomet: Same remark as in the `ClassInstanceProvider`

impl/src/main/java/org/ehcache/impl/internal/classes/ClassInstanceProvider.java
ljacomet: And one more advantage of `ConcurrentWeakIdentityHashMap` is that you are not holding a strong reference to the instance.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1567,21 +1567,34 @@ public int rank(final Set<ResourceType<?>> resourceTypes, final Collection<Servi
...
       OnHeapStore<K, V> onHeapStore = new OnHeapStore<K, V>(storeConfig, timeSource, keyCopier, valueCopier, sizeOfEngine, eventDispatcher);
-      createdStores.add(onHeapStore);
+      createdStores.put(onHeapStore, copiers);
       return onHeapStore;
     }
 
     @Override
     public void releaseStore(Store<?, ?> resource) {
-      if (!createdStores.remove(resource)) {
+      if (!createdStores.containsKey(resource)) {
```
ljacomet: I know this is lifecycle code and not performance sensitive, by why no do the remove to a local var and check it for `null` to throw?

impl/src/test/java/org/ehcache/impl/internal/spi/copy/DefaultCopyProviderTest.java
```diff
@@ -84,6 +90,39 @@ public void testCreateValueCopierWithSerializer() {
...
+    public T copy(final T obj) {
+      return obj;
+    }
+
+    @Override
+    public void close() throws IOException {
+      invoked = true;
+    }
+
+    boolean getInvoked() {
```
ljacomet: `isInvoked`

impl/src/test/java/org/ehcache/impl/internal/spi/serialization/DefaultSerializationProviderTest.java
```diff
@@ -142,13 +146,25 @@ public void testRemembersCreationConfigurationAfterStopStart() throws Unsupporte
...
     provider.releaseSerializer(serializer);
-    verify(serializer).close();
+    verify(serializer, times(0)).close();
+  }
+
+  @Test
+  public void testReleaseSerializerWithInstantiatedCloseableSerializerDoesClose() throws Exception {
+    DefaultSerializerConfiguration config = new DefaultSerializerConfiguration(CloseableSerializer.class, DefaultSerializerConfiguration.Type.KEY);
+    DefaultSerializationProvider provider = new DefaultSerializationProvider(null);
+//    CompactJavaSerializer<?> serializer = mock(CompactJavaSerializer.class);
```
ljacomet: Please remove dead code

impl/src/main/java/org/ehcache/impl/internal/spi/copy/DefaultCopyProvider.java
ljacomet: Not sure I follow 100% your thought path. Can you check what happens if you _explicitly_ configure an `IdentityCopier` or a `SerializingCopier`? Cause I am pretty sure you leak references by leaving them in the `ClassInstanceProvider` instantiated set.

AbfrmBlr: yes, so I have to maintain a set of provided default instances as well in `DefaultCopyProvider`

impl/src/main/java/org/ehcache/impl/internal/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -120,7 +128,7 @@ public void releaseSerializer(final Serializer<?> serializer) throws IOException
       throw new IllegalArgumentException("Given serializer:" + serializer.getClass().getName() + " is not managed by this provider");
     }
 
-    if(serializer instanceof Closeable) {
+    if(instantiated.contains(serializer) && serializer instanceof Closeable) {
```
ljacomet: I believe this should be `instantiated.remove(serialiser)`

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1567,21 +1567,34 @@ public int rank(final Set<ResourceType<?>> resourceTypes, final Collection<Servi
...
-      createdStores.add(onHeapStore);
+      createdStores.put(onHeapStore, copiers);
       return onHeapStore;
     }
 
     @Override
     public void releaseStore(Store<?, ?> resource) {
-      if (!createdStores.remove(resource)) {
+      List<Copier> copiers = null;
+      if ((copiers = createdStores.remove(resource)) == null) {
```
ljacomet: I would do the assignment at the variable declaration instead of inline in the if condition.

impl/src/test/java/org/ehcache/impl/internal/spi/copy/DefaultCopyProviderTest.java
```diff
@@ -84,6 +90,39 @@ public void testCreateValueCopierWithSerializer() {
     assertThat(copyProvider.createValueCopier(Long.class, mock(Serializer.class), config), instanceOf(SerializingCopier.class));
   }
 
+  @Test
```
ljacomet: This misses a test that close in well invoked on an instantiated copier type that implements `Closeable`

AbfrmBlr: There is a generic test case for that in `ClassInstanceProviderTest#testReleaseCloseableInstance` which tests `Closeable#close` on instantiated ones

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
AbfrmBlr: Will it be any different ?

ljacomet: Code readability matters. You declare the variable just the line before, why not assign it instead of having a more complicated construct in the if

Issue #357, opened 2015-04-01
ConfigurationParser rules about ResourcePool inheritance from templates need defining/fixing


chrisdennis:
The ConfigurationParser logic for collecting the resources used by a cache from it's XML config and template is a little weird.  Currently it collates all definitions from both together.  It seems we would want some kind of system where the cache config would take precedent over the template, but at what granularity I'm not yet sure (e.g. what happens if the cache config specifies a heap resource, and the template specifies a disk resource).  We also need to have the ConfigurationParser enforce that at least one of the template or cache configs contain resource definition - since this can't be done in XML without invoking the glory of Schematron - which is apparently frowned upon in this establishment.

alexsnaps:
Relate to #33

ljacomet:
Current status with regards to _defaults_:
* In a `CacheConfigurationBuilder`, we define a default resource pool containing `HEAP` resource, sized in entries at `Long.MAX_VALUE`
* In XML, if there is nothing defined, there are really no resources defined as we _always_ construct a `ResourcePools` even if fully empty.

I propose the following changes:
* `ResourcePoolsBuilder.build()` will throw an `IllegalStateException` if no pools have been defined
* XML will only configure `ResourcePools` if there is a definition, both for templates and cache declarations
* This means that there will now be a common default, the one from `CacheConfigurationBuilder`.

Comments on template inheritance at a later stage.

ljacomet:
After further discussion with @lorban and going back to #421, the changes above to align XML and builder based config will be done.
However, the default heap pool will be discarded for Ehcache, making sizing mandatory.
For JSR-107 when caches are created:
* From `javax.cache.configuration.CompleteConfiguration`, a default heap of `Long.MAX_VALUE` size will be used,
* From  an Ehcache config - XML, template override or programmatic wrapped - no defaults, sizing is mandatory.

ljacomet:
PR #952 opened for the changes mentioned above.

As for the cache vs. template behaviour, there is no plan to change current behaviour which is:
* When a cache definition is based on a template, any resource declaration will result in ignoring the template level definition.

This means that there is no notion of merging individual resource definitions.

PR #952, opened 2016-04-07
Issue #357 Consistency around ResourcePools

ljacomet:
* Align XML behaviour on programmatic one
* Remove any default but for JSR-107 caches

### DIFF
--- a/impl/src/main/java/org/ehcache/config/builders/CacheConfigurationBuilder.java
+++ b/impl/src/main/java/org/ehcache/config/builders/CacheConfigurationBuilder.java
@@ -19,7 +19,6 @@ package org.ehcache.config.builders;
 import org.ehcache.config.CacheConfiguration;
 import org.ehcache.config.EvictionVeto;
 import org.ehcache.config.ResourcePools;
-import org.ehcache.config.units.EntryUnit;
 import org.ehcache.config.units.MemoryUnit;
 import org.ehcache.core.config.BaseCacheConfiguration;
 import org.ehcache.core.config.store.StoreEventSourceConfiguration;
@@ -44,7 +43,6 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 
-import static org.ehcache.config.builders.ResourcePoolsBuilder.newResourcePoolsBuilder;
 import static org.ehcache.impl.config.store.heap.DefaultSizeOfEngineConfiguration.DEFAULT_MAX_OBJECT_SIZE;
 import static org.ehcache.impl.config.store.heap.DefaultSizeOfEngineConfiguration.DEFAULT_OBJECT_GRAPH_SIZE;
 import static org.ehcache.impl.config.store.heap.DefaultSizeOfEngineConfiguration.DEFAULT_UNIT;
@@ -63,27 +61,44 @@ public class CacheConfigurationBuilder<K, V> implements Builder<CacheConfigurati
   private Expiry<? super K, ? super V> expiry;
   private ClassLoader classLoader = null;
   private EvictionVeto<? super K, ? super V> evictionVeto;
-  private ResourcePools resourcePools = newResourcePoolsBuilder().heap(Long.MAX_VALUE, EntryUnit.ENTRIES).build();
+  private ResourcePools resourcePools;
   private Class<? super K> keyType;
   private Class<? super V> valueType;
 
   /**
    * Creates a new instance ready to produce a {@link CacheConfiguration} with key type {@code <K>} and with value type
-   * {@code <V>}.
+   * {@code <V>} and which will use the {@link ResourcePools configured resources}.
    *
    * @param keyType the key type
    * @param valueType the value type
+   * @param resourcePools the resources to use
    * @param <K> the key type
    * @param <V> the value type
    * @return a {@code CacheConfigurationBuilder}
    */
-  public static <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilder(Class<K> keyType, Class<V> valueType) {
-    return new CacheConfigurationBuilder<K, V>(keyType, valueType);
+  public static <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilder(Class<K> keyType, Class<V> valueType, ResourcePools resourcePools) {
+    return new CacheConfigurationBuilder<K, V>(keyType, valueType, resourcePools);
   }
 
-  private CacheConfigurationBuilder(Class<K> keyType, Class<V> valueType) {
+  /**
+   * Creates a new instance ready to produce a {@link CacheConfiguration} with key type {@code <K>} and with value type
+   * {@code <V>} and which will use the {@link ResourcePools configured resources}, passed as a {@link ResourcePoolsBuilder}.
+   *
+   * @param keyType the key type
+   * @param valueType the value type
+   * @param resourcePoolsBuilder the resources to use, as a builder
+   * @param <K> the key type
+   * @param <V> the value type
+   * @return a {@code CacheConfigurationBuilder}
+   */
+  public static <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilder(Class<K> keyType, Class<V> valueType, Builder<? extends ResourcePools> resourcePoolsBuilder) {
+    return new CacheConfigurationBuilder<K, V>(keyType, valueType, resourcePoolsBuilder.build());
+  }
+
+  private CacheConfigurationBuilder(Class<K> keyType, Class<V> valueType, ResourcePools resourcePools) {
     this.keyType = keyType;
     this.valueType = valueType;
+    this.resourcePools = resourcePools;
   }
 
   private CacheConfigurationBuilder(CacheConfigurationBuilder<? super K, ? super V> other) {
--- a/impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
+++ b/impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
@@ -181,6 +181,9 @@ public class ResourcePoolsBuilder implements Builder<ResourcePools> {
    */
   @Override
   public ResourcePools build() {
+    if (resourcePools.isEmpty()) {
+      throw new IllegalStateException("No resource pools defined");
+    }
     return new ResourcePoolsImpl(resourcePools);
   }
 
--- a/impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
+++ b/impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
@@ -18,6 +18,7 @@ package org.ehcache.config.builders;
 
 import org.ehcache.config.ResourcePool;
 import org.ehcache.config.SizedResourcePool;
+import org.ehcache.config.units.EntryUnit;
 import org.ehcache.core.config.SizedResourcePoolImpl;
 import org.ehcache.config.ResourcePools;
 import org.ehcache.core.config.ResourcePoolsImpl;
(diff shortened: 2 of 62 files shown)


xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -339,7 +339,7 @@ public ListenersConfig listenersConfig() {
           }
 
           @Override
-          public Iterable<ResourcePool> resourcePools() {
+          public Collection<ResourcePool> resourcePools() {
```
cljohnso: I have no problem with it but what prompted the change from `Iterable` to `Collection`?

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -364,18 +366,35 @@ public URL getURL() {
   /**
    * Creates a new {@link CacheConfigurationBuilder} seeded with the cache-template configuration
    * by the given {@code name} in the XML configuration parsed using {@link #parseConfiguration()}
+   * <P>
+   *   Note that this version does not specify resources, which are mandatory to create a
+   *   {@link CacheConfigurationBuilder}. So in case the template does not define resources, this will throw.
```
cljohnso: "So in case the template" -> "So if the template" would more correct wording.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
ljacomet: Being able to query its size, to know when it is empty.

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -364,18 +366,35 @@ public URL getURL() {
...
+   * </P>
    *
    * @param name the unique name identifying the cache-template element in the XML
+   * @param keyType the type of keys for the {@link CacheConfigurationBuilder} to use, would need to
```
cljohnso: "would need to match" -> "must match"

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -364,18 +366,35 @@ public URL getURL() {
...
+   *                match the {@code key-type} declared in the template if declared in XML
+   * @param valueType the type of values for the {@link CacheConfigurationBuilder} to use, would need to
```
cljohnso: "would need to match" -> "must match"

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -364,18 +366,35 @@ public URL getURL() {
   /**
    * Creates a new {@link CacheConfigurationBuilder} seeded with the cache-template configuration
    * by the given {@code name} in the XML configuration parsed using {@link #parseConfiguration()}
```
cljohnso: The lead sentence should end with a period (http://www.oracle.com/technetwork/articles/java/index-137868.html#firstsentence).

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -402,8 +422,48 @@ public URL getURL() {
...
                                                                                          final Class<K> keyType,
-                                                                                         final Class<V> valueType)
+                                                                                         final Class<V> valueType,
+                                                                                         final ResourcePools resourcePools)
       throws InstantiationException, IllegalAccessException, ClassNotFoundException {
+    return internalCacheConfigurationBuilderFromTemplate(name, keyType, valueType, resourcePools);
+  }
+
+  /**
+   * Creates a new {@link CacheConfigurationBuilder} seeded with the cache-template configuration
```
cljohnso: The lead sentence should end with a period (http://www.oracle.com/technetwork/articles/java/index-137868.html#firstsentence).

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -402,8 +422,48 @@ public URL getURL() {
...
+   * by the given {@code name} in the XML configuration parsed using {@link #parseConfiguration()}
+   *
+   * @param name the unique name identifying the cache-template element in the XML
+   * @param keyType the type of keys for the {@link CacheConfigurationBuilder} to use, would need to
```
cljohnso: "would need to match" -> "must match"

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -402,8 +422,48 @@ public URL getURL() {
...
+   *                match the {@code key-type} declared in the template if declared in XML
+   * @param valueType the type of values for the {@link CacheConfigurationBuilder} to use, would need to
```
cljohnso: "would need to match" -> "must match"

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -419,7 +479,19 @@ public URL getURL() {
       throw new IllegalArgumentException("CacheTemplate '" + name + "' declares value type of " + cacheTemplate.valueType());
     }
 
-    CacheConfigurationBuilder<K, V> builder = CacheConfigurationBuilder.newCacheConfigurationBuilder(keyType, valueType);
+    if (resourcePools == null && cacheTemplate.resourcePools().isEmpty()) {
+      throw new IllegalStateException("Template defines not resources, and none were provided");
```
cljohnso: "defines **no** resources"

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -419,7 +479,19 @@ public URL getURL() {
       throw new IllegalArgumentException("CacheTemplate '" + name + "' declares value type of " + cacheTemplate.valueType());
     }
 
-    CacheConfigurationBuilder<K, V> builder = CacheConfigurationBuilder.newCacheConfigurationBuilder(keyType, valueType);
+    if (resourcePools == null && cacheTemplate.resourcePools().isEmpty()) {
```
cljohnso: Strictly speaking, one can have a `ResourcePools` instance that is empty and, to check for that condition, you'd need `resourcePools.getResourceTypeSet().isEmpty()`.

ljacomet:
Pushed an update addressing remarks.

impl/src/main/java/org/ehcache/config/builders/CacheConfigurationBuilder.java
```diff
@@ -63,27 +61,44 @@
...
+   * {@code <V>} and which will use the {@link ResourcePools configured resources}, passed as a {@link ResourcePoolsBuilder}.
+   *
+   * @param keyType the key type
+   * @param valueType the value type
+   * @param resourcePoolsBuilder the resources to use, as a builder
+   * @param <K> the key type
+   * @param <V> the value type
+   * @return a {@code CacheConfigurationBuilder}
+   */
+  public static <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilder(Class<K> keyType, Class<V> valueType, Builder<ResourcePools> resourcePoolsBuilder) {
```
chrisdennis: I think this should be `Builder<? extends ResourcePools>`

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
```diff
@@ -181,6 +194,9 @@ public ResourcePoolsBuilder disk(long size, MemoryUnit unit, boolean persistent)
    */
   @Override
   public ResourcePools build() {
+    if (resourcePools.isEmpty()) {
+      throw new IllegalStateException("No resource pools defined");
+    }
```
chrisdennis: Should this validation occur inside the constructor?

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
chrisdennis: It does, doesn't it?  It's just a long line

impl/src/main/java/org/ehcache/config/builders/CacheConfigurationBuilder.java
cljohnso: Isn't that overkill?  `ResourcePools` is an interface and the builder is only prepared to deal with what is specified by `ResourcePools` and not some object that carries more "stuff".

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -402,8 +422,51 @@ public URL getURL() {
...
+   * @throws IllegalArgumentException if {@code keyType} or {@code valueType} don't match the declared type(s) of the template
+   * @throws ClassNotFoundException if a {@link java.lang.Class} declared in the XML couldn't be found
+   * @throws InstantiationException if a user provided {@link java.lang.Class} couldn't get instantiated
+   * @throws IllegalAccessException if a method (including constructor) couldn't be invoked on a user provided type
+   */
+  @SuppressWarnings("unchecked")
+  public <K, V> CacheConfigurationBuilder<K, V> newCacheConfigurationBuilderFromTemplate(final String name,
+                                                                                         final Class<K> keyType,
+                                                                                         final Class<V> valueType,
+                                                                                         final Builder<ResourcePools> resourcePoolsBuilder)
```
chrisdennis: `? extends` again

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
ljacomet: It does now, but the update was on next line so did not consider this comment as outdated.

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
ljacomet: Inside which constructor?

impl/src/main/java/org/ehcache/config/builders/CacheConfigurationBuilder.java
cljohnso: I should add that the downstream constructs are also only prepared to deal with `ResourcePools` ...

chrisdennis: I can pass an arbitrary subtype to the non-builder consuming method, no reason why I shouldn't be able to here.

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
chrisdennis: Inside the ResourcePoolsImpl one that is called further down.

impl/src/main/java/org/ehcache/config/builders/CacheConfigurationBuilder.java
cljohnso: You can already pass anything that implements `ResourcePools` -- what does the `extends` buy you in this case?

chrisdennis: If the Builder is typed as Builder<ArbitraryResourcePoolsSubtype> then this will not compile... even though the built result could be passed to the other method.

ljacomet: Fixing

impl/src/main/java/org/ehcache/config/builders/ResourcePoolsBuilder.java
ljacomet: Right, fixing

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
ljacomet: Fixing

PR #955, opened 2016-04-08
Integration tests now use Galvan to run against the actual built kit

chrisdennis:
The thing I have most concern with here is the effect this has on test runtimes.


chrisdennis:
I think this is failing because the cloudless machines don't have a valid toolchains.xml.  I'm going to make some changes so that this fails in a clearer way, and re-push.  Not sure what to do about the actual problem however.

AbfrmBlr:
test this please

ljacomet:
test this please

ljacomet:
Now that this PR builds, what's next before merging @chrisdennis ?

chrisdennis:
I have one more thing to play with to see if I can reduce test runtimes still further.  I will run that experiment today... other than that nothing.

chrisdennis:
@ljacomet I think this is as good as we can get runtime wise for the moment.  The remaining test time is dominated by the election period when the servers first start up.  We can look in to filing an issue about this in Terracotta-OSS/galvan but for now this can be considered eligible for merging.

ljacomet:
Once rebased, good to go

PR #971, opened 2016-04-14
Couple doc fixes

anthonydahanne:


anthonydahanne:
@ljacomet I've removed the commit about tiering

Issue #976, opened 2016-04-15
Align concurrency related fields in XML and code


ljacomet:
* write-behind concurrency stays as is
* event dispatcher parallelism becomes dispatcher-concurrency
* disk store writer-threads becomes writer-concurrency

PR #977, opened 2016-04-15
Issue #976 Rename attributes in XSD related to concurrency

ljacomet:
* Adapt SPI method names matching these XML config elements

### DIFF
--- a/core/src/main/java/org/ehcache/core/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/core/EhcacheManager.java
@@ -485,17 +485,17 @@ public class EhcacheManager implements PersistentCacheManager, InternalCacheMana
       }
     }
 
-    int eventParallelism;
+    int dispatcherConcurrency;
     StoreEventSourceConfiguration eventSourceConfiguration = ServiceLocator.findSingletonAmongst(StoreEventSourceConfiguration.class, config
         .getServiceConfigurations()
         .toArray());
     if (eventSourceConfiguration != null) {
-      eventParallelism = eventSourceConfiguration.getOrderedEventParallelism();
+      dispatcherConcurrency = eventSourceConfiguration.getDispatcherConcurrency();
     } else {
-      eventParallelism = StoreEventSourceConfiguration.DEFAULT_EVENT_PARALLELISM;
+      dispatcherConcurrency = StoreEventSourceConfiguration.DEFAULT_DISPATCHER_CONCURRENCY;
     }
 
-    Store.Configuration<K, V> storeConfiguration = new StoreConfigurationImpl<K, V>(config, eventParallelism, keySerializer, valueSerializer);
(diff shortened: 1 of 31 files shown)


core/src/main/java/org/ehcache/core/internal/store/StoreConfigurationImpl.java
```diff
@@ -37,21 +37,21 @@
...
   private final Serializer<V> valueSerializer;
-  private final int orderedEventParallelism;
+  private final int dispatcherConcurrency;
 
   /**
    * Creates a new {@code StoreConfigurationImpl} based on the provided parameters.
    *
    * @param cacheConfig the cache configuration
-   * @param orderedEventParallelism the level of parallelism for ordered events
+   * @param dispatcherConcurrency the level of parallelism for ordered events
```
AbfrmBlr: Should we use concurrency instead of parallelism in doc also. just for being consistent

core/src/main/java/org/ehcache/core/internal/store/StoreConfigurationImpl.java
```diff
@@ -63,14 +63,14 @@ public StoreConfigurationImpl(CacheConfiguration<K, V> cacheConfig, int orderedE
    * @param classLoader the class loader
    * @param expiry the expiry policy
    * @param resourcePools the resource pools
-   * @param orderedEventParallelism the level or parallelism for ordered events
+   * @param dispatcherConcurrency the level or parallelism for ordered events
```
AbfrmBlr: same as above

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -738,15 +738,15 @@ public void fatalError(SAXParseException exception) throws SAXException {
     final Iterable<Listener> listeners;
 
     private XmlListenersConfig(final ListenersType type, final ListenersType... others) {
-      this.parallelismLevel = type.getParallelismLevel().intValue();
-      String threadPool = type.getThreadPool();
+      this.parallelismLevel = type.getDispatcherConcurrency().intValue();
```
gibsong: Please change variable name parallelismLevel to dispatcherConcurrency.

core/src/main/java/org/ehcache/core/internal/store/StoreConfigurationImpl.java
gibsong: Yes, I would say for consistency that the doc should match.  So replace parallelism with concurrency.

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -490,9 +490,9 @@ public void close() throws Exception {
         .getServiceConfigurations()
         .toArray());
     if (eventSourceConfiguration != null) {
-      eventParallelism = eventSourceConfiguration.getOrderedEventParallelism();
+      eventParallelism = eventSourceConfiguration.getDispatcherConcurrency();
```
gibsong: There are 4 occurrences of the variable name eventParallelism that should be changed.

PR #981, opened 2016-04-15
Issue #975 Docs improvements

aurbroszniowski:


docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -15,7 +15,7 @@ endif::notBuildingForSite[]
 
 While Ehcache is a Java cache, it cannot always store its mapping as Java objects.
 
-The on-heap store is capable of storing cached objects either by reference (where the given key and value references
+The <<caching-concepts.adoc#storage-tiers,on-heap store>> is capable of storing cached objects either by reference (where the given key and value references
```
ljacomet: That does not work, as the end result are html files. See other links between sections across pages for the proper syntax.

docs/src/docs/asciidoc/user/caching-concepts.adoc
```diff
@@ -40,7 +40,7 @@ and technical decision based upon the requirements and assumptions of your appli
 
 
 === Storage Tiers
-You can configure Ehcache to use various data storage areas.  When a cache is configured to use more than one data
+[[storage-tiers]]You can configure Ehcache to use various data storage areas.  When a cache is configured to use more than one data
```
ljacomet: We normally do not define explicit section ids but use the generated ones.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -148,7 +151,7 @@ Registering a serializer that lacks such constructor at the cache manager level
 
 === Copiers
 
-As the on-heap store is capable of storing plain java objects as such,
+As the <<caching-concepts.adoc#storage-tiers,on-heap store>> is capable of storing plain java objects as such,
```
ljacomet: same link remark

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -94,9 +94,12 @@ So if you're planning to write your own implementation you have to keep in mind
 object.getClass().equals( mySerializer.read(mySerializer.serialize(object)).getClass() )
 ```
 
-This is especially important when you are planning to write a serializer for an abstract type, like a serializer of type
-`com.pany.MyInterface` would need to deserialize a `com.pany.MyClassImplementingMyInterface` when the serialized object was of class `com.pany.MyClassImplementingMyInterface`
-and at the same time return an object of type `com.pany.AnotherClassImplementingMyInterface` when the serialized object was of class `com.pany.AnotherClassImplementingMyInterface`.
+This is especially important when you are planning to write a serializer for an abstract type,
+
```
ljacomet: That seems a superfluous empty line

docs/src/docs/asciidoc/user/serializers-copiers.adoc
ljacomet: Works - my bad -I learned something

docs/src/docs/asciidoc/user/caching-concepts.adoc
ljacomet: Indeed a good practice to explicitly create them - but needs to be on top of the section definition, not below it.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
ljacomet: And works as well here ...

Issue #993, opened 2016-04-19
Server Store API and TCK


chrisdennis:

ljacomet:
Thinking a bit about the data structure and the kind of operations we will want to do on it, I am wondering if we do not want something like a peek next on the `Element` at client side. It would look for the next `Element` having an equal key.

One example:
* I reach an expired value in the chain
* It is effectively expired unless the next element for this key is an expiry duration update, coming from a flush.

PR #998, opened 2016-04-21
Server side store api for cache data storage #993

AbfrmBlr:
This PR is for early feedback on the api of the server side data structure only.

### DIFF
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.common.store;
+
+import java.util.Iterator;
+
+/**
+ * Chain of element made of blobs
+ * https://en.wikipedia.org/wiki/Chain.
+ *
+ * Technically, this is like a list of {@link Element}s.
+ * This is server side data structure to store binary representation
+ * of Cache data.
+ *
+ * This structure follows the insertion order, like an ordered list.
+ * The {@link Iterator} returned by this class allows to traverse the chain
+ * starting from first element.
+ */
+public interface Chain extends Iterable<Element> {
+
+  /**
+   * Returns the iterator to iterate the {@link Chain} of
+   * {@link Element}s in backwards direction i.e. starting
+   * from last one.
+   *
+   * @return an Iterator.
+   */
+  Iterator<Element> descendingIterator();
+
+  /**
+   * Returns true if Chain is empty else false.
+   *
+   * @return whether the Chain is empty
+   */
+  boolean isEmpty();
+}
+
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
@@ -26,6 +26,8 @@ import java.util.Iterator;
  * This is server side data structure to store binary representation
  * of Cache data.
  *
+ * This interface just allows iteration over elements at client side.
+ *
  * This structure follows the insertion order, like an ordered list.
  * The {@link Iterator} returned by this class allows to traverse the chain
  * starting from first element.
@@ -39,7 +41,7 @@ public interface Chain extends Iterable<Element> {
    *
    * @return an Iterator.
    */
-  Iterator<Element> descendingIterator();
+  Iterator<Element> reverseIterator();
 
   /**
    * Returns true if Chain is empty else false.
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/store/ChainBuilder.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/store/ChainBuilder.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.common.store;
+
+/**
+ * Builds {@link Chain}s
+ */
+public interface ChainBuilder {
+
+  /**
+   * Builds chain using elements provided
+   *
+   * @param elements to be part of the new Chain
+   * @return the new {@link Chain}
+   */
+  Chain build(Element... elements);
+
+  /**
+   * Returns an empty {@link Chain} with zero {@link Element}s.
+   *
+   * @return empty Chain
+   */
+  Chain emptyChain();
+}
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/store/Element.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/store/Element.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.common.store;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Building blocks of {@link Chain}
+ */
+public interface Element {
+
+  /**
+   * The data payload stored as element.
+   * This is binary representation of key/value pairs
+   * or key/func(value) pairs
+   *
+   * @return the payload associated with this {@link Element}
+   */
+  ByteBuffer getPayload();
+}
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/store/ElementBuilder.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/store/ElementBuilder.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.common.store;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Builds {@link Element}s
+ */
+public interface ElementBuilder {
+
+  /**
+   * Gets an updated {@link Element} from existing element
+   * with new payLoad
+   *
+   * @param element the old element
+   * @param payLoad the new payLoad
+   * @return the Element with a new payLoad
+   */
+  Element getElement(Element element, ByteBuffer payLoad);
+}
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/store/ServerStore.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/store/ServerStore.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.server.store;
+
+import org.ehcache.clustered.common.store.Chain;
+import org.ehcache.clustered.common.store.Element;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Defines Server Side data structure for storing
+ * cache data.
+ * This is conceptually a multi-map which has hash as the key
+ * and a {@link Chain} as the mapping
+ * Chain is list of {@link Element}s such that the same key has
+ * multiple mappings attached.
+ *
+ * Physical representation looks something like this :
+ *
+ *  hash -> |payLoad1| - |payLoad2| - |payLoad3|
+ *  hash -> |payLoad1| - |payLoad2|
+ *
+ *  This only allows append & replaceAtHead operations rather than
+ *  normal map operations.
+ *
+ */
+public interface ServerStore {
+
+  /**
+   * Returns the Chain associated with the provided hash.
+   * An empty Chain if no mapping exists.
+   *
+   * @param key hashcode of the key
+   * @return the {@link Chain} associated with the hash
+   */
+  Chain get(long key);
+
+  /**
+   * Appends the provided binary to Chain associated with key atomically.
+   * While appending, the payLoad is stored in {@link Element}.
+   *
+   * @param key to which the payLoad has to be appended
+   * @param payLoad to be appended
+   */
+  void append(long key, ByteBuffer payLoad);
+
+  /**
+   * The Chain associated with key, previous to append is returned.
+   * An empty Chain if no mapping existed previously.
+   * The operation atomically appends payLoad and returns the previously associated Chain
+   * with the key.
+   *  Following block of instructions are atomically performed.
+   *  {
+   *    Chain prevChain = get(key);
+   *    append(key, payLoad);
+   *    return prevChain;
+   *  }
+   *
+   * @param key to which the payLoad has to be appended
+   * @param payLoad to be appended
+   * @return the Chain associated with the key before payLoad was appended
+   */
+  Chain getAndAppend(long key, ByteBuffer payLoad);
+
+  /**
+   * Replaces the provided Chain with the equivalent Chain present at the head.
+   * This operation is not guaranteed to succeed.
+   * The replaceAtHead is successful iff the Chain associated with the key has
+   * a sub-sequence of elements present as expected..
+   *
+   * If below mapping is present:
+   *
+   *    hash -> |payLoadA| - |payLoadB| - |payLoadC|
+   *
+   * And replaceAtHead(hash, |payLoadA| - |payLoadB| - |payLoadC|, |payLoadC'|) is invoked
+   * then this operation will succeed & the final mapping would be:
+   *
+   *    hash -> |payLoadC'|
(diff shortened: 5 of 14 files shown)


clustered/common/src/main/java/org/ehcache/clustered/common/store/Link.java
```diff
@@ -0,0 +1,49 @@
...
+   * generated by the server.
+   * This number is unique in a Chain =>
+   *  Same sequence number cannot be allotted to more than one link
+   *  in same chain.
+   *
+   * It is always an ever increasing positive number.
+   *
+   * @return the sequence number associated with this {@link Link}
+   */
+  long getSequenceNumber();
```
chrisdennis: The sequence numbers shouldn't need to be exposed through this interface.  The fact that sequence numbers even exist is just an implementation detail that we will use to optimize the replace operation.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ServerStore.java
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * Chain to be replaced, should also adhere to the rule of ascending order of Links,
+   * & all the sequence numbers in that chain should fall in the range of provided sequenceNumbersExpected
+   * as the sequence number is generated by server else operation will be ignored.
+   *
+   * @param key for which Chain has to be replaced
+   * @param update the Chain to be replaced for the provided sub-sequence
+   * @param sequenceNumbersExpected of the original Chain which this operation intends to replace
+   */
+  void replace(long key, Chain update, long... sequenceNumbersExpected);
```
chrisdennis: At this level we just want to expose the replace arguments as (long, Chain, Chain)... the fact that we're using sequence numbers internally should be hidden from the API consumer.

clustered/common/src/main/java/org/ehcache/clustered/common/store/Link.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * The data payload stored as link.
+   * This is binary representation of key/value pairs
+   * or key/func(value) pairs
+   *
+   * @return the payload associated with this {@link Link}
+   */
+  byte[] getPayload();
```
chrisdennis: Given that the network comms code will be using ByteBuffer, and offheap will also be using ByteBuffers it probably makes more sense for us to use ByteBuffer in the API as well.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ServerStore.java
chrisdennis: Given what we're talking about as potential solutions for the data dependent passive replication problem we may want to think about returning a boolean here indicating success or failure.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ServerStore.java
```diff
@@ -0,0 +1,113 @@
...
+   * @return the Chain associated with the key before payLoad was appended
+   */
+  Chain getAndAppend(long key, byte[] payLoad);
+
+  /**
+   * Replaces the provided Chain with the equivalent Chain present as per
+   * provided sequence Numbers for the provided key.
+   * This operation is not guaranteed to succeed.
+   * The replace is successful iff the Chain associated with the key has
+   * a sub-sequence of links present with the provided list of sequence numbers.
```
chrisdennis: Can the sequence to be replaced be present anywhere in the stored chain, or can we be more specific than that?  What about multiple occurrences?

clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
```diff
@@ -0,0 +1,61 @@
...
+ *      long curNum = link.getSequenceNumber();
+ *      if (curNum <= sequenceNum) {
+ *        throw new AssertionError("Sequence Number should be in ascending order");
+ *      }
+ *      sequenceNum = curNum;
+ *    }
+ *
+ *
+ */
+public interface Chain extends Iterable<Link>, Serializable {
```
chrisdennis: Serializable?  Why?

AbfrmBlr:
@chrisdennis PR updated

clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
```diff
@@ -0,0 +1,44 @@
...
+ * https://en.wikipedia.org/wiki/Chain.
+ *
+ * Technically, this is like a list of {@link Link}s.
+ * This is server side data structure to store binary representation
+ * of Cache data.
+ *
+ * This structure follows the insertion order, like an ordered list.
+ * The {@link Iterator} returned by this class allows to traverse the chain
+ * in ascending order.
+ */
```
chrisdennis: Ascending/descending order is a bit confusing in a data structure that isn't ordered by a comparator.

clustered/common/src/main/java/org/ehcache/clustered/common/store/Link.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.ehcache.clustered.common.store;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Building blocks of {@link Chain}
+ */
+public interface Link {
```
chrisdennis: It's weird to call this class link... I would tend to think of the chain as composed of elements connected with links.  In that sense this is the "Element".

clustered/server/src/main/java/org/ehcache/clustered/server/store/ServerStore.java
```diff
@@ -0,0 +1,115 @@
...
+   *
+   * then replaceAtHead(hash, |payLoadC'|, |payLoadA| - |payLoadB| - |payLoadC|) will be ignored.
+   *
+   * Chain to be replaced, should also adhere to the rule of ascending order of Links.
+   *
+   * @param key for which Chain has to be replaced
+   * @param update the new Chain to be replaced
+   * @param expect of the original Chain which this operation intends to replaceAtHead
+   */
+  void replaceAtHead(long key, Chain update, Chain expect);
```
chrisdennis: This ordering is a bit unconventional.  More normal for this kind of API would be expected then update.

clustered/common/src/main/java/org/ehcache/clustered/common/store/ElementBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Gets an updated {@link Element} from existing element
+   * with new payLoad
+   *
+   * @param element the old element
+   * @param payLoad the new payLoad
+   * @return the Element with a new payLoad
+   */
+  Element getElement(Element element, ByteBuffer payLoad);
```
chrisdennis: What properties of the old element are preserved in the new element?  How are the two related?

clustered/common/src/main/java/org/ehcache/clustered/common/store/ChainBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+   * @return the new {@link Chain}
+   */
+  Chain build(Element... elements);
+
+  /**
+   * Returns an empty {@link Chain} with zero {@link Element}s.
+   *
+   * @return empty Chain
+   */
+  Chain emptyChain();
```
chrisdennis: Isn't this just ChainBuilder.build();

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,227 @@
...
+  private static ByteBuffer getPayload(long key) {
+    ByteBuffer byteBuffer = ByteBuffer.allocate(8).putLong(key);
+    return byteBuffer;
+  }
+
+  @Test
+  public void testGetNoMappingExists() {
+    ServerStore store = newStore();
+    Chain chain = store.get(1);
+    assertThat(chain.isEmpty(), is(true));
```
chrisdennis: Probably want to assert the behavior of the iterator as well

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,227 @@
...
+    Element element2 = linkIterator.next();
+    assertThat(readPayLoad(element2.getPayload()), is(Long.valueOf(22)));
+
+  }
+
+  @Test
+  public void testGetAndAppendNoMappingExists() {
+    ServerStore store = newStore();
+    Chain chain = store.getAndAppend(1, getPayload(1));
+    assertThat(chain.isEmpty(), is(true));
```
chrisdennis: Again might want to check the iterator behavior here.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  @Test
+  public void testAppendNoMappingExists() {
+    ServerStore store = newStore();
+    store.append(1, getPayload(1));
+    Chain chain = store.get(1);
+    assertThat(chain.isEmpty(), is(false));
+    for (Element element : chain) {
+      assertThat(readPayLoad(element.getPayload()), is(Long.valueOf(1)));
```
chrisdennis: Iterator could return multiple items erroneously and pass.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+    chain = store.get(1);
+    Iterator<Element> linkIterator = chain.iterator();
+    Element element1 = linkIterator.next();
+    assertThat(readPayLoad(element1.getPayload()), is(Long.valueOf(1)));
+    Element element2 = linkIterator.next();
+    assertThat(readPayLoad(element2.getPayload()), is(Long.valueOf(22)));
+
```
chrisdennis: Probably should assert that linkIterator.hasNext returns correctly here.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,227 @@
...
+    Chain existingMapping = store.get(1);
+
+    store.append(1, getPayload(11));
+
+    store.replaceAtHead(1, existingMapping, chainBuilder.build(compactChain(existingMapping, getPayload(111))));
+    Chain chain = store.get(1);
+    Iterator<Element> elements = chain.iterator();
+
+    assertThat(readPayLoad(elements.next().getPayload()), is(Long.valueOf(111)));
```
chrisdennis: Isn't this wrong?  The list before the replaceAtHead was: [1, 11], and the replace was ([1], [111]), so the result should be [111, 11], right?  So when iterator in reverse order you'll see [11] then [111].  Or am I getting confused?

clustered/common/src/main/java/org/ehcache/clustered/common/store/ChainBuilder.java
AbfrmBlr: It is, I just intended to add an explicit method for that.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
AbfrmBlr: But I am using normal iterator to traverse from start to end.

clustered/common/src/main/java/org/ehcache/clustered/common/store/ElementBuilder.java
AbfrmBlr: They may not be, totally depends on implementation, but I think while adding this method I was thinking too much about old ascending and descending stuff & sequence numbers. I think we can just have build(ByteBuffer payload) in this builder as well.

AbfrmBlr: I have some confusion here. This api is supposed toe be used at client side only.May be I will move it to `client` from `common`. Now for the client the two elements are related in the sense that the new element will use the same id(implementation specific) but a different payload. Thus that can be identified at server also unless we decide to just send a chain of dummy elements(with just payloads) and then create server side representation for elements during replace operation.

ljacomet: I do not understand the role of this method. From the data structure we plan to use, most often we will end up wanting to replace a sub `Chain` of `Element`s with a `Chain` of a single `Element`. And so I do not see where this would be used.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,222 @@
...
+      store.append(i, getPayload(i));
+    }
+  }
+
+  private static long readPayLoad(ByteBuffer byteBuffer) {
+    byteBuffer.flip();
+    return byteBuffer.getLong();
+  }
+
+  private static ByteBuffer getPayload(long key) {
```
ljacomet: `createPayload` for better readability

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,222 @@
...
+    populateStore(store);
+
+    Chain existingMapping = store.get(1);
+
+    store.append(1, getPayload(11));
+
+    store.replaceAtHead(1, existingMapping, chainBuilder.build(compactChain(existingMapping, getPayload(111))));
+    Chain chain = store.get(1);
+
+    assertChainHasOnly(chain, 111, 11);
```
ljacomet: You always combine this assert and the one on the next line - it could be extracted in a single method.

ljacomet: Something like `assertChainAndReverseChainOnlyHave(chain, values...)`

clustered/common/src/main/java/org/ehcache/clustered/common/store/ElementBuilder.java
AbfrmBlr: @ljacomet  that's the question I am asking, do we say that client only sends a payload in form of dummy element, because Element is a wrapper around payload with some more info in to it for Chain. If we got to compare two Chains at the server side, there has to be some identifier in implementation to enable that. 
So the current method provides possibility to copy that extra info(otherwise not available at client side) in the new element with the payload.

chrisdennis: 1. We should never change the payload attached to a specific sequence number.
2. The API (client or server side) shouldn't bleed any details regarding the use of sequence numbers.  Lets not build a leaky abstraction here.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ServerStoreTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+  private static long readPayLoad(ByteBuffer byteBuffer) {
+    byteBuffer.flip();
+    return byteBuffer.getLong();
+  }
+
+  private static ByteBuffer getPayload(long key) {
+    ByteBuffer byteBuffer = ByteBuffer.allocate(8).putLong(key);
```
chrisdennis: You either need to do an absolute put here, or flip the buffer before returning it.  At the moment this is returning buffers with zero remaining bytes, so they'll all be equal to each other.

clustered/common/src/main/java/org/ehcache/clustered/common/store/ChainBuilder.java
ljacomet: Drop it, adds nothing

AbfrmBlr:
PR updated

clustered/server/src/test/java/org/ehcache/clustered/server/store/impl/ReferenceStoreImpl.java
```diff
@@ -0,0 +1,152 @@
...
+  private final int LOCK_COUNT = 16;
+
+  public ReferenceStoreImpl() {
+    for (int i = 0; i < LOCK_COUNT; i++) {
+      locks.add(new ReentrantReadWriteLock());
+    }
+  }
+
+  private ReadWriteLock getLock(long key) {
+    return locks.get((int)key%LOCK_COUNT);
```
lorban: What if the key is negative?

clustered/server/src/test/java/org/ehcache/clustered/server/store/impl/ReferenceStoreImpl.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  public Chain get(long key) {
+    Lock lock = getLock(key).readLock();
+    try {
+      lock.lock();
```
lorban: It's harmless here since you're only using `j.u.c.l.ReentrantReadWriteLock` but your pattern is wrong: the lock should be taken out of the `try - catch` block.

clustered/common/src/main/java/org/ehcache/clustered/common/store/Element.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.ehcache.clustered.common.store;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Building blocks of {@link Chain}
+ */
+public interface Element {
```
lorban: Shouldn't there be some notion of `function` encoding in this interface?

clustered/server/src/test/java/org/ehcache/clustered/server/store/ChainBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.ehcache.clustered.server.store;
+
+import org.ehcache.clustered.common.store.Chain;
+import org.ehcache.clustered.common.store.Element;
+
+/**
+ * Builds {@link Chain}s
+ */
+public interface ChainBuilder {
```
lorban: What's the point of such `Builder` interface in the test sources?

clustered/server/src/test/java/org/ehcache/clustered/server/store/ElementBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+package org.ehcache.clustered.server.store;
+
+import org.ehcache.clustered.common.store.Element;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Builds {@link Element}s
+ */
+public interface ElementBuilder {
```
lorban: What's the point of such `Builder` interface in the test sources?

clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Technically, this is like a list of {@link Element}s.
+ * This is server side data structure to store binary representation
+ * of Cache data.
+ *
+ * This structure follows the insertion order, like an ordered list.
+ * The {@link Iterator} returned by this class allows to traverse the chain
+ * starting from first element.
+ */
+public interface Chain extends Iterable<Element> {
```
lorban: I have a feeling that this interface is either not complete enough or useless as it is not possible to implement `ServerStore` solely against this interface.

clustered/server/src/test/java/org/ehcache/clustered/server/store/impl/ReferenceStoreImpl.java
AbfrmBlr: I know its ignored since it's a reference implementation just for the sake of tests to run.

clustered/common/src/main/java/org/ehcache/clustered/common/store/Element.java
AbfrmBlr: I don't think so

clustered/server/src/test/java/org/ehcache/clustered/server/store/ChainBuilder.java
AbfrmBlr: Both the builders were added with intention to enable client side creation of Chains and Elements. But I am not very sure about how they will be used until we figure out the ServerStoreProxy and its impl. FOr now Ihave just moved them to test package and it surely helps in implementing a ServerStore test which does not depend on the reference implementation

clustered/common/src/main/java/org/ehcache/clustered/common/store/Chain.java
AbfrmBlr: Since this is used at the client side to traverse the Chain of elements, who should be just able to iterate through it, for now it is enough but needs a clear mention in java doc.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ChainBuilder.java
AbfrmBlr: Will leave these builders as it is for now, till ServerStoreProxy is nailed. Later if not required these can be removed totally and replaced by construction in tests.

clustered/common/src/main/java/org/ehcache/clustered/common/store/Element.java
AbfrmBlr: After discussing with @lorban , I think this needs to be re-looked and pondered over for future implications.. Having this information at server side can enable lot of optimizations in future if chain/element can somehow discriminate between different keys in same hash.

jhouserizer: Couldn't facilitation of such optimizations be added to the abstraction (or as an additional abstraction) at the point when we'd be adding such optimizations?

Let's not over-complicate things for now - keep the scope constrained.

PR #999, opened 2016-04-25
Gradle 2.13

ljacomet:


buildSrc/src/main/groovy/EhDistribute.groovy
```diff
@@ -70,7 +70,7 @@ class EhDistribute implements Plugin<Project> {
 
     project.jar.doFirst {
       manifest = new DefaultInheritManifest(getServices().get(FileResolver.class))
-      manifest.inheritFrom project.shadowJar.manifest
+      manifest.inheritFrom "$project.buildDir/tmp/shadowJar/MANIFEST.MF"
```
chrisdennis: This seems more brittle than before... what happened?

ljacomet: Changes in the internal gradle API that prevent the former to work. It now only will accept something that resolves as a path.

PR #1001, opened 2016-04-26
Issue 870: fix BulkCacheWritingException under load

lorban:
The EhcacheWithLoaderWriter removeAll() and putAll() implementations now have the logic to deal with Store.bulkCompute() calling the function with the same arguments many times.

I've also modified the BulkCacheWritingException class to report a more useful error message.


AbfrmBlr:
LGTM

chrisdennis:
The fix here looks good, only remaining question is whether we want to play around with the git workflow for bugs, using this as a guinea pig.

ljacomet:
Merged manually after rebasing on parent of master and release/3.0

PR #1010, opened 2016-04-27
Make preamble scrollable vertically

GaryWKeim:


ljacomet:
Looks good, just need a rebase.

PR #1011, opened 2016-04-29
Remove mavenLocal() from repo list

ljacomet:
Its presence causes more grief than anything. Developers needing to
pull locally built artifacts can add it temporarily.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,7 +147,6 @@ subprojects {
   repositories {
     mavenCentral()
     maven { url "http://repo.terracotta.org/maven2" }
-    mavenLocal()
   }
 
   sourceSets {


ljacomet:
As discussed with @chrisdennis and @cljohnso this fix uses the [daggy fix](http://wiki.monotone.ca/DaggyFixes/) approach.

lorban:
Except that in some cases you want gradle to use the local maven repo. Couldn't we add a command-line property, like `-P useLocalMavenRepo=true` to enable that on-demand?

ljacomet:
Then I would rather devs use the provided init feature of Gradle. See https://docs.gradle.org/current/userguide/init_scripts.html#N14A03

lorban:
LGTM

ljacomet:
Well, the whole point of a daggy fix is that it does not get rebased, instead if it fails to merge cleanly there is a merge on it taking care of the conflicts.

PR #1012, opened 2016-04-29
Remove mavenLocal() from repo list

ljacomet:
Its presence causes more grief than anything. Developers needing to
pull locally built artifacts can add it temporarily.


ljacomet:
As discussed with @chrisdennis and @cljohnso this fix uses the [daggy fix](http://wiki.monotone.ca/DaggyFixes/) approach.

Issue #995, opened 2016-04-19
ehcache v3 set setTimeToIdleSeconds without use xml config


marti1125:
I didn't find an example like this http://www.programcreek.com/java-api-examples/index.php?api=net.sf.ehcache.config.CacheConfiguration

ljacomet:
Hi,

Thanks for your interest in Ehcache 3.

For this kind of questions, we recommend using the [ehcache-users](https://groups.google.com/forum/#!forum/ehcache-users) google group.

Now, to answer your question:
Ehcache 3 is a complete rework of the API.

What you are looking for can be done as follows:
```java
CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, Person.class,
    ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES))
    .withExpiry(Expirations.timeToIdleExpiration(new Duration(10, TimeUnit.SECONDS)));
```
Please refer to the [rest of the documentation](http://www.ehcache.org/documentation/3.0/getting-started.html) for interactions between cache configurations and cache managers.

Thanks to your question I realised this is a topic left out of the Getting Started documentation and that is a mistake that needs fixing.

marti1125:
@ljacomet Thanks!! =D how to I can help you?

ljacomet:
Keep on reporting missing documentation topics, that would help for sure.

marti1125:
I would like to add eternal false http://techblog.constantcontact.com/software-development/solving-a-classic-ehcache-problem-retrieving-expired-data/

lorban:
In Ehcache 3, there is no such thing as 'eternal'. Simply configuring an expiry (using `CacheConfigurationBuilder.withExpiry(...)`) will make your configured expiry policy taken into account by the cache.

PR #1016, opened 2016-05-02
Issue #995 Expiration documentation

ljacomet:
* Expiry documentation
* Addition of `Duration.of` static method

### DIFF
--- a/docs/src/docs/asciidoc/user/index.adoc
+++ b/docs/src/docs/asciidoc/user/index.adoc
@@ -43,6 +43,7 @@ Each topic below corresponds to a menu item at the left.
 |===
 | Topic | Description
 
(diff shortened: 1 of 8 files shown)


docs/src/docs/asciidoc/user/index.adoc
```diff
@@ -43,6 +43,7 @@ Each topic below corresponds to a menu item at the left.
 |===
 | Topic | Description
 
+|link:expiry.html[Expiry]|Data freshness in Ehcache 3
```
lorban: I would add the word `expiry` or `expiration` to this description, and drop the ehcache 3 reference. Something along the lines of `Data freshness and expiration`.

Issue #1015, opened 2016-05-01
Per key metadata sharing across clients in clustered ehcache


rkavanap:
For a multi-tiered clustered cache manager, that has a cache that has either TTI and TTL configured, there are two challenges with respect to per-key metadata that needs to be addressed..

When a client does a get on a key locally from its caching tier, other clients doing a get on the same key later must somehow see the correct access time for this key locally in order to accurately compute 'time to idle' and decide whether an entry needs to be expired or not.

For both TTL and TTI computations with clients having different clocks, the computations must ensure that the clocks are synchronized across multiple VMs/systems either by using a single time reference (say of the server) for absolute time with clients computing time relative to this absolute time or by ensuring that the clocks are synchronized using internal or external clock synchronization mechanisms.

rkavanap:
A solution proposal for this issue is illustrated here to solicit comments and stimulate discussions.

### Timestamped Links, Metadata appends and Fault before flush

One potential solution for this issue is to time stamp every link in the blob chain on the server and then the client can use the time stamp on each link to recompute the lastaccesstime and the createdtime for the key.

Sub optimally on every get (or optimally when expiry is nearing) that hits the caching tier, a metadata append operation can be send asynchronously to the cluster tier, so that the server can add a link to the chain. Optimally, this append can just have the key and operation type in the blob, as the time stamp on the link can be put by the server when the link is appended to the blob chain. These metadata blob links in the chain along with other functions in the chain will help other clients to accurately compute the metadata such as lastaccesstime and creationtime etc.

If we want expiry computations to be more accurate in a clustered environment even in corner cases where there is frequent gets from an interested client and infrequent gets from another bored client, this solution can be expanded as follows: 

When a TTI is configured, a semi-interested or bored client that 'wrongly sees' an expiry for a key on a caching tier (due to infrequent hits on the caching tier by itself) can first trigger a 'fault', resolve and recompute the lastaccesstime and issue a 'flush' if and only if the expiry was deemed accurate on the second pass. 

_The above solution will be illustrated using sequence diagrams etc in another comment or doc_

### Pros and Cons of this approach:

1. This approach can be extended easily to ensure that the absolute time reference is always that of the server (and not the client). This means that only the active and its passive needs to be time synchronized and client VMs need not be time synchronized (at least for this requirement).

[...]

PR #1020, opened 2016-05-02
Issue #1005 Add PUT operation and the codec for the same

albinsuresh:
A few assumptions made during the operation definition:
- The design was limited to simple get and put operations only.
- A sequence of PUT operations get resolved to the last PUT operation. A need for a different type to represent the resolved value does not exist at the moment as there is not metadata associated with operations. Hence PUT operation can be reused. A new type can be added if the need arises in future.
- Since the initial iteration is not gonna have expiry support the metadata has been completely omitted. Addition of metadata for expiry calculations will depend upon the decision taken on #1015.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseOperation.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseOperation.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.client.internal.store.operations;
+
+/**
+ * Base class that represents a {@link org.ehcache.Cache} operation
+ *
+ * @param <K> key type
+ */
+public abstract class BaseOperation<K> {
+
+  protected final K key;
+
+  public BaseOperation(final K key) {
+    this.key = key;
+  }
+
+  public abstract OperationCode getOpCode();
+
+  public K getKey() {
+    return key;
+  }
+}
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/OperationCodecFactory.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/OperationCodecFactory.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.client.internal.store.operations.codecs;
+
+import org.ehcache.spi.serialization.Serializer;
+
+/**
+ * Base codec factory for all operations
+ */
+public abstract class OperationCodecFactory {
+
+  public abstract <K, V> OperationCodec<K> getOperationCodec(Serializer<K> keySerializer, Serializer<V> valueSerializer);
+
+  /**
+   * Codec factory for put operation
+   */
+  public static class PutOperationCodecFactory extends OperationCodecFactory {
+
+    public <K, V> OperationCodec<K> getOperationCodec(Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+      return new PutOperationCodec<K, V>(keySerializer, valueSerializer);
+    }
+  }
+}
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/OperationsCodec.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/OperationsCodec.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.client.internal.store.operations.codecs;
+
+import org.ehcache.clustered.client.internal.store.operations.BaseOperation;
+import org.ehcache.clustered.client.internal.store.operations.OperationCode;
+
+import java.nio.ByteBuffer;
+
+public class OperationsCodec<K, V> {
+
+  public static final int BYTE_SIZE_BYTES = 1;
+  public static final int INT_SIZE_BYTES = 4;
+  public static final int LONG_SIZE_BYTES = 8;
+
+  private final OperationCodecProvider<K, V> codecProvider;
+
+  public OperationsCodec(final OperationCodecProvider<K, V> codecProvider) {
+    this.codecProvider = codecProvider;
+  }
+
+  public ByteBuffer encode(BaseOperation<K> operation) {
+    OperationCode opCode = operation.getOpCode();
+    return codecProvider.getOperationCodec(opCode).encode(operation);
+  }
+
+  public BaseOperation<K> decode(ByteBuffer buffer) throws ClassNotFoundException {
+    OperationCode opCode = OperationCode.valueOf(buffer.get());
+    buffer.rewind();
+    return codecProvider.getOperationCodec(opCode).decode(buffer);
+  }
+}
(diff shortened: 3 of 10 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseOperation.java
```diff
@@ -0,0 +1,65 @@
...
+package org.ehcache.clustered.client.internal.store.operations;
+
+/**
+ * Base class that represents a {@link org.ehcache.Cache} operation
+ *
+ * @param <K> key type
+ */
+public abstract class BaseOperation<K> {
+
+  public enum Code {
```
ljacomet: I fail to see the point of this enum. It is awkward in the sense that being declared in the `BaseOperation` it will have to be close to all sub classes, thus braking encapsulation. And even for serialisation, it does not really help more than simply having directly a byte field inside the operation.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/PutOperationCodec.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public PutOperation<K, V> decode(ByteBuffer buffer) throws ClassNotFoundException {
+
+    BaseOperation.Code opCode = BaseOperation.Code.valueOf(buffer.get());
+    if (opCode != PUT) {
+      throw new AssertionError(this.getClass().getName() + " can not decode operation of type " + opCode);
+    }
+
+    byte[] keyBlob = new byte[buffer.getInt()];
```
ljacomet: Consider slicing the ByteBuffer instead of allocating another byte[], a bit more math to compute bounds, but less allocations in the end.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/OperationsCodec.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final OperationCodecFactory<K, V> codecFactory;
+
+  public OperationsCodec(final OperationCodecFactory<K, V> codecFactory) {
+    this.codecFactory = codecFactory;
+  }
+
+  public ByteBuffer encode(BaseOperation<K> operation) {
+    BaseOperation.Code opCode = operation.getOpCode();
+    switch (opCode) {
```
ljacomet: Having `switch` here feels like encapsulation breakage. Not that I have a specific recommendation just now.

ljacomet: Scratch that - `OperationCodecFactory` could be based on the operation concrete type, using a `Map<Class<? extends BaseOperation>,? extends OperationCodec>`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseOperation.java
albinsuresh: I agree. All I needed was a simple byte in the `BaseOperation` class to represent the type of operation. But the reason I chose the enum was to clearly map the operation type with a value. If no such mapping exists anywhere, there will be if conditions/switch blocks in the decode methods of codecs that checks for different values before decoding. Code like this:
```java
byte code = buffer.get();
switch(code) {
  case 0:
    //do stuff for PUT
  case 1:
    // do stuff for another operation
  default:
    // something else
}
```

So the mapping will end up defined in the conditional blocks in the decode methods. That is what I was trying to avoid. Define the mapping somewhere centrally. And use it while decoding. And enum provided the added benefit of readability at places where the decoding is done.

So that was the rationale behind the addition of enums. Does it make sense?

lorban: What I think Louis has in mind is to augment the `OperationCodecFactory` class to contain a pair of maps: one `Map<Class<? extends BaseOperation>,? extends OperationCodec>` for encoding where you can lookup the codec using the operation and another one `Map<Byte, ? extends OperationCodec>` for decoding where you can lookup the codec using the op byte.

That way, you'd get rid of the need to use switch blocks as that would centralize that logic in a single location.

ljacomet: Or if you believe the enum is the right way to go, push it all the way. It becomes the central place where operation characteristics, codecs and others are recored and the rest of the code becomes more generic.

albinsuresh:
Addressed some of the review feedback:
- Improved the decoding logic by avoiding additional allocations
- Did some refactoring to improve the abstractions

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/OperationCodecFactory.java
```diff
@@ -0,0 +1,57 @@
...
+      operationCodeVsCodecMap.put(code, getNewCodec(code));
+    }
+    return map;
+  }
+
+  public OperationCodec<K> getOperationCodec(OperationCode code) {
+    return operationCodeVsCodecMap.get(code);
+  }
+
+  private OperationCodec<K> getNewCodec(OperationCode code) {
```
lorban: This looks rather convoluted still. You could replace that switch with a `getCodec()` method in the `OperationCode` enum.

albinsuresh: Yeah I had though about it. That'd mean that I'll have to move the codec initialization to the enum. The codecs would require the key and value serializers for initialization. So that'd mean that I'll have to pass both the serializers at the time of enum constant initialization. I shouldn't be doing that, right?

lorban: Good point, let's not do that and leave that the way it is - unless someone else has a better suggestion.

ljacomet: This could still be improved by having the enum return an `OperationCodecFactory` which would return the proper `OperationCodec` through a method taking the different `Serializer`s.

albinsuresh: Well things could get a bit weird if I embark on that path. Or may be I'm not understanding your suggestion properly: 

If I add a method at the enum level returning an `AbstractOperationCodecFactory` which has a structure something like this: 

```java
abstract class AbstractOperationCodecFactory<K, V> {
  public abstract OperationCodec<K> getOperationCodec(Serializer<K> keySerializer, Serializer<V> valueSerializer);
}
```
and a concrete implementation, like for example `PutOperationCodecFactory` with a structure like this:

```java
public class PutOperationCodecFactory<K, V> extends AbstractOperationCodecFactory<K, V> {

  OperationCodec<K> getOperationCodec(Serializer<K> keySerializer, Serializer<V> valueSerializer) {
    return new PutOperationCodec<K, V>(keySerializer, valueSerializer);
  }
}
```
If we take this path, then for every encode method call, we'll end up calling:

Enum.getOperationCodec(keySer, valueSer).encode(operation)

Every _getOperationCodec_ call is gonna return a new instance of the codec with the provided serializer pairs and finally we end up with too many `OperationCodec` instances. So I'm not sure if this is the right thing to do.

Now if we decide to cache the instance of codec on the first _getOperationCodec_ call and then decide to reuse it on subsequent calls( singleton like approach), we'll end up with an enum that'll be tied to a single pair of serializers which in-turn ties the enum to a single cache. I don't think that's what we want either.

These were the two approaches that came to my mind. Did you have something totally different in mind? May be I'm missing something.

ljacomet: I am not sure I follow some of your logic and maybe there is something I am missing.
In the context of this method `private OperationCodec<K> getNewCodec(OperationCode code)` instead of the switch, you do:
```java
return code.getOperationCodecFactory().newCodec(keySerializer, valueSerializer);
```
And that is it, the rest of this class remains the same, you just encapsulated the codec selection inside the enum without changing any of the lifecycle existing as of now.

However, it is true I did not do a complete lifecycle analysis and maybe there is something not right there.

albinsuresh: "In the context of the _getNewCodec_ method". Okay. Now I get it. Have updated the PR.

albinsuresh:
In the last iteration I've done some nasty renaming which has resulted in a slightly convoluted diff view:
The old `OperationCodecFactory` has been renamed as `OperationCodecProvider`. And the name `OperationCodecFactory` was reused as the name of the new abstract base class that is extended by all operation specific codec factories like `PutOperationCodecFactory`. So just watch out for these.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/PutOperationCodec.java
```diff
@@ -0,0 +1,96 @@
...
+  private final Serializer<V> valueSerializer;
+
+  public PutOperationCodec(final Serializer<K> keySerializer, final Serializer<V> valueSerializer) {
+    this.keySerializer = keySerializer;
+    this.valueSerializer = valueSerializer;
+  }
+
+  public ByteBuffer encode(BaseOperation<K> operation) {
+    if (!(operation instanceof PutOperation)) {
+      throw new AssertionError(this.getClass().getName() + " can only decode " + PutOperation.class.getName());
```
lorban: I would rather throw `IllegalArgumentException` here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/PutOperationCodec.java
```diff
@@ -0,0 +1,96 @@
...
+    buffer.put(valueBuf);
+    buffer.flip();
+    return buffer;
+  }
+
+  public PutOperation<K, V> decode(ByteBuffer buffer) throws ClassNotFoundException {
+
+    OperationCode opCode = OperationCode.valueOf(buffer.get());
+    if (opCode != PUT) {
+      throw new AssertionError(this.getClass().getName() + " can not decode operation of type " + opCode);
```
lorban: As above, I would rather throw `IllegalArgumentException` here.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/operations/codecs/PutOperationCodecTest.java
```diff
@@ -0,0 +1,76 @@
...
+    blob.flip();
+
+    PutOperationCodec<Long, String> codec =
+        new PutOperationCodec<Long, String>(new LongSerializer(), new StringSerializer());
+
+    PutOperation<Long, String> putOperation = codec.decode(blob);
+    assertEquals(key, putOperation.getKey());
+    assertEquals(value, putOperation.getValue());
+  }
+}
```
lorban: I would add some more tests:
 - 2 tests that check that the code throws when a wrong operation is being encoded / decoded
 - a test that decodes what's been encoded then asserting that you get the original key/value pair

lorban:
Please rebase, then it's good to go.

Issue #949, opened 2016-04-07
Implement offheap server data structure for clustered caches


chrisdennis:
Implement the map of lists structure that will back the clustered ehcache instances.  As an initial start an implementation of the following interface will be required:

```java
interface HashOperationStore {

  Chain get(int hash);

  Chain getAndAppend(int hash, ByteBuffer value);

  void append(int hash, ByteBuffer value);

  void replaceAtHead(int hash, Chain expected, Iterable<ByteBuffer> replacement);
}

interface Chain extends Iterable<ByteBuffer> {}
```

The `Chain` interface is required to allow for the use of sequence numbers in the stored values to accelerate the replaceAtHead operation (sequence number match is a safe substitute for a full binary comparison).  Other than the sequence number detail, the initial implementation of this interface should use a pessimistic locking approach... i.e. a copy under read lock during the get operation, and all mutation performed under write lock.  For now this is allows for a significantly simpler implementation, and we can then move toward something more complex as the need arises (or is proven through performance testing).

chrisdennis:
My suggestion for implementation here is to create a simple locked implementation using the AbstractOffHeapClockCache and then simply shard that implementation in the normal way.

PR #1022, opened 2016-05-03
Issue #949 Initial implementation of the chain map

chrisdennis:
What is here right now needs some more work, but I wanted to get some review on it early before going too much further.  Things that are left to do:
 - ~~Cleaner 'activeChain' solution to prevent or accomodate mutation of the chain being worked on during compaction/stealing.~~
 - ~~Segmentation layer above this - fairly trivial.~~
 - ~~Fast path element comparison using sequence numbers - everything is already structured to take this addition, but it still needs to be done.~~

The segmentation layer is really just a temporary stand-in at the moment.  Depending on the details of how we integrate with the Voltron concurrency keys, and how passive replication gets ironed out things here might need to be changed.  This isn't a big deal... but hence I've really not spent too much energy on that at this point.


clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorage.java
```diff
@@ -0,0 +1,568 @@
...
+class OffHeapChainStorage<K> implements StorageEngine<K, InternalChain> {
+
+  private static final int ELEMENT_LENGTH = 0;
+  private static final int ELEMENT_NEXT = 4;
+  private static final int ELEMENT_HEADER_SIZE = 12;
+
+  private static final int CHAIN_KEY_LENGTH = 0;
+  private static final int CHAIN_KEY_HASH = 4;
+  private static final int CHAIN_TAIL = 8;
+  private static final int CHAIN_HEADER_SIZE = 16;
```
lorban: All those constants are a bit poorly named and/or documented. For instance, `CHAIN_KEY_LENGTH = 0` doesn't make much sense but I guess you meant something like `CHAIN_KEY_POSITION = 0`?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/ChainMap.java
```diff
@@ -0,0 +1,229 @@
...
+import org.ehcache.clustered.common.store.Chain;
+import org.ehcache.clustered.common.store.Element;
+import org.ehcache.clustered.common.store.Util;
+import org.terracotta.offheapstore.MapInternals;
+
+import org.terracotta.offheapstore.ReadWriteLockedOffHeapClockCache;
+import org.terracotta.offheapstore.paging.PageSource;
+import org.terracotta.offheapstore.storage.portability.Portability;
+
+public class ChainMap<K> implements MapInternals {
```
lorban: `ChainMap` sounds like a great abstract name for an interface, not for an impl. I would at least use the `OffHeap` prefix.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorage.java
chrisdennis: As you probably guessed, they are all offsets in to the offheap data structures. I'll look in to rewording them to make them clearer/cleaner.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/ChainMap.java
chrisdennis: Can do, will probably rework this to be an implementation of ServerStore as well when I do the segmentation work.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
```diff
@@ -0,0 +1,255 @@
...
+
+  public boolean replaceAtHead(K key, Chain expected, Chain replacement) {
+    final Lock lock = heads.writeLock();
+    lock.lock();
+    try {
+      InternalChain chain = heads.get(key);
+      if (chain == null) {
+        if (expected.iterator().hasNext()) {
+          return false;
+        } else {
```
ljacomet: Fail to see the use case for this. Isn't is simpler to throw `IllegalArgumentException` when receiving an empty `expected`?

chrisdennis: I guess the only argument here is completeness... we handle the empty expected inside the storage engine too.  If you'd rather us not support it, that can be arranged.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -0,0 +1,623 @@
...
+      }
+    }
+
+    @Override
+    public boolean replace(Chain expected, Chain replacement) {
+      if (expected.isEmpty()) {
+        if (replacement.isEmpty()) {
+          return true;
+        } else {
+          insertHeader(replacement);
```
ljacomet: Feels wrong with regards to the high level data structure discussion that `replace` can be made to perform like `prepend`

chrisdennis: See my other comment before... if we want to make empty expected sequences legal then I'm okay with that.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/SequencedElement.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.server.offheap;
+
+import org.ehcache.clustered.common.store.Element;
+
+public interface SequencedElement extends Element {
```
ljacomet: Might be too early - but doesn't this need to be in common so that the client entity can do optimisation in messages?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
```diff
@@ -0,0 +1,255 @@
...
+import org.ehcache.clustered.common.store.Chain;
+import org.ehcache.clustered.common.store.Element;
+import org.ehcache.clustered.common.store.Util;
+import org.terracotta.offheapstore.MapInternals;
+
+import org.terracotta.offheapstore.ReadWriteLockedOffHeapClockCache;
+import org.terracotta.offheapstore.paging.PageSource;
+import org.terracotta.offheapstore.storage.portability.Portability;
+
+public class OffHeapChainMap<K> implements MapInternals {
```
ljacomet: Does it need to be public?

chrisdennis: I guess not - will reduce to package protected.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/SequencedElement.java
chrisdennis: Yes, probably - I'll move it there now... if it needs massaging later then it can be done when someone uses it.

AbfrmBlr:
LGTM

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
lorban: I tend to agree with @ljacomet : an empty `expected` chain should never happen in real life, except when there's a bug. I'd prefer to go the fail-fast route too.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -0,0 +1,624 @@
...
+
+  private final OffHeapStorageArea storage;
+  private final Portability<? super K> keyPortability;
+  private final Set<AttachedInternalChain> activeChains = new HashSet<AttachedInternalChain>();
+
+  private StorageEngine.Owner owner;
+  private long nextSequenceNumber = 0;
+
+  public OffHeapChainStorageEngine(PageSource source, Portability<? super K> keyPortability, boolean thief, boolean victim) {
+    this.storage = new OffHeapStorageArea(PointerSize.LONG, new StorageOwner(), source, 4096, thief, victim);
```
lorban: I'm a bit picky here, but I wonder what you had in mind when you decided to make `thief` and `victim` ctor arguments while hardcoding the page size to 4k.

Good reason? No particular reason?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
lorban: I'm also worried that the entire chain could be replaced with an empty one... isn't that something that should be prevented too?

lorban: Looking more carefully at the code, we should prevent replacement with an empty chain. Look at `createAttachedChain` called by `replaceHeader`: it calls `iterator.next().getPayload()` which would throw `NoSuchElementException` if passed an empty replacement chain.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
chrisdennis: Okay, will do this.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
chrisdennis: `replaceHeader` is never called with an empty replacement chain.  If the chain is empty we either return true (if the expected is also empty) or call removeHeader(expected).  Replacing a sequence with empty is also I assume important for removes.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
chrisdennis: No particular reason.  I'll expose these further out, and switch to more reasonable defaults, but I think at least for now defaulting inside the offheap specific code makes sense.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
lorban: Shouldn't a remove rather destroy the entire chain rather than replacing it with an empty one?

chrisdennis: We don't know (until we get inside the server) whether or not another operation has been appended to the chain since we pulled it that reinserts the mapping.

Issue #1026, opened 2016-05-03
Name cluster connection with "connection.name" to have a user-name in M&M


mathieucarbou:
The server team as set a new property to use when creating a voltron connection. You can set `connection.name` (optional) to set the user-name for a logical connection to a cluster.

M&M will use this name to display the clients.

@chrisdennis / @ljacomet suggested Ehcache could set this property to something like `EHCACHE:<cm-name>` when creating the connection.

CC @cljohnso

ljacomet:
I see PR #1031 already takes care of the default.
In the context of this issue though, we want to think about a public API allowing users to name their connection, which would then override the default.

chrisdennis:
Is this really what we want?  Maybe I've gotten out of sync on this issue.  To my mind:

"Ehcache:NameOfMyCacheStorageInTheCluster" - tells me what the connection is doing
"HotelBookingJettyInstance-43" - tells me what the greater purpose of the connection is.

I had hoped this wouldn't have to be an either/or decision on which of these we expose.

mathieucarbou:
Form latest discussion, we defined the `connection.name` to be a user-defiend string that can be anything but that will represent for what the cluster connection is used for.

Ehcache would then use the name: `Ehcache:my-cache-manager-name`

This is sufficient to know that the connection will only be used by Ehcache in the context of my-cache-manager-name.

If several webapp on the same app server / JVM is using the same name `Ehcache:my-cache-manager-name`, will be able to differentiate these different clients thanks to the `connection.uuid`

In M&M a client is identified by this user string: `PID@ADDRESS:NAME:CONN-UUID`

After that, we will add the ability to set some tags. The tags will allow us to regroup clients based on tags for logical views. So tags could be: `HotelBookingJettyInstance-43, appserver-node-1` for example

mathieucarbou:
Does the above explanation makes sense @ljacomet @chrisdennis  ?
(CC @anthonydahanne)

chrisdennis:
Yes.

mathieucarbou:
Closed by #1031 ?

PR #1031, opened 2016-05-05
Issue #1026 Name ehcache cluster connections

albinsuresh:


clustered/client/src/main/java/org/ehcache/clustered/client/service/DefaultClusteringService.java
```diff
@@ -86,7 +86,9 @@ public ClusteringServiceConfiguration getConfiguration() {
   @Override
   public void start(final ServiceProvider<Service> serviceProvider) {
     try {
-      clusterConnection = ConnectionFactory.connect(clusterUri, new Properties());
+      Properties properties = new Properties();
+      properties.put("connection.name", "EHCACHE:" + this.entityIdentifier);
```
ljacomet: What is the entityIdentifier looking like? the whole URI? only the name part?
Also not sure I like the all caps for Ehcache.

chrisdennis: The entityIdentifier is the just the "name" part.

lorban: I would *at the very least* use constants for the "EHCACHE" identifier as well as the ":" separator.

albinsuresh:
Addressed the review feedback and added a minimal unit test.

ljacomet:
Good to merge once rebased

albinsuresh:
test this please

PR #1033, opened 2016-05-06
updated toolchain path to work on any OS.  Didn't work on windows

gibsong:
For those of you that are unfortunate enough to have to use windows you might need this fix.

### DIFF
--- a/buildSrc/src/main/groovy/MavenToolchain.groovy
+++ b/buildSrc/src/main/groovy/MavenToolchain.groovy
@@ -5,12 +5,12 @@ class MavenToolchain {
 
   static def mavenToolchainDefinitions = {
     String userHome = System.getProperty("user.home");
-    File toolchain = new File(userHome, ".m2/toolchains.xml")
+    File toolchain = new File(userHome, ".m2" + File.separator + "toolchains.xml")
     if (toolchain.isFile()) {
       def xmlSlurper = new XmlSlurper()
       return new XmlSlurper().parse(toolchain)
     } else {
-      return null;
+      throw new Exception("toolchain file not found!!! " + toolchain);
     }
   }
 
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/MavenToolchain.groovy
```diff
@@ -26,11 +26,11 @@ class MavenToolchain {
 
   private static def exe = OperatingSystem.current().isWindows() ? '.exe' : ''
 
-  static def javaHome = { v -> 
+  static def javaHome = { v ->
```
lorban: Please try to avoid all whitespaces-only changes in your PRs.

buildSrc/src/main/groovy/MavenToolchain.groovy
```diff
@@ -26,11 +26,11 @@ class MavenToolchain {
...
     def jdk = toolchains.get(v);
     if (jdk == null) {
 
-      throw new RuntimeException("JDK $v not available - check your toolchains.xml")
+            throw new RuntimeException("JDK $v not available - check your toolchains.xml")
```
lorban: Same as above.

buildSrc/src/main/groovy/MavenToolchain.groovy
ljacomet: Although we have a check style rule ... that seems to no apply to build files :wink: so I am fine with that one staying in.

gibsong: sorry about the unnecessary whites spaces.  I will try to avoid this happening in the future.  Louis, I see you don't mind the one above.  Would you like me to update the one below?

ljacomet: Yes please, and a rebase to make it good to merge

Issue #763, opened 2016-02-18
Server Side CacheManager Entity - Cache Creation


chrisdennis:
Implement a state object to represent the server side existence of a cache.  This includes the retrieval and/or creation of the associated core storage areas as defined in the cache's configuration (see #764).

chrisdennis:
Blocked on various tasks in Voltron/Platform on the core storage service.

chrisdennis:
Terracotta-OSS/terracotta-platform#49 and Terracotta-OSS/terracotta-platform#50 are now complete - this is therefore unblocked.

PR #1039, opened 2016-05-09
Issue #763 Add support for clustered cache objects

cljohnso:
This commit stream adds the ServerStoreProxy and ServerStore along with everything that supports their management (creation, configuration, access, de-access, destruction). 

Notes:

* The commit stream lacks adequate unit tests ... this stream will be updated to include unit tests to ensure each of the management operations functions properly.
* The commit stream introduces a second ServerStore.  This version and the version added in Issue #993 (PR #998) will be reconciled in updates to this commit stream.

As submitted, this PR needs rebasing.  This will be addressed as the above outstanding work is addressed.


cljohnso:
Rebased.

The multiple `ServerStore` versions should be resolved before this commit stream is merged.

cljohnso:
test this please

clustered/build.gradle
```diff
@@ -16,6 +16,7 @@
 
 ext {
     coordinatorVersion = '1.0-SNAPSHOT'
+    offheapResourceVersion = '1.0-SNAPSHOT'
```
ljacomet: This is expected to be the same version as above - that is terracotta-platform version. Can you properly indicate that?

api/src/main/java/org/ehcache/spi/service/PersistableResourceService.java
```diff
@@ -49,6 +50,15 @@
   Collection<ServiceConfiguration<?>> additionalConfigurationsForPool(String alias, ResourcePool pool) throws CachePersistenceException;
 
   /**
+   * Create the persistence space with the given name.
+   *
+   * @param name the name of the persistence context
+   * @param config the configuration for the associated cache
+   * @throws CachePersistenceException if the persistence space cannot be created
+   */
+  void create(String name, CacheConfiguration<?, ?> config) throws CachePersistenceException;
```
ljacomet: So the addition of this method raises a number of questions:
* What is its expected usage? because there are none that I can find.
* However, it made me realise that the method `additionalConfigurationsForPool` in that same interface kind of plays that role ... amongst others. SRP violation possible there. What would be our lifecycle strategy for creating persistence spaces?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -26,29 +28,61 @@
...
+  private static ResourceType defineResource(String name, MemoryUnit unit, long size) {
+    final ResourceType resource = new ResourceType();
+    resource.setName(name);
+    resource.setUnit(unit);
+    resource.setValue(BigInteger.valueOf(size));
+    return resource;
+  }
+
+  private static PassthroughServer server;
+  static {
```
ljacomet: Why the static init and then use of `reset()` in test setup methods?
It would be more clear using a `start` and `stop` called from setup and tear down methods.

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClusteringService.java
```diff
@@ -17,14 +17,47 @@
...
+   * @param cacheIdentifier the {@code ClusteredCacheIdentifier} for the cache for which a
+   *                        {@code ServerStoreProxy} is requested
+   * @param storeConfig the configuration used for the {@link Store} for which the {@code ServerStoreProxy}
+   *                    is requested
+   * @param <K> the cache-exposed key type
+   * @param <V> the cache-exposed value type
+   *
+   * @return a new {@code ServerStoreProxy}
+   */
+  <K, V> ServerStoreProxy<K, V> getServerStoreProxy(ClusteredCacheIdentifier cacheIdentifier, final Store.Configuration<K, V> storeConfig);
```
ljacomet: Why is that handled here and not through the Ehcache client entity?
Maybe I am not aligned with this, but the `ClusteringService` feels more like a connection level thingy than a clustered store specific. That is I would see the potential entry points for getting entities on this API - Ehcache one, Monitoring one and potentially more to come.

ljacomet: Or maybe there is just a naming issue. `ClusteringService` being to general for the role this interface has.

ljacomet:
@cljohnso Could you share that sequence diagram you have been building to describe clustered?

clustered/client/src/main/java/org/ehcache/clustered/client/service/ServerStoreProxyImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Override
+  public Class<K> getKeyType() {
+    return keyType;
+  }
+
+  @Override
+  public Class<V> getValueType() {
+    return valueType;
+  }
```
AbfrmBlr: Why do we need getKeyType and getValueType here, this info is already present in `ServerStoreConfiguration`.

cljohnso: The content of the `ServerStoreConfiguration` is generally used for compatibility support but, given the API recently inherited from `ServerStore` is **not** key/value type dependent, we can probably discard the key & value type fields and their getters.

cljohnso: After chatting with @chrisdennis and @ljacomet, I'm likely to discard the `ServerStoreConfiguration` as well -- there won't be a need to perform a compatibility check against an existing proxy.

api/src/main/java/org/ehcache/spi/service/PersistableResourceService.java
cljohnso: The `create` method is going to be removed from the `PersistableResourceService` interface -- seems it can be local to `DefaultLocalPersistenceService` and _may_ actually be removable entirely.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: I don't know the origin of this -- was something I inherited from @chrisdennis ... maybe he can explain.

cljohnso:
test this please

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: This class and its usage pattern existed before this PR.  It does make sense to alter the usage pattern for this and, if it makes sense (from an effort point of view) for me to change this for the unit tests I'm working on for this commit stream, I'll go ahead and do it then.. Otherwise I'll open an issue for it.

Issue #958, opened 2016-04-11
StringSerializer uses String.getBytes("UTF-8") which will not correctly encode all Strings


chrisdennis:
StringSerializer currently uses `String.getBytes("UTF-8")` to convert the String char sequences in to UTF-8 sequences.  This method however will not encode unpaired surrogates correctly - it expects the char sequence to be a valid UTF-16 sequences.  We can fix this by adopting a strategy similar to that used by DataOutputStream or ObjectOutputStream (see https://docs.oracle.com/javase/8/docs/api/java/io/DataInput.html#modified-utf-8).  See https://jira.terracotta.org/jira/browse/DEV-2800 for a previous occurrence of the same issue in another TC product, and some more explanation on the behavior here.

PR #1045, opened 2016-05-12
Issue #958 Use the same modified UTF-8 encoding as DataInput/OutputSt…

chrisdennis:
…ream for string encoding


ljacomet:
Might break persistent cache storage - waiting for @chrisdennis to test compatibility between the two.

PR #1049, opened 2016-05-13
Clustered store get and put implementation

albinsuresh:
This is my initial work on how to do the chain resolution. `OperationResolver` and `OperationResolverTest` are the pieces of interest here. The `Operation` classes were also modified to handle the resolution in a "functional" way. I took the liberty of adding the remove and putIfAbsent operations to better demonstrate the resolve logic that I have implemented. These can be removed before the final commit. Keeping those wouldn't hurt either.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -88,12 +88,12 @@ public class ClusteredStore<K, V> implements Store<K, V> {
   @Override
   public ValueHolder<V> get(final K key) throws StoreAccessException {
     Chain chain = storeProxy.get(key.hashCode());
-    Map.Entry<Operation<K>, Chain> entry = resolver.resolve(chain, key);
+    ResolvedChain<K> resolvedChain = resolver.resolve(chain, key);
 
-    Chain compactedChain = entry.getValue();
+    Chain compactedChain = resolvedChain.getCompactedChain();
     storeProxy.replaceAtHead(key.hashCode(), chain, compactedChain);
 
-    Operation<K> resolvedOperation = entry.getKey();
+    Operation<K> resolvedOperation = resolvedChain.getResolvedOperation(key);
     V value = null;
     if(resolvedOperation != null && resolvedOperation instanceof KeyValueOperation) {
       value = ((KeyValueOperation<K, V>)resolvedOperation).getValue();
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -16,19 +16,17 @@
 
 package org.ehcache.clustered.client.internal.store;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.ehcache.Cache;
 import org.ehcache.clustered.client.config.ClusteredResourceType;
 import org.ehcache.clustered.client.internal.store.operations.KeyValueOperation;
 import org.ehcache.clustered.client.internal.store.operations.Operation;
-import org.ehcache.clustered.client.internal.store.operations.OperationResolver;
+import org.ehcache.clustered.client.internal.store.operations.ChainResolver;
 import org.ehcache.clustered.client.internal.store.operations.PutOperation;
 import org.ehcache.clustered.client.internal.store.operations.codecs.OperationCodecProvider;
 import org.ehcache.clustered.client.internal.store.operations.codecs.OperationsCodec;
 import org.ehcache.clustered.client.service.ClusteringService;
 import org.ehcache.clustered.client.service.ClusteringService.ClusteredCacheIdentifier;
 import org.ehcache.clustered.common.store.Chain;
-import org.ehcache.clustered.common.store.Element;
 import org.ehcache.config.ResourceType;
 import org.ehcache.core.CacheConfigurationChangeListener;
 import org.ehcache.core.internal.store.StoreSupport;
@@ -77,17 +75,16 @@ public class ClusteredStore<K, V> implements Store<K, V> {
   private final ServerStoreProxy storeProxy;
   private final Store<K, V> underlyingStore;
   private final OperationsCodec<K, V> codec;
-  private final OperationResolver<K, V> resolver;
+  private final ChainResolver<K, V> resolver;
 
   private ClusteredStore(ServerStoreProxy serverStoreProxy, Store<K, V> underlyingStore,
-                         final OperationsCodec<K, V> codec, final OperationResolver<K, V> resolver) {
+                         final OperationsCodec<K, V> codec, final ChainResolver<K, V> resolver) {
     this.storeProxy = serverStoreProxy;
     this.underlyingStore = underlyingStore;
     this.codec = codec;
     this.resolver = resolver;
   }
 
-  @SuppressFBWarnings("DLS_DEAD_LOCAL_STORE")
   @Override
   public ValueHolder<V> get(final K key) throws StoreAccessException {
     Chain chain = storeProxy.get(key.hashCode());
@@ -101,9 +98,7 @@ public class ClusteredStore<K, V> implements Store<K, V> {
     if(resolvedOperation != null && resolvedOperation instanceof KeyValueOperation) {
       value = ((KeyValueOperation<K, V>)resolvedOperation).getValue();
     }
-//    return new ClusteredValueHolder<V>(value);
-
-    return underlyingStore.get(key);
+    return new ClusteredValueHolder<V>(value);
   }
 
   @Override
@@ -117,9 +112,7 @@ public class ClusteredStore<K, V> implements Store<K, V> {
     PutOperation<K, V> operation = new PutOperation<K, V>(key, value);
     ByteBuffer payload = codec.encode(operation);
     storeProxy.append(key.hashCode(), payload);
-//    return PutStatus.PUT; // TODO: 17/05/16 Do we need to differentiate between different statuses?
-
-    return underlyingStore.put(key, value);
+    return PutStatus.PUT; // TODO: 17/05/16 Do we need to differentiate between different statuses?
   }
 
   @Override
@@ -266,7 +259,7 @@ public class ClusteredStore<K, V> implements Store<K, V> {
       OperationCodecProvider<K, V> codecProvider =
           new OperationCodecProvider<K, V>(storeConfig.getKeySerializer(), storeConfig.getValueSerializer());
       OperationsCodec<K, V> codec = new OperationsCodec<K, V>(codecProvider);
-      OperationResolver<K, V> resolver = new OperationResolver<K, V>(codec);
+      ChainResolver<K, V> resolver = new ChainResolver<K, V>(codec);
       Store<K, V> store = new ClusteredStore<K, V>(serverStoreProxy, underlyingStore, codec, resolver);
 
       createdStores.put(store, underlyingStoreProvider);
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -16,10 +16,19 @@
 
 package org.ehcache.clustered.client.internal.store;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.ehcache.Cache;
 import org.ehcache.clustered.client.config.ClusteredResourceType;
+import org.ehcache.clustered.client.internal.store.operations.KeyValueOperation;
+import org.ehcache.clustered.client.internal.store.operations.Operation;
+import org.ehcache.clustered.client.internal.store.operations.OperationResolver;
+import org.ehcache.clustered.client.internal.store.operations.PutOperation;
+import org.ehcache.clustered.client.internal.store.operations.codecs.OperationCodecProvider;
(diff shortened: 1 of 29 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -231,7 +242,11 @@ public void clear() throws StoreAccessException {
 
       ClusteredCacheIdentifier cacheId = findSingletonAmongst(ClusteredCacheIdentifier.class, (Object[]) serviceConfigs);
       ServerStoreProxy serverStoreProxy = clusteringService.getServerStoreProxy(cacheId, storeConfig);
-      Store<K, V> store = new ClusteredStore<K, V>(serverStoreProxy, underlyingStore);
+      OperationCodecProvider<K, V> codecProvider = null;
+//          new OperationCodecProvider<K, V>(storeConfig.getKeySerializer(), storeConfig.getValueSerializer());
```
albinsuresh: Ignore this for now

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/OperationResolver.java
```diff
@@ -0,0 +1,95 @@
...
+    Chain newChain = null;  // TODO: 13/05/16  build the chain using the list of elements and the resolvedElement
+    return newChain;
+  }
+
+  /*
+    Iterate the list of operations in the reverse order and find the first PUT or REMOVE from the end.
+    The entries prior to that can be ignored as a PUT or REMOVE overrides the effect of the execution
+    of all those preceding operations.
+   */
+  Operation<K> resolve(List<Operation<K>> opList) {
```
lorban: Transforming the `Chain` into a `List<Operation>` before processing the list looks fairly expensive. Ideally, the `resolve` method should work directly with the `Chain`.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/OperationResolver.java
```diff
@@ -0,0 +1,95 @@
...
+    if(opList.isEmpty()) {
+      return null;
+    }
+
+    ListIterator<Operation<K>> opsIterator = opList.listIterator(opList.size()-1);
+    Operation<K> current = opsIterator.next();
+    while (current.getOpCode() != OperationCode.PUT &&
```
lorban: That's a nice optimization but it should be better abstracted, ie: the `Operation` itself should be able to tell if it is going to ignore the previous value or not instead of having to maintain a set of `if (opcode != ZZZ)` conditions here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/OperationCode.java
```diff
@@ -18,11 +18,15 @@
...
 import org.ehcache.clustered.client.internal.store.operations.codecs.OperationCodecFactory;
 
+import com.sun.org.apache.regexp.internal.RE;
+
 import static org.ehcache.clustered.client.internal.store.operations.codecs.OperationCodecFactory.PutOperationCodecFactory;
 
 public enum OperationCode {
 
-  PUT((byte)1, new PutOperationCodecFactory());
+  PUT((byte)1, new PutOperationCodecFactory()),
```
lorban: Now that I look at this I wonder if there aren't too many unneeded abstraction layers here. I don't think we need a per-operation codec: one codec should suffice for all operations as the encoding / decoding is always only about the opcode, the key and the value. Having a single `OperationCodec` class would certainly simplify things a lot.

albinsuresh:
Lazy decoding is one optimization that I could do while processing the `Chain`. That is, I don't really need to deserialize the value part of the operations as the value is not really used in the resolution process, at least for now(Though this will change with the addition of value based CAS operations). But I'm gonna delay these optimizations for a later point unless instructed otherwise.

albinsuresh:
Some major refactoring was done to improve the abstractions:
- Interfaces were added to represent operations(both key based and key-value based operations)
- Since the encoding/decoding logic for every key-value based operations are similar the common logic was extracted out to a base class to simplify the concrete codec implementation for different operations like put, putIfAbsent etc(all key-value based operations). The same was done for key based operations like remove.

The resolve logic was simplified to do the resolution as the chain is processed sequentially. The optimization done in the previous iteration; by reverse iterating the operations and starting the resolution only from the first put/remove from the end; has been rolled back and will be revisited later.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/OperationResolver.java
```diff
@@ -0,0 +1,73 @@
...
+   * @param chain a heterogeneous {@code Chain}
+   * @param key a key
+   * @return an entry with the resolved operation for the provided key as the key
+   * and the compacted chain as the value
+   * @throws ClassNotFoundException
+   */
+  @SuppressFBWarnings({ "DLS_DEAD_LOCAL_STORE", "UC_USELESS_OBJECT" })
+  public Map.Entry<Operation<K>, Chain> resolve(Chain chain, K key) {
+    Operation<K> resolvedOperation = null;
+    List<Element> elements = new ArrayList<Element>();
```
albinsuresh: Once the Chain builder is available, I won't have to create this list and then convert that to a chain. I can use the builder directly to build the compacted chain.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/RemoveOperation.java
```diff
@@ -0,0 +1,39 @@
...
+    return OperationCode.REMOVE;
+  }
+
+  @Override
+  public Operation<K> apply(final Operation<K> kOperation) {
+    return null;
+  }
+
+  @Override
+  public String toString() {
```
lorban: Useless override

impl/src/main/java/org/ehcache/impl/internal/store/AbstractValueHolder.java
```diff
@@ -29,6 +29,8 @@
  */
 public abstract class AbstractValueHolder<V> implements Store.ValueHolder<V> {
 
+  public static final TimeUnit TIME_UNIT = TimeUnit.MILLISECONDS;
```
lorban: An abstract value holder cannot have a notion of default time unit; this can only be done in non-abstract implementations, i.e.: this should move to `ClusteredValueHolder`.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredValueHolder.java
```diff
@@ -0,0 +1,41 @@
...
+  private final V value;
+
+  public ClusteredValueHolder(V value) {
+    super(0, 0);
+    this.value = value;
+  }
+
+  @Override
+  protected TimeUnit nativeTimeUnit() {
+    return TIME_UNIT;
```
lorban: that constant should be declared in this class

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/BaseKeyValueOperationCodec.java
```diff
@@ -0,0 +1,105 @@
...
+    buffer.position(buffer.limit());
+    buffer.limit(buffer.capacity());
+    int valueSize = buffer.getInt();
+    buffer.limit(buffer.position() + valueSize);
+    ByteBuffer valueBlob = buffer.slice();
+
+    try {
+      return newOperation(keySerializer.read(keyBlob), valueSerializer.read(valueBlob));
+    } catch (ClassNotFoundException e) {
+      throw new SerializerException(e);
```
lorban: The `SerializerException` is completely unrelated to the operation codecs - you should throw something else

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/BaseOperationCodec.java
```diff
@@ -0,0 +1,70 @@
...
+  protected abstract Operation<K> newOperation(K key);
+
+  @Override
+  public Operation<K> decode(final ByteBuffer buffer) {
+    OperationCode opCode = OperationCode.valueOf(buffer.get());
+    validateOperation(opCode);
+    try {
+      return newOperation(keySerializer.read(buffer));
+    } catch (ClassNotFoundException e) {
+      throw new SerializerException(e);
```
lorban: Again, the SerializerException is completely unrelated to the operation codecs - you should throw something else

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/BaseOperationCodec.java
```diff
@@ -0,0 +1,70 @@
...
+import org.ehcache.spi.serialization.Serializer;
+import org.ehcache.spi.serialization.SerializerException;
+
+import java.nio.ByteBuffer;
+
+import static org.ehcache.clustered.client.internal.store.operations.codecs.OperationsCodec.BYTE_SIZE_BYTES;
+
+public abstract class BaseOperationCodec<K> implements OperationCodec<K> {
+
+  protected final Serializer<K> keySerializer;
```
lorban: why is that field `protected`?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/codecs/BaseOperationCodec.java
```diff
@@ -0,0 +1,70 @@
...
+    OperationCode opCode = OperationCode.valueOf(buffer.get());
+    validateOperation(opCode);
+    try {
+      return newOperation(keySerializer.read(buffer));
+    } catch (ClassNotFoundException e) {
+      throw new SerializerException(e);
+    }
+  }
+
+  protected void validateOperation(OperationCode code) {
```
lorban: why is that method `protected`? I doubt this should be overridable.

lorban:
The `BaseOperationCodec` and `BaseKeyValueOperationCodec` abstractions look bogus to me.

First, I don't get why you make `OperationCodec` implement `Function` as this doesn't bring anything special on the table. I would at the very least override the `apply` definition to rename the parameter to `previousOperation` or something like that as implementors look a bit confusing at first look with the parameter simply called `operation`.

Second, the only `BaseOperationCodec` that will ever be is for remove. I don't think this justifies the extra mechanism. `RemoveOperationCodec` could subclass `BaseKeyValueOperationCodec` and return a hardcoded `null` as its value which would make the code more straightforward.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
```diff
@@ -0,0 +1,43 @@
...
+package org.ehcache.clustered.client.internal.store.operations;
+
+/**
+ * Base class that represents {@link org.ehcache.Cache} operations
+ * that involves a key and a value.
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+public abstract class BaseKeyValueOperation<K, V> extends BaseOperation<K> implements KeyValueOperation<K, V> {
```
lorban: The generic type should actually be `BaseKeyValueOperation<K, ClusteredValueHolder<V>>` which means that the `OperationCodec`s should also be responsible for encoding/decoding the `ValueHolder` metadata on top of the key and the value.

lorban:
I would also name the `OperationCodec.getValue()` method differently, something like `getResult()` as this method is supposed to return the resulting value that the store contains and I find `getValue()` confusing.

albinsuresh:
For now `RemoveOperationCodec` is the only subclass of `BaseOperationCodec`. But I created that additional base class thinking about the future addition of the Metadata update operation which also would be a key-based operation. Both those operations *could* extend `BaseKeyValueOperationCodec` with hard null values. I just thought that this additional abstraction is cleaner.

Is the comment only about having multiple versions of codecs for key based operations and key-value based operations? Or does it apply to the multiple versions of operations; `Operation<K>` and `KeyValueOperation<K,V>`; as well? That is, are you suggesting that we have a single abstraction to represent all operations: key based, key-value based and key-value-value based(like the 3-arg remove) operations?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
albinsuresh: Since the initial implementation of clutered-tier only cache is not gonna use any of the metadata we'll keep the codecs handle keys and values only. This change can be done once we start using the metadata.

albinsuresh:
Integrated store proxy with clustered store.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -66,16 +74,31 @@
...
+    this.codec = codec;
+    this.resolver = resolver;
   }
 
   @Override
   public ValueHolder<V> get(final K key) throws StoreAccessException {
-    // TODO: Make appropriate ServerStoreProxy call
-    return underlyingStore.get(key);
+    Chain chain = storeProxy.get(key.hashCode());
+    Map.Entry<Operation<K>, Chain> entry = resolver.resolve(chain, key);
```
ljacomet: I had missed this usage of Map.Entry as a tuple. I am pretty sure there is a type hiding there which would make sense ... or an indication that something is not right and needs refactoring.

albinsuresh:
Introduced the `ResolvedChain` interface to better represent the result of a chain resolution. Fixed some failing tests as well.

Issue #940, opened 2016-04-06
Clustered Ehcache Operations Network Messaging Research Spike


chrisdennis:
Research spike on the necessary message set for the clustered Ehcache operation message set. This should be a relatively simple exercise, but I want the focus here to be on deciding how the messaging is done (proxy or native).

AbfrmBlr:
I think we should use native approach for this.
The `voltron-proxy`, as it stands today is not in a state where it can used since the server to client communication has changed since the proxy was designed.
It would be a considerable effort to incorporate that and then make Ehcache's Server Side entity a proxied one.

Instead we can define a set of operations for the client side proxy of ServerStore.

General Structure of Message over the would like this :

Client invoking Server Side api :

`[Cache Identifier] [Type of Operation] [Operation Arguments]`

`Cache Identifier` :- Unique identifier for the cache(gets associated with cache during cache creation)

`Type of Operation` :- The kind of operation (encoded in bits) which client wants to do on server store.
					There are four operations `get`, `getAndAppend`, `append` & `replaceAtHead` which can be denoted in 2 bits.

`Operation Arguments` :- arguments required for the respective operation. The Codec should be able to
					   decode operation arguments based on the operation provided.

Server Side Response in lieu of the Request sent :

Get and GetAndAppend require response to be sent back to the client. 
Though the `InvocationBuilder` at platform level allows client to decide to wait/or to not wait for server to 
ack the message at different stages(recived, complete), some operations require response as well to continue.
All those responses would be sent as `EhcacheEntityResponse` to the client.  

`[Cache Identifier] [Chain of Elements in binary]`

`Chain of Elements in binary` :- As of now both the operations need Chain as the response back

AbfrmBlr:
ReplaceAtHead as the special case :-

Replace Request :-
The request message needs special attention as this has to send `Chain` of `Element`s back to the server. 
All other operations require just handling of long and byte buffers which is simple to encode/decode.
In this case since replace has to be followed up by a get or getAndAppend, the expectation for client is to send the 
original `Chain`. Since server already has that knowledge of payload associated with the chain, client can just send an equivalent to
original chain which just has sequence numbers rather than the whole chain.
Here there is a need of a builder kind of construct to be made available at client side so the client is able to send a equivalent chain. 

Replace Response :-
Based on the Passive replication approach taken `replaceAtHead` will also require `boolean` as the response which can be easily accommodated as well.

AbfrmBlr:
[...]

PR #1050, opened 2016-05-13
Issue 994

AbfrmBlr:
This can be merged if found ok.
The work that follows this PR is 

- Op specific Codec integration in `EhcacheCodec`
- tests
- `ClusteredStore` clean up regarding to `underlyingStore`.(not very sure about it though)

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
@@ -108,7 +108,7 @@ public class EhcacheClientEntity implements Entity {
   public EhcacheEntityResponse invoke(EhcacheEntityMessage message, boolean waitUntilReplicated) throws Exception {
     InvokeFuture<EhcacheEntityResponse> result = null;
     if (waitUntilReplicated) {
-      result = endpoint.beginInvoke().message(message).ackCompleted().invoke();
+      result = endpoint.beginInvoke().message(message).replicate(true).ackCompleted().invoke();
     } else {
       result = endpoint.beginInvoke().message(message).invoke();
     }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
@@ -55,7 +55,7 @@ public class EhcacheClientEntity implements Entity {
 
   public void validate(ServerSideConfiguration config) throws IllegalArgumentException {
     try {
-      invoke(EhcacheEntityMessage.validate(config));
+      invoke(EhcacheEntityMessage.validate(config), false);
     } catch (IllegalArgumentException e) {
       throw e;
     } catch (Throwable e) {
@@ -65,7 +65,7 @@ public class EhcacheClientEntity implements Entity {
 
   public void configure(ServerSideConfiguration config) throws IllegalStateException {
     try {
-      invoke(EhcacheEntityMessage.configure(config));
+      invoke(EhcacheEntityMessage.configure(config), false);
     } catch (IllegalStateException e) {
       throw e;
     } catch (Throwable e) {
@@ -75,7 +75,7 @@ public class EhcacheClientEntity implements Entity {
 
   public void createCache(String name, ServerStoreConfiguration serverStoreConfiguration) throws CachePersistenceException {
     try {
-      invoke(EhcacheEntityMessage.createServerStore(name, serverStoreConfiguration));
+      invoke(EhcacheEntityMessage.createServerStore(name, serverStoreConfiguration), false);
     } catch (Exception e) {
       throw unwrapException(e, CachePersistenceException.class);
     }
@@ -83,7 +83,7 @@ public class EhcacheClientEntity implements Entity {
 
   public void validateCache(String name, ServerStoreConfiguration serverStoreConfiguration) throws CachePersistenceException {
     try {
-      invoke(EhcacheEntityMessage.validateServerStore(name , serverStoreConfiguration));
+      invoke(EhcacheEntityMessage.validateServerStore(name , serverStoreConfiguration), false);
     } catch (Exception e) {
       throw unwrapException(e, CachePersistenceException.class);
     }
@@ -91,7 +91,7 @@ public class EhcacheClientEntity implements Entity {
 
   public void releaseCache(String name) throws CachePersistenceException {
     try {
-      invoke(EhcacheEntityMessage.releaseServerStore(name));
+      invoke(EhcacheEntityMessage.releaseServerStore(name), false);
     } catch (Exception e) {
       throw unwrapException(e, CachePersistenceException.class);
     }
@@ -99,37 +99,20 @@ public class EhcacheClientEntity implements Entity {
 
   public void destroyCache(String name) throws CachePersistenceException {
     try {
-      invoke(EhcacheEntityMessage.destroyServerStore(name));
+      invoke(EhcacheEntityMessage.destroyServerStore(name), false);
     } catch (Exception e) {
       throw unwrapException(e, CachePersistenceException.class);
     }
   }
 
-  public EhcacheEntityResponse invoke(EhcacheEntityMessage message) throws Exception {
-    InvokeFuture<EhcacheEntityResponse> result = endpoint.beginInvoke().message(message).invoke();
-    boolean interrupted = false;
-    try {
-      while (true) {
-        try {
-          EhcacheEntityResponse response = result.get();
-          if (Type.FAILURE.equals(response.getType())) {
-            throw ((Failure) response).getCause();
-          } else {
-            return response;
-          }
-        } catch (InterruptedException e) {
-          interrupted = true;
-        }
-      }
-    } finally {
-      if (interrupted) {
-        Thread.currentThread().interrupt();
-      }
+  public EhcacheEntityResponse invoke(EhcacheEntityMessage message, boolean waitUntilReplicated) throws Exception {
+    InvokeFuture<EhcacheEntityResponse> result = null;
+    if (waitUntilReplicated) {
+      result = endpoint.beginInvoke().message(message).ackCompleted().invoke();
+    } else {
+      result = endpoint.beginInvoke().message(message).invoke();
     }
-  }
 
-  public EhcacheEntityResponse invokeCompleted(EhcacheEntityMessage message) throws Exception {
-    InvokeFuture<EhcacheEntityResponse> result = endpoint.beginInvoke().message(message).ackCompleted().invoke();
     boolean interrupted = false;
     try {
       while (true) {
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
@@ -105,8 +105,31 @@ public class EhcacheClientEntity implements Entity {
     }
   }
 
-  private EhcacheEntityResponse invoke(EhcacheEntityMessage message) throws Exception {
+  public EhcacheEntityResponse invoke(EhcacheEntityMessage message) throws Exception {
     InvokeFuture<EhcacheEntityResponse> result = endpoint.beginInvoke().message(message).invoke();
+    boolean interrupted = false;
+    try {
+      while (true) {
+        try {
+          EhcacheEntityResponse response = result.get();
+          if (Type.FAILURE.equals(response.getType())) {
+            throw ((Failure) response).getCause();
+          } else {
+            return response;
+          }
+        } catch (InterruptedException e) {
+          interrupted = true;
+        }
+      }
+    } finally {
+      if (interrupted) {
+        Thread.currentThread().interrupt();
+      }
+    }
+  }
+
+  public EhcacheEntityResponse invokeCompleted(EhcacheEntityMessage message) throws Exception {
+    InvokeFuture<EhcacheEntityResponse> result = endpoint.beginInvoke().message(message).ackCompleted().invoke();
     boolean interrupted = false;
     try {
       while (true) {
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ChainBuilder.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ChainBuilder.java
@@ -16,13 +16,10 @@
 package org.ehcache.clustered.client.internal.store;
 
 import org.ehcache.clustered.common.store.Chain;
-import org.ehcache.clustered.common.store.Element;
+import org.ehcache.clustered.common.store.Util;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -30,51 +27,27 @@ import java.util.List;
  */
 public class ChainBuilder {
 
-  private List<Element> elements = new ArrayList<Element>();
+  private List<ByteBuffer> buffers = new ArrayList<ByteBuffer>();
 
   public ChainBuilder() {
-
   }
 
-  private ChainBuilder(List<Element> elements) {
-    this.elements = elements;
+  private ChainBuilder(List<ByteBuffer> buffers) {
+    this.buffers = buffers;
   }
 
+  //TODO: optimize this & make this mutable
   public ChainBuilder add(final ByteBuffer payload) {
-    this.elements.add(new Element() {
-      @Override
-      public ByteBuffer getPayload() {
-        return payload;
-      }
-    });
-    return new ChainBuilder(this.elements);
+    List<ByteBuffer> newList = new ArrayList<ByteBuffer>();
+    newList.addAll(this.buffers);
+    newList.add(payload);
+    return new ChainBuilder(newList);
   }
 
   public Chain build() {
-    return new DummyChain(elements);
+    ByteBuffer[] elements = new ByteBuffer[buffers.size()];
+    buffers.toArray(elements);
+    return Util.getChain(false, elements);
   }
 
-  private static class DummyChain implements Chain {
-
-    private final List<Element> elementList;
-
-    private DummyChain(List<Element> elements) {
-      elementList = elements;
-    }
-
-    @Override
-    public Iterator<Element> reverseIterator() {
-      throw new UnsupportedOperationException("Not Supported");
-    }
-
-    @Override
-    public boolean isEmpty() {
-      return elementList.isEmpty();
-    }
-
-    @Override
-    public Iterator<Element> iterator() {
-      return elementList.iterator();
-    }
-  }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ChainBuilder.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ChainBuilder.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.client.internal.store;
+
+import org.ehcache.clustered.common.store.Chain;
+import org.ehcache.clustered.common.store.Element;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Builds {@link Chain}s
+ */
+public class ChainBuilder {
+
+  private List<Element> elements = new ArrayList<Element>();
+
+  public ChainBuilder() {
+
+  }
+
+  private ChainBuilder(List<Element> elements) {
+    this.elements = elements;
+  }
+
+  public ChainBuilder add(final ByteBuffer payload) {
+    this.elements.add(new Element() {
+      @Override
+      public ByteBuffer getPayload() {
+        return payload;
+      }
+    });
+    return new ChainBuilder(this.elements);
+  }
+
+  public Chain build() {
+    return new DummyChain(elements);
+  }
+
+  private static class DummyChain implements Chain {
+
+    private final List<Element> elementList;
+
+    private DummyChain(List<Element> elements) {
+      elementList = elements;
+    }
+
+    @Override
+    public Iterator<Element> reverseIterator() {
+      throw new UnsupportedOperationException("Not Supported");
+    }
+
+    @Override
+    public boolean isEmpty() {
+      return elementList.isEmpty();
+    }
+
+    @Override
+    public Iterator<Element> iterator() {
+      return elementList.iterator();
+    }
+  }
+}
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
@@ -21,11 +21,8 @@ import org.ehcache.clustered.common.messages.EhcacheEntityMessage;
 import org.ehcache.clustered.common.messages.EhcacheEntityResponse;
 import org.ehcache.clustered.common.store.Chain;
 import org.ehcache.clustered.common.store.ServerStore;
-import org.ehcache.core.spi.store.StoreAccessException;
 
 import java.nio.ByteBuffer;
-
-import static org.ehcache.clustered.common.Util.unwrapException;
 
 /**
  * Provides client-side access to the services of a {@code ServerStore}.
@@ -51,15 +48,18 @@ public class ServerStoreProxy implements ServerStore {
 
   @Override
   public Chain get(long key) {
+    EhcacheEntityResponse response;
     try {
-      EhcacheEntityResponse response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key), false);
-      if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
-        return ((EhcacheEntityResponse.GetResponse)response).getChain();
-      }
+      response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key), false);
     } catch (Exception e) {
-      unwrapException(e, StoreAccessException.class);
+      throw new ServerStoreProxyException(e);
     }
-    return null;
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for get operation was invalid : " +
+                                          response != null ? response.getType().toString() : "null message");
+    }
   }
 
   @Override
@@ -67,29 +67,33 @@ public class ServerStoreProxy implements ServerStore {
     try {
       entity.invoke(EhcacheEntityMessage.appendOperation(cacheId, key, payLoad), true);
     } catch (Exception e) {
-      unwrapException(e, StoreAccessException.class);
+      throw new ServerStoreProxyException(e);
     }
   }
 
   @Override
   public Chain getAndAppend(long key, ByteBuffer payLoad) {
+    EhcacheEntityResponse response;
     try {
-      EhcacheEntityResponse response = entity.invoke(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad), true);
-      if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
-        return ((EhcacheEntityResponse.GetResponse)response).getChain();
-      }
+      response = entity.invoke(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad), true);
     } catch (Exception e) {
-      unwrapException(e, StoreAccessException.class);
+      throw new ServerStoreProxyException(e);
     }
-    return null;
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for getAndAppend operation was invalid : " +
+                                          response != null ? response.getType().toString() : "null message");
+    }
   }
 
   @Override
   public void replaceAtHead(long key, Chain expect, Chain update) {
+    // TODO: Optimize this method to just send sequences for expect Chain
     try {
       entity.invoke(EhcacheEntityMessage.replaceAtHeadOperation(cacheId, key, expect, update), true);
     } catch (Exception e) {
-      unwrapException(e, StoreAccessException.class);
+      throw new ServerStoreProxyException(e);
     }
   }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
@@ -51,9 +51,8 @@ public class ServerStoreProxy implements ServerStore {
 
   @Override
   public Chain get(long key) {
-    EhcacheEntityResponse response = null;
     try {
-      response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key));
+      EhcacheEntityResponse response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key), false);
       if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
         return ((EhcacheEntityResponse.GetResponse)response).getChain();
       }
@@ -66,7 +65,7 @@ public class ServerStoreProxy implements ServerStore {
   @Override
   public void append(long key, ByteBuffer payLoad) {
     try {
-      entity.invokeCompleted(EhcacheEntityMessage.appendOperation(cacheId, key, payLoad));
+      entity.invoke(EhcacheEntityMessage.appendOperation(cacheId, key, payLoad), true);
     } catch (Exception e) {
       unwrapException(e, StoreAccessException.class);
     }
@@ -74,9 +73,8 @@ public class ServerStoreProxy implements ServerStore {
 
   @Override
   public Chain getAndAppend(long key, ByteBuffer payLoad) {
-    EhcacheEntityResponse response = null;
     try {
-      response = entity.invokeCompleted(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad));
+      EhcacheEntityResponse response = entity.invoke(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad), true);
       if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
         return ((EhcacheEntityResponse.GetResponse)response).getChain();
       }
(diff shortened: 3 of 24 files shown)


AbfrmBlr:
One more thing builders for `Chain` and `Element`s, as I think they would be required.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public ServerStoreOp operation() {
+      return ServerStoreOp.GET;
+    }
+  }
+
+  public static class GetAndAppendMessage extends ServerStoreOpMessage {
+
+    private final ByteBuffer payload;
```
lorban: `ByteBuffer` is not serializable, while `EntityMessage`s have to AFAICT.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public ByteBuffer getPayload() {
+      return payload;
+    }
+  }
+
+  public static class AppendMessage extends ServerStoreOpMessage {
+
+    private final ByteBuffer payload;
```
lorban: `ByteBuffer` is not serializable.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public ByteBuffer getPayload() {
+      return payload;
+    }
+  }
+
+  public static class ReplaceAtHeadMessage extends ServerStoreOpMessage {
+
+    private final Chain expect;
+    private final Chain update;
```
lorban: `Chain` is most probably not serializable.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -127,4 +127,27 @@ private EhcacheEntityResponse invoke(EhcacheEntityMessage message) throws Except
       }
     }
   }
+
+  public EhcacheEntityResponse invokeCompleted(EhcacheEntityMessage message) throws Exception {
```
lorban: I get the idea of this `invokeCompleted` method but I have two problems with it:
1) the name is quite poor
2) it duplicates 99% of the `invoke` method's code

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +51,47 @@ public String getCacheId() {
 
   @Override
   public Chain get(long key) {
-    // TODO: Implement ServerStoreProxyImpl.get
-    throw new UnsupportedOperationException("ServerStoreProxyImpl.get not implemented");
+    EhcacheEntityResponse response = null;
```
lorban: Why is this variable outside the `try` block?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +51,47 @@ public String getCacheId() {
...
+    try {
+      response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key));
+      if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+        return ((EhcacheEntityResponse.GetResponse)response).getChain();
+      }
+    } catch (Exception e) {
+      unwrapException(e, StoreAccessException.class);
```
lorban: the result of `unwrapException` must still be thrown

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheEntityResponse.java
```diff
@@ -75,4 +77,26 @@ public Exception getCause() {
     }
   }
 
+  public static GetResponse response(Chain chain) {
+    return new GetResponse(chain);
+  }
+
+  public static class GetResponse extends EhcacheEntityResponse {
+
+    private final Chain chain;
```
lorban: `Chain` is probably not serializable.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -127,4 +127,27 @@ private EhcacheEntityResponse invoke(EhcacheEntityMessage message) throws Except
...
+    InvokeFuture<EhcacheEntityResponse> result = endpoint.beginInvoke().message(message).ackCompleted().invoke();
+    boolean interrupted = false;
+    try {
+      while (true) {
+        try {
+          EhcacheEntityResponse response = result.get();
+          if (Type.FAILURE.equals(response.getType())) {
+            throw ((Failure) response).getCause();
```
chrisdennis: What if the cause is an InterruptedException?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+public abstract class ServerStoreOpMessage extends EhcacheEntityMessage {
+  public enum ServerStoreOp {
+    GET,
+    GETANDAPPEND,
+    APPEND,
+    REPLACE
+  }
+
+  private final String cacheId;
```
chrisdennis: Do we really need to send the whole cacheId every time?  Even if we don't want to optimize this now we should probably record this inefficiency in an issue.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: I know this, but it need not be serializable at all.
As I said the codec would extract the appropriate bytes and send it to the server.
Neither this nor `Chain` needs to be serializable. The changes in codec will take care of it as explained n #940

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: Same as earlier

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: Same as earlier

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
AbfrmBlr: yup missed that

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: no we do not need to send this all the time, we can surely optimize this now also

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheEntityResponse.java
AbfrmBlr: As explained in first reply

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
AbfrmBlr: `InvokeFuture.get` throws it.

chrisdennis: Thats nice... but what if the cause of the remote operation failure is an InterruptedException... the loop will never terminate.  You need to remove the response processing outside the try catch don't you?

AbfrmBlr: Then we will be getting an `EntityException` from the platform for this invocation and that is why we have `throws`. This behaves the same way `invoke` behaves and I thought it was a decided way of handling endpoint invokes.

lorban:
This is looking alright for now, but I wouldn't merge that without the EhcacheCodec changes nor tests.

AbfrmBlr:
I have updated the PR just to report/make clear the progress on the issue, & to provide the overview about codec approach as well.
The codecs are added and integrated for all message types.
The code still considerable refactoring and tests are on the way as well.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -55,7 +55,7 @@ public void close() {
 
   public void validate(ServerSideConfiguration config) throws IllegalArgumentException {
     try {
-      invoke(EhcacheEntityMessage.validate(config));
+      invoke(EhcacheEntityMessage.validate(config), false);
```
ljacomet: Why this change and the others in this file?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
ljacomet: Could this be addressed?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
ljacomet: I believe I understand the change now

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +51,45 @@ public String getCacheId() {
 
   @Override
   public Chain get(long key) {
-    // TODO: Implement ServerStoreProxyImpl.get
-    throw new UnsupportedOperationException("ServerStoreProxyImpl.get not implemented");
+    try {
+      EhcacheEntityResponse response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key), false);
+      if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+        return ((EhcacheEntityResponse.GetResponse)response).getChain();
+      }
```
ljacomet: else??

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +51,45 @@ public String getCacheId() {
...
 
   @Override
   public Chain getAndAppend(long key, ByteBuffer payLoad) {
-    // TODO: Implement ServerStoreProxyImpl.getAndAppend
-    throw new UnsupportedOperationException("ServerStoreProxyImpl.getAndAppend not implemented");
+    try {
+      EhcacheEntityResponse response = entity.invoke(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad), true);
+      if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+        return ((EhcacheEntityResponse.GetResponse)response).getChain();
+      }
```
ljacomet: else??

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -226,6 +224,41 @@ public void loadExisting() {
...
+      }
+    } catch (Exception e) {
+      return EhcacheEntityResponse.failure(e);
+    }
+  }
+
+  private EhcacheEntityResponse invokeServerStoreOperation(ServerStoreOpMessage message) {
+    try {
+      switch (message.operation()) {
+        case GET: return EhcacheEntityResponse.response(stores.get(message.getCacheId()).get(message.getKey()));
```
ljacomet: All the `stores.get` should be wrapped so that they throw a meaningful exception in case there is no such store. Otherwise the caller will just get a useless NPE that happened on the other side of the network ...

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheCodec.java
```diff
@@ -37,46 +34,30 @@
 
   @Override
   public byte[] encodeMessage(EhcacheEntityMessage message) {
-    return marshall(message);
+    if (message.getType() == EhcacheEntityMessage.Type.LIFECYCLE_OP) {
+      return LifeCycleOpCodec.encode((LifecycleMessage) message);
```
ljacomet: I fail to see benefit in having static codecs though I have plenty of downsides coming to mind. Care to chime in?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
AbfrmBlr: if the response is null then the get returns null. I think rather it should throw Exception since even for absent keys the caller will get an empty chain, not null

AbfrmBlr:
PR updated. Its done with all codecs and tests.
I have added a passthrough unit test `ServerStoreProxyTest` to verify client-server interaction as well w.r.t to store interactions only.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
ljacomet: And also what happens if the type is not the expected one?

ljacomet:
Build fails because it times out.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
AbfrmBlr: I meant for that also it should throw, because both cases should never happen, So if ti does not falls in if clause it should be an Exception

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpCodec.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  private static final byte MSG_TYPE_OFFSET = 1;
+  private static final byte STORE_OP_CODE_OFFSET = 1;
+  private static final byte CACHE_ID_LEN_OFFSET = 4;
+  private static final byte KEY_OFFSET = 8;
+  private static final byte CHAIN_LEN_OFFSET = 4;
+
+  public static byte[] encode(ServerStoreOpMessage message) {
```
AbfrmBlr: Had a discussion with @albinsuresh @rkavanap @itsroshan & sairam and no one seems to like this switch/case thing. This needs refactoring.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ChainBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  private ChainBuilder(List<Element> elements) {
+    this.elements = elements;
+  }
+
+  public ChainBuilder add(final ByteBuffer payload) {
+    List<Element> newList = new ArrayList<Element>();
+    newList.addAll(this.elements);
+    newList.add(new Element() {
```
rkavanap: copying and rebuilding the list everytime an element is added to the chain? Maybe chains are not long to see the adverse impact of this now. But when long chains are build this could be seen as inefficient. Something to think about in future refactorings.

lorban: @rkavanap : this is our standard builder pattern: the builder itself is immutable and always returns a modified copy of itself. though I agree that since this builder isn't meant for end-user consumption, many array allocations and copies might be overkill so we may consider using a mutable builder in this case.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +48,50 @@ public String getCacheId() {
...
+    EhcacheEntityResponse response;
+    try {
+      response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key), false);
+    } catch (Exception e) {
+      throw new ServerStoreProxyException(e);
+    }
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for get operation was invalid.");
```
lorban: More details please: at least include the type of response you got in the exception's message.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +48,50 @@ public String getCacheId() {
...
+    EhcacheEntityResponse response;
+    try {
+      response = entity.invoke(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad), true);
+    } catch (Exception e) {
+      throw new ServerStoreProxyException(e);
+    }
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for getAndAppend operation was invalid.");
```
lorban: Same as above: more details in the exception's message please.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ *
+ */
+public class ChainCodec {
+
+  private static final byte NON_SEQUNCED_CHAIN = 0;
```
lorban: typo.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+            @Override
+            public ByteBuffer getPayload() {
+              return elementPayload.duplicate();
+            }
+          });
+        }
+      }
+    }
+
+    return new Chain() {
```
lorban: It's not the first time I see such inline immutable on-heap `Chain` implementation (in the builder too?), please consider extracting that to an external class and re-use it.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+  private static final byte NON_SEQUNCED_CHAIN = 0;
+  private static final byte SEQUENCED_CHAIN = 1;
+  private static final byte SEQ_NUM_OFFSET = 8;
+  private static final byte ELEMENT_PAYLOAD_OFFSET = 4;
+
+  public static byte[] encode(Chain chain) {
+    ByteBuffer msg = null;
+    boolean firstIteration = true ;
+    for (Element element : chain) {
+      if (firstIteration) {
```
lorban: I'd rather use a classic `for (int i = 0; ...)` and test for `if (i == 0)` rather than a `boolean firstIteration`

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+          buffer.put(SEQUENCED_CHAIN);
+        } else {
+          buffer.put(NON_SEQUNCED_CHAIN);
+        }
+        buffer.flip();
+        msg = combine(buffer, encodeElement(element));
+        continue;
+      }
+      if (msg == null) {
+        throw new IllegalArgumentException("Message cannot be null");
```
lorban: Does that mean we don't support encoding empty `Chain`s?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+      }
+      msg = combine(msg, encodeElement(element));
+
+    }
+    return msg != null ? msg.array() : new byte[0];
```
lorban: Does that mean we support encoding empty `Chain`s?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.nio.ByteBuffer;
+
+/**
+ */
+public abstract class ServerStoreOpMessage extends EhcacheEntityMessage {
+  public enum ServerStoreOp {
+
+    GET((byte) 0),
+    GETANDAPPEND((byte) 1),
```
albinsuresh: GET_AND_APPEND?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+      if (firstIteration) {
+        firstIteration = false;
+        ByteBuffer buffer = ByteBuffer.allocate(1);
+        if (element instanceof SequencedElement) {
+          buffer.put(SEQUENCED_CHAIN);
+        } else {
+          buffer.put(NON_SEQUNCED_CHAIN);
+        }
+        buffer.flip();
+        msg = combine(buffer, encodeElement(element));
```
lorban: Encoding a `Chain` seems to require quite a few `ByteBuffer` allocations and copies which looks fairly expensive. I wonder if pre-iterating the `Chain` once to calculate the final `byte[]` size wouldn't be a lot cheaper despite the fact that the cain would need to be iterated twice.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheEntityResponse.java
```diff
@@ -15,20 +15,30 @@
...
-public abstract class EhcacheEntityResponse implements EntityResponse, Serializable {
-  private static final long serialVersionUID = 4559645892362408528L;
+public abstract class EhcacheEntityResponse implements EntityResponse {
 
   public enum Type {
-    SUCCESS,
-    FAILURE;
+    SUCCESS((byte) 0),
+    FAILURE((byte) 1),
+    GET_RESPONSE((byte) 2);
```
lorban: I would call that differently as not only `get` can return a response.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
AbfrmBlr: Its supported see `ChainCodecTest#testEmptyChain`

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ResponseCodec.java
```diff
@@ -0,0 +1,70 @@
...
+      case GET_RESPONSE:
+        GetResponse getResponse = (GetResponse)response;
+        byte[] encodedChain = ChainCodec.encode(getResponse.getChain());
+        int chainLen = encodedChain.length;
+        buffer = ByteBuffer.allocate(OP_CODE_OFFSET + chainLen);
+        buffer.put(EhcacheEntityResponse.Type.GET_RESPONSE.getOpCode());
+        buffer.put(encodedChain);
+        return buffer.array();
+      default:
+        throw new UnsupportedOperationException("The operation is not supported");
```
lorban: please include more details about the `EhcacheEntityResponse`.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ResponseCodec.java
```diff
@@ -0,0 +1,70 @@
...
+    switch (opCode) {
+      case 0:
+        return EhcacheEntityResponse.success();
+      case 1:
+        Exception exception = (Exception)LifeCycleOpCodec.unmarshall(payArr);
+        return EhcacheEntityResponse.failure(exception);
+      case 2:
+        return EhcacheEntityResponse.response(ChainCodec.decode(payArr));
+      default:
+        throw new UnsupportedOperationException("The operation is not supported");
```
lorban: please report the unsupported `opCode` value in the exception's message.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ResponseCodec.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  public static EhcacheEntityResponse decode(byte[] payload) {
+    ByteBuffer buffer = ByteBuffer.wrap(payload);
+    byte opCode = buffer.get();
+    byte[] payArr = new byte[buffer.remaining()];
+    buffer.get(payArr);
+    switch (opCode) {
+      case 0:
```
lorban: I'm sure there's a better way than using hardcoded literals here.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpCodec.java
```diff
@@ -0,0 +1,130 @@
...
+        int expectChainLen = replaceBuf.getInt();
+        byte[] encodedExpectChain = new byte[expectChainLen];
+        replaceBuf.get(encodedExpectChain);
+        int updateChainLen = replaceBuf.getInt();
+        byte[] encodedUpdateChain = new byte[updateChainLen];
+        replaceBuf.get(encodedUpdateChain);
+        return EhcacheEntityMessage.replaceAtHeadOperation(cacheId, key, ChainCodec.decode(encodedExpectChain),
+            ChainCodec.decode(encodedUpdateChain));
+      default:
+        throw new UnsupportedOperationException("This operation code is not supported" + opCode);
```
lorban: Please pay attention to the message's formatting: you miss at least a space and ideally a space-colon-space (`' : '`) before concatenating the opcode.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpCodec.java
```diff
@@ -0,0 +1,130 @@
...
+    int cacheIdLen = msg.getInt();
+    byte[] idArr = new byte[cacheIdLen];
+    msg.get(idArr);
+    String cacheId = new String(idArr, UTF_8);
+    long key = msg.getLong();
+    byte opCode = msg.get();
+    byte[] remaining = new byte[msg.remaining()];
+    msg.get(remaining);
+    switch (opCode) {
+      case 0:
```
lorban: Just say no to hardcoded literals.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
lorban: Renaming that to `GET_AND_APPEND` would make that name more readable.

clustered/common/src/test/java/org/ehcache/clustered/common/messages/Util.java
```diff
@@ -0,0 +1,110 @@
...
+      @Override
+      public ByteBuffer getPayload() {
+        return payload.asReadOnlyBuffer();
+      }
+    };
+  }
+
+  public static Chain getChain(boolean isSequenced, ByteBuffer... buffers) {
+    final List<Element> elements = new ArrayList<Element>();
+    final AtomicLong counter = new AtomicLong();
```
lorban: Why using a `AtomicLong` instead of a plain simple `long`?

clustered/common/src/test/java/org/ehcache/clustered/common/messages/Util.java
```diff
@@ -0,0 +1,110 @@
...
+    for (final ByteBuffer buffer : buffers) {
+      if (isSequenced) {
+        elements.add(getElement(counter.incrementAndGet(), buffer));
+      } else {
+        elements.add(getElement(buffer));
+      }
+
+    }
+    return new Chain() {
```
lorban: Yet another immutable on-heap `Chain` implementation.

clustered/common/src/test/java/org/ehcache/clustered/common/messages/Util.java
```diff
@@ -0,0 +1,110 @@
...
+
+      @Override
+      public Iterator<Element> iterator() {
+        return list.iterator();
+      }
+    };
+  }
+
+  private static Element getElement(final long sequence, final ByteBuffer payload) {
+    return new SequencedElement() {
```
lorban: It seems like `SequencedElement` and `Element` also deserve their own immutable on-heap class.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -27,10 +28,12 @@
 
   private final ServerStoreConfiguration storeConfiguration;
   private final PageSource pageSource;
+  private final OffHeapServerStore store;
 
   ServerStoreImpl(ServerStoreConfiguration storeConfiguration, PageSource pageSource) {
     this.storeConfiguration = storeConfiguration;
     this.pageSource = pageSource;
+    this.store = new OffHeapServerStore(pageSource, 16);
```
lorban: I'd at least use a constant instead of that hardcoded `16`.

clustered/common/src/test/java/org/ehcache/clustered/common/messages/Util.java
AbfrmBlr: I use it as counter, and that needs to final since it gets incremented in anonymous class. I don's see it as any problem

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ServerStoreProxyTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+  private static void assertChainHas(Chain chain, long... payLoads) {
+    Iterator<Element> elements = chain.iterator();
+    for (long payLoad : payLoads) {
+      assertThat(readPayLoad(elements.next().getPayload()), is(Long.valueOf(payLoad)));
+    }
+    assertThat(elements.hasNext(), is(false));
+  }
+
+  private static Chain getChain(ByteBuffer... buffers) {
```
rkavanap: Maybe some test cases that uses the ChainBuilder here?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
AbfrmBlr: But how would I know Chain's length.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private static ByteBuffer encodeElement(Element element) {
+    if (element instanceof SequencedElement) {
+      ByteBuffer buffer = ByteBuffer.allocate(SEQ_NUM_OFFSET + ELEMENT_PAYLOAD_OFFSET + element.getPayload().remaining());
+      buffer.putLong(((SequencedElement)element).getSequenceNumber());
+      buffer.putInt(element.getPayload().remaining());
+      buffer.put(element.getPayload());
+      buffer.flip();
+      return buffer;
+    }
```
rkavanap: else could be better here..also some duplicate code can be avoided (from 139-142) if that part of the code is made common. Can we also abstract the allocate difference between sequenced and normal elements and use a template pattern here?. Again possibly a future refactoring.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ChainCodec.java
```diff
@@ -0,0 +1,150 @@
...
+    return msg != null ? msg.array() : new byte[0];
+  }
+
+  public static Chain decode(byte[] payload) {
+    final List<Element> elements = new ArrayList<Element>();
+    if (payload.length != 0) {
+      ByteBuffer buffer = ByteBuffer.wrap(payload);
+      boolean isSequenced = buffer.get() == 1;
+      if (isSequenced) {
+        while (buffer.hasRemaining()) {
```
rkavanap: again looks like refactoring to a template pattern (between sequenced and normal) could help in reducing code duplicate and keeping code more readable

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheCodec.java
```diff
@@ -37,46 +34,30 @@
...
+      return ServerStoreOpCodec.encode((ServerStoreOpMessage) message);
+    }
   }
 
   @Override
   public EhcacheEntityMessage decodeMessage(byte[] payload) throws MessageCodecException {
-    return (EhcacheEntityMessage) unmarshall(payload);
+    ByteBuffer payloadBuf = ByteBuffer.wrap(payload);
+    if (payloadBuf.get() == 1) {
+      return LifeCycleOpCodec.decode(payloadBuf);
```
rkavanap: this seems to lack symmetry w.r.t how the first byte signifying the type is coded. Also, payloadBuf.get() == 1 seems to be hardcoded and affecting readability and increasing brittleness,.can we have a const or enum here that specifies the type?.

clustered/common/src/test/java/org/ehcache/clustered/common/messages/Util.java
AbfrmBlr: Not for `SequencedElement`

AbfrmBlr:
rebased and reviews taken care of

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +48,52 @@ public String getCacheId() {
...
+    EhcacheEntityResponse response;
+    try {
+      response = entity.invoke(EhcacheEntityMessage.getOperation(cacheId, key), false);
+    } catch (Exception e) {
+      throw new ServerStoreProxyException(e);
+    }
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for get operation was invalid : " +
```
ljacomet: There is a NPE hiding here: `+` is higher priority than `!=`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -44,25 +48,52 @@ public String getCacheId() {
...
+    EhcacheEntityResponse response;
+    try {
+      response = entity.invoke(EhcacheEntityMessage.getAndAppendOperation(cacheId, key, payLoad), true);
+    } catch (Exception e) {
+      throw new ServerStoreProxyException(e);
+    }
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for getAndAppend operation was invalid : " +
```
ljacomet: Same as above

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
AbfrmBlr: No there is no NPE here. `?` operator is the highest precedence.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
AbfrmBlr: niether here

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
AbfrmBlr: That's stupid. parentheses are missing

Issue #1047, opened 2016-05-12
NullPointerException when omitting sizeof related elements in XML


ljacomet:
The following XML inside a cache declaration will cause a NPE:

```xml
<ehcache:heap-store-settings>
  <ehcache:max-object-graph-size>500</ehcache:max-object-graph-size>
</ehcache:heap-store-settings>
```
because the XSD optional element `max-object-size` is not declared.

This was caused by the changes introduced in #703.

PR #1053, opened 2016-05-13
Issue #1047 NullPointerException when omitting sizeof related elements in XML

chrisdennis:


ljacomet:
Looks good but does not address the cache manager level it seems

chrisdennis:
Right you are... working on it now.

Issue #1007, opened 2016-04-27
Server Side Caching: invalidation


chrisdennis:

rkavanap:
# Design Notes on Cache Invalidation

There are three main problem areas to consider when broadcasting or multicasting a change (such as an invalidation) to clients:

  **Problem 1. Which clients to notify the change?**
  
  Should we broadcast the change to all clients connected to the store or should we multicast to only those clients that has expressed an interest in that particular mapping. A client's interest is known by the server when the client is doing a get on the store for a given hash. 
  
  For multicast, we will need to track the client list for a given hash. Several space efficient techniques are possible, including techniques that may result is some false positives. Some of these techniques are considered in this design notes.
    
  **Problem 2. Cache consistency configuration**
  
  For strong consistency, the client's caching tier that issued the change must be notified after all the other clients has acknowledged the change notification to the server. The mutation API on the client (such as Put) must wait for this notification before returning an acknowledgement to its client for the mutation. For eventual consistency, the client issueing the change does not have to wait. Also for eventual consistency, there is no need for server to get acknowledgements from the client. It can be a pure 'fire and forget' interface as the underlying transport offers guaranteed delivery as long as the client is up and connected. If the client is disconnected, it will invalidate all its mappings anyway. 
  
  Due to the above, the server side must know whether the store is configured for strong or eventual consistency and act accordingly. Also since the server side deals with only blobs we need to decide on what data to send when the change is broadcasted. Should we send just the hash(key) and expect that the client somehow knows how to derive the key(s) from hash. Or should we extract the key from the blob and send?. Or should we extract the entire blob and send when broadcasting the change. 
  
  All these questions are answered in the design notes after making some assumptions.
  
  **Problem 3: Constrained resource usage on the Server**
  
  The change broadcast communication from the server to the clients must be completely asynchronous to avoid holding costly server's stage threads waiting for responses from the client. Voltron mechanisms are available to achieve this asynchronous communication.
  
  Voltron dictates that all communication from server MUST be asynchronous. This is to avoid server threads waiting and consuming resources of the stage affecting the overall scalability and responsiveness of the system.
  
  **Problem 4: Active/Passive failover impact for the invalidations.**
  
  How does a active/passive failover impact the invalidations. This is still work in progress.

Considering the effort required to handle all the above, the work will be broken into two iterations, Iteration 1 and Iteration 2.

**Iteration 1 Scope:**

[...]

PR #1061, opened 2016-05-18
 #1007: implement CHMv8.removeByHash()

lorban:
Prerequisite of #1007, but obviously only a small subset of it. Could be merged separately of the remaining work or not, at convenience.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/concurrent/ConcurrentHashMap.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/concurrent/ConcurrentHashMap.java
@@ -30,6 +30,7 @@ import java.lang.reflect.Type;
 import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/ehcache/impl/internal/concurrent/ConcurrentHashMap.java
```diff
@@ -1163,6 +1164,40 @@ else if (t.removeTreeNode(p))
...
+              break;
+          else if (f.hash == MOVED)
+              tab = helpTransfer(tab, f);
+          else {
+              Map<K, V> nodes = null;
+              synchronized (f) {
+                  if (tabAt(tab, i) == f) {
+                      nodes = nodesAt(f);
+                      setTabAt(tab, i, null);
+                      invalidated.putAll(nodes);
```
ljacomet: Two map creations + putAll. I believe the code should pay more attention to that ... even if it means using the map in a side effect way as an argument.

lorban: ok, let me take care of that

ljacomet:
Looks good to me - needs a rebase though

PR #1062, opened 2016-05-18
Issue #763 EhcacheActiveEntity unit tests and corrections

cljohnso:
This PR introduces unit tests for `EhcacheActiveEntity` and corrections for issues exposed by @lorban.

Additional unit tests to follow in additional PRs.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
@@ -53,7 +53,6 @@ import org.terracotta.offheapstore.paging.Page;
 import org.terracotta.offheapstore.paging.PageSource;
 import org.terracotta.offheapstore.paging.UpfrontAllocatingPageSource;
 
-import static java.util.Collections.unmodifiableMap;
 import static org.terracotta.offheapstore.util.MemoryUnit.GIGABYTES;
 import static org.terracotta.offheapstore.util.MemoryUnit.MEGABYTES;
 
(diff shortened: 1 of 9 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -333,6 +379,17 @@ private EhcacheEntityResponse validate(ValidateCacheManager message) {
...
   private EhcacheEntityResponse createServerStore(ClientDescriptor clientDescriptor, CreateServerStore createServerStore) {
+    ClientState clientState = this.clientStateMap.get(clientDescriptor);
+    if (clientState == null) {
+      return failure(new IllegalStateException("Client " + clientDescriptor + " is not connected to the Clustered Store Manager"));
+    }
+    if (!clientState.isAttached()) {
+      return failure(new IllegalStateException("Client " + clientDescriptor + " is not attached to the Clustered Store Manager"));
+    }
+    if (!isConfigured()) {
+      return failure(new IllegalStateException("Clustered Store Manager is not configured"));
```
GaryWKeim: Could you include the clientDescriptor in this message like, "while attempting to create server-side storage for client {}"

cljohnso: This exception is actually transmitted back to the requesting client making the client ID somewhat redundant.  (It really didn't need to be provided in the exception messages above it.)  We should probably log these in the server before returning them and the client ID _needs_ to be in the log messages.

In general, we need  to "normalize" server- and client-side logging and exception messages to include information needed to track fault contributions back to clients (in particular).

I'll adjust this one (and the others like it) as an interim move until we take care of the more general logging/exception issue.

cljohnso: @ljacomet @chrisdennis @lorban  Opinion requested ...

It's clear that the client needs to be identified when logging -- at least server-side.  Is there any benefit to identifying the client in exception detail messages in exceptions that are sent back to the client?  Right now, some messages identify the client and some do not.  We need to pick something and normalized on it.

cljohnso:
Rebased ...

cljohnso:
Rebased with adjustments for PR #1050

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
ljacomet: I recommend opening an issue for this topic and having the discussion there.

cljohnso: Opened issue #1067.

PR #1063, opened 2016-05-18
Remove dependency on M&M messaging stuff

mathieucarbou:


anthonydahanne:
LGTM

management/src/main/java/org/ehcache/management/CollectorService.java
```diff
@@ -29,6 +27,7 @@
  * @author Mathieu Carbou
  */
 public interface CollectorService extends StatisticCollector, Service {
-
```
GaryWKeim: Thank you.

PR #1064, opened 2016-05-18
:construction: Fix #1030 Update handling of over size mappings

### LINKED ISSUE #1030: Introduce InvalidationValve
```java
interface InvalidationValve<K> {
  void invalidateAll();
}
```
Used for allowing `AuthoritativeTier` to request `CachingTier` to flush all its content.

Outcome from #919 and #941

ljacomet:
* Introduce sub interface of EvictionAdvisor that can be disabled
* Wire said interface in all the offheap related types that knew
about EvictionAdvisor
* Simplify OverSizeMappingException handling following what's
currently done in offheap
* Trigger the InvalidationValve and disable EvictionAdvisor

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
@@ -24,7 +24,6 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.Lock;
 
 import org.ehcache.Cache;
 import org.ehcache.config.EvictionAdvisor;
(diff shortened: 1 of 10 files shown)


ljacomet:
test this please

impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -1117,6 +1058,29 @@ private void notifyInvalidation(final K key, final ValueHolder<V> p) {
     }
   }
 
+  private OffHeapValueHolder<V> computeWithRetry(K key, BiFunction<K, OffHeapValueHolder<V>, OffHeapValueHolder<V>> computeFunction, boolean fault) throws StoreAccessException {
```
lorban: I wonder why `computeWithRetry` only tries to retry once on `OversizeMappingException` while the old code used to rety in a `while (true)` loop. Can you elaborate?

ljacomet: Answered that in #1030 - let me know if more detail is needed.

lorban: ok, makes sense - thanks

PR #1068, opened 2016-05-20
Issue 994 cleanup of underlying heap store

AbfrmBlr:
This PR just removes underlying heap store.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -74,13 +74,12 @@ import static org.ehcache.core.internal.service.ServiceLocator.findSingletonAmon
 // TODO: Remove underlyingStore when ServerStore/ServerStoreProxy is complete
 public class ClusteredStore<K, V> implements Store<K, V> {
 
-  private final ServerStoreProxy storeProxy;
   private final OperationsCodec<K, V> codec;
   private final ChainResolver<K, V> resolver;
 
-  private ClusteredStore(ServerStoreProxy serverStoreProxy,
-                         final OperationsCodec<K, V> codec, final ChainResolver<K, V> resolver) {
-    this.storeProxy = serverStoreProxy;
+  private volatile ServerStoreProxy storeProxy;
+
+  private ClusteredStore(final OperationsCodec<K, V> codec, final ChainResolver<K, V> resolver) {
     this.codec = codec;
     this.resolver = resolver;
   }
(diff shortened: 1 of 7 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -249,66 +247,36 @@ public void clear() throws StoreAccessException {
...
 
     @Override
     public void initStore(final Store<?, ?> resource) {
-      Store.Provider underlyingStoreProvider = createdStores.get(resource);
-      if (underlyingStoreProvider == null) {
-        throw new IllegalArgumentException("Given store is not managed by this provider: " + resource);
-      }
-
-      underlyingStoreProvider.initStore(((ClusteredStore)resource).underlyingStore);
+      //no op
```
ljacomet: Impossible - init cannot be a no-op since release will get rid of the `storeProxy`. Lifecycle needs to be preserved here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -249,66 +247,36 @@ public void clear() throws StoreAccessException {
...
     @Override
     public void releaseStore(final Store<?, ?> resource) {
-      Store.Provider underlyingStoreProvider = createdStores.remove(resource);
-      if (underlyingStoreProvider == null) {
-        throw new IllegalArgumentException("Given store is not managed by this provider: " + resource);
-      }
-
       ClusteredStore clusteredStore = (ClusteredStore)resource;
-      underlyingStoreProvider.releaseStore(clusteredStore.underlyingStore);
       this.clusteringService.releaseServerStoreProxy(clusteredStore.storeProxy);
```
ljacomet: All other `Store.Provider` do some level of accounting - same should be done here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
AbfrmBlr: I thought of it, but the only way to to do that is to preserve info about cacheID, serviceconfig and storeconfig, until created stores are initialized. Does this idea make sense ?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
AbfrmBlr: yes

ljacomet:
Looks good once rebased

AbfrmBlr:
Its rebased now.

ljacomet:
I missed that you did not remove the comments at the top of `ClusteredStore`. Could you open a follow-up PR for that?

PR #1032, opened 2016-05-05
#1002: User facing doc for clustered ehcache creation

kedar031:
Please also review the terminology used in the context

PR #1069, opened 2016-05-20
Issue #763 Rewrite UnitTestConnectionService

cljohnso:
This commit moves `UnitTestConnectionService` to support test-specific server configuration -- for all intents and purposes, `UnitTestConnectionService` is rewritten.  All existing uses of `UnitTestConnectionService` are changed to make use of the updates.

### DIFF
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
@@ -27,6 +27,7 @@ import org.ehcache.config.builders.CacheManagerBuilder;
 import org.ehcache.config.builders.ResourcePoolsBuilder;
 import org.ehcache.config.units.EntryUnit;
 import org.ehcache.config.units.MemoryUnit;
+import org.junit.After;
 import org.junit.Test;
 
 import java.net.URI;
@@ -45,7 +46,16 @@ public class GettingStarted {
 
   @Before
   public void resetPassthroughServer() throws Exception {
-    UnitTestConnectionService.reset();
+    UnitTestConnectionService.add("http://localhost:9510/my-application?auto-create",
+        new UnitTestConnectionService.PassthroughServerBuilder()
+            .resource("primary-server-resource", 64, MemoryUnit.MB)
+            .resource("secondary-server-resource", 64, MemoryUnit.MB)
+            .build());
+  }
+
+  @After
+  public void removePassthroughServer() throws Exception {
+    UnitTestConnectionService.remove("http://localhost:9510/my-application?auto-create");
   }
 
   @Test
@@ -84,7 +94,7 @@ public class GettingStarted {
     // tag::clusteredCacheManagerWithServerSideConfigExample
     final CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder
             = CacheManagerBuilder.newCacheManagerBuilder()
-            .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://example.com:9540/my-application?auto-create"))
+            .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://localhost:9510/my-application?auto-create"))
                     .defaultServerResource("primary-server-resource")
                     .resourcePool("resource-pool-a", 128, MemoryUnit.B));
     final PersistentCacheManager cacheManager = clusteredCacheManagerBuilder.build(false);
@@ -108,7 +118,7 @@ public class GettingStarted {
     // tag::clusteredCacheManagerWithServerSideConfigExample
     final CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder
         = CacheManagerBuilder.newCacheManagerBuilder()
-        .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://example.com:9540/my-application?auto-create"))
+        .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://localhost:9510/my-application?auto-create"))
             .defaultServerResource("primary-server-resource"));
     final PersistentCacheManager cacheManager = clusteredCacheManagerBuilder.build(false);
     cacheManager.init();
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
@@ -17,17 +17,37 @@
 package org.ehcache.clustered.client.internal;
 
 
+import java.io.IOException;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.math.BigInteger;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 import org.ehcache.clustered.server.EhcacheServerEntityService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terracotta.connection.Connection;
 import org.terracotta.connection.ConnectionException;
 import org.terracotta.connection.ConnectionService;
 import org.terracotta.consensus.entity.CoordinationServerEntityService;
 import org.terracotta.consensus.entity.client.ClientCoordinationEntityService;
+import org.terracotta.entity.EntityClientService;
+import org.terracotta.entity.EntityMessage;
+import org.terracotta.entity.EntityResponse;
+import org.terracotta.entity.ServerEntityService;
+import org.terracotta.entity.ServiceProvider;
+import org.terracotta.entity.ServiceProviderConfiguration;
 import org.terracotta.offheapresource.OffHeapResourcesConfiguration;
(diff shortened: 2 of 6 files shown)


clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -84,7 +94,7 @@ public void clusteredCacheManagerWithDynamicallyAddedCacheExample() throws Excep
     // tag::clusteredCacheManagerWithServerSideConfigExample
     final CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder
             = CacheManagerBuilder.newCacheManagerBuilder()
-            .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://example.com:9540/my-application?auto-create"))
+            .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://localhost:9510/my-application?auto-create"))
```
cljohnso: Changed to `localhost` to conform with comments in PR #1032

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -108,7 +118,7 @@ public void clusteredCachePutGet() throws Exception {
     // tag::clusteredCacheManagerWithServerSideConfigExample
     final CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder
         = CacheManagerBuilder.newCacheManagerBuilder()
-        .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://example.com:9540/my-application?auto-create"))
+        .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("http://localhost:9510/my-application?auto-create"))
```
cljohnso: Changed to `localhost` to conform with comments in PR #1032

cljohnso:
`UnitTestConnectionService` is changed to use a registration system as suggested in PR #1039.  While working on this, I noticed that `PassthroughServer` was leaving a thread after stop -- after discussing the issue with @jd0-sag opened Terracotta-OSS/terracotta-apis#90.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
-    List<ResourceType> resourcesList = resources.getResource();
-    resourcesList.add(defineResource("defaultResource", MemoryUnit.MB, 128L));
-    resourcesList.add(defineResource("primary-server-resource", MemoryUnit.MB, 64L));
-    resourcesList.add(defineResource("secondary-server-resource", MemoryUnit.MB, 64L));
-    OffHeapResourcesConfiguration ohrpConfig = new OffHeapResourcesConfiguration(resources);
-    newServer.registerServiceProvider(new OffHeapResourcesProvider(), ohrpConfig);
-
-    newServer.start();
-    return newServer;
+  private static final Map<URI, ServerDescriptor> servers = new HashMap<URI, ServerDescriptor>();
```
chrisdennis: static so should be capitalized.

cljohnso: True -- was static and lower-case before I got to it but I'll change it.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
+  /**
+   * Creates a "key" {@code URI} by dropping the <i>user-info</i>, <i>path</i>, <i>query</i>, and <i>fragment</i>
+   * portions of the {@code URI}.
+   *
+   * @param requestURI the {@code URI} for which the key is to be generated
+   *
+   * @return a {@code URI} instance with the <i>user-info</i>, <i>path</i>, <i>query</i>, and <i>fragment</i> discarded
+   */
+  private static URI createKey(URI requestURI) {
+    try {
```
chrisdennis: This seems weird... it feels like a detail of the way Ehcache uses URIs is bleeding in to the test framework.  I'd rather see this be done in a test class (statically in code).

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
-    return resource;
+  /**
+   * Adds a {@link PassthroughServer} if, and only if, a mapping for the URI supplied does not
+   * already exist.  The server is started as it is added.
+   *
+   * @param uri the URI, in string form, for the server; only the <i>scheme</i>, <i>host</i>, and <i>port</i>
+   *            contribute to the server identification
+   * @param server the {@code PassthroughServer} instance to use for connections to {@code uri}
+   */
+  public static void add(String uri, PassthroughServer server) {
```
chrisdennis: Why the need for this method?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
+    add(URI.create(uri), server);
```
chrisdennis: This should really be `new URI(uri)` if we're going to have this method shouldn't it?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
+  public static PassthroughServer remove(URI uri) {
+    URI keyURI = createKey(uri);
+    ServerDescriptor serverDescriptor = servers.remove(keyURI);
+    if (serverDescriptor != null) {
+      serverDescriptor.server.stop();
+      LOGGER.info("Stopped PassthroughServer at {}", keyURI);
+
+      for (Connection connection : serverDescriptor.getConnections().keySet()) {
+        try {
+          LOGGER.warn("Force close {}", formatConnectionId(connection));
```
chrisdennis: Shouldn't this be the pass through servers responsibility?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
+          connection.close();
+        } catch (IOException e) {
```
chrisdennis: Logging?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: Actually, this reflects what a "real" server would observe -- the path, query, and fragment aren't really "seen" by the connection infrastructure.  And this *is* an Ehcache class ...

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
+    } else {
+      return Collections.emptyList();
+    }
   }
 
   @Override
   public boolean handlesURI(URI uri) {
-    String hostName = uri.getHost();
-    return hostName.endsWith("example.com") || hostName.equals("localhost");
+    return servers.containsKey(createKey(uri));
```
chrisdennis: Not sure we should be calling createKey here.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: Are you questioning this versus the URI version?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -17,90 +17,384 @@
...
   }
 
   @Override
   public Connection connect(URI uri, Properties properties) throws ConnectionException {
-    connectionProperties = properties;
-    return server().connectNewClient();
+
+    URI keyURI = createKey(uri);
```
chrisdennis: This call seems suspicious too.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: No, it's not -- based on a discussion with @jd0-sag this morning.  I don't really care if the connections are closed or not except that not closing them leaves a thread-per-connection lying about.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: Don't really care since all I'm interest in is dropping a running thread.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
chrisdennis: If this were a real server the connections would get torn down from the server end though... shouldn't the passthrough server behave the same way?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
cljohnso: The servers are registered using the result of `createKey` ... the lookup needs to be that way as well.

PR #1070, opened 2016-05-20
Not a PR: Issue #1007: clustered store invalidation

lorban:
A few notes:
 * config for strong/eventual has been added
 * calling the Montreal Design's invalidation valve hasn't been done yet since the clustered store cannot participate in tiering yet
 * there are System.out calls all around the mechanism, they'll obviously be replaced with logs but for now it's handy to see the traces end-to-end when running a test
 * Strong/eventual discrimination (and the entire invalidation mechanism) only makes sense when clients are using the clustered store with tiering, otherwise strong is no different than eventual. Should that config be part of the tiered store instead of the clustered store? And should we add some extra logic that doesn't fire the invalidation when tiering isn't used (i.e.: 'untiered' consistency)?


lorban:
test this please

PR #1071, opened 2016-05-24
Issue 1079 ConcurrentWeakIdentityHashMap may get null value for mutable Key

zhanhb:
The hash code should be System.identityHashCode, detail http://hg.openjdk.java.net/jdk9/dev/jdk/file/d27a73776c05/src/java.base/share/classes/java/util/IdentityHashMap.java

### DIFF
--- a/core/src/main/java/org/ehcache/core/internal/util/ConcurrentWeakIdentityHashMap.java
+++ b/core/src/main/java/org/ehcache/core/internal/util/ConcurrentWeakIdentityHashMap.java
@@ -191,6 +191,9 @@ public class ConcurrentWeakIdentityHashMap<K, V> implements ConcurrentMap<K, V> 
 
     private WeakReference(final T referent, final ReferenceQueue<? super T> q) {
       super(referent, q);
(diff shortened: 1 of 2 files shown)


terracotta-org:

ljacomet:
Hello,

Thanks for this PR.
Before accepting it, we need you to sign a contributor agreement - see http://www.ehcache.org/community/contribute.html

Also, could you raise an issue so that this commit can be linked to it.
And expect a comment (soon) on the strategy we have in place for fixing issues that apply to multiple branches, as is the case with yours here.

ljacomet:
Regarding the bug fixing approach - have a look at https://github.com/ehcache/ehcache3/wiki/dev.git#initiate-a-pull-request

zhanhb:
I agree . The source I committed is under no license, You can make any change on it.

zhanhb:
Maybe the code should be
```java
super(referent, q);
if(referent == null) throw new NullPointerException();
hashCode = System.identityHashCode(referent);
```

Issue #994, opened 2016-04-19
Server Store Proxy Impl


chrisdennis:
Implementation of the ServerStoreProxy interface, including the necessary Voltron messaging code (marshall and unmarshall) to interface with the server side ServerStore implementation.

AbfrmBlr:
Had a discussion with @albinsuresh @rkavanap @itsroshan, @ljacomet  & sairam and no one seems to like this switch/case thing. This needs refactoring and statics needs to go.Will be taken in further PRs.

PR #1072, opened 2016-05-24
Remove TODO from ClusteredStore for proxy #994

AbfrmBlr:
Removing the comment as well, since all the operations are having specific `TODO`s.
And we have a `TODO` for tiering as well


albinsuresh:
Should we merge this commit separately? As part of my #1006 refactoring work I'm already touching the `ClusteredStore` class. I could include this removal as well in my PR.

ljacomet:
Obsolete through PR #1073

PR #1073, opened 2016-05-24
Issue 1006 refactoring

albinsuresh:
The following refactoring was done.
- `Operation` interface has been modified to hold values as well. This eliminated the need for `KeyValueOperation` interface.
- Removed all the codec classes.
- All the encoding/decoding logic has been moved to individual operation classes.
- Added conditional remove and replace operations.

I added the conditional replace operation to be sure that the new type hierarchy would work for this 3-arg operation as well. Then there were only the replace and conditional remove operations left. So I added those as well. But I have not written the clustered store implementations using these operation classes.


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -87,15 +73,17 @@ private ClusteredStore(final OperationsCodec<K, V> codec, final ChainResolver<K,
...
-    Operation<K> resolvedOperation = resolvedChain.getResolvedOperation(key);
-    V value = null;
-    if(resolvedOperation != null && resolvedOperation instanceof KeyValueOperation) {
-      value = ((KeyValueOperation<K, V>)resolvedOperation).getValue();
+      Operation<K, V> resolvedOperation = resolvedChain.getResolvedOperation(key);
+      if(resolvedOperation != null) {
+        value = resolvedOperation.getValue();
+      }
     }
     return new ClusteredValueHolder<V>(value);
```
ljacomet: With your changes, this is no longer correct. A `ValueHolder` should never contain a `null`. You need to return `null` instead.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseOperation.java
```diff
@@ -16,26 +16,140 @@
...
+  protected K key;
+  protected V value;
+
+  public BaseOperation() {
+    this.key = null;
+    this.value = null;
+  }
 
   public BaseOperation(final K key) {
+    this(key, null);
```
ljacomet: This constructor makes no sense, as it will always throw the NPE because the value is null.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ConditionalReplaceOperation.java
```diff
@@ -0,0 +1,156 @@
...
+  public ConditionalReplaceOperation(final K key, final V oldValue, final V value) {
+    super(key, value);
+    if(oldValue == null) {
+      throw new NullPointerException("Old value can not be null");
+    }
+    this.oldValue = oldValue;
+  }
+
+  ConditionalReplaceOperation(final ByteBuffer buffer, final Serializer<K> keySerializer, final Serializer<V> valueSerializer) {
+    OperationCode opCode = OperationCode.valueOf(buffer.get());
```
ljacomet: The fact that you have to write all this here, plus the default no arg constructor on `BaseOperation` plus the protected fields are an indication that something is wrong in the abstraction you defined.
I think a lot of these problems would go away if the BaseOperation did not deal at all with values and that is left to subclasses. The the fact that you have one or two values can be properly encapsulated.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/Operation.java
```diff
@@ -17,10 +17,17 @@
...
-public interface Operation<K> extends Function<Operation<K>, Operation<K>> {
+import java.nio.ByteBuffer;
+
+public interface Operation<K, V> extends Function<Operation<K, V>, Operation<K, V>> {
 
   OperationCode getOpCode();
 
   K getKey();
+
+  V getValue();
```
ljacomet: As discussed at one point with @lorban, this name feels wrong. We want to get to the operation result here, not the value it was constructed with.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/PutIfAbsentOperation.java
```diff
@@ -32,10 +40,11 @@ public OperationCode getOpCode() {
    * for the same key.
    */
   @Override
-  public Operation<K> apply(final Operation<K> previousOperation) {
+  public Operation<K, V> apply(final Operation<K, V> previousOperation) {
     if(previousOperation == null) {
       return this;
     } else {
+      assertSameKey(previousOperation);
       return previousOperation;
```
ljacomet: This is wrong I believe, because if the application of the previousOperation results in `null` then the `PutIfAbsentOperation` is applicable.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/Operation.java
ljacomet: Given my comment on `PutIfAbsentOperation`, I am starting to be convinced that an `Operation` should not even have this method. We would introduce a different abstraction which is a `ResolvedOperation` which would then have a `getResult()` and be an extension of an operation ... which is a problem with regards to the OpCode ... or maybe not

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/PutIfAbsentOperation.java
albinsuresh: That's exactly what lines 44-45 takes care of. If the application of _previousOperation_ results in a null, for example in a sequence like PUT, REMOVE, PUT_IF_ABSENT, the resolution goes as follow: 
PUT applied on top of null(initial state) -> PUT
REMOVE applied on top of the the previous result(PUT in this case) -> null
PUT_IF_ABSENT applied on top of null(the previous result) -> PUT_IF_ABSENT

So finally the PUT_IF_ABSENT succeeded. And that's exactly what the expected result is.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseOperation.java
albinsuresh: That's up-to the subclass. This is the constructor that `RemoveOperation` uses and to avoid that NPE it extends the _checkValueNonNull_ method. This is exactly the reason why I made this method protected so that sub-classes that don't care about the values can use this constructor and extend the _checkValueNonNull_ method to avoid NPE.

The other choice that I had was to keep the _checkValueNonNull_ in the base class do nothing and the sub-classes that requires a non-null value extend it in such a way that it throws. But then I'd have had to duplicate that error throwing logic to 5 different classes: PUT, PUT_IF_ABSENT, REMOVE_CONDITIONAL, REPLACE, REPLACE_CONDITIONAL. So I went for minimal codu duplication and did it this way. I understand that it wasn't a LOT of duplicated code. But still tried to avoid as much as I can.

albinsuresh:
Introduced the `Result` interface that represents that result of applying an `Operation`

albinsuresh:
The last commit is kinda optional. I just personally feel that the operations hierarchy is cleaner this way though it resulted in some code duplication. But if others disagree then it can be undone.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
```diff
@@ -16,41 +16,110 @@
...
+    return getOpCode() + "# key: " + key + ", value: " + value;
   }
 
   @Override
   public boolean equals(final Object obj) {
-    if(!super.equals(obj)) {
+    if(obj == null) {
       return false;
     }
     if(!(obj instanceof BaseKeyValueOperation)) {
```
ljacomet: If you change this test with a `getClass()` equality, yo can then skip the opcode check and do the cast with confidence.

ljacomet: Although now that I think of it - do we need equals and hashCode for these types?

albinsuresh: The equals method is only used in tests. The `ChainResolverTest` heavily depends on this. The hashCode was added just to satisfy findbugs.

albinsuresh:
Dropped the `BaseOperation` class and moved all its logic to `RemoveOperation` it was the only child. Improved the hashCode methods as well.

PR #1074, opened 2016-05-24
Issue 994 refactoring

AbfrmBlr:


clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -78,16 +97,52 @@ public long getKey() {
...
+      super(buffer);
+    }
+
     @Override
     public ServerStoreOp operation() {
       return ServerStoreOp.GET;
     }
+
+    @Override
+    public byte[] encode() {
```
rkavanap: Is this required?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheEntityResponse.java
```diff
@@ -17,14 +17,19 @@
...
 
+import java.nio.ByteBuffer;
+
 /**
  *
  * @author cdennis
  */
 public abstract class EhcacheEntityResponse implements EntityResponse {
 
+  private static final byte OP_CODE_OFFSET = 1;
```
ljacomet: Given its use, this is not an offset but a size

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheCodec.java
```diff
@@ -34,30 +34,28 @@
 
   @Override
   public byte[] encodeMessage(EhcacheEntityMessage message) {
-    if (message.getType() == EhcacheEntityMessage.Type.LIFECYCLE_OP) {
-      return LifeCycleOpCodec.encode((LifecycleMessage) message);
-    } else {
-      return ServerStoreOpCodec.encode((ServerStoreOpMessage) message);
-    }
+    return message.encode();
```
ljacomet: I do not get that change. Why did you decided to have messages carry their encoding / decoding logic?

I understand I gave that advice for the work @albinsuresh is doing, but in my mind the context here is different.

AbfrmBlr: Messages only know about their structure of data, so to me I think it makes sense to have encode/decode in messages itself. Also the encode/decode of messages is only done by platform using the codec interface, so if the message is already there, just call encode on that message.

ljacomet: The problem in this context is that we may end up having to support evolution in messages to guarantee backward compatibility. Having the logic isolated in codec classes makes more sense in that context.
Could others chime in?

lorban: I agree with Louis: keeping the codecs separated from the messages makes sense in this context as it will ease backward compatibility when we start supporting multiple versions of the messages.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -78,16 +97,48 @@ public long getKey() {
 
   public abstract ServerStoreOp operation();
 
+  public byte[] encode() {
+    ByteBuffer encodedMsg = ByteBuffer.allocate(MSG_TYPE_SIZE + STORE_OP_CODE_SIZE + CACHE_ID_LEN_SIZE + KEY_SIZE + getCacheId().length());
+    encodedMsg.put(EhcacheEntityMessage.Type.SERVER_STORE_OP.getOpCode());
+    encodedMsg.put(this.operation().getStoreOpCode());
+    encodedMsg.putInt(this.cacheId.length());
+    encodedMsg.put(this.cacheId.getBytes(UTF_8));
```
lorban: The above two lines assume that UTF-8 will encode all characters as single bytes which is only true for ASCII characters. You should rather do:

```
byte[] cacheIdBytes = cacheId.getBytes(UTF_8);
encodedMsg.putInt(cacheIdBytes.length);
encodedMsg.put(cacheIdBytes);
```

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -99,6 +150,13 @@ public ServerStoreOp operation() {
       this.payload = payload;
     }
 
+    GetAndAppendMessage(ByteBuffer buffer) {
+      super(buffer);
+      byte[] remaining = new byte[buffer.remaining()];
+      buffer.get(remaining);
+      this.payload = ByteBuffer.wrap(remaining).asReadOnlyBuffer();
```
lorban: Is there a reason to perform a deep copy of the `ByteBuffer` here?

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -107,6 +165,17 @@ public ServerStoreOp operation() {
     public ByteBuffer getPayload() {
       return payload;
     }
+
+    @Override
+    public byte[] encode() {
+      byte[] encodedMsg = super.encode();
```
lorban: This method looks very inefficient to me: multiple `byte[]` allocations and copies look like this will noticeably impact performance.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheEntityResponse.java
```diff
@@ -82,10 +113,14 @@ public static Failure failure(Exception cause) {
 
     private final Exception cause;
 
-    private Failure(Exception cause) {
+    Failure(Exception cause) {
       this.cause = cause;
     }
 
+    Failure(ByteBuffer failure) {
+      this.cause = (Exception)Util.unmarshall(failure.array());
```
lorban: This looks dodgy to me: calling `ByteBuffer.array()` only works with heap byte buffers which you don't know if you have such one here.

AbfrmBlr: AFAICT this constructor is package local and is only being called by providing `ByteBuffer.wrap()` which uses `HeapByteBuffer`. So it should be fine.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: Since the buffer gets passed to the super ctor, from line 155 onward I am just reading the rest in the buffer to a byte array and then setting is as payload.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: Will add a TODO to optimize later

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
AbfrmBlr: Cool

clustered/common/src/main/java/org/ehcache/clustered/common/messages/EhcacheCodec.java
AbfrmBlr: Why will we have to support multiple versions of messages ?

lorban: Think of the future when multiple versions of the server have been released. When that happens, we'll have to support multiple versions of the client connecting on the server to allow hot upgrades.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
lorban: I'm worried that the optimization would require to flatten the hierarchy. Let's not postpone that.

albinsuresh:
`EhcacheEntityMessage` entity messages are divided into two subtypes: `ServerStoreOpMessage`s and `LifeCycleOpMessage`s. And this reflects in the encoded messages as well. We have a byte reserved to identify this top-level split and another byte specifying the actual operation. Do we really need this extra byte in the message payload itself. It is just one byte but one that gets added to all cache operation messages making it a bit costly. Can't we get rid of this top level type byte somehow? Looks like an implementation details leaking out to the messages. I know this is a perf improvement and could be looked at later.

AbfrmBlr:
Decoupled the codecs and messages.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/LifeCycleMessageFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.ehcache.clustered.common.ServerSideConfiguration;
+import org.ehcache.clustered.common.ServerStoreConfiguration;
+
+/**
+ *
+ */
+public class LifeCycleMessageFactory {
+
+  public EhcacheEntityMessage validateCacheManger(ServerSideConfiguration configuration){
```
cljohnso: This should be `validateStoreManager` not `validateCacheManager` -- the cache manager is not a clustered concept.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/LifeCycleMessageFactory.java
```diff
@@ -0,0 +1,50 @@
...
+    return new LifecycleMessage.ValidateCacheManager(configuration);
+  }
+
+  public EhcacheEntityMessage configureCacheManager(ServerSideConfiguration configuration) {
```
cljohnso: This should be `configureStoreManager` instead of `configureCacheManager` -- the cache manager is not a clustered concept.  (The same for the message class.)

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -39,9 +40,11 @@
 public class EhcacheClientEntity implements Entity {
 
   private final EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint;
+  private final LifeCycleMessageFactory messageFactory;
 
-  public EhcacheClientEntity(EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint) {
+  public EhcacheClientEntity(EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint, LifeCycleMessageFactory messageFactory) {
```
cljohnso: I may have missed some discussion about this but what't the point of passing in a `LifeCycleMessageFactory`?  Unless we're going to configure it (which would require some odd interactions with the server in order to work properly), we're only ever going to use one type of this is a client -- the one against which the current version of the client is written.  I'd prefer direct allocation over another level of indirection without a purpose.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ServerStoreProxy.java
```diff
@@ -31,10 +31,12 @@
 
   private final String cacheId;
   private final EhcacheClientEntity entity;
+  private final ServerStoreMessageFactory messageFactory;
 
-  public ServerStoreProxy(String cacheId, EhcacheClientEntity entity) {
+  public ServerStoreProxy(String cacheId, EhcacheClientEntity entity, ServerStoreMessageFactory messageFactory) {
```
cljohnso: Like the comment on `EhcacheClientEntity`, I don't see the point of passing this in -- unless we're going to use a *single* instance for all stores under a cache manager (which, based on the implementation of `ServerStoreMessageFactory`, is presently not possible).

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpMessage.java
```diff
@@ -53,16 +54,34 @@ public static ServerStoreOp getServerStoreOp(byte storeOpCode) {
           throw new IllegalArgumentException("Store operation not defined for : " + storeOpCode);
       }
     }
+
   }
 
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
```
cljohnso: The encode/decode operations based on this are likely to be a lot faster if you simple transfer the `char[]`from the String.  Using the UTF-8 `Charset` is _expensive_ and subject to some _unannounced_ data modification if the String happens to have some "bad" characters.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -110,10 +108,15 @@
       new ConcurrentHashMap<String, Set<ClientDescriptor>>();
 
   private final ServerStoreCompatibility storeCompatibility = new ServerStoreCompatibility();
+  private final EhcacheEntityResponseFactory responseFactory;
 
-  EhcacheActiveEntity(ServiceRegistry services, byte[] config) {
+  EhcacheActiveEntity(ServiceRegistry services, byte[] config, EhcacheEntityResponseFactory responseFactory) {
```
cljohnso: Again, I don't get passing in a response factory.  I seriously doubt we're going to ever have a different set of message objects for this -- a different _codec_ maybe but not messages.

AbfrmBlr:
Rebased and removed all message factories from constructor visibility..

clustered/common/src/main/java/org/ehcache/clustered/common/messages/LifeCycleMessageCodec.java
```diff
@@ -0,0 +1,44 @@
...
+import java.nio.ByteBuffer;
+
+/**
+ *
+ */
+class LifeCycleMessageCodec {
+
+  public byte[] encode(LifecycleMessage message) {
+    byte[] encodedMsg = Util.marshall(message);
+    ByteBuffer buffer = ByteBuffer.allocate(1 + encodedMsg.length);
```
ljacomet: Please replace that 1 by a constant indicating what it is.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/LifeCycleMessageCodec.java
```diff
@@ -0,0 +1,44 @@
...
+    buffer.put(EhcacheEntityMessage.Type.LIFECYCLE_OP.getOpCode());
+    buffer.put(encodedMsg);
+    return buffer.array();
+  }
+
+  public EhcacheEntityMessage decode(byte[] payload) {
+    ByteBuffer message = ByteBuffer.wrap(payload);
+    byte[] encodedMsg = new byte[message.capacity() - 1];
+    message.get();
```
ljacomet: that's weird. Given the opcode found there, I would double check it is correct instead of dropping it.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/LifeCycleMessageCodec.java
```diff
@@ -0,0 +1,44 @@
...
+    byte[] encodedMsg = Util.marshall(message);
+    ByteBuffer buffer = ByteBuffer.allocate(1 + encodedMsg.length);
+    buffer.put(EhcacheEntityMessage.Type.LIFECYCLE_OP.getOpCode());
+    buffer.put(encodedMsg);
+    return buffer.array();
+  }
+
+  public EhcacheEntityMessage decode(byte[] payload) {
+    ByteBuffer message = ByteBuffer.wrap(payload);
+    byte[] encodedMsg = new byte[message.capacity() - 1];
```
ljacomet: that `1` is probably the same as above ...

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ResponseCodec.java
```diff
@@ -50,20 +52,20 @@
...
-        return EhcacheEntityResponse.success();
+        return EhcacheEntityResponse.Success.INSTANCE;
       case FAILURE:
-        Exception exception = (Exception)LifeCycleOpCodec.unmarshall(payArr);
-        return EhcacheEntityResponse.failure(exception);
+        Exception exception = (Exception)Util.unmarshall(payArr);
+        return new EhcacheEntityResponse.Failure(exception);
       case GET_RESPONSE:
-        return EhcacheEntityResponse.response(ChainCodec.decode(payArr));
+        return new EhcacheEntityResponse.GetResponse(ChainCodec.decode(payArr));
```
ljacomet: Again another static hiding there. Now maybe this can be simply flattened in this class as I believe no one else uses the `ChainCodec`

clustered/common/src/main/java/org/ehcache/clustered/common/messages/LifecycleMessage.java
```diff
@@ -25,6 +25,8 @@
  */
 public abstract class LifecycleMessage extends EhcacheEntityMessage implements Serializable {
 
+  public static final int LIFECYCLE_MSG_OP_CODE = 1;
```
ljacomet: This should be a `byte` and not an `int`

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpCodec.java
```diff
@@ -15,74 +15,63 @@
...
-        encodedMsg.put(message.getCacheId().getBytes(UTF_8));
-        encodedMsg.putLong(message.getKey());
-        encodedMsg.put(message.operation().getStoreOpCode());
+        putCacheIdKeyAndOpCode(encodedMsg, message.getCacheId(), message.getKey(), message.operation().getStoreOpCode());
         encodedMsg.put(appendMessage.getPayload());
         return encodedMsg.array();
       case GET_AND_APPEND:
         GetAndAppendMessage getAndAppendMessage = (GetAndAppendMessage)message;
-        encodedMsg = ByteBuffer.allocate(MSG_TYPE_OFFSET + STORE_OP_CODE_OFFSET + CACHE_ID_LEN_OFFSET + KEY_OFFSET + cacheIdLen + getAndAppendMessage
+        encodedMsg = ByteBuffer.allocate(STORE_OP_CODE_SIZE + CACHE_ID_LEN_SIZE + KEY_SIZE + 2 * cacheIdLen + getAndAppendMessage
```
ljacomet: this `STORE_OP_CODE_SIZE + CACHE_ID_LEN_SIZE + KEY_SIZE` is repeated for all messages - could be declared as a constant

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ServerStoreOpCodec.java
```diff
@@ -15,74 +15,63 @@
...
-  private static final byte KEY_OFFSET = 8;
-  private static final byte CHAIN_LEN_OFFSET = 4;
+  private final ChainCodec chainCodec;
 
-  public static byte[] encode(ServerStoreOpMessage message) {
+  ServerStoreOpCodec() {
+    this.chainCodec = new ChainCodec();
+  }
+
+  public byte[] encode(ServerStoreOpMessage message) {
```
ljacomet: The whole method would read better if each case was simply a call to a method doing the necessary work per message type.

ljacomet:
Still have a number of remarks on the internals of a few classes. However, structure looks ok and since there are dependencies on it, merging now.

Issue #1043, opened 2016-05-10
ClusteredStore to implement AuthoritativeTier


ljacomet:
Once we have clustered put and get working, we need to enable local caching tiers to live on top of the clustered store.

PR #1078, opened 2016-05-24
Issue #1043

ljacomet:
* Tiering support for `ClusteredStore`
* Generalise `TieredStore.Provider` logic


core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -141,6 +145,8 @@
      * @param resource the caching tier to initialise
      */
     void initCachingTier(CachingTier<?, ?> resource);
+
+    int rankCachingTier(Set<ResourceType<?>> resourceTypes, Collection<ServiceConfiguration<?>> serviceConfigs);
```
lorban: Javadoc is missing.

ljacomet: Good catch - fixing

Issue #1080, opened 2016-05-25
Disk store type checking incorrectly uses loadClass(...) instead of forName(...)


chrisdennis:
The OffHeapDiskStore code persist the names of it's key and value types to allow for strict type checking on recovery - it writes out the return of `klazz.getName()`.  It then passes this string to `classloader.loadClass(name)` to recover the type, but this will not work for array types.  We should switch this to `Class.forName(name, false, classloader)`.

PR #1081, opened 2016-05-25
Issue #1080 Disk store type checking incorrectly uses loadClass(...) instead of forName(...)

chrisdennis:


ljacomet:
Fix looks good but not too happy with the merge strategy.
Given the effect on the graph, can we agree to:
* merge final destination into fix branch

Issue #1002, opened 2016-04-26
User facing doc for clustered CacheManager creation


ljacomet:
Using the existing format of asciidoc + callouts to a sample that is executed as part of the test suite.

ljacomet:
Had a discussion with @kedar031 about my comments on #1032 about adding bits about starting the server.
To confirm the idea there is to use the ehcache clustered kit as produced by the build currently and to have limited documentation on the server itself, the bare minimum for a developers to have a clustered cache manager running - which at this stage does not include any server side configuration of resources for example.

PR #1084, opened 2016-05-26
Correcting the URI protocol used in user facing documentation

kedar031:


ljacomet:
Even though the issue has been closed, could you link the commit to #1002

Issue #1077, opened 2016-05-24
Public facing API for consistency configuration


ljacomet:

lorban:
Writing some `GettingStarted` should be part of this task too.

PR #1091, opened 2016-05-27
Issue #1007: clustered store invalidation

lorban:
* basic config for strong/eventual has been added, the real deal should be done in #1077 
 * 3-tier invalidation doesn't work yet as it requires a new release of the offheap store
 * tests are lacking but coming


clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -16,32 +16,78 @@
...
+
   private final EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint;
+  private final Map<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>> responseListeners = new ConcurrentHashMap<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>>();
 
   public EhcacheClientEntity(EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint) {
     this.endpoint = endpoint;
+    endpoint.setDelegate(new EndpointDelegate() {
+      @Override
+      public void handleMessage(EntityResponse messageFromServer) {
+        fireResponseEvent((EhcacheEntityResponse) messageFromServer);
```
ljacomet: I would like to have a bit more error handling here - although it should never happen. But a class cast exception without context seems a bit bland.

lorban: good point, let me wrap that with a `if (... instanceof ...)`

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -30,7 +30,7 @@
 
   private final ServerStoreConfiguration storeConfiguration;
   private final PageSource pageSource;
-  private final OffHeapServerStore store;
+  private final ServerStore store;
```
albinsuresh: I fail to understand why this was changed.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
```diff
@@ -0,0 +1,149 @@
...
+      if (countDownLatch == null) {
+        break;
+      }
+      countDownLatch.await();
+    }
+
+    try {
+      return c.apply();
+    } finally {
+      latch.await();
```
ljacomet: This wait in finally seems risky. What if the operation failed to reach the server for example?
Also, I would rather have `await` with timeout and logging, even if we effectively end up waiting forever.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
lorban: Correct.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
```diff
@@ -0,0 +1,155 @@
...
+        for (InvalidationListener listener : invalidationListeners) {
+          listener.onInvalidationRequest(key);
+        }
+
+        try {
+          LOGGER.debug("CLIENT: ack'ing invalidation of hash {} from cache {} (ID {})", key, cacheId, invalidationId);
+          entity.invoke(EhcacheEntityMessage.clientInvalidateHashAck(cacheId, key, invalidationId), true);
+        } catch (Exception e) {
+          //TODO: what should be done here?
+          LOGGER.error("error acking client invalidation of hash " + key + " on cache " + cacheId, e);
```
AbfrmBlr: Why not use {} approach here as well?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
```diff
@@ -0,0 +1,155 @@
...
+      if (countDownLatch == null) {
+        break;
+      }
+      countDownLatch.await();
+    }
+
+    try {
+      T result = c.apply();
+      latch.await();
+      LOGGER.debug("CLIENT: key {} invalidated on all clients, unblocking append");
```
AbfrmBlr: Variable to be logged is missing

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
lorban: Right, I'll do that

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
lorban: Nice catch, thanks

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
lorban: That's an overlook on my part, let me clean this up.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -293,6 +346,52 @@ private EhcacheEntityResponse invokeServerStoreOperation(ServerStoreOpMessage me
...
+      clientInvalidated(invalidationHolder.clientDescriptorWaitingForInvalidation, invalidationId);
+    }
+  }
+
+  private void clientInvalidated(ClientDescriptor clientDescriptor, int invalidationId) {
+    InvalidationHolder invalidationHolder = clientsWaitingForInvalidation.get(invalidationId);
+
+    if (stores.get(invalidationHolder.cacheId).getStoreConfiguration().getConsistency() == Consistency.STRONG) {
+      invalidationHolder.clientsHavingToInvalidate.remove(clientDescriptor);
+      if (invalidationHolder.clientsHavingToInvalidate.isEmpty()) {
```
ljacomet: Is this not racy in case invalidations happen really close to each other? Or is it protected at a higher level?

docs/src/docs/uml/invalidation_strong.txt
```diff
@@ -0,0 +1,68 @@
+@startuml
```
ljacomet: If this file gets its extension changed to `.puml` then the intellij plugin picks it up automatically.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
lorban: Each invalidation has a unique `invalidationId` that is used as key of the `clientsWaitingForInvalidation` map, so all invalidation requests are isolated from each other.

As long as the `invalidationHolder.clientsHavingToInvalidate` `Set` is thread-safe, I fail to see how this code could be racy: the `Set` is filled up upon construction and entries are being removed while invalidation acks are received. Once the last ack is received (i.e.: the `Set` is empty), the `Map` gets cleaned up.

docs/src/docs/uml/invalidation_strong.txt
lorban: Good to know, thanks.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
lorban: Okay, now I see the race: if the last two clients' ack messages are being handled in parallel, they could both see the `clientsHavingToInvalidate` `Set` empty and both call `clientCommunicator.sendNoResponse`.

I just need to check the return value of `remove()` and only call `clientCommunicator.sendNoResponse` when it's not null to solve this problem.

Thanks!

clustered/server/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
   compile group: 'org.terracotta', name: 'offheap-store', version: parent.offheapVersion
   compile group: 'org.slf4j', name: 'slf4j-api', version: parent.slf4jVersion
   provided "org.terracotta:entity-server-api:$parent.entityApiVersion"
+    provided "org.terracotta:standard-cluster-services:$parent.terracottaApisVersion"
```
ljacomet: Formatting

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/NoInvalidationServerStoreProxy.java
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      response = entity.invoke(messageFactory.getOperation(key), false);
+    } catch (Exception e) {
+      throw new ServerStoreProxyException(e);
+    }
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for get operation was invalid : " +
+                                          response != null ? response.getType().toString() : "null message");
```
ljacomet: That's wrong - was fixed in master in the ServerStoreProxy and so fix needs to be applied here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/NoInvalidationServerStoreProxy.java
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      response = entity.invoke(messageFactory.getAndAppendOperation(key, payLoad), true);
+    } catch (Exception e) {
+      throw new ServerStoreProxyException(e);
+    }
+    if (response != null && response.getType() == EhcacheEntityResponse.Type.GET_RESPONSE) {
+      return ((EhcacheEntityResponse.GetResponse)response).getChain();
+    } else {
+      throw new ServerStoreProxyException("Response for getAndAppend operation was invalid : " +
+                                          response != null ? response.getType().toString() : "null message");
```
ljacomet: That's wrong - was fixed in master in the ServerStoreProxy and so fix needs to be applied here.

ljacomet:
Now that this PR is based on top of master having the clear operation, I believe it lacks the triggering of complete invalidation on clear.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -296,6 +365,53 @@ private EhcacheEntityResponse invokeServerStoreOperation(ServerStoreOpMessage me
...
+
+    InvalidationHolder invalidationHolder = null;
+    if (stores.get(cacheId).getStoreConfiguration().getConsistency() == Consistency.STRONG) {
+      invalidationHolder = new InvalidationHolder(originatingClientDescriptor, clientsToInvalidate, cacheId, key);
+      clientsWaitingForInvalidation.put(invalidationId, invalidationHolder);
+    }
+
+    LOGGER.debug("SERVER: requesting {} client(s) invalidation of hash {} in cache {} (ID {})", clientsToInvalidate.size(), key, cacheId, invalidationId);
+    for (ClientDescriptor clientDescriptorThatHasToInvalidate : clientsToInvalidate) {
+      LOGGER.debug("SERVER: asking client {}" + clientDescriptorThatHasToInvalidate + " to invalidate hash {} from cache {} (ID {})", clientDescriptorThatHasToInvalidate, key, cacheId, invalidationId);
```
ljacomet: That logging line mixes concatenation and variables

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -295,8 +298,22 @@ public void initStore(final Store<?, ?> resource) {
...
+      clusteredStore.storeProxy = clusteringService.getServerStoreProxy(storeConfig.getCacheIdentifier(), storeConfig.getStoreConfig(), storeConfig.getConsistency());
+      clusteredStore.storeProxy.addInvalidationListener(new ServerStoreProxy.InvalidationListener() {
+        @Override
+        public void onInvalidationRequest(long hash) {
+          if (clusteredStore.invalidationValve != null) {
+            try {
+              LOGGER.debug("CLIENT: calling invalidation valve");
+              clusteredStore.invalidationValve.invalidateAllWithHash(hash);
+            } catch (StoreAccessException sae) {
+              //TODO: what should be done here?
```
ljacomet: Most likely delegate to the resilience strategy ... so can remain a TODO for now

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxy.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private void awaitOnLatch(CountDownLatch countDownLatch) throws InterruptedException {
+    int totalAwaitTime = 0;
+    int backoff = INVALIDATION_BACKOFF_SECONDS;
+    while (!countDownLatch.await(backoff, TimeUnit.SECONDS)) {
+      totalAwaitTime += backoff;
+      if (totalAwaitTime >= MAX_INVALIDATION_WAIT_SECONDS) {
+        throw new IllegalStateException("Server did not send an InvalidationDone message within " + totalAwaitTime + "s");
```
ljacomet: I am not sure we want a max time here. In a very slow but still moving system, we may hit that limit.
I would rather have something like waiting for 1 second, then for 2 then 4 then 10 and from then on always 10. With the logging at each wake up. Makes sense?

lorban: Makes sense. Let me modify that.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
lorban: I'll update the TODO accordingly.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
lorban: Good catch!

clustered/server/build.gradle
lorban: That's because of stupid intellij refusing me to manually format the gradle files.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/NoInvalidationServerStoreProxy.java
lorban: Looks like I messed up the rebasing. I'll go revisit it entirely.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/NoInvalidationServerStoreProxy.java
lorban: Same as above.

ljacomet:
test this please

lorban:
Handling of `clear` will happen in another PR.

Issue #1090, opened 2016-05-27
Reduced logging during Ehcache startup


ljacomet:
While we have a very conservative view on logging during cache operations, we wanted to be more explicit while starting Ehcache.
However we have come to realise that the level of logs outputted during startup of a `CacheManager` is too high.

Desired `INFO` level output:
* Location of config file, when applicable
* Started caches

PR #1092, opened 2016-05-27
Issue #1090 Reduce log level to debug

ljacomet:
Includes some logging related code cleanups.

### DIFF
--- a/core/src/main/java/org/ehcache/core/EhcacheManager.java
+++ b/core/src/main/java/org/ehcache/core/EhcacheManager.java
@@ -93,15 +93,16 @@ public class EhcacheManager implements PersistentCacheManager, InternalCacheMana
     }
   }
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheManager.class);
+
   private final DefaultConfiguration configuration;
   private final ClassLoader cacheManagerClassLoader;
-  private final boolean useLoaderInAtomics;
 
-  private final Logger logger = LoggerFactory.getLogger(this.getClass());
+  private final boolean useLoaderInAtomics;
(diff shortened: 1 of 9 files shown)


ljacomet:
Wow, managed to botch the branch origin ... will re-open a new one once fixed.

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -463,7 +450,7 @@ public void close() throws Exception {
             throw new RuntimeException(e);
           }
         }
-        this.getLogger().debug("Could not create serializers for " + alias, e);
+        LOGGER.debug("Could not create serializers for " + alias, e);
```
chrisdennis: This might or might not be serious... what logging do we see now if someone misconfigures a cache that needs serializers?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -258,7 +258,7 @@ public void destroy(String name) throws CachePersistenceException {
   @Override
   public void destroyAll() {
     if(recursiveDeleteDirectoryContent(rootDirectory)){
-      LOGGER.info("Destroyed all file based persistence context");
+      LOGGER.debug("Destroyed all file based persistence context");
```
chrisdennis: plural?

chrisdennis:
I think you're going to need to make changes to logging in offheap-store otherwise you might find all your good work undone when using offheap.

GaryWKeim:
Can we just use the {} mechanism everywhere? Sometimes it's used, sometimes not.

ljacomet:
Opened PRs to fix the remarks made here.

Issue #1075, opened 2016-05-24
Implement remaining CRUD ops for ClusteredStore


ljacomet:

ljacomet:
* `containsKey(K key)`
* `remove(K key)`
* `clear()`

ljacomet:
A few things to add:
* Statistics
* Proper return value - thinking put here but may apply to others

ljacomet:
Still need clean up on the clear message that _wastes_ 8 bytes for a long key that is not used.

albinsuresh:
The `Clear` message and `ClientInvalidationHashAck` message that got added in #1007 has an extra long field in their types for the key that they got inherited from the `ServerStoreOpMessage` in the hierarchy. Even though the extra 8 bytes is not sent over the network at least for the`Clear` message the message hierarchy requires some refactoring to eliminate it from the type level itself. Working on fixing this hierarchy.

PR #1095, opened 2016-05-27
Issue #1075 Completing CRUD operations on clustered store

albinsuresh:
The message encoding of `ServerStoreOpMessage.ClearMessage` can  definitely be improved. Currently it is unnecessarily encoding a key along with the *cacheId*. This is really not required but to fix that I'd have had to refactor some code in the codec class. Since @AbfrmBlr is already working on refactoring that part of the code I did not want to cause a conflict. So I'll fix this once his work gets merged.

Now containsKey could have been implemented like _return get(key) != null_ as @lorban suggested.
But just to avoid the new `ClusteredValueHolder` creation and the meta-data updates that would happen with a get in future, I have done it differently(which caused some code duplication). If no one really cares about that difference then I could revert. I was just trying to keep it as different as I can from a store.get()

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -70,8 +70,17 @@ public class ClusteredStore<K, V> implements AuthoritativeTier<K, V> {
 
   @Override
   public ValueHolder<V> get(final K key) throws StoreAccessException {
+    V value = getInternal(key);
+    if(value == null) {
+      return null;
+    } else {
+      return new ClusteredValueHolder<V>(value);
+    }
+  }
+
+  private V getInternal(K key) {
     Chain chain = storeProxy.get(key.hashCode());
-    V value = null;
+    V value;
     if(!chain.isEmpty()) {
       ResolvedChain<K, V> resolvedChain = resolver.resolve(chain, key);
 
@@ -87,13 +96,16 @@ public class ClusteredStore<K, V> implements AuthoritativeTier<K, V> {
     } else {
       return null;
     }
-    return new ClusteredValueHolder<V>(value);
+    return value;
   }
 
   @Override
   public boolean containsKey(final K key) throws StoreAccessException {
-    // TODO: Make appropriate ServerStoreProxy call
-    throw new UnsupportedOperationException("Implement me");
+    if(getInternal(key) == null) {
+      return false;
+    } else {
+      return true;
+    }
   }
 
   @Override
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -155,7 +155,7 @@ public class ClusteredStore<K, V> implements AuthoritativeTier<K, V> {
 
   @Override
   public void clear() throws StoreAccessException {
-    // TODO: Make appropriate ServerStoreProxy call
+    storeProxy.clear();
   }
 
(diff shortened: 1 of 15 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -87,13 +97,16 @@ private ClusteredStore(final OperationsCodec<K, V> codec, final ChainResolver<K,
...
     }
-    return new ClusteredValueHolder<V>(value);
+    return value;
   }
 
   @Override
   public boolean containsKey(final K key) throws StoreAccessException {
-    // TODO: Make appropriate ServerStoreProxy call
-    throw new UnsupportedOperationException("Implement me");
+    if(getInternal(key) == null) {
```
lorban: you can simplify that as 

`return getInternal(key) != null`

albinsuresh: Aah yes. How silly of me!

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -132,16 +132,25 @@ public void clusteredCachePutGet() throws Exception {
       Cache<Long, String> cache = cacheManager.createCache("clustered-cache", config);
       assertThat(cache.get(1L), is(nullValue()));
       cache.put(1L, "The one");
+      assertThat(cache.containsKey(2L), is(false));
```
ljacomet: I do not like having asserts in code that used in docs. The callout should indicate the expected result and you may have asserts outside of what's included to make sure nothing breaks.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ServerStoreProxyTest.java
```diff
@@ -160,6 +160,19 @@ public void testReplaceAtHeadSuccessFull() {
     assertChainHas(anotherReplace, 800L, 4000L, 40000L);
   }
 
+  @Test
+  public void testClear() {
+    serverStoreProxy.append(1L, createPayload(100L));
+
+    Chain chain = serverStoreProxy.get(1);
+    assertThat(chain.isEmpty(), is(false));
```
ljacomet: These assert are not needed here, the fact that append works should be tested elsewhere already.

clustered/server/src/test/java/org/ehcache/clustered/server/store/impl/ReferenceStoreImpl.java
```diff
@@ -145,6 +145,28 @@ public void replaceAtHead(long key, Chain expect, Chain update) {
     }
   }
 
+  private void writeLockAll() {
+    for (ReadWriteLock lock : locks) {
```
ljacomet: is lock and iteration order guaranteed here?

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -115,8 +115,8 @@ public void clusteredCacheManagerWithDynamicallyAddedCacheExample() throws Excep
   }
 
   @Test
-  public void clusteredCachePutGet() throws Exception {
-    // tag::clusteredCachePutGet[]
+  public void clusteredCacheCRUD() throws Exception {
+    // tag::clusteredCacheCRUD[]
```
ljacomet: tag name change but no `.adoc` file touched - forgotten or not existing yet?

albinsuresh:
Addressed all the review comments and did a minor refactoring.

PR #1101, opened 2016-05-30
[release/3.0] #1090 Minor logging improvements

ljacomet:
* Use slf4j substitution instead of string concatenation
* Fix plural

From remarks on PR #1092


ljacomet:
As discussed, did one final pass to replace string concatenation

Issue #1107, opened 2016-06-01
Stateful serializer and clustered cache


ljacomet:
There needs to be a revision of how stateful serializers are handling state.

In a clustered cache, any modification to the serializer state needs to be shared with the cluster before the mapping makes it into the clustered store. Without that, clients at the other end may fail to read entries.

In this context, the limited implicit contract at `Serializer` constructor level is insufficient.

PR #1110, opened 2016-06-01
[Design Review] Issue #1107: Serializer for clustered

ljacomet:
* Refactor `PersistableResourceService` and `LocalPersistenceService`
* Add `StateRepository`
* Update implicit `Serializer` constructor contract

I am interested in having comments on the high level design.

Future task in this context:
* Implement the `StateRepository` for disk and clustered
* Rework the `CompactPersistentJavaSerializer` and the `DefaultSerializationProvider`


api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -0,0 +1,38 @@
...
+   * <P>
+   *   If the map existed already, it is returned with its content fully available.
+   * </P>
+   *
+   * @param name the map name
+   * @param <K> the map key type, must be {@code Serializable}
+   * @param <V> the map value type, must be {@code Serializable}
+   * @return a map
+   */
+  <K, V> ConcurrentMap<K, V> getPersistentConcurrentMap(String name);
```
chrisdennis: Persistent is a property of the implementation, isn't it?

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
```diff
@@ -28,8 +28,9 @@
  * <dl>
  *   <dt><code><i>Serializer</i>(ClassLoader loader)</code>
  *   <dd>This constructor is used to initialize the serializer for transient caches.
```
chrisdennis: This constructor is used to initialize 'stateless' serializers.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
```diff
@@ -28,8 +28,9 @@
...
- *   <dt><code><i>Serializer</i>(ClassLoader loader, org.ehcache.core.spi.service.FileBasedPersistenceContext context)</code>
- *   <dd>This constructor is used to initialize the serializer for persistent caches.
+ *   <dt><code><i>Serializer</i>(ClassLoader loader, org.ehcache.spi.persistence.StateRepository stateRepository)</code>
+ *   <dd>This constructor is used to initialize the serializer for persistent caches and allows them to store any relevant
```
chrisdennis: This constructor is used to initialize stateful serializer.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
ljacomet: No, it is the constructor we use when no persistent tier is configured. Which is what makes this implicit constructor contract even more weird. A stateful serializer still needs to have this constructor if it is fit for transient caches.

api/src/main/java/org/ehcache/spi/serialization/Serializer.java
ljacomet: No, it is the constructor we use when a persistent tier is configured. Which is what makes this implicit constructor contract even more weird. A statelessl serializer still needs to have this constructor if it is fit for persistent caches.

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
ljacomet: Well, it depends if we want to always provide a `StateRepository` or only in the context of a persistent cache.

ljacomet:
Conversation with @lorban @chrisdennis @cljohnso @AbfrmBlr @albinsuresh concluded this is good wrt design - closing to work on proper implementation

PR #1111, opened 2016-06-01
[Design Review] Issue #1077 Consistency configuration API

ljacomet:
* XML proposal
* Builders proposal

This is decoupled from the resource configuration because it is not really a resource constraint but a mutation visibility guarantee.
This is currently decoupled from what @lorban did in PR #1091 but would of course be based or wired to work in that context.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.client.config.builders;
+
+import org.ehcache.clustered.client.config.ClusteredStoreConfiguration;
+import org.ehcache.config.Builder;
+
+/**
+ * ClusteredStoreConfigurationBuilder
+ */
+public class ClusteredStoreConfigurationBuilder implements Builder<ClusteredStoreConfiguration> {
+
+  public static ClusteredStoreConfigurationBuilder withConsistency(ClusteredStoreConfiguration.Consistency consistency) {
+    return new ClusteredStoreConfigurationBuilder();
+  }
+
+  @Override
+  public ClusteredStoreConfiguration build() {
+    return new ClusteredStoreConfiguration();
+  }
+}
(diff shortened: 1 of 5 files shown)


clustered/client/src/test/resources/configs/resource-sample.xml
```diff
@@ -32,6 +32,7 @@
       <eh:offheap unit="MB">512</eh:offheap>
       <tc:cluster-fixed unit="GB" from="secondary">128</tc:cluster-fixed>
     </eh:resources>
+    <tc:clustered-store consistency="STRONG"/>
```
chrisdennis: I'm not sure how much I like the block capitals here...

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -114,4 +116,28 @@ public void clusteredCacheManagerWithDynamicallyAddedCacheExample() throws Excep
     // end::clusteredCacheManagerWithDynamicallyAddedCacheExample[]
   }
 
+  @Test
+  public void explicitConsistencyConfiguration() throws Exception {
+    final CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder
+            = CacheManagerBuilder.newCacheManagerBuilder()
+            .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("terracotta://localhost:9510/my-application?auto-create"))
+                    .defaultServerResource("primary-server-resource")
+                    .resourcePool("resource-pool-a", 128, MemoryUnit.B));
```
chrisdennis: Should we support a default consistency at the CM level too?

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+import org.ehcache.clustered.client.config.ClusteredStoreConfiguration;
+import org.ehcache.config.Builder;
+
+/**
+ * ClusteredStoreConfigurationBuilder
+ */
+public class ClusteredStoreConfigurationBuilder implements Builder<ClusteredStoreConfiguration> {
+
+  public static ClusteredStoreConfigurationBuilder withConsistency(ClusteredStoreConfiguration.Consistency consistency) {
+    return new ClusteredStoreConfigurationBuilder();
```
chrisdennis: This isn't supposed to be working right?

clustered/client/src/test/resources/configs/resource-sample.xml
ljacomet: indeed, let me change that to lowercase

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
ljacomet: Seems overkill but ready to hear arguments. We should however pick a default when nothing is selected at the cache level.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
ljacomet: Nope, this PR is just a shell to discuss the public API bits.

ljacomet:
Conversation with @lorban @chrisdennis @cljohnso @AbfrmBlr @albinsuresh concluded this is good wrt design - closing to work on proper implementation

PR #1114, opened 2016-06-02
Issue #1075: Correcting the return value of ClusteredStore.put

albinsuresh:
and wiring stats for CRUD ops

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -38,12 +38,14 @@ import org.ehcache.core.spi.store.Store;
 import org.ehcache.core.spi.store.events.StoreEventSource;
 import org.ehcache.core.spi.store.StoreAccessException;
 import org.ehcache.core.spi.store.tiering.AuthoritativeTier;
+import org.ehcache.core.statistics.StoreOperationOutcomes;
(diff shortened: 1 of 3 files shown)


lorban:
I'm fine with the way you implemented that, *if* we want such implementation as I find that `put`'s cost is going through the roof just to differentiate between a new put and an updating put to calculate accurate statistics.

albinsuresh:
@lorban I had the same concern when I had to do it for remove operation as well. But the decision was taken at that point to honor the store contract. So we did it that way. But yeah you're right. The price that we're paying is huge. May be others can chime in and we can finalize on how accurate we want clustered stats to be.

ljacomet:
@albinsuresh Let's rebase and then I merge.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -54,29 +56,43 @@
...
  */
 public class ClusteredStore<K, V> implements AuthoritativeTier<K, V> {
 
+  private static final String STATISTICS_TAG = "clustered-store";
+
   private final OperationsCodec<K, V> codec;
   private final ChainResolver<K, V> resolver;
 
-  private volatile ServerStoreProxy storeProxy;
+  volatile ServerStoreProxy storeProxy;
```
ljacomet: Had missed that change on previous reviews - I am sure there is a way to avoid needing this for tests.

PR #1117, opened 2016-06-02
Issue #1077 Consistency configuration

ljacomet:


clustered/common/src/main/java/org/ehcache/clustered/common/Consistency.java
```diff
@@ -16,11 +16,20 @@
 package org.ehcache.clustered.common;
 
 /**
+ *
+ * Enumeration of the different consistency levels supported in clustered caches.
+ *
  * @author Ludovic Orban
```
GaryWKeim: These @author tags are to be removed, yes?

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -79,4 +79,36 @@ include::{sourcedir}/clustered/client/src/test/java/org/ehcache/clustered/client
...
+
+=== Specifying consistency level
+
+Ehcache offers two level of consistency:
+
+Eventual::
+In _eventual_ consistency, the clustering solution will make sure that when a mutative operation returns, invalidation for the corresponding entries found on other clients has been initiated.
+This is the default level if not configured explicitly.
+Strong::
+In _strong_ consistency, the clustering solution will make sure that when a mutative operation returns, the corresponding entries found on other clients have been invalidated.
```
AbfrmBlr: Invalidation of keys on clients seems to be an internal thing, I think we should document strong vs eventual by explaining the visibility of change available to other clients.

clustered/common/src/main/java/org/ehcache/clustered/common/Consistency.java
ljacomet: Indeed - removing

docs/src/docs/asciidoc/user/clustered-cache.adoc
ljacomet: That's a good point - will rephrase these and javadoc.

ljacomet:
Updated following comments and also completed the Javadoc work.
Did also move the XML parsing classes to internal package.

ljacomet:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+package org.ehcache.clustered.client.config.builders;
+
+import org.ehcache.clustered.client.config.ClusteredStoreConfiguration;
+import org.ehcache.clustered.common.Consistency;
+import org.ehcache.config.Builder;
+
+/**
+ * {@link Builder} for the {@link ClusteredStoreConfiguration}.
+ */
+public class ClusteredStoreConfigurationBuilder implements Builder<ClusteredStoreConfiguration> {
```
albinsuresh: Do we really need a builder for this configuration? As there is only one parameter for the consistency configuration.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -165,4 +165,17 @@
       </xs:annotation>
     </xs:attribute>
   </xs:attributeGroup>
+
+  <xs:element name="clustered-store" type="tc:clustered-store-type" substitutionGroup="eh:service-configuration"/>
```
albinsuresh: Do we want to call this element clustered-"store"? Store is more of an internal concept, right? How about having the config something like this:
`<consistency>STRONG</consistency>` and `<consistency>EVENTUAL</consistency>` ?

clustered/client/src/test/java/org/ehcache/clustered/client/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -16,6 +16,7 @@
 
 package org.ehcache.clustered.client.config.xml;
 
+import org.ehcache.clustered.client.internal.config.xml.ClusteringServiceConfigurationParser;
```
albinsuresh: The parser moved to the internal package but the test did not.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
ljacomet: We have builders for all internal things - why make an exception?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
ljacomet: This is to be consistent with other store level configurations - have a look at the default Ehcache XSD.

clustered/client/src/test/java/org/ehcache/clustered/client/config/xml/ClusteringServiceConfigurationParserTest.java
ljacomet: Good catch - will fix

lorban:
Needs rebasing, otherwise LGTM

lorban:
You may want to look at `DefaultClusteringService.getServerStoreProxy()` as it assumes the consistency can be null and sets the default to `EVENTUAL`. I don't like duplicating defaults so if one can go, that'as all for the best.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredStoreConfigurationBuilder.java
albinsuresh: Configurations for serializers and copiers are exposed directly, no? How is this different from those cache level configurations?

ljacomet:
@lorban Indeed, had missed that bit - fixed, including specifying a value in all the tests

PR #1120, opened 2016-06-03
 #1007 ServerStore.clear() invalidation handling, both strong and eventual

lorban:
I added a couple more TODO's in `ServerStoreOpMessage` as we now have two messages that do not need a key but do have to send 8 bytes on the wire to act as that key.


ljacomet:
PR looks good but the comment above should be taken care of - probably by adding an interface at some level. Commenting on the issue that this still needs taking care of.

Issue #1121, opened 2016-06-03
Wire invalidation valve into offheap store


lorban:
More precisely into  `LowerCachingTier.invalidateAllWithHash()`, which the `AbstractOffHeapStore` currently is the only implementation.

PR #1125, opened 2016-06-03
 #1121 implement AbstractOffHeapStore.invalidateAllWithHash()

lorban:


clustered/client/src/test/java/org/ehcache/clustered/client/BasicClusteredCacheTest.java
```diff
@@ -88,7 +88,7 @@ public void testClusteredCacheTwoClients() throws Exception {
         newCacheManagerBuilder()
             .with(cluster(CLUSTER_URI))
             .withCache("clustered-cache", newCacheConfigurationBuilder(Long.class, String.class,
-                ResourcePoolsBuilder.newResourcePoolsBuilder().heap(100, EntryUnit.ENTRIES)
+                ResourcePoolsBuilder.newResourcePoolsBuilder().heap(100, EntryUnit.ENTRIES).offheap(1, MemoryUnit.MB)
```
chrisdennis: Right now this test doesn't actually do anything with the offheap since the heap tier is large enough to cache all the values used in the test.  Would probably be good to add a new test that used a very small heap tier so that we definitely need to be invalidating things out of the offheap tier in order to get correct reads.

core/src/main/java/org/ehcache/core/statistics/LowerCachingTierOperationsOutcome.java
```diff
@@ -49,6 +49,20 @@
...
+   */
+  enum InvalidateAllWithHashOutcome implements LowerCachingTierOperationsOutcome {
+    /**
+     * entries invalidated, without errors
+     */
+    SUCCESS,
+    /**
+     * entries invalidated, with errors
+     */
+    FAILURE
```
chrisdennis: Why do we have this outcome if it's never used?

lorban: Because of stupid copy/paste. Let me clean that.

clustered/client/src/test/java/org/ehcache/clustered/client/BasicClusteredCacheTest.java
lorban: Right, let's do that.

PR #1127, opened 2016-06-03
Fix #919 Wire emergency valve

### LINKED ISSUE #919: Montreal design 'Emergency Valve' is not currently wired in.
Currently if there is a size inversion and everything in the offheap tier is 'pinned' because it's all cached in the heap tier, there is no functional emergency valve to trigger clearing of the heap tier.  We should think about fixing this.

ljacomet:


anthonydahanne:
an emergency valve could help Clifford
House On fire !

PR #1128, opened 2016-06-03
don't embed management  in Ehcache Clustered

anthonydahanne:
This module brings external dependencies (terracotta management-model) that could cause classpath clashing issues
More concretely, when we launch an ehcache client from our ide, some management model classes are loaded from ehcache-clustered (since it shades them) instead of the "official" management-model jar

### DIFF
--- a/clustered/dist/build.gradle
+++ b/clustered/dist/build.gradle
@@ -28,7 +28,6 @@ dependencies {
   compile project(':api')
   compile project(':core')
   compile project(':impl')
-  compile project(':management')
   compile project(':107')
   compile project(':xml')
   compile project(':transactions')


mathieucarbou:
Good to me! 
@chrisdennis @ljacomet : I'll let you merge if it is OK for you :-)

ljacomet:
This PR makes no sense to me. From what I read the goal is to not embed terracotta `management-model` but the way to achieve this is not by removing the management module of Ehcache from clustered jar.
What needs to be done is keep the management module in without its `management-model` dependency and declare that dependency as external.
Now doing this means we increase the number of required dependencies for ehcache clustered users and thus this needs to be discussed further.

mathieucarbou:
Yes the real goal is to remove `management-model` from the shaded jar :-)

The ehcache mamnagement module is just usable for standalone at the moment, and there will be new modules for clustered: so I don't see why it should be within the clustered shaded jar ?

I sent you an email also last week about a proposal on how to better integrate management and Ehcache. We would need now to work on some Ehcache modules to add support for management, but we'd like to avoid impacting Ehcache project directly since these new modules would heavily depend on some Terracotta-OSS/terracotta-management modules.

And to not impact Ehcache, I was wondering whether we should create `ehcache/ehcache3-management` project, that would contain:

- standalone-management (ehcache management module)
- clustered-management (for client-side monitoring, depending on the management entity module)
- ehcache-entity-management (for server-side monitoring - exposing in voltron)

So using ehcache standalone with this project, or the big shaded clustered jar with this project, which would be optional, makes more sense to me than having some management stuff embedded with missing dependencies :-)

mathieucarbou:
@anthonydahanne : FYI: I've opened Terracotta-OSS/terracotta-management#43 : by the time Ehcache team takes a decision we can set the serialVersionUID field everywhere to avoid future impacts. We'll increment it only when we change some serialized fields...

ljacomet:
Ok, then relevant discussion / proposition: https://groups.google.com/forum/#!topic/ehcache-dev/hg8LLwxNZfw

PR #1133, opened 2016-06-06
Replaces #1128 to solve classpath issue.

mathieucarbou:
Replaces #1128 to solve classpath issue.

- management jars must be the same everywhere
- Ehcache master can point to management's snapshots

### DIFF
--- a/management/build.gradle
+++ b/management/build.gradle
@@ -20,3 +20,7 @@ dependencies {
   compile project(':api'), project(':core'), project(':impl')
   compile group: 'org.terracotta.management', name: 'management-registry', version: parent.managementVersion
 }
+
+repositories {
+  maven { url "http://snapshots.terracotta.org/" }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ import scripts.*
 ext {
   baseVersion = '3.1.0-SNAPSHOT'
   offheapVersion = '2.2.2'
-  managementVersion = '2.5.0'
+  managementVersion = '2.6.0-SNAPSHOT'
   statisticVersion = '1.1.0'
   jcacheVersion = '1.0.0'
   slf4jVersion = '1.7.7'


mathieucarbou:
@anthonydahanne @ljacomet:

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all dependencies for configuration ':management:compileClasspath'.
> Could not find org.terracotta.management:management-registry:2.6.0-SNAPSHOT.
  Required by:
      org.ehcache.modules:management:3.1.0-SNAPSHOT
```

It seems like the snapshots of tc-management are not synced with Sonatype ? The build cannot find them...

mathieucarbou:
@ljacomet : commit 904aa64 fixes the build locally but I don't know if it is the right way to do it in Gradle ?

management/build.gradle
```diff
@@ -20,3 +20,9 @@ dependencies {
   compile project(':api'), project(':core'), project(':impl')
   compile group: 'org.terracotta.management', name: 'management-registry', version: parent.managementVersion
 }
+
+repositories {
+  mavenCentral()
```
ljacomet: This is not needed as the parent already declares it.

management/build.gradle
```diff
@@ -20,3 +20,9 @@ dependencies {
...
+  maven { url "http://repo.terracotta.org/maven2" }
+  maven { url "http://snapshots.terracotta.org/" }
```
ljacomet: This raises an interesting question. Ehcache 3 dependencies should be coming from Maven Central only - otherwise we break the implicit contract.
So if we start having to depend on the terracotta-management product in client libraries, it needs to be published to Sonatype snapshots and made ready for Central publication on release.
@mathieucarbou @anthonydahanne makes sense?

mathieucarbou: Yes it make sense... We'll need to open an issue thought to make sure we are ready for central publication...

anthonydahanne:
@mathieucarbou @ljacomet it's already the case : http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.terracotta.management%22
but there's a catch : some of these modules depend on voltron snapshots, not even published on OSS Sonatype, but on tc snapshots...
let's chat about this together to find out what an acceptable solution would be.
Either Voltron team publishes the needed snapshots to central
Or we only publish the M&M modules needed by ehcache3 and not relying on voltron to central

Issue #859, opened 2016-03-15
Define Server Side Cache Configuration Contents and Compatibility Rules


chrisdennis:
This includes functionality for validating the configuration against the configuration provided by a client attempting to retrieve the cache.  The server-side cache configuration type will be a folding of both cluster specific configuration pulled from the clustered "resource" for this cache, in addition to some aspects of the general configuration of this cache (it may indeed be simplest to embed the actual cache configuration object within this type - if that is possible).   We need to decide what properties are part of the server side cache configuration, and what the rules are for compatibility on those properties.

ljacomet:
See #1130 for one invalid use case that needs a proper user level exception,

Issue #1130, opened 2016-06-03
Changing a cache from using a fixed pool to a shared one gives internal error message


GaryWKeim:
```
Caused by: java.lang.ClassCastException: org.ehcache.clustered.common.ServerStoreConfiguration$PoolAllocation$Shared cannot be cast to org.ehcache.clustered.common.ServerStoreConfiguration$PoolAllocation$Fixed
	at org.ehcache.clustered.common.ServerStoreCompatibility.verify(ServerStoreCompatibility.java:50)
	at org.ehcache.clustered.server.EhcacheActiveEntity.validateServerStore(EhcacheActiveEntity.java:717)
```

ljacomet:
This should be tackled by #859 which is still in progress.

PR #1145, opened 2016-06-07
Issue #859 Define Server Side Cache Configuration Contents and Compat…

gibsong:
…ibility Rules

### DIFF
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/DefaultClusteringServiceTest.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/DefaultClusteringServiceTest.java
@@ -55,6 +55,7 @@ import org.junit.Test;
 import org.terracotta.connection.ConnectionPropertyNames;
 import org.terracotta.consensus.entity.CoordinationServerEntityService;
 import org.terracotta.consensus.entity.client.ClientCoordinationEntityService;
+import org.terracotta.entity.ClientDescriptor;
 import org.terracotta.exception.EntityNotFoundException;
 
 import java.net.URI;
@@ -67,6 +68,7 @@ import java.util.Set;
 import static org.ehcache.clustered.client.config.ClusteredResourceType.Types.FIXED;
 import static org.ehcache.clustered.client.config.ClusteredResourceType.Types.SHARED;
 import static org.ehcache.clustered.client.internal.service.TestServiceProvider.providerContaining;
+import org.ehcache.clustered.common.ClusteredStoreCreationException;
 import static org.ehcache.config.ResourceType.Core.DISK;
 import static org.ehcache.config.ResourceType.Core.HEAP;
 import static org.ehcache.config.ResourceType.Core.OFFHEAP;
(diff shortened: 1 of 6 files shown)


ljacomet:
This PR lacks testing from the Ehcache API level which would show that our exception handling in the face of `auto-create` is wrong.

The way `org.ehcache.clustered.client.internal.service.DefaultClusteringService#getServerStoreProxy` works is a problem as validation failure will always be hidden by creation failures in `auto-create` mode.

cljohnso:
@ljacomet, I believe your assertion is that a validation failure with `auto-create` is hidden by a creation failure isn't correct.  The exception that gets returned from the server in this case is a `ClusteredStoreValidationException` (which subclasses `RuntimeException`).  That `RuntimeException` is rethrown from `EhcacheClientEntity.validateServerStore` and is emitted from `DefaultServerStore.getServerStoreProxy`.  (This is confirmed by a couple of tests in `DefaultServerStoreProxyTest`.)  We probably need to have it wrapped in `EhcacheClientEntity` so we get the full client-side stack trace in the exception -- I think the way it is now only gets the server-side stack trace -- but I believe the test path you seek is covered.

cljohnso:
I looked at the tests in `DefaultClusteringServiceTest` again and they don't actually exercise the logic in `ServerStoreCompatibility` ... some tests need to be added in `DefaultClusteringServiceTest` (or a supplementary class) to ensure proper operation/visibility of validation failures.

cljohnso:
I cobbled together a sample and @ljacomet is right ... the `ClusteredStoreValidationException` is hidden under a `CachePersistenceException` by `EhcacheClientEntity.invokeInternal`.  A scattering of changes in `DefaultClusteringService` and `EhcacheClientEntity` are needed to properly expose the `ClusteredStoreValidationException`.  We should discuss ...

cljohnso:
@ljacomet indicated issue #1130 should be handled in this PR -- this necessitates some changes in `ServerStoreCompatibility` to ensure the server-side and client-specified pool allocations are of the same type before casting.

gibsong:
I was able to expose the ClusteredStoreValidationException via a unit test.  I still need to write some more tests to see if there are any other bugs.  A discussion would help find additional cases I might not be seeing.

gibsong:
ok I can combine #1130 with this PR.  I have already made the fix to prevent a user from changing a PoolAllocation from fixed to shared and vice versa.

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
```diff
@@ -70,12 +73,24 @@ public void verify(ServerStoreConfiguration serverConfiguration, ServerStoreConf
...
     isCompatible &= compareField(sb, "valueSerializerType", serverConfiguration.getValueSerializerType(), clientConfiguration.getValueSerializerType());
+    isCompatible &= compareConsistencyField(sb, serverConfiguration.getConsistency(), clientConfiguration.getConsistency());
 
     if (!isCompatible) {
       throw new ClusteredStoreValidationException(sb.toString());
     }
   }
 
+  private static boolean compareConsistencyField(StringBuilder sb, Consistency serverConsistencyValue, Consistency clientConsistencyValue)
+  {
```
ljacomet: Code formatting mismatch

ljacomet:
Please squash some of these commits - goal is to have merged commits that enable `git bisect`, that is at each commit, tests should pass.

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
```diff
@@ -70,12 +73,24 @@ public void verify(ServerStoreConfiguration serverConfiguration, ServerStoreConf
     isCompatible &= compareField(sb, "actualValueType", serverConfiguration.getActualValueType(), clientConfiguration.getActualValueType());
     isCompatible &= compareField(sb, "keySerializerType", serverConfiguration.getKeySerializerType(), clientConfiguration.getKeySerializerType());
     isCompatible &= compareField(sb, "valueSerializerType", serverConfiguration.getValueSerializerType(), clientConfiguration.getValueSerializerType());
+    isCompatible &= compareConsistencyField(sb, serverConfiguration.getConsistency(), clientConfiguration.getConsistency());
```
lorban: There doesn't seem to be a test for this check

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
```diff
@@ -0,0 +1,485 @@
...
+import org.ehcache.clustered.common.ServerStoreConfiguration.PoolAllocation.Fixed;
+import org.ehcache.clustered.common.ServerStoreConfiguration.PoolAllocation.Shared;
+import static org.junit.Assert.fail;
+import org.junit.Test;
+
+/**
+ *
+ * @author GGIB
+ */
+public class ServerStoreCompatibilityTest
```
lorban: You should observe the existing code formatting. There is some freedom but not around curly braces placement.

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
```diff
@@ -0,0 +1,485 @@
...
+    ServerStoreCompatibility serverStoreCompatibility = new ServerStoreCompatibility();
+
+    try
+    {
+      serverStoreCompatibility.verify(serverConfiguration, clientConfiguration);
+      fail("Expected ClusteredStoreValidationException");
+    }
+    catch(ClusteredStoreValidationException e)
+    {
+      System.out.println("errorMsg: " + e.getMessage());
```
lorban: No `System.out.println()` please.

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
```diff
@@ -0,0 +1,485 @@
...
+      Assert.assertTrue("test failed", e.getMessage().equals(ERROR_MESSAGE_BASE + "storedKeyType existing: " + serverConfiguration.getStoredKeyType() + ", desired: " + clientConfiguration.getStoredKeyType()));
```
lorban: Better use `assertThat(... , is(true))`

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
```diff
@@ -0,0 +1,485 @@
...
+
+    ServerStoreCompatibility serverStoreCompatibility = new ServerStoreCompatibility();
+
+    try
+    {
+      serverStoreCompatibility.verify(serverConfiguration, clientConfiguration);
+    }
+    catch(Exception e)
+    {
+      fail("No Exception Expected!");
```
lorban: It's much better to declare the test method with `throws Exception` than catching it and failing the test without knowing what the exception was.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/DefaultClusteringServiceTest.java
```diff
@@ -1350,6 +1352,463 @@ public void testFullDestroyAll() throws Exception {
...
+        .resourcePool("sharedPrimary", 16, MemoryUnit.MB, "serverResource1")
+        .resourcePool("sharedSecondary", 16, MemoryUnit.MB, "serverResource2")
+        .resourcePool("sharedTertiary", 32, MemoryUnit.MB)
+        .build();
+    DefaultClusteringService creationServiceGood = new DefaultClusteringService(creationConfigGood);
+
+    try {
+      creationServiceGood.start(null);
+    } catch (Exception e) {
+      fail("There should not be an error on start");
```
lorban: Make the method `throws Exception` instead of catching and failing.

gibsong:
I checked in this code but had some failing tests.  They appeared to be only failing on Windows and were not related to my code changes.  They were failing on the master branch.

The following tests were failing:
1.
class: EhcacheSettingsProviderTestTest
test: test_standalone_ehcache()

2. (i had to comment out the above test and then this failed)
class: TerminatedServerTest
test: testTerminationBeforeCacheManagerClose()

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
gibsong: I actually have a test in ServerStoreCompatibilityTest, it's called testConsitencyMismatch()

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
gibsong: sorry, bad habit formatting that way.  I changed it

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
gibsong: its been removed

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
gibsong: I changed to check to use assertThat().  Why is this better?  Sorry this is my first time with junit.  More familiar with testng.

clustered/common/src/test/java/org/ehcache/clustered/common/ServerStoreCompatibilityTest.java
gibsong: you are right.  I made the change.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/DefaultClusteringServiceTest.java
gibsong: i updated it as requested.  Nice catch.

.gitattributes
```diff
@@ -1,6 +1,8 @@
 # Set the default behavior to LF, because checkstyle enforces it
 * text eol=lf
 
+*.bat eol=crlf
```
ljacomet: No idea why these changes appear here but they should not. They are already on master following the Gradle 2.14 upgrade

gibsong: This pull request was accidentally closed. I then pushed my PR changes, which included the Gradle upgrade, and re-opened the PR.  Could this be why they are showing?  Maybe it did a diff from the closed PR code to the new re-opened PR code.

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
gibsong: fixed formatting

PR #1149, opened 2016-06-08
Issue 1076

AbfrmBlr:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
```diff
@@ -25,8 +25,10 @@
 
   private final K key;
   private final V value;
+  private final long timeStamp;
+  private final byte isFirst;
```
albinsuresh: I would have declared this field as a boolean one and taken care of the value conversion(to 0 or 1) during the encoding/decoding.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -134,10 +140,10 @@ public boolean containsKey(final K key) throws StoreAccessException {
   @Override
   public PutStatus put(final K key, final V value) throws StoreAccessException {
     putObserver.begin();
-    PutOperation<K, V> operation = new PutOperation<K, V>(key, value);
+    PutOperation<K, V> operation = new PutOperation<K, V>(key, value, timeSource.getTimeMillis(), true);
```
albinsuresh: An additional helper constructor in the operation classes that would handle the _isFirst_ flag implicitly would have been nice.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -44,24 +53,50 @@ public ChainResolver(final OperationsCodec<K, V> codec) {
    * @param key a key
    * @return an entry with the resolved operation for the provided key as the key
    * and the compacted chain as the value
-   * @throws ClassNotFoundException
    */
-  public ResolvedChain<K, V> resolve(Chain chain, K key) {
+  public ResolvedChain<K, V> resolve(Chain chain, K key, long now) {
```
albinsuresh: You may wanna declare this additional parameter in the javadoc above. And as I read the javadoc for the return, it is wrong. My bad. I forgot to correct it when I changed the return type. While you're at it could you correct that too, please.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -44,24 +53,50 @@ public ChainResolver(final OperationsCodec<K, V> codec) {
...
+          expiration = expiry.getExpiryForCreation(key, result.getValue());
+          isNew = true;
+        } else {
+          expiration = expiry.getExpiryForUpdate(key, new ValueSupplier<V>() {
+            @Override
+            public V value() {
+              return previousResult != null ? previousResult.getValue() : null;
+            }
+          }, result.getValue());
+        }
```
albinsuresh: Looks like you forgot to set the _isNew_ flag to *false* here in this else block.

albinsuresh: Verified that it's a bug by running a test that does 2 sequential puts and a subsequent get on a cache with a decent expiry time. The get call is creating the new `PutOperation` in line 99 with the _isNew_ flag set to true.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
AbfrmBlr: How would that help?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
AbfrmBlr: Yes that's a bug

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
AbfrmBlr: cool

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
ljacomet: Well, readability of code. It would make all logic cleaner but for codec methods. Also, you are adding a byte for a bit. Why not use the most significant bit of the timestamp since that one can never be negative?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -44,24 +53,50 @@ public ChainResolver(final OperationsCodec<K, V> codec) {
...
-      Operation<K, V> operation = codec.decode(payload);
+      operation = codec.decode(payload);
+      final Result<V> previousResult = result;
+      Duration expiration;
       if(key.equals(operation.getKey())) {
         result = operation.apply(result);
+        if(result == null) {
+          continue;
+        }
+        if((previousResult == null & operation.isFirst())) {
```
ljacomet: missing an `&` here

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/BaseKeyValueOperation.java
AbfrmBlr: Removing this byte can be done by masking it in opcode as well and I have that as TODO already. So it does not make much difference

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -44,24 +53,50 @@ public ChainResolver(final OperationsCodec<K, V> codec) {
...
+          expiration = expiry.getExpiryForCreation(key, result.getValue());
+          isNew = true;
+        } else {
+          expiration = expiry.getExpiryForUpdate(key, new ValueSupplier<V>() {
+            @Override
+            public V value() {
+              return previousResult != null ? previousResult.getValue() : null;
```
ljacomet: How could previousResult be null if you are doing an expiry for update check?

Returning null is not legal per the Expiry contract.

ljacomet:
Looking at the added logic based on flags and the fact that invoking `Expiry.getExpiryForUpdate` remains a problem, why not improve the timestamp we store.
Let's keep sending a timestamp in the appended operations. But when we do a replace, we could store the effective expiry time, to not have to recompute it again on the next `resolve`. Of course, we need to differentiate between these two times but it is trivial to store the effective expiry time as its negative value.
I believe this would simplify somewhat the logic around the flags `isFirst` and `isNew` and take care of never invoking expiry methods with a chance for null in params.

AbfrmBlr:
Cool, I will give a try to that

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -42,26 +53,72 @@ public ChainResolver(final OperationsCodec<K, V> codec) {
...
+        }
       } else {
         payload.flip();
         chainBuilder = chainBuilder.add(payload);
       }
     }
-    Operation<K, V> resolvedOperation = null;
-    if(result != null) {
-      resolvedOperation = new PutOperation<K, V>(key, result.getValue());
+    if(result != null & operation != null) {
```
ljacomet: Why is the operation check needed here?
Also, you should use `&&` here and not the bit manipulation operator.

AbfrmBlr: oh ya, that's a left over when operation had status.will clear that up

albinsuresh:
@AbfrmBlr As we discussed immediate expiry of operations is still not handled. That is, in the current PR all the expiry calculations are done at the chain resolver level. But during a mutative operation like *put* the chain resolver only resolves the "previous" operations that it gets from the cluster and all the expiry calculations are done for these previous operations and not the operation that is being executed. So you may wanna add expiry checks at the store level as well for the operation that is being performed. Otherwise we'll miss the immediate expiry cases and it will affect the store level stats for the operations. I hope you're plannnig on handling this in your follow-up PR.

Issue #1144, opened 2016-06-07
Bulk operations for clustered caches


ljacomet:
* putAll
* getAll
* removeAll

ljacomet:
Had a chat with @albinsuresh and currently recommend the following approach:
* Extract the bulk compute related functions in `Ehcache` to classes instead of anonymous inner ones
* Based on the type of the received function, act accordingly

This is clearly not ideal but we know we will have to do something around generic functions and probably move to more specific ones in order to make clustered `compute*` work. But given the fact that we are not yet integrating with `CacheLoaderWriter` I want the search for the most generic solution to be postponed to a later stage.

PR #1154, opened 2016-06-10
Issue #1144 Clustered store bulk operations

albinsuresh:
This is a bloody hack of the store's bulk compute methods tailored to work with only cache's get, put and remove methods ONLY. The stats are not wired in. There is no test coverage except for an integration test that I wrote to check if things would work with this hack. Raising this PR just to validate my approach. Don't wanna waste time doing this if this isn't what's expected.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
@@ -34,6 +34,7 @@ import org.ehcache.clustered.common.Consistency;
 import org.ehcache.clustered.common.store.Chain;
 import org.ehcache.config.ResourceType;
 import org.ehcache.core.CacheConfigurationChangeListener;
+import org.ehcache.core.Ehcache;
 import org.ehcache.core.internal.util.ConcurrentWeakIdentityHashMap;
 import org.ehcache.core.spi.function.BiFunction;
 import org.ehcache.core.spi.function.Function;
@@ -45,7 +46,6 @@ import org.ehcache.core.spi.store.tiering.AuthoritativeTier;
 import org.ehcache.core.statistics.StoreOperationOutcomes;
 import org.ehcache.core.spi.time.TimeSource;
 import org.ehcache.core.spi.time.TimeSourceService;
-import org.ehcache.expiry.Expiry;
 import org.ehcache.impl.internal.events.NullStoreEventDispatcher;
 import org.ehcache.spi.service.ServiceProvider;
 import org.ehcache.spi.service.Service;
@@ -57,11 +57,13 @@ import org.terracotta.statistics.observer.OperationObserver;
 import java.nio.ByteBuffer;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static org.ehcache.core.exceptions.StorePassThroughException.handleRuntimeException;
 import static org.ehcache.core.internal.service.ServiceLocator.findSingletonAmongst;
 import static org.terracotta.statistics.StatisticBuilder.operation;
 
@@ -123,23 +125,23 @@ public class ClusteredStore<K, V> implements AuthoritativeTier<K, V> {
     }
   }
 
-  private V getInternal(K key) {
-    Chain chain = storeProxy.get(key.hashCode());
-    V value;
-    if(!chain.isEmpty()) {
-      ResolvedChain<K, V> resolvedChain = resolver.resolve(chain, key, timeSource.getTimeMillis());
+  private V getInternal(K key) throws StoreAccessException {
+    V value = null;
+    try {
+      Chain chain = storeProxy.get(key.hashCode());
+      if(!chain.isEmpty()) {
+        ResolvedChain<K, V> resolvedChain = resolver.resolve(chain, key, timeSource.getTimeMillis());
 
-      Chain compactedChain = resolvedChain.getCompactedChain();
-      storeProxy.replaceAtHead(key.hashCode(), chain, compactedChain);
+        Chain compactedChain = resolvedChain.getCompactedChain();
+        storeProxy.replaceAtHead(key.hashCode(), chain, compactedChain);
 
-      Result<V> resolvedResult = resolvedChain.getResolvedResult(key);
-      if(resolvedResult != null) {
-        value = resolvedResult.getValue();
-      } else {
-        return null;
+        Result<V> resolvedResult = resolvedChain.getResolvedResult(key);
+        if (resolvedResult != null) {
+          value = resolvedResult.getValue();
+        }
       }
-    } else {
-      return null;
+    } catch (RuntimeException re) {
+      handleRuntimeException(re);
     }
     return value;
   }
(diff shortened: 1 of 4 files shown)


albinsuresh:
I'll be adding one more commit to this PR that fixes exception contract of all the store methods wrapping all `RuntimeException`s to `StoreAccessException`s. I have the changes ready. But not pushing it now as that'd screw up the whole formatting of `ClusteredStore` class.

albinsuresh:
test this please

albinsuresh:
Fixed the stats bug in bulk ops and added tests to verify the same.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1031,4 +969,116 @@ public V getValue() {
...
   }
 
+  // The compute function that will return the keys to their NEW values, taking the keys to their old values as input;
+  // but this could happen in batches, i.e. not necessary containing all of the entries of the Iterable passed to this method
+  public static class PutAllFunction<K, V> implements Function<Iterable<? extends Map.Entry<? extends K, ? extends V>>, Iterable<? extends Map.Entry<? extends K, ? extends V>>> {
+
+    private final Logger logger;
+    private final Map<K, V> entriesToRemap;
+    private final Expiry<? super K, ? super V> expiry;
+    private final AtomicInteger actualPutCount = new AtomicInteger();
```
lorban: Wouldn't a `volatile int` be good enough?

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1031,4 +969,116 @@ public V getValue() {
...
+        } catch (RuntimeException re) {
+          logger.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
+          return true;
+        }
+      }
+
+      return Duration.ZERO.equals(duration);
+    }
+
+    public AtomicInteger getActualPutCount() {
```
lorban: Even if you stick with the `AtomicInteger` field, I'd make this method return `int`.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -152,44 +155,66 @@ public boolean containsKey(final K key) throws StoreAccessException {
...
+        putObserver.end(StoreOperationOutcomes.PutOutcome.PUT);
+        break;
+      case UPDATE:
+        putObserver.end(StoreOperationOutcomes.PutOutcome.REPLACED);
+        break;
+      case NOOP:
+        putObserver.end(StoreOperationOutcomes.PutOutcome.NOOP);
+        break;
+      default:
+        throw new IllegalArgumentException("Invalid put status: " + status);
```
lorban: I'm being picky here, but this should throw `AssertionError` instead of `IllegalArgumentException` as this exception can never be caused by user code, but only by an internal bug.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -152,44 +155,66 @@ public boolean containsKey(final K key) throws StoreAccessException {
...
+      Chain chain = storeProxy.getAndAppend(key.hashCode(), payload);
+      ResolvedChain<K, V> resolvedChain = resolver.resolve(chain, key, timeSource.getTimeMillis());
+      if(resolvedChain.getResolvedResult(key) == null) {
+        return PutStatus.PUT;
+      } else {
+        return PutStatus.UPDATE;
+      }
+    } catch (RuntimeException re) {
+      handleRuntimeException(re);
+      return null;
```
lorban: As bizzare as it may sound, I would return `PutStatus.NOOP` here. This will make zero difference at runtime but would help code checkers understand this method will never actually return null.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ClusteredStoreTest.java
```diff
@@ -39,18 +42,27 @@
...
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.*;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class ClusteredStoreTest {
 
+  private static final RuntimeException RUNTIME_EXCEPTION = new RuntimeException();
```
lorban: That's a weird construct for a test. Why doing this over inlining `new RuntimeException` everywhere it is used?

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -326,24 +327,10 @@ public void clear() {
...
-        // put all the entries to get ordering correct
-        for (K key : keys) {
-          computeResult.put(key, null);
-        }
-
-        return computeResult.entrySet();
-      }
-    };
-
+    GetAllFunction<K, V> getAllFunction = new GetAllFunction<K, V>();
```
lorban: I wouldn't bother with a variable, just pass the `new GetAllFunction<K, V>()` to the `bulkComputeIfAbsent`. That would make the code an tad cleaner.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -463,32 +422,11 @@ public void removeAll(final Set<? extends K> keys) throws BulkCacheWritingExcept
...
-            actualRemoveCount.incrementAndGet();
-          }
-          results.put(key, null);
-          }
-
-          return results.entrySet();
-        }
-      };
 
+    RemoveAllFunction<K, V> removeAllFunction = new RemoveAllFunction<K, V>();
```
lorban: I would declare this variable inside the `try / catch` block too.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1031,4 +969,116 @@ public V getValue() {
...
+    }
+
+    public AtomicInteger getActualPutCount() {
+      return actualPutCount;
+    }
+  }
+
+  public static class RemoveAllFunction<K, V> implements Function<Iterable<? extends Map.Entry<? extends K, ? extends V>>, Iterable<? extends Map.Entry<? extends K, ? extends V>>> {
+
+    private final AtomicInteger actualRemoveCount = new AtomicInteger();
```
lorban: Same as above: wouldn't `volatile int` do the job?

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1031,4 +969,116 @@ public V getValue() {
...
+        if (existingValue != null) {
+          actualRemoveCount.incrementAndGet();
+        }
+        results.put(key, null);
+      }
+
+      return results.entrySet();
+    }
+
+    public AtomicInteger getActualRemoveCount() {
```
lorban: Same as above: I'd make this method return `int` even if you keep the `AtomicInteger`.

core/src/main/java/org/ehcache/core/Ehcache.java
albinsuresh: This atomic integer is accessed from the `ClusteredStore.bulkCompute` method and incremented while doing the looped puts. That's why I kept it as this atomic integer so that i can easily access it and increment it from the store. But I could replace that by making it a volatile int and adding an additional setter method for it that can be used from `ClusteredStore.bulkCompute`. Should I do that? Isn't this a cleaner way?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
albinsuresh: I agree.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
albinsuresh: Sure. Will do.

core/src/main/java/org/ehcache/core/Ehcache.java
albinsuresh: Same case as in `PutAllFunction`

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ClusteredStoreTest.java
albinsuresh: To be honest this was a usage that I saw in the `BaseOnHeapStoreTest` and just used it the same way. Now there are multiple tests that throws `RuntimeException`. So it made sense to have it defined once and reuse it everywhere rather than throwing new exceptions. Either way would work here to be honest. But still I'm wondering why this is considered a bad practice.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -401,39 +388,11 @@ public void putAll(final Map<? extends K, ? extends V> entries) throws BulkCache
...
-            actualPutCount.incrementAndGet();
-            mutations.put(key, newValue);
-          }
-        }
-
-        // Finally return the values to be installed in the Cache's Store
-        return mutations.entrySet();
-      }
-    };
+    PutAllFunction<K, V> putAllFunction = new PutAllFunction<K, V>(logger, entriesToRemap, runtimeConfiguration.getExpiry());
```
albinsuresh: Sure

core/src/main/java/org/ehcache/core/Ehcache.java
albinsuresh: Sure.

core/src/main/java/org/ehcache/core/Ehcache.java
albinsuresh: Same as in the `PutAllFunction` case.

core/src/main/java/org/ehcache/core/Ehcache.java
albinsuresh: Sure.

core/src/main/java/org/ehcache/core/Ehcache.java
lorban: I was thinking of adding an extra `incrementActualPutCount()` method in `PutAllFunction` so you could write `putAllFunction.incrementActualPutCount()` instead of `putAllFunction.getActualPutCount().incrementAndGet()`.

But it's probably not worth it, so we can probably ignore this idea.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ClusteredStoreTest.java
ljacomet: I would need to double check, but I believe reusing an exception instance means its stack trace will be filled only once, the first time it is thrown (or accessed) ... but after that you will always see the same.

PR #1155, opened 2016-06-10
Issue 1105: very basic concurrency strategy

lorban:


ljacomet:
You have a merge in that history - pretty sure that's not what you wanted.

lorban:
I can't even explain how that merge ended up in there.

AbfrmBlr:
I think we should try to differentiate between reads and mutations in concurrency startegy so that reads are not blocked on mutations which typically require passive replication (and FRS), making them slow.
If we do that I think we can get some advantage in read perf when we have passives and considerable mutations being done on cache.
I think we can do some perf tests to assert if moving reads to a different queue would really help.

ljacomet:
@AbfrmBlr Created #1178 following your suggestion
@lorban looks good to me once rebased

ljacomet:
Merged command line after a rebase

Issue #1123, opened 2016-06-03
Temporarily replace optimized serialization with plain java one


lorban:
Until we get #1107 done, we should replace the optimized serializer with a non-optimized one that is stateless.

PR #1159, opened 2016-06-13
Issue #1107 StateRepository implementation for DiskStore

ljacomet:
This is the implementation of the `StateRepository` concept for the `DiskStore`. It leverages the newly added SPI, properly handles the lifecycle of disk based `StateRepository`.

Upcoming work will add the same support for clustered.


ljacomet:
Took care of the compilation failures ... do not rebase then believe everything worked because it merged fine ...

lorban:
I'm not very keen on merging this PR as that would reintroduce the problem #1123 fixed. Shouldn't we wait until the clustered part of the work is done too?

api/src/main/java/org/ehcache/spi/persistence/PersistableResourceService.java
```diff
@@ -65,6 +64,22 @@
   PersistenceSpaceIdentifier create(String name, CacheConfiguration<?, ?> config) throws CachePersistenceException;
 
   /**
+   * Creates a named {@link StateRepository state repository} in the context of the given
```
chrisdennis: Create seems like the wrong verb here... the semantics seem to be "Get or create"?

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -0,0 +1,38 @@
...
+   * <P>
+   *   If the map existed already, it is returned with its content fully available.
+   * </P>
+   *
+   * @param name the map name
+   * @param <K> the map key type, must be {@code Serializable}
+   * @param <V> the map value type, must be {@code Serializable}
+   * @return a map
+   */
+  <K, V> ConcurrentMap<K, V> getPersistentConcurrentMap(String name);
```
chrisdennis: I think it was Ludovic who suggested bounding K and V to Serializable... are we no longer doing that?

api/src/main/java/org/ehcache/spi/persistence/package-info.java
```diff
@@ -15,7 +15,6 @@
  */
 
 /**
- * {@link org.ehcache.spi.persistence.PersistableResourceService PersistableResourceService} and other persistence
- * related SPI.
+ * {@link org.ehcache.spi.persistence.StateRepository StateRepository} and other persistence related SPI.
```
chrisdennis: Why did we change this?  Isn't PersistableResourceService still the "top level" class in this system?

ljacomet:
@lorban That is not correct, unless I missed something. Currently our `DefaultClusteringService` does not add a `PersistenceSpaceIdentifier` to the set of configurations, which means it is identified as _transient_ and not persistent and thus will properly load the stateless serializer.

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
ljacomet: Indeed, I considered it then did not do it .. let me look once more.

api/src/main/java/org/ehcache/spi/persistence/package-info.java
ljacomet: That's a rebase mistake.

api/src/main/java/org/ehcache/spi/persistence/PersistableResourceService.java
ljacomet: Will update javadoc indeed.

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
```diff
@@ -0,0 +1,120 @@
...
+        public boolean accept(File dir, String name) {
+          return name.endsWith(MAP_FILE_SUFFIX);
+        }
+      })) {
+        FileInputStream fis = new FileInputStream(file);
+        try {
+          ObjectInputStream oin = new ObjectInputStream(fis);
+          try {
+            ConcurrentMap<?, ?> map = (ConcurrentMap<?, ?>) oin.readObject();
+            knownMaps.put(file.getName().replace(MAP_FILE_SUFFIX, ""), map);
```
chrisdennis: Non case preserving filesystems will bite you here.  The map name will need to be stored in the file itself instead, and then name will need something like an md5 to avoid clashes due to casing.

lorban:
@ljacomet You're right. I once again missed that there is a `TransientProvider` and a `PersistentProvider`.

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -65,7 +66,19 @@
    * @see Serializer
    */
   public CompactJavaSerializer(ClassLoader loader) {
+    this(loader, new TransientStateRepository());
+  }
+
+  public CompactJavaSerializer(ClassLoader loader, StateRepository stateRepository) {
     this.loader = loader;
+    this.readLookup = stateRepository.getPersistentConcurrentMap("readLookup");
+    if (!readLookup.isEmpty()) {
```
chrisdennis: Is this check really necessary?

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -171,8 +184,7 @@ private int getOrAddMapping(ObjectStreamClass desc) throws IOException {
    */
   @Override
   public void close() {
```
chrisdennis: Could we remove the Closeable interface now then?

impl/src/main/java/org/ehcache/impl/serialization/TransientStateRepository.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * TransientStateRepository
+ */
+class TransientStateRepository implements StateRepository {
+
+  @Override
+  public <K, V> ConcurrentMap<K, V> getPersistentConcurrentMap(String name) {
```
chrisdennis: These should get saved in a data structure so that repeated calls with the same name return the same map.

impl/src/main/java/org/ehcache/impl/serialization/TransientStateRepository.java
```diff
@@ -0,0 +1,35 @@
...
+import org.ehcache.spi.persistence.StateRepository;
+
+import java.io.ObjectStreamClass;
+import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * TransientStateRepository
+ */
+class TransientStateRepository implements StateRepository {
```
chrisdennis: This is so close to what I really want... :cry:

impl/src/main/java/org/ehcache/impl/serialization/TransientStateRepository.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.impl.serialization;
+
+import org.ehcache.impl.internal.concurrent.ConcurrentHashMap;
+import org.ehcache.spi.persistence.StateRepository;
+
+import java.io.ObjectStreamClass;
```
chrisdennis: Unnecessary imports.

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -52,7 +53,7 @@
 
   private final AtomicInteger nextStreamIndex = new AtomicInteger(0);
```
chrisdennis: I think this is actually my bug and not yours... but since you're in here... this needs to be updated after reading the mappings from the file.

ljacomet: Yes! I was puzzled by this and wanted to chat with you about it - will fix.

impl/src/main/java/org/ehcache/impl/internal/spi/serialization/DefaultSerializationProvider.java
```diff
@@ -274,16 +303,19 @@ private AbstractProvider(Map<Class<?>, Class<? extends Serializer<?>>> serialize
         throw new RuntimeException(e);
       }
     }
+  }
 
-    @Override
-    public void stop() {
-      // no-op
-    }
+  private static class SerializerTuple<T> {
```
chrisdennis: This class should really be immutable, but that said I disagree with the way this stuff is being lifecycled, or at least the way the lifecycle is being driven.  Maybe we can have a chat at some point?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -282,6 +313,15 @@ public FileBasedPersistenceContext createPersistenceContextWithin(PersistenceSpa
     }
   }
 
+  private boolean isKnownPersistenceSpaceIdentifier(PersistenceSpaceIdentifier<?> space) {
+    for (PersistenceSpaceIdentifier<LocalPersistenceService> persistenceSpace : knownPersistenceSpaces.values()) {
```
chrisdennis: Isn't this just: `knownPersistenceSpaces.containsValue(space)`?

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -65,7 +66,19 @@
    * @see Serializer
    */
   public CompactJavaSerializer(ClassLoader loader) {
+    this(loader, new TransientStateRepository());
+  }
+
+  public CompactJavaSerializer(ClassLoader loader, StateRepository stateRepository) {
     this.loader = loader;
+    this.readLookup = stateRepository.getPersistentConcurrentMap("readLookup");
```
chrisdennis: Should we have a name that ties to the FQCN instead of (or maybe in addition to) the field name or purpose here: e.g "CompactJavaSerializer-ObjectStreamClassIndex"

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -65,7 +66,19 @@
...
+    if (!readLookup.isEmpty()) {
+      for (Entry<Integer, ObjectStreamClass> entry : readLookup.entrySet()) {
+        if (writeLookup.putIfAbsent(new SerializableDataKey(disconnect(entry.getValue()), true), entry.getKey()) != null) {
+          throw new AssertionError("Corrupted data " + readLookup.toString());
```
chrisdennis: the `toString()` call is superfluous right?

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
ljacomet: No

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
ljacomet: Yes

impl/src/main/java/org/ehcache/impl/serialization/TransientStateRepository.java
ljacomet: Indeed

impl/src/main/java/org/ehcache/impl/config/serializer/DefaultSerializationProviderConfiguration.java
```diff
@@ -104,6 +105,14 @@ public DefaultSerializationProviderConfiguration(DefaultSerializationProviderCon
...
 
+    if(persistentConstructorPresent = isConstructorPresent(serializerClass, ClassLoader.class, StateRepository.class)) {
+      if (!overwrite && persistentSerializers.containsKey(serializableClass)) {
+        throw new IllegalArgumentException("Duplicate persistent serializer for class : " + serializableClass.getName());
+      } else {
+        persistentSerializers.put(serializableClass, serializerClass);
+      }
+    }
+
     if(persistentConstructorPresent = isConstructorPresent(serializerClass, ClassLoader.class, FileBasedPersistenceContext.class)) {
```
ljacomet: Indeed

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
ljacomet: Correct, changed it at one point but can now be reverted back

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
ljacomet: Actually, will open an issue and fix elsewhere since it needs to be fixed on 3.0 as well.

AbfrmBlr:
tests are failing

ljacomet:
Pushed an updated version - still unresolved comments so do not merge for now.

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -0,0 +1,39 @@
...
+   * <P>
+   *   If the map existed already, it is returned with its content fully available.
+   * </P>
+   *
+   * @param name the map name
+   * @param <K> the map key type, must be {@code Serializable}
+   * @param <V> the map value type, must be {@code Serializable}
+   * @return a map
+   */
+  <K extends Serializable, V extends Serializable> ConcurrentMap<K, V> getPersistentConcurrentMap(String name);
```
lorban: I don't like the `getPersistentConcurrentMap` method name as the persistence is an implementation detail, as demonstrated by your `TransientStateRepository` impl.

How about something like `getStateMap` instead?

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
lorban: `DefaultLocalPersistenceService` already has some logic that tries to create universally correct file names. Ideally, it should be reused.

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
lorban: +1 otherwise it will become messy when multiple disparate classes do use the state repository.

AbfrmBlr:
test this please

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
ljacomet: `TransientStateRepository` is really an internal detail for the `CompactJavaSerializer`. The whole point of the `StateRepository` is to provide maps with the same persistence level as the authority.

ljacomet:
Pushed an update that should resolve the lifecycle concerns that @chrisdennis had.
Did not yet address the concerns about file names.

ljacomet:
Updated to resolve the naming problem:
* Strictly limited valid names for internal ones
* `FileBasedStateRepository` encodes the name of the map in the file instead of as its name

Issue #1157, opened 2016-06-13
Server eviction should be notified to all interested clients


lorban:
`ServerStore.append()`, `ServerStore.getAndAppend()` (and maybe `ServerStore.replaceAtHead()` too) might make the server evict mappings. When that happens, clients that have faulted those mappings should be notified and invalidate them from the higher tiers.

PR #1162, opened 2016-06-14
 #1157 notify all interested clients when a chain eviction occurs on the server

lorban:


clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
```diff
@@ -28,17 +31,48 @@
...
 
 class OffHeapChainMap<K> implements MapInternals {
 
+  interface ChainMapEvictionListener<K> {
+    void onEviction(K key);
+  }
+
   private final ReadWriteLockedOffHeapClockCache<K, InternalChain> heads;
   private final OffHeapChainStorageEngine chainStorage;
+  private final List<ChainMapEvictionListener<K>> evictionListeners = new CopyOnWriteArrayList<ChainMapEvictionListener<K>>();
```
ljacomet: Why not a single one? Do we really need multiple here?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -29,14 +32,29 @@
 public class OffHeapServerStore implements ServerStore {
 
   private final List<OffHeapChainMap<Long>> segments;
+  private final List<ServerStoreEvictionListener> listeners = new CopyOnWriteArrayList<ServerStoreEvictionListener>();
```
ljacomet: Why not a single one? Seeing other places, it looks like we stack three collections on top of each other - that seems overkill.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ResponseCodec.java
```diff
@@ -94,6 +96,15 @@
         buffer.putInt(((ClientInvalidateAll) response).getInvalidationId());
         return buffer.array();
       }
+      case SERVER_INVALIDATE_HASH: {
+        ServerInvalidateHash serverInvalidateHash = (ServerInvalidateHash) response;
+        byte[] cacheIdBytes = serverInvalidateHash.getCacheId().getBytes(Charset.forName("UTF-8"));
```
ljacomet: We have inconsistent cache id encoding between this class and `ServerStoreOpCodec`. Would be good to pick one and only one.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
lorban: You're right, there will never be a need for multiple listeners. Let me rationalize all that.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
lorban: Nope, you're right.

clustered/common/src/main/java/org/ehcache/clustered/common/messages/ResponseCodec.java
lorban: I'm going to commonize the code that encodes strings, but the more I look at the message / codec code the more I think this should all be handled with something like Protobuf.

PR #1166, opened 2016-06-14
Fixes #1165 Proper state recovery in compact serializer

### LINKED ISSUE #1165: CompactPersistentJavaSerializer load bug
When the `CompactPersistentJavaSerializer` loads its state, it does not set the `nextStreamIndex` `AtomicInteger` to a proper value. This will lead to index override which will cause failures.

ljacomet:
Opened PR for review - GitHub says it cannot merge but git command line does.


ljacomet:
replaced by #1167 and #1168

PR #1169, opened 2016-06-15
Fix #1160 - null classloader

### LINKED ISSUE #1160: cacheManager.getRuntimeConfiguration().getClassloader() returns null
Don't know whether it's a bug or doc issue (to me it's a bug, getClassloader() should always return I think from a user perspective - this is what I would think it does).

But now it returns null. The javadoc says: 

"The ClassLoader must not be null."

mathieucarbou:


ljacomet:
@mathieucarbou We need to chat about this.

PR #1170, opened 2016-06-15
Fix #1161 - recursiveDeleteDirectoryContent() always true

### LINKED ISSUE #1161: DefaultLocalPersistenceService.recursiveDeleteDirectoryContent(File) always returns true
This method is used in `DefaultLocalPersistenceService.destroyAll()`, which then always records a "positive" action _"Destroyed all file based persistence contexts"_

Content of this method is:

```
  private static boolean recursiveDeleteDirectoryContent(File file) {
    File[] contents = file.listFiles();
    if (contents == null) {
      throw new IllegalArgumentException("File " + file.getAbsolutePath() + " is not a directory");
    } else {
      boolean deleteSuccessful = true;
      for (File f : contents) {
        deleteSuccessful |= tryRecursiveDelete(f);
      }
      return deleteSuccessful;
    }
  }
```

I guess probably a fix could be:

```
  private static boolean recursiveDeleteDirectoryContent(File file) {
    File[] contents = file.listFiles();
    if (contents == null) {
      throw new IllegalArgumentException("File " + file.getAbsolutePath() + " is not a directory");
    } if (contents.length == 0) {
      return true;
    } else {
      boolean deleteSuccessful = false;
      for (File f : contents) {
        deleteSuccessful |= tryRecursiveDelete(f);
      }
      return deleteSuccessful;
    }
  }
```

Let me know if you are ok with this fix - I can send a PR.

mathieucarbou:


ljacomet:
@mathieucarbou We need to chat about this too.

PR #1174, opened 2016-06-15
Issue 1160 3.0

mathieucarbou:


core/src/main/java/org/ehcache/core/config/DefaultConfiguration.java
```diff
@@ -77,7 +78,7 @@ public DefaultConfiguration(ClassLoader classLoader, ServiceCreationConfiguratio
   public DefaultConfiguration(Map<String, CacheConfiguration<?, ?>> caches, ClassLoader classLoader, ServiceCreationConfiguration<?>... services) {
     this.services = unmodifiableCollection(Arrays.asList(services));
     this.caches = new ConcurrentHashMap<String, CacheConfiguration<?, ?>>(caches);
-    this.classLoader = classLoader;
+    this.classLoader = classLoader == null ? ClassLoading.getDefaultClassLoader() : classLoader;
```
ljacomet: A similar check should be made on the copy constructor, or even have it delegate to this one to simplify.

mathieucarbou: Api doc of `Configuration.getClassloader()` says: _The {@code ClassLoader} must not be null._
so should we instead add a null check like it's done in `XmlConfiguration` ?

ljacomet: That's what I am saying. But I also suggest refactoring the constructors so that you end up having the check in a single one. And similarly the fact that cache config map is in a concurrent map and that service configs is an immutable collection.
Makes sense?

mathieucarbou: Oh sorry! Misunderstanding ;-) By null-check I meant:

```
if(classloader == null) throw new NPE()
```

The idea behind that is that Configuration.getClassloader() contract says classloader should not be null; So we can make sure of that in the copy constructor.

Also, the idea of this null check is to have a real copy constructor: if we init the classloader to `ClassLoading.getDefaultClassLoader()` if `Configuration.getClassloader()` is null, then it means that the classloader is not "copied" from the Configuration object and `DefaultConfiguration.getClassloader()` would be different than `Configuration.getClassloader()`

So perhaps it would be better to keep this where it is now:

```
this.classLoader = classLoader == null ? ClassLoading.getDefaultClassLoader() : classLoader
```

... and just add a null check in the copy ctor ?

```
if(classloader == null) throw new NPE()
```

ljacomet: Agreed

ljacomet:
@mathieucarbou Before you go merge like crazy, make sure you update the issue-1160 branch and I will check if the command line manages to resolve the merge.

mathieucarbou:
@ljacomet : I don't know if it the right thing to do.

Because the fix for `this.classLoader = classLoader == null ? ClassLoading.getDefaultClassLoader() : classLoader` (originally in `Configuration` class) does not have the same base commit as the fix for the copy ctor.

So I did to branches from 2 different base commits:

- one for the default init.
- one for the null check

and i merged both into the PR branches issue-1160-master and issue-1160-3.0

ljacomet:
@mathieucarbou In that case, I would recommend you choose the release/3.0 branch point as the source commit on which to base your fix, that will make for a simpler history.

mathieucarbou:
ok I'll try that. because as I saw, even if I place my fix just after the introduction of DefaultConfiguration class, I saw it has been moved so I'll still have to resolve a merge also. Really, not easy this process when fixing things a long time ago ;-)

PR #1177, opened 2016-06-16
Fix EhcacheSettingsProviderTest on Windows (CR)

mathieucarbou:
I would need some Windows guy to test this patch...

https://patch-diff.githubusercontent.com/raw/ehcache/ehcache3/pull/1177.patch


kcleereman:
This fixes the problem on my machine

management/src/test/java/org/ehcache/management/providers/settings/EhcacheSettingsProviderTest.java
```diff
@@ -114,9 +113,9 @@ public void test_standalone_ehcache() throws IOException {
 
     cacheManager.init();
 
-    System.out.println(mapper.writeValueAsString(sharedManagementService.getCapabilities()));
+    //System.out.println(mapper.writeValueAsString(sharedManagementService.getCapabilities()));
```
ljacomet: Can that be removed instead of commented out?

mathieucarbou: oh yes I forgot!

PR #1179, opened 2016-06-17
Issue 1140

AbfrmBlr:


docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -83,7 +83,28 @@ include::{sourcedir}/clustered/client/src/test/java/org/ehcache/clustered/client
...
+
+=== Clustered Storage Tier
+
+[source,java,indent=0]
+----
+include::{sourcedir}/clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java[tag=clusteredCacheTieredExample]
+----
+
+<1> Configuring heap tier for cache.
+<2> Configuring clustered tier of fixed size from server resource using `ClusteredResourcePoolBuilder`.
```
albinsuresh: from a "server off-heap resource". That's the term that we decided to use. It should be consistent.

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -113,4 +134,10 @@ The equivalent XML configuration is as follows:
 include::{sourcedir}/clustered/client/src/test/resources/configs/docs/ehcache-clustered.xml[tag=consistencySample]
 ----
 
-<1> Specify the consistency level through a custom service configuration from the `clustered` namespace.
\ No newline at end of file
+<1> Specify the consistency level through a custom service configuration from the `clustered` namespace.
+
+=== Clustered Cache Expiry
+
+Expiry in clustered caches work with an exception that `Expiry#getExpiryForAccess` is ignored in clustered tier, though the same
```
albinsuresh: I don't think we want to be blunt about the fact that it is ignored. I think best effort would be the right terms to use to disguise the reality. @ljacomet WDYT?

AbfrmBlr:
test this please

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -83,7 +83,28 @@ include::{sourcedir}/clustered/client/src/test/java/org/ehcache/clustered/client
 
 == Clustered caches
 
-IMPORTANT: Content needed here!
+Distributed Caching can be enabled on a cache by adding Clustered storage as a tier to the cache.
```
ljacomet: This is a more general introduction to clustering and should live at the top of the page instead of inside the clustered caches section.

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -13,6 +13,8 @@ endif::notBuildingForSite[]
 
 == Introduction
 
+Distributed Caching can be enabled on a cache by adding Clustered storage as a tier to the cache.
```
ljacomet: I think this addition makes sense but then the whole section needs to be reworked. Whatever the order, there are two separate things:
* Have a started server
* Have the proper client configuration
  * Cache manager
  * Cache with clustered tier
And maybe before that there can be a reference to the benefits of clustering.

PR #1180, opened 2016-06-17
Fix #1132 Move auto-create from the URI to the server side configuration

### LINKED ISSUE #1132: Move auto-create declaration
Currently, the way to declare you want entity and clustered store(s) to be created automatically is done through a query parameter on the connections string:

```xml
terracotta://example.com:9540/cachemanager?auto-create
```

This flag should be moved to another location, namely the `server-side-config` element.

It would allow the following:

```xml
<tc:cluster>
  <tc:connection url="terracotta://example.com:9540/cachemanager"/>
</tc:cluster>
```
indicates that we want to connect to an existing cache manager entity and we are ready to accept whatever configuration it has.

```xml
<tc:cluster>
  <tc:connection url="terracotta://example.com:9540/cachemanager"/>
  <tc:server-side-config auto-create="true">
    ...
  </tc:server-side-config>
</tc:cluster>
```
indicates that we want to connect to a cache manager entity having the specified config (including an empty one) and potentially create it.

It would thus close the current hole where it seems valid to have `auto-create` with no server side config.

chrisdennis:


gibsong:
Look like there are still 4 occurrences of url's that use "?auto-create"
1. ClusteredStoreTest.tearDown()
2. consistency.xml
3. ehcache-clustered.xml
4. resource-sample.xml

Issue #1152, opened 2016-06-09
Passive Capable CoordinationService Replacement


chrisdennis:
In order to support active-passive ehcache instances we must have a passive capable version of the functionality currently provided by the CoordinationService.  After some discussion with @cljohnso we decided it would be simpler to replace the coordination service with a more cut down facility before making the entity passive capable.

PR #1181, opened 2016-06-17
Issue #1152 Replace Coordination Service with new ReadWriteLock Impl

chrisdennis:


clustered/client/src/main/java/org/ehcache/clustered/lock/client/VoltronReadWriteLock.java
```diff
@@ -0,0 +1,142 @@
...
+import org.terracotta.connection.Connection;
+import org.terracotta.connection.entity.EntityRef;
+import org.terracotta.exception.EntityAlreadyExistsException;
+import org.terracotta.exception.EntityNotFoundException;
+import org.terracotta.exception.EntityNotProvidedException;
+import org.terracotta.exception.EntityVersionMismatchException;
+
+/**
+ *
+ * @author cdennis
```
ljacomet: Code style: no author tags

clustered/client/src/main/java/org/ehcache/clustered/lock/client/VoltronReadWriteLockClient.java
```diff
@@ -0,0 +1,141 @@
...
+    });
+  }
+
+  @Override
+  public void close() {
+    endpoint.close();
+  }
+
+  public boolean tryLock(HoldType type) {
+    while (true) {
```
ljacomet: Never loops

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -91,58 +79,72 @@ public void abandonLeadership(String entityIdentifier) {
...
+    Hold existingMaintenance = maintenanceHolds.get(identifier);
+    Hold localMaintenance = null;
+    if (existingMaintenance == null) {
+      localMaintenance = createAccessLockFor(identifier).tryWriteLock();
+    }
+    if (existingMaintenance == null && localMaintenance == null) {
+      String message = "Unable to create entity for cluster id " + identifier
+          + ": another client owns the maintenance lease";
+      LOGGER.error(message);
+      throw new EhcacheEntityCreationException(message);
```
ljacomet: Seems weird to both log and throw - why is throw not enough?

ljacomet: Especially as the code path in `DefaultClusteringService.start` will most likely print this message when `auto-create` is enabled and the entity already exists and is in use.

chrisdennis:
@ljacomet corrections applied

Issue #1138, opened 2016-06-07
Enable "thin" cluster server configuration


cljohnso:
Current configuration, either through XML or programmatic configuration, requires that the `/ehcache:config/ehcache:service/tc:cluster/tc:connection` element (`ClusteringServiceConfiguration`) be present.  This is necessary in order to identify a cluster server to which the client must connect for clustered caches.

If a shared resource pool is used in a cache definition, the `/ehcache:config/ehcache:service/tc:cluster/tc:server-side-config/tc:shared-pool` element describing that shared resource pool must be present and must match the definition used to create that pool in the server.  If a clustered cache using fixed resources is defined (`/ehcache:config/ehcache:cache/ehcache:resources/tc:cluster-fixed`) and that definition _omits_ the `from` attribute, then the `/ehcache:config/ehcache:cache/ehcache:resources/tc:cluster-fixed` element must also be present.

Is is desired that, for a client simply _using_ clustered caches and not _defining_ them, the `connection` element be the only required element in the `service/cluster` configuration.

----

## Discussion

1. The `auto-create` query argument needs to be moved off of the cluster URI.
    This is covered in issue #1132.
2. Change the `EhcacheClientEntity.validate` method (and the underlying mechanisms) to return the `ServerSideConfig` for the current server-resident configuration.  The validate message still needs to present the possibly thin/empty `ServerSideConfig` expressed in the client configuration and the server still needs to validate the client expectations against the existing server-resident configuration.  This permits partial or omitted client-provided configuration for use (not create).  Changing the name from `validate` to `retrieve` should be considered.
3. A `ClusteringServiceConfiguration` may need a constructor that omits the `defaultServerResource` and `pools` parameters; for a "thin" configuration, neither of these parameters is required.  The `ClusteringServiceConfigurationBuilder` appears to already support the proper "thinness".

PR #1189, opened 2016-06-20
Issue #1138 Move all configuration resolution to the server

chrisdennis:


chrisdennis:
Talked with @ljacomet, going to revise these rules to be stricter.

Issue #1136, opened 2016-06-07
Normalize validation of cluster URL/URI


cljohnso:
A. Validation of the _scheme_ used in the cluster URL/URI is inconsistent:

  1. `ehcache-clustered-ext.xsd` enforces the _scheme_ for `/ehcache:config/ehcache:service/tc:cluster/tc:connection/@url` to be **`terracotta`**
  2. Neither `ClusteringServiceConfigurationBuilder`, `ClusteringServiceConfiguration`, nor `DefaultClusteringService` validates the scheme

B. The URL/URI _path_ is not validated for presence in either the XSD or clustered configuration classes.  Since a _path_ is required for identifying the clustered store manager in a cluster server, the _path_ should probably be **required**.

cljohnso:
This is a superset of issue #890.  The two issues should likely be combined.

PR #1191, opened 2016-06-21
validate URI scheme in clusterserviceconfig #1136

AbfrmBlr:
Scheme is already validated in XSD. same approach is used here.
URI is already mandatory in `ClusteringServiceConfigurationBuilder`


clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -40,6 +41,8 @@
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<PersistentCacheManager> {
 
+  private static final String CLUSTER_URI_PATTERN = "terracotta://([^\\]\\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(/[^\\?#]*)?(\\?[^#]*)?(#.*)?";
```
ljacomet: Why? Why not simply validating the URI `scheme` is `terracotta:`?

AbfrmBlr: oh yes

albinsuresh: Something unrelated to this issue exactly: Do we really want to keep this scheme prefix as `terracotta`? Isn't that a bit too long? `tc://` like old times would have been much better I feel.

ljacomet: `terracotta:` is used all over the place, let's not change that so late - we may open up to a shorter version later.

ljacomet:
Still have `http` in `clustered/client/src/test/resources/configs/resource-sample.xml`

AbfrmBlr:
No tests are failing, this means its not being used. that too it should show error for xsd's

ljacomet:
Indeed - fine with me to delete the file

AbfrmBlr:
test this please

PR #1198, opened 2016-06-23
Adapting UnitTestConnectionService with the new changes in Passthroug…

albinsuresh:
This fixes the compilation failure on Ehcache builds caused by the changes in method signatures in  `PassthroughServer`. There are failing integration tests. But AFAIK someone else is looking into that.

I'm not quite familiar with the procedure to follow to address such build failures. Should an issue be created and referred by this PR?


chrisdennis:
I merged this even though the tests are failing because the test failure is due to an unrelated issue... working on that now.

Issue #1184, opened 2016-06-18
Improve exception message when a clustered cachemanager entity doesn't exist


aurbroszniowski:
Using 

  ```
  Ehcache Clustered Implementation-Version: 3.1.0-SNAPSHOT 081007b2d8a41f018abe55f85407069
   dc8e53e9a
  ```

  and 

  ```
  Terracotta 5.0.0-SNAPSHOT, as of 2016-06-14 at 09:28:52 EDT (Revision 2a5deb444144e54c75f6996454cc2470e6a1308f from UNKNOWN)
  ```
  
When trying to programmatically create a Clustered CacheManager, and not setting the autocreate flag of the ClusteringServiceConfigurationBuilder to true, if the Cache Manager entity doesn't exist yet, then the exception is

```
org.ehcache.StateTransitionException: org.terracotta.exception.EntityNotFoundException: Entity: org.ehcache.clustered.client.internal.EhcacheClientEntity:myCachemanager not found

	at org.ehcache.core.StatusTransitioner$Transition.failed(StatusTransitioner.java:235)
	at org.ehcache.core.EhcacheManager.init(EhcacheManager.java:585)
	at org.ehcache.config.builders.CacheManagerBuilder.build(CacheManagerBuilder.java:72)
	at com.terracotta.qa.performance.clustered.Ehcache3HeapOffheapTsa.test(Ehcache3HeapOffheapTsa.java:114)
	
Caused by: java.lang.IllegalStateException: org.terracotta.exception.EntityNotFoundException: Entity: org.ehcache.clustered.client.internal.EhcacheClientEntity:myCachemanager not found
	at org.ehcache.clustered.client.internal.service.DefaultClusteringService.start(DefaultClusteringService.java:140)
	at org.ehcache.core.internal.service.ServiceLocator.startAllServices(ServiceLocator.java:306)
	at org.ehcache.core.EhcacheManager.init(EhcacheManager.java:557)
	... 31 more
Caused by: org.terracotta.exception.EntityNotFoundException: Entity: org.ehcache.clustered.client.internal.EhcacheClientEntity:myCachemanager not found
	at com.tc.objectserver.handler.ProcessTransactionHandler.addMessage(ProcessTransactionHandler.java:246)
	at com.tc.objectserver.handler.ProcessTransactionHandler.access$200(ProcessTransactionHandler.java:59)
[...]

PR #1214, opened 2016-06-23
Issue #1184 Better error message on clustering service start failures

chrisdennis:
I've gone with what I understood to be the accepted terminology here "server storage manager".  That is I think the most important thing to review here.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
@@ -129,12 +129,19 @@ class DefaultClusteringService implements ClusteringService {
       try {
         entity = entityFactory.retrieve(entityIdentifier, configuration.getServerConfiguration());
       } catch (EntityNotFoundException e) {
-        /*
-         * If the connection failed because of a creation failure, re-throw the creation failure.
-         */
-        throw new IllegalStateException(failure == null ? e : failure);
+        if (failure == null) {
+          throw new IllegalStateException("The server-side storage manager does not exist."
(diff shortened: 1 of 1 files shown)


AbfrmBlr:
LGTM

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -132,12 +132,19 @@ public void start(final ServiceProvider<Service> serviceProvider) {
       try {
         entity = entityFactory.retrieve(entityIdentifier, configuration.getServerConfiguration());
       } catch (EntityNotFoundException e) {
-        /*
-         * If the connection failed because of a creation failure, re-throw the creation failure.
-         */
-        throw new IllegalStateException(failure == null ? e : failure);
+        if (failure == null) {
+          throw new IllegalStateException("The server-side storage manager does not exist."
```
albinsuresh: The entity is defined as _Ehcache clustered entity_ in our clustered cache doc. So I guess that's what we should be using here.

albinsuresh:
I see that the word _Clustered Store Manager_ is used all over the place in the `EhcacheActiveEntity` class. Either that should change or the term that is documented should change. I'm not saying that it should be in this PR. May be another one.

ljacomet:
I vote against using `Store` in any user documentation. That term has way too much meaning associated with it. So my preference goes to _Ehcache clustered entity_.

ljacomet:
Still believe this adds something after the rework from @lorban so good to go once rebased

lorban:
Agreed; this change makes some errors more intelligible. Let's leave terminology to #1243.

PR #1215, opened 2016-06-24
Issue #1141 enables thin cluster cache configuration

### LINKED ISSUE #1141: Enable "thin" cluster cache configuration
This issue is a companion to and depends on issue #1138.

Defining a clustered cache presently requires either the presence of a `/ehcache:config/ehcache:cache/ehcache:resources/tc:cluster-shared` or `/ehcache:config/ehcache:cache/ehcache:resources/tc:cluster-fixed` element in an XML configuration or the `ResourcePoolsBuilder.with(ClusteredResourcePoolBuilder.<type>)` where `<type>` is either `fixed` or `shared` for programmatic configuration.  And, depending on the configuration values, the clustered service configuration must be available.

For a "thin" clustered cache configuration, the cache must still be marked as _clustered_ but the fixed or shared resource should be omittable.  What is proposed here is the addition of a `clustered` resource type in addition to the `cluster-fixed` and `cluster-shared` resource types.

Support for this requires addition of a `/ehcache:config/ehcache:cache/ehcache:resources/tc:clustered` element to the XSD along with supporting changes in `ClusteredResourceConfigurationParser` and matching changes `ClusteredResourcePoolBuilder`.  An extension of `ClusteredResourcePool` is needed along with a new `ClusteredResourceType`.  A new `ServerStoreConfiguration.PoolAllocation` _may_ be desired.  These additions cascade into other clustering classes.

The validation performed by `EhcacheActiveEntity.validateServerStore` needs to be changed to accept the new _clustered_ pool type.

Although not currently needed to support a "thin" clustered cache configuration, the `EhcacheClientEntity.validateCache` method (and underlying mechanism) should be changed to support return of the `ServerStoreConfiguration` in use by the server (to be consistent with the change to `EhcacheClientEntity.validate` in issue #1138).

gibsong:
Fixes #1141 removes potential null ptr exception and updates javadoc

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredResourcePoolBuilder.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredResourcePoolBuilder.java
@@ -16,11 +16,13 @@
 
 package org.ehcache.clustered.client.config.builders;
 
-import org.ehcache.clustered.client.config.FixedClusteredResourcePool;
 import org.ehcache.clustered.client.config.SharedClusteredResourcePool;
-import org.ehcache.clustered.client.internal.config.FixedClusteredResourcePoolImpl;
+import org.ehcache.clustered.client.config.ClusteredResourcePool;
+import org.ehcache.clustered.client.internal.config.DedicatedClusteredResourcePoolImpl;
 import org.ehcache.clustered.client.internal.config.SharedClusteredResourcePoolImpl;
+import org.ehcache.clustered.client.internal.config.ClusteredResourcePoolImpl;
 import org.ehcache.config.units.MemoryUnit;
+import org.ehcache.clustered.client.config.DedicatedClusteredResourcePool;
 
 /**
  * Constructs a {@link org.ehcache.config.ResourcePool ResourcePool} for a clustered resource.
@@ -32,25 +34,25 @@ public final class ClusteredResourcePoolBuilder {
   }
 
   /**
-   * Creates a new clustered resource pool using fixed clustered resources.
+   * Creates a new clustered resource pool using dedicated clustered resources.
    *
    * @param size       the size
    * @param unit       the unit for the size
    */
-  public static FixedClusteredResourcePool fixed(long size, MemoryUnit unit) {
-    return new FixedClusteredResourcePoolImpl(null, size, unit);
+  public static DedicatedClusteredResourcePool clusteredDedicated(long size, MemoryUnit unit) {
+    return new DedicatedClusteredResourcePoolImpl(null, size, unit);
   }
 
   /**
-   * Creates a new clustered resource pool using fixed clustered resources.
+   * Creates a new clustered resource pool using dedicated clustered resources.
    *
-   * @param fromResource the name of the server-based resource from which this fixed resource pool
-   *                     is reserved; may be {@code null}
+   * @param fromResource the name of the server-based resource from which this dedicated resource pool
+                     is reserved; may be {@code null}
    * @param size       the size
    * @param unit       the unit for the size
    */
-  public static FixedClusteredResourcePool fixed(String fromResource, long size, MemoryUnit unit) {
-    return new FixedClusteredResourcePoolImpl(fromResource, size, unit);
+  public static DedicatedClusteredResourcePool clusteredDedicated(String fromResource, long size, MemoryUnit unit) {
+    return new DedicatedClusteredResourcePoolImpl(fromResource, size, unit);
   }
 
   /**
@@ -59,7 +61,14 @@ public final class ClusteredResourcePoolBuilder {
    * @param sharedResource the non-{@code null} name of the server-based resource pool whose space is shared
    *                       by this pool
    */
-  public static SharedClusteredResourcePool shared(String sharedResource) {
+  public static SharedClusteredResourcePool clusteredShared(String sharedResource) {
     return new SharedClusteredResourcePoolImpl(sharedResource);
+  }
+
+  /**
+   * Creates a resource pool that inherits the resources already configured on the server.
+   */
+  public static ClusteredResourcePool clustered() {
+    return new ClusteredResourcePoolImpl();
   }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteredResourceConfigurationParser.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteredResourceConfigurationParser.java
@@ -16,8 +16,9 @@
 
 package org.ehcache.clustered.client.internal.config.xml;
 
-import org.ehcache.clustered.client.internal.config.FixedClusteredResourcePoolImpl;
+import org.ehcache.clustered.client.internal.config.DedicatedClusteredResourcePoolImpl;
 import org.ehcache.clustered.client.internal.config.SharedClusteredResourcePoolImpl;
+import org.ehcache.clustered.client.internal.config.ClusteredResourcePoolImpl;
 import org.ehcache.config.ResourcePool;
 import org.ehcache.config.units.MemoryUnit;
 import org.ehcache.xml.CacheResourceConfigurationParser;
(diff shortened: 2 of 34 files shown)


gibsong:
Note: I did not implement EhcacheClientEntity.validateCache to return ServerStoreConfiguration

clustered/client/src/main/java/org/ehcache/clustered/client/config/ThinClusteredResourcePool.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.client.config;
+
+/**
+ *
+ * @author GGIB
```
ljacomet: Please drop `@author` tag, we are not using them

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredResourcePoolBuilder.java
```diff
@@ -62,4 +64,11 @@ public static FixedClusteredResourcePool fixed(String fromResource, long size, M
   public static SharedClusteredResourcePool shared(String sharedResource) {
     return new SharedClusteredResourcePoolImpl(sharedResource);
   }
+
+  /**
+   * Creates a resource pool that inherits the resources already configured on the server.
+   */
+  public static ThinClusteredResourcePool thin() {
```
ljacomet: I really dislike that name. What about `clustered()` simply?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteredResourceConfigurationParser.java
```diff
@@ -84,6 +85,10 @@ public ResourcePool parseResourceConfiguration(final Element fragment) {
 
       return new FixedClusteredResourcePoolImpl(from, size, sizeUnits);
     }
+    else if("clustered".equals(elementName))
+    {
```
ljacomet: Code style

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -27,7 +27,6 @@
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.transform.stream.StreamSource;
-
```
ljacomet: Get rid of that totally unrelated change

clustered/client/src/test/resources/configs/thin-cluster-cache-invalid-element.xml
```diff
@@ -0,0 +1,19 @@
...
+  <eh:service>
+    <tc:cluster>
+      <tc:connection url="terracotta://example.com/myCacheManager`+"/>
+      <tc:server-side-config auto-create="false"/>
+    </tc:cluster>
+  </eh:service>
+
+  <eh:cache alias="thinCache">
+    <eh:resources>
+      <tc:clustered>xxx</tc:clustered>
```
ljacomet: clearly, if we use `clustered` in XML and `thin` in code, there is a name issue.

chrisdennis:
A number of fractured side discussions between @GaryWKeim @cljohnso @AbfrmBlr @ljacomet and myself led to the idea of "clustered", "clustered-shared" and "clustered-dedicated" as suitable terminology for the three clustered resource types.  If nobody has any objections I suggest that we standardize on those names as part of this PR.

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
```diff
@@ -76,6 +77,26 @@ public void verify(ServerStoreConfiguration serverConfiguration, ServerStoreConf
     if (!isCompatible) {
       throw new ClusteredStoreValidationException(sb.toString());
     }
+
+  }
+
+  private static boolean comparePoolAllocationType(StringBuilder sb, PoolAllocation serverPoolAllocation, PoolAllocation clientPoolAllocation) {
+    if ((serverPoolAllocation == null && clientPoolAllocation == null)
```
chrisdennis: What does it mean for both to be null?  Surely this method should never be called with both being null?

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
```diff
@@ -42,27 +42,28 @@ public void verify(ServerStoreConfiguration serverConfiguration, ServerStoreConf
 
     PoolAllocation serverPoolAllocation = serverConfiguration.getPoolAllocation();
     PoolAllocation clientPoolAllocation = clientConfiguration.getPoolAllocation();
```
chrisdennis: It feels to me like the methods of ServerStoreCompatibility should really be instance methods of the various types it validates.  Does that make sense?

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreConfiguration.java
```diff
@@ -159,6 +159,13 @@ public String getResourcePoolName() {
         return resourcePoolName;
       }
     }
+
+    /**
+     * Creates a Pool Allocation which inherits the Shared or Fixed Pool Allocation from a cache which is already configured on the server.
+     */
+    final class Thin implements PoolAllocation {
```
chrisdennis: This class (and the other inner classes here) should probably all be static, right?

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
gibsong: yes that makes sense.   I will try to refactor code into instance methods.  This should help clean it up a bit too.

clustered/common/src/main/java/org/ehcache/clustered/common/ServerStoreCompatibility.java
gibsong: I removed the null checks.  It appears that this case can't happen.  If I am wrong please correct me.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredResourcePoolBuilder.java
```diff
@@ -32,25 +34,25 @@ private ClusteredResourcePoolBuilder() {
...
-   * @param fromResource the name of the server-based resource from which this fixed resource pool
-   *                     is reserved; may be {@code null}
+   * @param fromResource the name of the server-based resource from which this dedicated resource pool
+                     is reserved; may be {@code null}
    * @param size       the size
    * @param unit       the unit for the size
    */
-  public static FixedClusteredResourcePool fixed(String fromResource, long size, MemoryUnit unit) {
-    return new FixedClusteredResourcePoolImpl(fromResource, size, unit);
+  public static DedicatedClusteredResourcePool dedicated(String fromResource, long size, MemoryUnit unit) {
```
ljacomet: I am in favour of `clusteredDedicated` here

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteredResourcePoolBuilder.java
```diff
@@ -62,4 +64,11 @@ public static FixedClusteredResourcePool fixed(String fromResource, long size, M
   public static SharedClusteredResourcePool shared(String sharedResource) {
```
ljacomet: I am in favour of `clusteredShared` here

ljacomet:
test this please

ljacomet:
Verified this runs tests and passes checks locally so merging.

Issue #1192, opened 2016-06-21
Refactor local persistence service


rkavanap:
Refactor local persistence service in such a way that the root directory handling and sub-space creation is separated from disk based persistence service..

ljacomet:
With the changes introduced by PR #1216, the disks tier problems fixed in #1410 are re-introduced. Some more thinking needs to go into fixing that for good.

PR #1216, opened 2016-06-24
Issue #1192 Separate local persistence from file based persistence

rkavanap:
Refactor to separate local persistence service from file based persistence.

### DIFF
--- a/core/src/main/java/org/ehcache/core/spi/service/DiskResourceService.java
+++ b/core/src/main/java/org/ehcache/core/spi/service/DiskResourceService.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.core.spi.service;
+
+import org.ehcache.CachePersistenceException;
+import org.ehcache.spi.persistence.PersistableResourceService;
+
+/**
+ * Resource service handling file level operations for disk tiers.
+ */
+public interface DiskResourceService extends PersistableResourceService {
+
+  /**
+   * Creates a new persistence context within the given space.
+   *
+   * @param identifier space to create within
+   * @param name name of the context to create
+   * @return a {@link FileBasedPersistenceContext}
+   */
+  FileBasedPersistenceContext createPersistenceContextWithin(PersistenceSpaceIdentifier<?> identifier, String name) throws CachePersistenceException;
+}
--- a/core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
+++ b/core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
@@ -17,22 +17,63 @@
 package org.ehcache.core.spi.service;
 
 import org.ehcache.CachePersistenceException;
-import org.ehcache.spi.persistence.PersistableResourceService;
+import org.ehcache.spi.service.MaintainableService;
+
+import java.io.File;
 
 /**
- * Service to provide persistence context to caches requiring it.
- * <P>
- * Will be used by caches with a disk store, whether or not the data should survive a program restart.
- * <P/>
+ * Service that provides isolated persistence spaces to any service that requires it
+ * under the local root directory.
  */
-public interface LocalPersistenceService extends PersistableResourceService {
+public interface LocalPersistenceService extends MaintainableService {
 
   /**
-   * Creates a new persistence context within the given space.
+   * Creates a logical safe directory space for the owner and returns an identifying space Id.
    *
-   * @param identifier space to create within
-   * @param name name of the context to create
-   * @return a {@link FileBasedPersistenceContext}
+   * @param owner Service owner that owns the safe space.
+   * @param name Identifying name for the space.
+   *
+   * @return Opaque Identifier that can be used to identify the safe space.
    */
-  FileBasedPersistenceContext createPersistenceContextWithin(PersistenceSpaceIdentifier<?> identifier, String name) throws CachePersistenceException;
+  SafeSpaceIdentifier createSafeSpaceIdentifier(String owner, String name);
+
+  /**
+   * Creates the safe space represented by {@code safeSpaceId}, if it does not exist in the underlying physical space.
+   *
+   * @param safeSpaceId Identifier to the created logical space on which the physical space needs to be created
+   * @throws CachePersistenceException If the space cannot be created or found, due to system errors
+   */
+  void createSafeSpace(SafeSpaceIdentifier safeSpaceId) throws CachePersistenceException;
+
+  /**
+   * Destroys the safe space.
+   *
+   * @param safeSpaceId Safe space identifier.
+   * @param verbose Log more information.
+   */
+  void destroySafeSpace(SafeSpaceIdentifier safeSpaceId, boolean verbose);
+
+  /**
+   * Destroys all safe spaces provided to this owner.
+   *
+   * @param owner owner of safe spaces.
+   */
+  void destroyAll(String owner);
+
+  /**
+   * Identifier to the logical safe space
+   */
+  interface SafeSpaceIdentifier {
+    /**
+     * Represents the root directory of the given logical safe space.
+     * <P>
+     *   Note that the directory represented by {@code File} may or may not be created in the physical space.
+     *   The existence of the physical space depends on whether the {@code createSafeSpace} method was invoked
+     *   for the space at some time in the past or not.
+     * </P>
+     *
+     * @return Root directory of the safe space.
+     */
+    File getRoot();
+  }
 }
--- a/impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
+++ b/impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
@@ -29,6 +29,7 @@ import org.ehcache.config.EvictionAdvisor;
 import org.ehcache.config.ResourcePools;
 import org.ehcache.config.ResourceType;
 import org.ehcache.core.internal.store.StoreConfigurationImpl;
+import org.ehcache.core.spi.service.DiskResourceService;
 import org.ehcache.core.spi.store.heap.SizeOfEngine;
 import org.ehcache.impl.events.CacheEventDispatcherImpl;
 import org.ehcache.core.internal.store.StoreSupport;
@@ -58,7 +59,6 @@ import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
 import org.ehcache.spi.serialization.SerializationProvider;
 import org.ehcache.spi.serialization.Serializer;
 import org.ehcache.spi.serialization.UnsupportedTypeException;
-import org.ehcache.core.spi.service.LocalPersistenceService;
 import org.ehcache.core.spi.store.heap.SizeOfEngineProvider;
 import org.ehcache.spi.service.Service;
 import org.ehcache.spi.service.ServiceConfiguration;
@@ -210,20 +210,20 @@ public class UserManagedCacheBuilder<K, V, T extends UserManagedCache<K, V>> imp
       if (id == null) {
         throw new IllegalStateException("Persistent user managed caches must have an id set");
       }
-      final LocalPersistenceService persistenceService = serviceLocator.getService(LocalPersistenceService.class);
+      final DiskResourceService diskResourceService = serviceLocator.getService(DiskResourceService.class);
       if (!resourcePools.getPoolForResource(ResourceType.Core.DISK).isPersistent()) {
         try {
-          persistenceService.destroy(id);
+          diskResourceService.destroy(id);
         } catch (CachePersistenceException cpex) {
           throw new RuntimeException("Unable to clean-up persistence space for non-restartable cache " + id, cpex);
         }
       }
       try {
-        final PersistableResourceService.PersistenceSpaceIdentifier<?> identifier = persistenceService.getPersistenceSpaceIdentifier(id, cacheConfig);
+        final PersistableResourceService.PersistenceSpaceIdentifier<?> identifier = diskResourceService.getPersistenceSpaceIdentifier(id, cacheConfig);
         lifeCycledList.add(new LifeCycledAdapter() {
           @Override
           public void close() throws Exception {
-            persistenceService.releasePersistenceSpaceIdentifier(identifier);
+            diskResourceService.releasePersistenceSpaceIdentifier(identifier);
           }
         });
         serviceConfigsList.add(identifier);
@@ -304,13 +304,13 @@ public class UserManagedCacheBuilder<K, V, T extends UserManagedCache<K, V>> imp
     eventDispatcher.setStoreEventSource(store.getStoreEventSource());
 
     if (persistent) {
-      LocalPersistenceService persistenceService = serviceLocator
-          .getService(LocalPersistenceService.class);
-      if (persistenceService == null) {
+      DiskResourceService diskResourceService = serviceLocator
+          .getService(DiskResourceService.class);
+      if (diskResourceService == null) {
         throw new IllegalStateException("No LocalPersistenceService could be found - did you configure one?");
       }
 
-      PersistentUserManagedEhcache<K, V> cache = new PersistentUserManagedEhcache<K, V>(cacheConfig, store, persistenceService, cacheLoaderWriter, eventDispatcher, id);
+      PersistentUserManagedEhcache<K, V> cache = new PersistentUserManagedEhcache<K, V>(cacheConfig, store, diskResourceService, cacheLoaderWriter, eventDispatcher, id);
       registerListeners(cache, serviceLocator, lifeCycledList);
       for (LifeCycled lifeCycled : lifeCycledList) {
         cache.addHook(lifeCycled);
--- a/impl/src/main/java/org/ehcache/impl/internal/persistence/DefaultDiskResourceServiceFactory.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/persistence/DefaultDiskResourceServiceFactory.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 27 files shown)


core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,71 @@
...
+   * @return Identifier that can be used to identify the safe space.
+   */
+  String prepareSafeSpace(String owner, String name);
+
+  /**
+   * Create the safe spaces in actual physical space for the owner.
+   *
+   * @param safeSpaceId prepared logical space on which the physical space
+   */
+  void createSafeSpace(String safeSpaceId) throws CachePersistenceException;
```
ljacomet: Please, let's not use String there but instead some opaque id with a specific type.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -327,6 +328,9 @@ public int rankAuthority(ResourceType<?> authorityResource, Collection<ServiceCo
       if (serviceProvider == null) {
         throw new NullPointerException("ServiceProvider is null in OffHeapDiskStore.Provider.");
       }
+      if (diskPersistenceService == null) {
```
ljacomet: That error here makes no sense - do the check in `start` instead

impl/src/main/java/org/ehcache/impl/persistence/DefaultFileBasedPersistenceService.java
```diff
@@ -0,0 +1,262 @@
...
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Default implementation of the {@link FileBasedPersistenceService} which can be used explicitly when
+ * {@link org.ehcache.PersistentUserManagedCache persistent user managed caches} are desired.
+ *
+ * @author RKAV refactored for issue-1192
```
ljacomet: Drop `@author` tag - we are not using them

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,71 @@
...
+
+  /**
+   * Prepare a logical safe directory space for the owner and return an identifying space Id.
+   *
+   * @param owner Service owner that owns the safe space.
+   * @param name Identifying name for the space.
+   *
+   * @return Identifier that can be used to identify the safe space.
+   */
+  String prepareSafeSpace(String owner, String name);
```
ljacomet: What exactly is the `owner` concept providing?

impl/src/main/java/org/ehcache/impl/persistence/DefaultFileBasedPersistenceService.java
```diff
@@ -0,0 +1,262 @@
...
+  public FileBasedPersistenceContext createPersistenceContextWithin(PersistenceSpaceIdentifier<?> identifier, String name)
+      throws CachePersistenceException {
+    if (containsSpace(identifier)) {
+      return persistenceService.createContextWithin(identifier.toString(), name);
+    } else {
+      throw new CachePersistenceException("Unknown space: " + identifier);
+    }
+  }
+
+  //////PRIVATE/////////
```
ljacomet: Please match code style - we do not use such separators

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
rkavanap: In case there are both frs and disk (and maybe hybrid) under the same cache manager. So when I destroy one it does not destroy the other..

ljacomet:
Given the release of 3.1.0, this PR cannot be merged before master has been moved to 3.2.0-SNAPSHOT

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -16,74 +16,44 @@
...
-    ILLEGALS.add('"');
-    ILLEGALS.add('|');
-    ILLEGALS.add('?');
-    ILLEGALS.add('*');
-    ILLEGALS.add('.');
-  }
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultLocalPersistenceService.class);
 
-  private final ConcurrentMap<String, PersistenceSpace> knownPersistenceSpaces = new ConcurrentHashMap<String, PersistenceSpace>();
+  private final ConcurrentMap<String, SafeSpace> createdSpaces = new ConcurrentHashMap<String, SafeSpace>();
```
albinsuresh: Is it really required to maintain this map internally? The user of this service should be maintaining the reference to the `SafeSpaceIdentifier` instances that it prepares? Isn't the `SafeSpaceIdentifier` just a logical representation of the `SafeSpace`? And the `SafeSpaceIdentifier` generation is idempotent using the same owner and identifier values, right? So I'm not sure if we need to maintain it here. Let the caller keep the references to the `SafeSpaceIdentifier` that it has prepared or prepare a new one which will provide the same logical `SafeSpaceIdentifier`. I'm viewing this `SafeSpaceIdentifier` as analogous to `File` which is just a logical representation of a physical file just like `SafeSpaceIdentifier` is for a `SafeSpace`. May be there is something that I'm missing.

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -153,363 +101,196 @@ public synchronized void stop() {
...
   }
 
   /**
    * {@inheritDoc}
    */
   @Override
-  public boolean handlesResourceType(ResourceType<?> resourceType) {
-    return ResourceType.Core.DISK.equals(resourceType);
-  }
+  public SafeSpaceIdentifier prepareSafeSpace(String owner, String identifier) {
```
albinsuresh: The name as _prepareSafeSpaceIdentifier_ or __createSafeSpaceIdentifier_, may be? Since it is just about creating the identifier? Just a personal opinion.

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -153,363 +97,168 @@ public synchronized void stop() {
...
-  private File getDirectoryFor(String identifier) {
-    File directory = new File(rootDirectory, safeIdentifier(identifier));
-
-    for (File parent = directory.getParentFile(); parent != null; parent = parent.getParentFile()) {
-      if (rootDirectory.equals(parent)) {
-        return directory;
-      }
+  @Override
+  public synchronized void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider) {
+    if (rootDirectory == null) {
```
ljacomet: Why is that not an error?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -153,363 +97,168 @@ public synchronized void stop() {
...
-  }
 
-  private static MessageDigest getSha1Digest() {
-    try {
-      return MessageDigest.getInstance("SHA-1");
-    } catch (NoSuchAlgorithmException e) {
-      throw new AssertionError("All JDKs must have SHA-1");
-    }
+  private SafeSpace createSafeSpaceLogical(String owner, String identifier) {
+    File ownerDirectory = new File(rootDirectory, owner);
```
ljacomet: Is owner considered _safe_ because it is provided by Ehcache services and not user input?

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,64 @@
...
+
+  /**
+   * Create a physical persistent context (directory) within the safe space.
+   *
+   * @param safeSpaceId Safe space identifier
+   * @return persistent context
    */
-  FileBasedPersistenceContext createPersistenceContextWithin(PersistenceSpaceIdentifier<?> identifier, String name) throws CachePersistenceException;
-}
+  FileBasedPersistenceContext createContextWithin(SafeSpaceIdentifier safeSpaceId, String name) throws CachePersistenceException;
```
ljacomet: As discussed with @rkavanap this method does not belong here

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,64 @@
...
+  void destroyAll(String owner);
+
+  /**
+   * Destroy safe space given owner and identifier.
+   *
+   * @param owner owner of safe spaces
+   * @param name name of safe spaces
+   * @param verbose log more
+   */
+  void destroySafeSpace(String owner, String name, boolean verbose);
```
ljacomet: Given that the `SafeSpaceIdentifier` can be obtained without any file based action, this makes little sense as one can always invoke the overloaded method.

ljacomet:
Still having issues with the naming here.

`LocalPersistenceService` and `FileBasedPersistenceService` are names too close to each other.

The first one represents the ability to obtain a `FileBasedPersistenceContext` to store information on disk. The second one is tailored for the needs of `DISK` resource and a close companion to the `DiskStore.Provider`.

Should they be named: `FileBasedPersistenceService` and DiskResourceService`? any other proposals?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
rkavanap: Yes. owner is typically hardcoded, pre-defined values used by Ehcache persistable resource services only. It is to identify a root for a given service that is separate from other services,

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
rkavanap: yes. this should be an error. Infact, this check is not even necessary as this object will not exist if rootDirectory is null...

rkavanap:
Addressed @ljacomet comments into 2 separate commits. 1 commit has the renames and 1 commit has other changes Will squash into a single commit post final review. 
 
Btw, I have not renamed LocalPersistenceService, since FileBasedPersistenceService is now renamed to DiskResourceService that reduced the confusion that @ljacomet stated above. But if @ljacomet  still thinks LocalPeristenceService should be renamed, it can be done in another commit (which can also then rename FileBasedStateRepository to DiskResourceStateRepository)..

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,50 @@
...
 
   /**
-   * Creates a new persistence context within the given space.
+   * Create the safe spaces in actual physical space for the owner.
    *
-   * @param identifier space to create within
-   * @param name name of the context to create
-   * @return a {@link FileBasedPersistenceContext}
+   * @param safeSpaceId prepared logical space on which the physical space
    */
```
ljacomet: missing javadoc on `name` param

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,50 @@
...
+   * @param owner Service owner that owns the safe space.
+   * @param name Identifying name for the space.
+   *
+   * @return Opaque Identifier that can be used to identify the safe space.
+   */
+  SafeSpaceIdentifier createSafeSpaceIdentifier(String owner, String name);
 
   /**
-   * Creates a new persistence context within the given space.
+   * Create the safe spaces in actual physical space for the owner.
```
ljacomet: Typo: Creates

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,50 @@
...
+  void destroySafeSpace(SafeSpaceIdentifier safeSpaceId, boolean verbose);
+
+  /**
+   * Destroy all safe spaces provided to this owner.
+   *
+   * @param owner owner of safe spaces.
+   */
+  void destroyAll(String owner);
+
+  // identifier that also returns the root of the space
```
ljacomet: Let's have proper javadoc here on the interface and its method. Especially as there is a catch on the `File` return, you cannot assume it exists ... and you should not be the one creating it ...

impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
```diff
@@ -207,7 +207,7 @@ T build(ServiceLocator serviceLocator) throws IllegalStateException {
       if (id == null) {
         throw new IllegalStateException("Persistent user managed caches must have an id set");
       }
-      LocalPersistenceService persistenceService = serviceLocator.getService(LocalPersistenceService.class);
+      DiskResourceService persistenceService = serviceLocator.getService(DiskResourceService.class);
```
ljacomet: lets rename the variable as well.

impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
```diff
@@ -296,8 +296,8 @@ public void close() throws Exception {
     eventDispatcher.setStoreEventSource(store.getStoreEventSource());
 
     if (persistent) {
-      LocalPersistenceService persistenceService = serviceLocator
-          .getService(LocalPersistenceService.class);
+      DiskResourceService persistenceService = serviceLocator
```
ljacomet: lets rename the variable as well

impl/src/main/java/org/ehcache/impl/internal/persistence/DefaultFileBasedPersistenceServiceFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.ehcache.core.spi.service.DiskResourceService;
+import org.ehcache.core.spi.service.ServiceFactory;
+import org.ehcache.impl.persistence.DefaultDiskResourceService;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+
+/**
+ *
+ */
+public class DefaultFileBasedPersistenceServiceFactory implements ServiceFactory<DiskResourceService> {
```
ljacomet: That needs a rename to match the renamed service

core/src/main/java/org/ehcache/core/spi/service/DiskResourceService.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.core.spi.service;
+
+import org.ehcache.CachePersistenceException;
+import org.ehcache.spi.persistence.PersistableResourceService;
+
+/**
+ * Service to provide file based persistence context to caches tied to disk based stores.
```
ljacomet: I would rephrase this class level doc (in place of the two paragraphs you have):

> Resource service handling file level operations for disk tiers.

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,63 @@
...
+import org.ehcache.spi.service.MaintainableService;
+
+import java.io.File;
 
 /**
- * Service to provide persistence context to caches requiring it.
- * <P>
- * Will be used by caches with a disk store, whether or not the data should survive a program restart.
- * <P/>
+ * Service that provides unique local persistence spaces (and safe sub spaces) to any service that requires it
```
ljacomet: > unique local persistence spaces (and safe sub spaces)

changed to
> isolated persistence spaces

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,63 @@
...
    *
-   * @param identifier space to create within
-   * @param name name of the context to create
-   * @return a {@link FileBasedPersistenceContext}
+   * @return Opaque Identifier that can be used to identify the safe space.
+   */
+  SafeSpaceIdentifier createSafeSpaceIdentifier(String owner, String name);
+
+  /**
+   * Creates or finds the safe space represented by {@code safeSpaceId} in the underlying physical space.
```
ljacomet: > Creates or finds ...

changed to
> Creates if it does not exist ...

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -17,22 +17,63 @@
...
+   * @param owner owner of safe spaces.
+   */
+  void destroyAll(String owner);
+
+  /**
+   * Identifier to the logical safe space
    */
-  FileBasedPersistenceContext createPersistenceContextWithin(PersistenceSpaceIdentifier<?> identifier, String name) throws CachePersistenceException;
-}
+  interface SafeSpaceIdentifier {
```
ljacomet: Should we rename that do `FileSpaceIdentifier`? Not sure the `Safe` brings anything.

impl/src/main/java/org/ehcache/impl/internal/persistence/DefaultDiskResourceServiceFactory.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.ehcache.impl.internal.persistence;
+
+import org.ehcache.core.spi.service.DiskResourceService;
+import org.ehcache.core.spi.service.ServiceFactory;
+import org.ehcache.impl.persistence.DefaultDiskResourceService;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+
+/**
```
ljacomet: Get rid of the empty javadoc block

core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
rkavanap: Isn't SafeSpaceIdentifier consistent with the rest of the calls (such as createSafeSpaceId..etc). So should I rename every method in the interface to FileSpace from SafeSpace?.

Isn't FileSpace pleonastic? or another way to look at it is space for file (which is directory). Since the service does a bit more than providing a directory (such as locking to prevent accidental use of the same directory under different cache managers and validating and unique naming etc.) thought SafeSpace conveys the meaning better.

rkavanap:
test this please

impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
```diff
@@ -0,0 +1,274 @@
...
+      return null;
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void destroy(String name) throws CachePersistenceException {
+    if (persistenceService == null) {
```
ljacomet: So because it is null, you have no option to start it even if you wanted to. And that explains the ignored tests.
Let me think about this ...

ljacomet: Let's go ahead with this merge but not close #1192 under which I want to address re-enabling the tests.
So go ahead, rebase and squash commits.

ljacomet: @henri-tremblay May want to give it some thought. Because this is now decomposed in two services, the solution we came up with for 3.1.2 no longer works. However, we should stick with the public behaviour IMO. Just find a different implementation to fix the problem.

Issue #1139, opened 2016-06-07
Clustered CacheManager configuration documentation


ljacomet:
Add documentation on the different configuration options:
* `offheap-resource` on the server side
* `default-resource` concept on cache manager
* `shared-pool` concept on cache manager

PR #1218, opened 2016-06-24
Issue #1139 Generalise image lookup for asciidocs

albinsuresh:


ljacomet:
This needs a slightly different solution in order to work with the website.

Issue #1219, opened 2016-06-24
Create after destroy throws IllegalStateException


AbfrmBlr:
Creating a cache with same alias after destorying the same throws java.lang.IllegalStateException: Cache  already present!

AbfrmBlr:
```java
CacheConfigurationBuilder<Long, String> cacheConfigurationBuilder = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
    ResourcePoolsBuilder.newResourcePoolsBuilder().disk(10, MemoryUnit.MB, true));
PersistentCacheManager persistentCacheManager = CacheManagerBuilder.newCacheManagerBuilder()
    .with(CacheManagerBuilder.persistence(getStoragePath() + File.separator + "myData")) 
    .withCache("persistent-cache", cacheConfigurationBuilder)
    .build(true);

Cache<Long, String> cache = persistentCacheManager.getCache("persistent-cache", Long.class, String.class);
cache.put(1L, "one");
assertThat(cache.get(1L), equalTo("one"));
persistentCacheManager.destroyCache("persistent-cache");

persistentCacheManager.createCache("persistent-cache", cacheConfigurationBuilder);
```

AbfrmBlr:
Creation of cache with same alias fails after destroying. It affects both disk and clustered caches

ljacomet:
Re-opened because the fix causes another issue.

It must be valid to invoke `destroyCache` for a _not known_ alias which will just try to get rid of it persistent storage.

cljohnso:
If it helps any ... the following is a stack trace from a failure I get from one of my (new) tests after the base PR for this issue is applied.  The operation being tested is `EhcacheManager.destroyCache(String)`; under conditions of the test, a timeout-related error is expected because the server is down.

```text
java.lang.IllegalStateException: Cache 'simple-cache' unknown!
	at org.ehcache.core.config.DefaultConfiguration.removeCacheConfiguration(DefaultConfiguration.java:136)
	at org.ehcache.core.EhcacheManager.removeAndCloseWithoutNotice(EhcacheManager.java:649)
	at org.ehcache.core.EhcacheManager.destroyCache(EhcacheManager.java:658)
	at org.ehcache.clustered.TerminatedServerTest$4.runTask(TerminatedServerTest.java:255)
	at org.ehcache.clustered.TerminatedServerTest$4.runTask(TerminatedServerTest.java:252)
[...]

PR #1220, opened 2016-06-24
Removing config from DefaultConfiguration #1219

AbfrmBlr:


AbfrmBlr:
Will raise the PR in accordance with daggy fix as this is a 3.0 bug as well

PR #1223, opened 2016-06-24
Removing config in destroy #1219

AbfrmBlr:


ljacomet:
Do not merge for now - not sure the bug report is valid.

ljacomet:
Looked a bit deeper, this is good to go once rebased

PR #1228, opened 2016-06-27
Issue 1221: rework client / server entity bits' error handling to provide more meaningful error messages

lorban:


clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -273,10 +279,20 @@ public void disconnected(ClientDescriptor clientDescriptor) {
...
+          return invokeLifeCycleOperation(clientDescriptor, (LifecycleMessage) message);
+        case SERVER_STORE_OP:
+          return invokeServerStoreOperation(clientDescriptor, (ServerStoreOpMessage) message);
+        default:
+          throw new IllegalMessageException("Unknown message : " + message);
+      }
+    } catch (ClusteredEhcacheException e) {
+      return responseFactory.failure(e);
+    } catch (Exception e) {
+      LOGGER.error("Unexpected exception raised during LifeCycle operation: " + e, e);
```
albinsuresh: Why is the error message only about LifeCycleOperation? Why was ServerStoreOperations excluded?

lorban: You're right; this message should be more generic as it applies to both `LifeCycleOperation` and `ServerStoreOperations`.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -300,22 +301,22 @@ public void destroy(String name) throws CachePersistenceException {
...
-      try {
-        this.entity.validateCache(cacheId, clientStoreConfiguration);
-      } catch (IllegalStateException e) {
+    try {
+      if (configuration.isAutoCreate()) {
         try {
-          this.entity.createCache(cacheId, clientStoreConfiguration);
-        } catch (CachePersistenceException ex) {
-          throw new ClusteredStoreCreationException("Error creating server-side cache for " + cacheId, ex);
+          this.entity.validateCache(cacheId, clientStoreConfiguration);
```
ljacomet: This is opened to races - we most likely need a loop around here to make sure we exit after having validated configuration always.

PR #1233, opened 2016-06-27
closes #1193 : Issue 1193

### LINKED ISSUE #1193: Assess current destroy support
* Destroy a cache having a clustered tier
  * No other clients can be connected
  * Still closes the local cache
  * Throws to indicate did not destroy
* Destroy a cache manager having a clustered entity
  * Requires no clients have fetched the entity or hold the maintenance lease
  * Still destroys all local state!
  * Throws on failure, CM back to un-initialized state

Need to know what remains to be done and how we handle corner cases

AbfrmBlr:
Added some tests to verify destroy behavior for cache and cachemanager for user facing apis

### DIFF
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.client;
+
+import org.ehcache.Cache;
+import org.ehcache.CachePersistenceException;
+import org.ehcache.PersistentCacheManager;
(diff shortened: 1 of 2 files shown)


ljacomet:
The `Closes` or other word to trigger issue closing is better put as a commit message, the second in this case.
Also could you fix the typos in the messages? destory -> destroy

clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
```diff
@@ -0,0 +1,171 @@
...
+  @Test
+  public void testDestroyCacheManagerWithMultipleClients() throws CachePersistenceException {
+    PersistentCacheManager persistentCacheManager1 = clusteredCacheManagerBuilder.build(true);
+    PersistentCacheManager persistentCacheManager2 = clusteredCacheManagerBuilder.build(true);
+
+    persistentCacheManager1.close();
+
+    try {
+      persistentCacheManager1.destroy();
+    } catch (StateTransitionException e) {
```
albinsuresh: Wouldn't it be safer to add an Assert.fail here after line 81?

clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
```diff
@@ -0,0 +1,171 @@
...
+        .withCache("test", CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
+            ResourcePoolsBuilder.newResourcePoolsBuilder()
+                .with(ClusteredResourcePoolBuilder.fixed("primary-server-resource", 32, MemoryUnit.MB))));
+
+    PersistentCacheManager persistentCacheManager1 = cacheManagerBuilder.build(true);
+    PersistentCacheManager persistentCacheManager2 = cacheManagerBuilder.build(true);
+
+    persistentCacheManager1.close();
+    try {
+      persistentCacheManager1.destroy();
```
albinsuresh: Same as above.

PR #1234, opened 2016-06-27
Issue #1231 Do not wait for replace completion before returning from get

chrisdennis:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -211,32 +211,38 @@ public EhcacheEntityResponse invoke(EhcacheEntityMessage message, boolean waitUn
     }
   }
 
-  private EhcacheEntityResponse invokeInternal(EhcacheEntityMessage message, boolean waitUntilReplicated)
+  private EhcacheEntityResponse invokeInternal(EhcacheEntityMessage message, boolean replicate)
```
ljacomet: If you change that param name, could you change it as well in `invoke(EhcacheEntityMessage, boolean)`?

PR #1235, opened 2016-06-27
ehcache 3 kit structure and content work.

jhouserizer:
Ehcache kit needs a README.txt file that is relevant to it - including pointing out where to find docs and samples.

Also it needs to include enough info in the 'legal' folder to help users sort out what is licensed which way.

### DIFF
--- a/ehcache-kit/src/assemble/README.txt
+++ b/ehcache-kit/src/assemble/README.txt
@@ -1,0 +1,26 @@
+Ehcache 3.1
+=========================
+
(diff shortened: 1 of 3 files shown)


jhouserizer:
The idea is that these contents would get "merged over" the contents of the terracotta-kit  contents.

ljacomet:
All this content should go under `clustered/clustered-dist` and indeed `src/assemble` might be a good location.

jhouserizer:
I haven't yet added any maven-related work - hoping akomakom can help with that.

ljacomet:
well ... no maven here 😉

PR #1236, opened 2016-06-27
ehcache 3.1 kit structure and content work.

jhouserizer:
I have not yet done any work to make this properly fall into the resulting ehcache 3.1 kit.

We want the contents of src/assemble to overlay what is currently going into the kit  (which includes things from `terracotta-oss\terracotta-kit`).

### DIFF
--- a/clustered/clustered-dist/src/assemble/README.txt
+++ b/clustered/clustered-dist/src/assemble/README.txt
@@ -1,0 +1,26 @@
+Ehcache 3.1
+=========================
+
+Welcome to version 3.1 of Ehcache with Terrcotta-based distributed caching. 
+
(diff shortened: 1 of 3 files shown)


akomakom:
So this should go into ./legal/ at the top level?   Should I trust the right column of the doc at this point or is that still in flux?

ljacomet:
@akomakom Yes, you can start trusting that - at least for the `legal` folder. Do you plan on taking this commit and include in a PR doing the packaging or do you want this to be committed first?

akomakom:
I'd say merge this if it's finished, I'll start a new one

ljacomet:
Rebased and merged command line

PR #1248, opened 2016-06-29
Issue #1219 rel 3.0

AbfrmBlr:


core/src/main/java/org/ehcache/core/config/DefaultConfiguration.java
```diff
@@ -132,9 +132,7 @@ public void addCacheConfiguration(final String alias, final CacheConfiguration<?
    * @throws IllegalStateException if the alias was not in use
```
ljacomet: Javadoc needs to be updated if you are removing the exception!

ljacomet:
test this please

PR #1251, opened 2016-06-29
Closes #1205 Prevent clients from interacting with a server without any resources configured

### LINKED ISSUE #1205: Default config of the kit's server doesn't contain any resource pool
That makes the server virtually useless - and even confusing - as you can create entities but can't get a single working cache.

albinsuresh:
Dependent on https://github.com/Terracotta-OSS/terracotta-platform/pull/96

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
@@ -41,6 +41,7 @@ import org.ehcache.clustered.common.exceptions.InvalidStoreManagerException;
 import org.ehcache.clustered.common.exceptions.LifecycleException;
 import org.ehcache.clustered.common.exceptions.ResourceBusyException;
 import org.ehcache.clustered.common.exceptions.ResourceConfigurationException;
+import org.ehcache.clustered.common.exceptions.ServerMisconfigurationException;
 import org.ehcache.clustered.common.messages.EhcacheEntityMessage;
 import org.ehcache.clustered.common.messages.EhcacheEntityResponse;
 
@@ -60,6 +61,7 @@ import org.terracotta.entity.ServiceConfiguration;
 import org.terracotta.entity.ServiceRegistry;
 import org.terracotta.offheapresource.OffHeapResource;
 import org.terracotta.offheapresource.OffHeapResourceIdentifier;
+import org.terracotta.offheapresource.OffHeapResources;
 import org.terracotta.offheapstore.buffersource.OffHeapBufferSource;
 import org.terracotta.offheapstore.paging.OffHeapStorageArea;
 import org.terracotta.offheapstore.paging.Page;
@@ -89,6 +91,7 @@ class EhcacheActiveEntity implements ActiveServerEntity<EhcacheEntityMessage, Eh
 
   private final UUID identity;
   private final ServiceRegistry services;
+  private final Set<String> offHeapResourceIdentifiers;
 
   /**
    * The name of the resource to use for dedicated resource pools not identifying a resource from which
@@ -158,11 +161,24 @@ class EhcacheActiveEntity implements ActiveServerEntity<EhcacheEntityMessage, Eh
     }
   }
 
+  private static class OffHeapResourcesServiceConfiguration implements ServiceConfiguration<OffHeapResources> {
+    @Override
+    public Class<OffHeapResources> getServiceType() {
+      return OffHeapResources.class;
+    }
+  }
+
   EhcacheActiveEntity(ServiceRegistry services, byte[] config) {
     this.identity = ClusteredEhcacheIdentity.deserialize(config);
     this.services = services;
     this.responseFactory = new EhcacheEntityResponseFactory();
     this.clientCommunicator = services.getService(new CommunicatorServiceConfiguration());
+    OffHeapResources offHeapResources = services.getService(new OffHeapResourcesServiceConfiguration());
+    if (offHeapResources == null) {
+      this.offHeapResourceIdentifiers = Collections.emptySet();
+    } else {
+      this.offHeapResourceIdentifiers = offHeapResources.getAllIdentifiers();
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


clustered/build.gradle
```diff
@@ -27,6 +27,7 @@ ext {
 
 subprojects {
   repositories {
+    mavenLocal()
```
albinsuresh: Will be reverted once the dependent platform PR gets merged.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -151,18 +153,31 @@
     }
   }
 
-  private static class CommunicatorServiceConfiguration implements ServiceConfiguration<ClientCommunicator> {
+  static class CommunicatorServiceConfiguration implements ServiceConfiguration<ClientCommunicator> {
```
albinsuresh: Did that for some testing purposes.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -279,6 +294,13 @@ public void disconnected(ClientDescriptor clientDescriptor) {
 
   @Override
   public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEntityMessage message) {
+    if(this.offHeapResourceIdentifiers.isEmpty()) {
```
albinsuresh: Decided to put this here and not just in configure to prevent ANY interaction with a useless server.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -279,6 +294,13 @@ public void disconnected(ClientDescriptor clientDescriptor) {
...
+      return responseFactory.failure(
+          new IllegalStateException("Server started without any offheap resources defined." +
```
albinsuresh: Used `IllegalStateException` here and not any other specific types that were recently added because I feel that in this case the server is really in an illegal state. But please do suggest if there is something better.

clustered/server/src/test/java/org/ehcache/clustered/server/EhcacheActiveEntityTest.java
```diff
@@ -2598,13 +2704,25 @@ private TestOffHeapResource getResource(String resourceName) {
       if (serviceConfiguration instanceof OffHeapResourceIdentifier) {
         final OffHeapResourceIdentifier resourceIdentifier = (OffHeapResourceIdentifier) serviceConfiguration;
         TestOffHeapResource offHeapResource = this.pools.get(resourceIdentifier);
-        if (offHeapResource == null && offHeapSize != 0) {
-          offHeapResource = new TestOffHeapResource(offHeapSize);
-          this.pools.put(resourceIdentifier, offHeapResource);
-        }
+//        if (offHeapResource == null && offHeapSize != 0) {
```
ljacomet: Please remove commented out code.

ljacomet:
This misses the client handling the exception and tearing down the invalid entity.

ljacomet:
Forget my previous comment - this has already been addressed before.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
lorban: `ResourceConfigurationException` sounds better to me. Let's stop using any form of `RuntimeException` across the wire, please. Those are a pain to handle on the client side.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -158,11 +160,24 @@
...
+    }
+  }
+
   EhcacheActiveEntity(ServiceRegistry services, byte[] config) {
     this.identity = ClusteredEhcacheIdentity.deserialize(config);
     this.services = services;
     this.responseFactory = new EhcacheEntityResponseFactory();
     this.clientCommunicator = services.getService(new CommunicatorServiceConfiguration());
+    OffHeapResources offHeapResources = services.getService(new OffHeapResourcesServiceConfiguration());
+    if(offHeapResources == null) {
```
GaryWKeim: Nit: should be "if ("

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -279,6 +294,13 @@ public void disconnected(ClientDescriptor clientDescriptor) {
 
   @Override
   public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEntityMessage message) {
+    if (this.offHeapResourceIdentifiers.isEmpty()) {
+      return responseFactory.failure(
+          new IllegalStateException("Server started without any offheap resources defined." +
```
lorban: Why not throwing `ResourceConfigurationException` here? And I really mean throwing, not returning a failure that wraps it.

albinsuresh: 2 reasons:
1. If I throw any exception from here it'll crash the server.
2. Can't throw `ResourceConfigurationException` which is a checked exception directly from this invoke method.

May be you've mistaken the place from which I'm throwing this. I'm throwing from _invoke()_ not _configure()_.

lorban: You're right, I didn't notice you were outside the try - catch block. But you could simply move that code within it.

Maybe we should keep `ResourceConfigurationException` for client-server configuration mismatch, but in that case another exception that extends from `ClusteredEhcacheException` should be used.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -158,11 +160,24 @@
...
+  }
+
   EhcacheActiveEntity(ServiceRegistry services, byte[] config) {
     this.identity = ClusteredEhcacheIdentity.deserialize(config);
     this.services = services;
     this.responseFactory = new EhcacheEntityResponseFactory();
     this.clientCommunicator = services.getService(new CommunicatorServiceConfiguration());
+    OffHeapResources offHeapResources = services.getService(new OffHeapResourcesServiceConfiguration());
+    if (offHeapResources == null) {
+      this.offHeapResourceIdentifiers = Collections.EMPTY_SET;
```
chrisdennis: `Collections.emptySet()`

ljacomet:
test this please

Issue #1253, opened 2016-06-29
Clean error message for invalid features combination


ljacomet:
Ehcache 3.1 will have the following invalid feature combinations:

* Clustered and events
* Clustered and transactions
* Clustered and `CacheLoaderWriter`

We need to make sure exceptions are thrown at initialisation time with a clear error message.

AbfrmBlr:
Clustered & `CacheLoaderWriter` throws `java.lang.UnsupportedOperationException: Implement me`

AbfrmBlr:
Clustered & events never complains

AbfrmBlr:
Clustered & XA fails during creation due to null pointer in `DefaultJournalProvider`.

```java
Caused by: java.lang.IllegalStateException: Cache 'xaCache' creation in EhcacheManager failed.
	at org.ehcache.core.EhcacheManager.createCache(EhcacheManager.java:285)
	at org.ehcache.core.EhcacheManager.init(EhcacheManager.java:564)
	... 31 more
Caused by: java.lang.NullPointerException
	at org.ehcache.transactions.xa.internal.journal.DefaultJournalProvider.getJournal(DefaultJournalProvider.java:57)
	at org.ehcache.transactions.xa.internal.XAStore$Provider.createStore(XAStore.java:886)
	at org.ehcache.core.EhcacheManager.getStore(EhcacheManager.java:491)
	at org.ehcache.core.EhcacheManager.createNewEhcache(EhcacheManager.java:308)
	at org.ehcache.core.EhcacheManager.createCache(EhcacheManager.java:257)
	... 32 more

```

AbfrmBlr:
Ideally all of the above cases should fail during cachemanager init itself during cache creation saying these cases are not supported yet

PR #1255, opened 2016-06-29
Validate cache unsupported combination #1253

AbfrmBlr:
This PR is for events and loaderwriter


lorban:
Nice, can you think of a way to also prevent the XA / ClusteredStore combination?

ljacomet:
Merged command line

PR #1260, opened 2016-06-29
Closes #1150 - management xml config

### LINKED ISSUE #1150: Add XML configuration support for management module

mathieucarbou:
Closes #1150 - management xml config

@GaryWKeim @anthonydahanne : there are 3 xml samples at the end: ehcache-management-1.xml, ehcache-management-2.xml, ehcache-management-3.xml.


management/src/test/resources/ehcache-management-4.xml
```diff
@@ -0,0 +1,29 @@
...
+
+  <service>
+    <mnm:management cacheManagerAlias="my-cache-manager-name">
+      <mnm:tags>
+        <mnm:tag>webapp-name</mnm:tag>
+        <mnm:tag>jboss-1</mnm:tag>
+        <mnm:tag>server-node-1</mnm:tag>
+      </mnm:tags>
+      <mnm:statistics>
+        <mnm:configuration averageWindowDuration="2"
```
GaryWKeim: Formatting and, these units are in seconds I assume?

mathieucarbou: `minutes`.
If you look at the sample 3, all the fields are there.

I declared some default in the XSD...

I hesitated between using optional fields with default values or required fields, but you need to set all.

What do you think ?

ljacomet:
The whole XML bit seems very alien compared to the rest of Ehcache. Could you have a look and try to align?
For example, when specifying time, we have a tag with its value the duration and an attribute `unit` which usually defaults to seconds.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -985,7 +985,7 @@ public String threadPool() {
 
   }
 
-  private static TimeUnit convertToJavaTimeUnit(org.ehcache.xml.model.TimeUnit unit) {
+  public static TimeUnit convertToJavaTimeUnit(org.ehcache.xml.model.TimeUnit unit) {
```
ljacomet: Just to get to that method ... let's move it to another class instead of making ConfigurationParser public!

mathieucarbou:
ok I'll have a look. I looked at tx and clustered but not at the plain ehcache xml files as examples.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
mathieucarbou: ok!

ljacomet:
If this PR ends up requiring any existing API or SPI changes, it will need to wait before master is moved to 3.2.0-SNAPSHOT

mathieucarbou:
@anthonydahanne @GaryWKeim @ljacomet : I updated the XSD and samples to better match other Ehcache's schemas.

Can you have a look at the [samples here in this PR](https://github.com/mathieucarbou/ehcache3/tree/1cce7ca0b54899055376a7a8f02e65a9ab721dda/management/src/test/resources) ?

Thanks :-)

management/src/main/java/org/ehcache/management/config/EhcacheStatisticsProviderConfiguration.java
```diff
@@ -22,63 +22,14 @@
...
-public class EhcacheStatisticsProviderConfiguration implements StatisticsProviderConfiguration {
-
-  private long averageWindowDuration;
-  private TimeUnit averageWindowUnit;
-  private int historySize;
-  private long historyInterval;
-  private TimeUnit historyIntervalUnit;
-  private long timeToDisable;
-  private TimeUnit timeToDisableUnit;
+public class EhcacheStatisticsProviderConfiguration extends DefaultStatisticsProviderConfiguration {
```
anthonydahanne: why did you extract a parent class ?

management/src/main/java/org/ehcache/management/ManagementRegistryServiceConfiguration.java
```diff
@@ -46,6 +48,8 @@
    */
   String getCollectorExecutorAlias();
 
+  Collection<String> getTags();
```
anthonydahanne: javadoc ? for example : 
  /**
   * Gets the tags set by the user on the Management service
   *
   * @return The static colector executor alias
   */

management/src/main/java/org/ehcache/management/config/EhcacheStatisticsProviderConfiguration.java
mathieucarbou: Because you can have several stat configuration : one per provider actually.

EhcacheStatisticsProviderConfiguration is one.

So when configuring through XML (see `<statisticConfiguration>`), you have to optionally specify the provider you want this config to be applied to.

management/src/main/java/org/ehcache/management/ManagementRegistryServiceConfiguration.java
mathieucarbou: will add

management/build.gradle
```diff
@@ -17,7 +17,7 @@
 apply plugin: EhDeploy
 
 dependencies {
-  compile project(':api'), project(':core'), project(':impl')
+  compile project(':api'), project(':core'), project(':impl'), project(':xml')
   compile group: 'org.terracotta.management', name: 'management-registry', version: parent.managementVersion
   testCompile "com.fasterxml.jackson.core:jackson-databind:2.6.3"
```
anthonydahanne: upgrade to 2.7.5 ?

management/src/test/java/org/ehcache/management/providers/settings/EhcacheSettingsProviderTest.java
```diff
@@ -106,7 +104,9 @@ public void test_standalone_ehcache() throws IOException {
         .using(new DefaultPersistenceConfiguration(ROOT.newFolder("test_standalone_ehcache")))
         .using(new DefaultManagementRegistryConfiguration()
             .addConfiguration(statisticsProviderConfiguration)
-            .setCacheManagerAlias("my-cm-1"))
+            .setCacheManagerAlias("my-cm-1")
+            .addTag("boo")
+            .addTags("foo", "baz"))
         .withCache("cache-1", cacheConfiguration1)
         .withCache("cache-2", cacheConfiguration2)
         .build(false);
```
anthonydahanne: so I know it's not from this PR, but I suggest we change the json comparison (because the current one is limited to comparing 1 line long json document, that are not really convenient to read...)
line #121 : 

    JsonNode expectedJsonObject = mapper.readTree(expected);
    JsonNode actualJsonObject = mapper.readTree(actual);
    assertEquals(expectedJsonObject, actualJsonObject);

and then line #138 : 

      return scanner.useDelimiter("\\A").next();

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryConfiguration.java
```diff
@@ -34,14 +36,15 @@
...
+  private final Map<Class<? extends ManagementProvider>, StatisticsProviderConfiguration<?>> statisticConfigurations = new HashMap<Class<? extends ManagementProvider>, StatisticsProviderConfiguration<?>>();
+  private final Collection<String> tags = new TreeSet<String>();
   private Context context = Context.empty();
   private String statisticsExecutorAlias;
   private String collectorExecutorAlias;
 
   public DefaultManagementRegistryConfiguration() {
     setCacheManagerAlias("cache-manager-" + COUNTER.getAndIncrement());
-    addConfiguration(new EhcacheStatisticsProviderConfiguration(1, TimeUnit.MINUTES, 100, 1, TimeUnit.SECONDS, 30, TimeUnit.SECONDS));
+    addConfiguration(new EhcacheStatisticsProviderConfiguration());
```
anthonydahanne: ah, I guess this explains why the DefaultStatisticsProviderConfiguration !

management/src/test/java/org/ehcache/management/providers/settings/EhcacheSettingsProviderTest.java
mathieucarbou: This has been discussed with @chrisdennis and @cljohnso : the formatted json format is not compatible with windows and makes the build fail... So we changed it to a 1-liner. The idea is able to pinpoint the exact diff in a line.

management/src/main/java/org/ehcache/management/registry/NodeListIterable.java
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      @Override
+      public void remove() {
+        throw new UnsupportedOperationException();
+      }
+    };
+  }
+
+  public static Iterable<Node> nodes(NodeList nodeList) {
```
anthonydahanne: not used

management/src/main/java/org/ehcache/management/registry/NodeListIterable.java
```diff
@@ -0,0 +1,74 @@
...
+    return new NodeListIterable<Node>(nodeList, Node.class);
+  }
+
+  public static Iterable<Element> elements(Element root, String ns, String localName) {
+    return new NodeListIterable<Element>(root.getElementsByTagNameNS(ns, localName), Element.class);
+  }
+
+  public static Iterable<Element> elements(Element root, String tagName) {
```
anthonydahanne: not used

management/src/main/resources/ehcache-management-ext.xsd
```diff
@@ -0,0 +1,89 @@
...
+        <xs:annotation>
+          <xs:documentation>
+            The average window used for statistic computing. Default is 60 seconds.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:element>
+      <xs:element name="history-interval" type="eh:time-type" minOccurs="0" maxOccurs="1" default="1">
+        <xs:annotation>
+          <xs:documentation>
+            Frequency at which the samples are computed. Default is each 1 second.
```
anthonydahanne: Default is every second OR is at each 1 second

mathieucarbou: will update

management/src/main/java/org/ehcache/management/config/EhcacheStatisticsProviderConfiguration.java
anthonydahanne: okee ! thanks

management/src/test/java/org/ehcache/management/providers/settings/EhcacheSettingsProviderTest.java
anthonydahanne: oh, ok...

anthonydahanne:
LGTM

mathieucarbou:
I removed the `@author` tags on newly added files

PR #1261, opened 2016-06-30
Clustered cache adoc updates

gibsong:


docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -196,6 +204,32 @@ include::{sourcedir31}/clustered/client/src/test/resources/configs/docs/ehcache-
 Expiry in clustered caches work with an exception that `Expiry#getExpiryForAccess` is handled on a best effort basis for clustered tiers,
 it may not be as accurate as in the case of local tiers. Though the same is honored accurately in on-heap tiers.
 
+=== Clustered Unknown Inheritance
+
+We have included an option which allows a cache to be created without specifying its clustered resource pool allocation.
+In order to do so you must already have a cache created as either Shared or Dedicated.
```
jhouserizer: You wouldn't create the *cache* as Shared or Dedicated, right?   Not the cache but the pool for the cache?

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -196,6 +204,32 @@ include::{sourcedir31}/clustered/client/src/test/resources/configs/docs/ehcache-
...
+When you create the second cache(with the same name) in a different cache manager you must configure it with a clustered unknown resource pool.
```
jhouserizer: Needs a space between "cache" and "(with the same name".   Also, again, is this language precise enough (is it the cache that needs the same name?  I thought caches didn't have names).

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -211,4 +211,47 @@ public void loadDocsXml() throws Exception {
...
+
+    CacheManagerBuilder<PersistentCacheManager> cacheManagerBuilderExpecting = CacheManagerBuilder.newCacheManagerBuilder()
+            .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("terracotta://localhost:9510/my-application"))
+                .expecting()  // <4>
+                .resourcePool("resource-pool", 32, MemoryUnit.MB, "primary-server-resource"));
+
+    final PersistentCacheManager cacheManager2 = cacheManagerBuilderExpecting.build(false);
+    cacheManager2.init();
+
+    CacheConfiguration<Long, String> cacheConfigUnknown = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
```
jhouserizer: Would it be better to say unspecified rather than unknown?

docs/src/docs/asciidoc/user/clustered-cache.adoc
jhouserizer: "clustered unknown resource pool"   is there a thing called "clustered unknown resource pool"? - the language makes me want to look in the API for a thing called that.  Or do we just mean with an unspecified resource pool?

docs/src/docs/asciidoc/user/clustered-cache.adoc
gibsong: you are right this isn't worded correctly.  I will fix it.

docs/src/docs/asciidoc/user/clustered-cache.adoc
gibsong: I will add the space.  I think it should actually be cache alias not name.

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
gibsong: actually this does sound way better, and makes more sense.  I will get rid of unknown and replace with unspecified.

Issue #1246, opened 2016-06-28
Release 3.1.0


ljacomet:
- [x] Tag release
- [x] Build and verify release
 - includes checking javadoc and source jars
- [x] Update and publish release on GitHub
- [x] Update and publish website
 - includes checking XSDs
- [x] Publish jars to Maven Central
- [x] Update readme / bump version on release branch
- [x] Email announcement

PR #1265, opened 2016-06-30
Issue #1246 Some more package dance

ljacomet:


ljacomet:
That added commit concludes the package dance. 👯

PR #1272, opened 2016-06-30
Issue #1256 Wrap Exception in server response in client-side exception

### LINKED ISSUE #1256: Restore client-side stack trace to server-side exceptions
Updates for issue #763 changed the way server-side exceptions returned to the client in response to clustered cache operations -- the server-side exception is wrapped in client-side exception to provide client-side stack trace information in addition to the server-side stack trace information.

Updates for issue #1221 reverted this behavior.

This issue is directed at re-enabling the wrapping of a server-side exception in a client-side exception to recover the client-side stack trace.

cljohnso:
This commit restores the behavior of wrapping a server-wide exception
received as a clustered operation response in a client-side exception.
This ensures that the client-side stack trace is available in the thrown
exception.

Closes #1256

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
@@ -258,7 +258,7 @@ public class EhcacheClientEntity implements Entity {
          */
         Exception cause = ((Failure)response).getCause();
         if (cause instanceof ClusteredEhcacheException) {
-          throw ((ClusteredEhcacheException) cause).copyInContext();
+          throw ((ClusteredEhcacheException) cause).withClientStackTrace();
         }
         throw new RuntimeException(message + " error: " + cause.toString(), cause);
       } else {
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
@@ -251,9 +251,14 @@ public class EhcacheClientEntity implements Entity {
     try {
       EhcacheEntityResponse response = waitFor(timeLimit, invokeAsync(message, replicate));
       if (Type.FAILURE.equals(response.getType())) {
+        /*
+         * The FAILURE cause is a server-side exception lacking client-side stack trace
+         * elements.  The server-side exception must be wrapped in a client-side exception
+         * to provide proper stack trace for analysis.
+         */
         Exception cause = ((Failure)response).getCause();
         if (cause instanceof ClusteredEhcacheException) {
-          throw (ClusteredEhcacheException) cause;
+          throw ((ClusteredEhcacheException) cause).copyInContext();
         }
         throw new RuntimeException(message + " error: " + cause.toString(), cause);
       } else {
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
@@ -17,6 +17,7 @@
 package org.ehcache.clustered.common.internal.exceptions;
 
 public abstract class ClusteredEhcacheException extends Exception {
+  private static final long serialVersionUID = 38615046229882871L;
 
   public ClusteredEhcacheException(String message) {
     super(message);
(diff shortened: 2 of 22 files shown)


clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
```diff
@@ -29,4 +30,14 @@ public ClusteredEhcacheException(String message, Throwable cause) {
   public ClusteredEhcacheException(Throwable cause) {
     super(cause);
   }
+
+  /**
+   * Makes a copy of this {@code ClusteredEhcacheException}, retaining the type,
+   * in the current context.  This method is used to prepare exceptions originating in
+   * tbe server to be re-thrown in the client.
```
albinsuresh: Minor typo

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -209,9 +209,14 @@ private EhcacheEntityResponse invokeInternal(EhcacheEntityMessage message, boole
     try {
       EhcacheEntityResponse response = waitFor(invokeAsync(message, replicate));
       if (Type.FAILURE.equals(response.getType())) {
+        /*
+         * The FAILURE cause is a server-side exception lacking client-side stack trace
+         * elements.  The server-side exception must be wrapped in a client-side exception
+         * to provide proper stack trace for analysis.
+         */
```
chrisdennis: This comment isn't need imho, that info is already captured in the javadoc of ClusteredEhcacheException.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
```diff
@@ -29,4 +30,14 @@ public ClusteredEhcacheException(String message, Throwable cause) {
...
+   *
+   * @return a new {@code ClusteredEhcacheException} of {@code this} subtype with {@code this}
+   *      instance as the cause
+   */
+  public abstract ClusteredEhcacheException copyInContext();
```
chrisdennis: I'm not enamoured with this method name, it doesn't seem to have the right connotations to me... but damned if I can think of a better one.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
cljohnso: I'm inclined to leave it .. if for nothing else, it serves as a reminder that server-side exceptions included in responses need to be "clientized" before re-throwing.  If someone introduces another server-side exception that gets re-tossed here, they won't necessarily be looking at `ClusteredEhcacheException`.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
cljohnso: Using `withClientStackTrace`

clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
```diff
@@ -29,4 +30,14 @@ public ClusteredEhcacheException(String message, Throwable cause) {
...
+
+  /**
+   * Makes a copy of this {@code ClusteredEhcacheException}, retaining the type, in the current
+   * client context.  This method is used to prepare exceptions originating in the server to be
+   * re-thrown in the client.
+   *
+   * @return a new {@code ClusteredEhcacheException} of {@code this} subtype with {@code this}
+   *      instance as the cause
+   */
+  public abstract ClusteredEhcacheException withClientStackTrace();
```
ljacomet: Do we want generics magic here so that the type can be the exception concrete type?

lorban:
There seems to be some overlap / dependency with #1279. You probably want to talk to @albinsuresh about this.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/ClusteredEhcacheException.java
cljohnso: Exceptions and Generics can't be mixed == "Error:(6, 42) java: a generic class may not extend java.lang.Throwable"

ljacomet: Well, will have to remember that one ... damn language exceptions 😉

cljohnso:
Issue #1279 was expected to be follow-on work to this issue.  The original expectation was that this work would have been included in 3.1 to correct the lack of client-side stack trace in re-thrown server-side exceptions and that #1279 would get to where we really wanted to be.  Since 3.1 has been released without #1272, this can be dropped in deference to #1279 as long as #1279 adheres to the "don't re-throw a server-side exception in the client" tenet.

cljohnso:
Blocked pending resolution with issue #1279 / PR #1283.

ljacomet:
From discussions with @AbfrmBlr, @albinsuresh, @cljohnso and @lorban, this goes in first and we will see what remains to be done for #1279 after.

Issue #1273, opened 2016-06-30
kit zip archive has duplicate README file


aurbroszniowski:
I downloaded the kit from
https://github.com/ehcache/ehcache3/releases/download/v3.1.0/ehcache-clustered-3.1.0-kit.zip

and extracted it in windows **(there was no previous installation, the directory was empty)**
I got a duplicate file error as seen in the screenshot

![readme-dplicate](https://cloud.githubusercontent.com/assets/1264251/16502856/b3adc8e4-3f10-11e6-8234-4329284ebc83.png)

ljacomet:
@GaryWKeim Assigning to you since you have a PR open for this.

GaryWKeim:
It will be easier if you handle this.

anthonydahanne:
I could give it  try

PR #1281, opened 2016-07-01
Explicitly remove README.txt from Voltron kit while assembling release kit

GaryWKeim:
…ling release kit.


ljacomet:
Please add a reference to #1273 in the commit message

ljacomet:
Good to go once rebased

ljacomet:
Some git mess there. Looks like it was merged and not rebased.

PR #1283, opened 2016-07-04
Closes #1279 Limit EhcacheEntityResponse.Failure to return ClusteredEhcacheException

### LINKED ISSUE #1279: Limit EhcacheEntityResponse.Failure to return ClusteredEhcacheException
The `EhcacheEntityResponse.Failure` message is used to return a server-thrown exception.  The message is presently set up to accept any `Exception` which then needs to be re-interpreted on the client to add the client context (most importantly, stack trace).  Updates for issue #1256 adds a method specifically for the adding the client-side stack trace to `ClusteredEhcacheException` instances.  From an error handling hygiene point of view, it would be best to have _only_ `ClusteredEhcacheException` subclasses returned forcing server code to return exceptions that can be more directly re-thrown in the client (or provide more contextual detail for other client-side handling).

albinsuresh:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -316,8 +317,13 @@ public void destroy(String name) throws CachePersistenceException {
         try {
           this.entity.validateCache(cacheId, clientStoreConfiguration);
         } catch (ClusteredTierValidationException ex) {
-          if (ex.getCause() instanceof InvalidStoreException) {
-            this.entity.createCache(cacheId, clientStoreConfiguration);
+          if(ex.getCause() instanceof ClusteredClientException) {
```
ljacomet: The wrapping / unwrapping that leads to these `instanceof` checks are a design smell IMO

lorban:
Why have you moved the server exceptions from `org.ehcache.clustered.common` into `org.ehcache.clustered.server` in the server module? This way, the client doesn't have access to the server exceptions and will fail to deserialize them.

lorban:
There seems to be some overlap with #1256. Both PRs should be aligned, you'll probably want to speak with @cljohnso about this.

albinsuresh:
Was wondering the same. With the split done into client and server side exceptions and the way server side exceptions are wrapped in a client side exception, #1256 will be resolved.

albinsuresh:
I request reviewers to look at the last commit separately and tell me if it doesn't make sense.

cljohnso:
Blocked pending resolution with issue  #1256 / PR #1272.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -143,48 +144,48 @@ public void close() {
...
+      throw new ClusteredTierManagerValidationException("Error validating server clustered tier manager", e.getCause());
     }
   }
 
   public void configure(ServerSideConfiguration config) throws ClusteredTierManagerConfigurationException {
     try {
       invokeInternal(messageFactory.configureStoreManager(config), true);
-    } catch (Exception e) {
-      throw new ClusteredTierManagerConfigurationException("Error configuring clustered tier manager", e);
+    } catch (ClusteredClientException e) {
```
ljacomet: This is very wrong - we end up wrapping a server exception in a client exception to unwrap it immediately and rewrap before throwing.

ljacomet:
Unblocking following comments on the #1279

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -256,11 +244,7 @@ private EhcacheEntityResponse invokeInternal(TimeoutDuration timeLimit, EhcacheE
          * elements.  The server-side exception must be wrapped in a client-side exception
          * to provide proper stack trace for analysis.
          */
-        Exception cause = ((Failure)response).getCause();
-        if (cause instanceof ClusteredEhcacheException) {
-          throw ((ClusteredEhcacheException) cause).withClientStackTrace();
-        }
-        throw new RuntimeException(message + " error: " + cause.toString(), cause);
+        throw ((Failure)response).getCause().withClientStackTrace();
```
ljacomet: Why not push that `withClientStackTrace()` into the `Response.decode` method directly?

albinsuresh: Sure.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -256,11 +244,7 @@ private EhcacheEntityResponse invokeInternal(TimeoutDuration timeLimit, EhcacheE
          * elements.  The server-side exception must be wrapped in a client-side exception
          * to provide proper stack trace for analysis.
          */
```
ljacomet: That comment is no longer relevant since it happens inside the codec

PR #1284, opened 2016-07-04
Closes #1280 Document the constructor contract change for persistent Serializers

### LINKED ISSUE #1280: Custom serializers constructor contract changed for 3.1, docs are not updated
In 3.1, custom Serializers have change from needed a constructor of:

`public MySerializer(ClassLoader classLoader, FileBasedPersistenceContext persistenceContext) ..`

to:

`public MySerializer(ClassLoader classLoader, StateRepository stateRepository) ...`

The current docs (http://www.ehcache.org/documentation/3.1/serializers-copiers.html) do not reflect this.

albinsuresh:


ljacomet:
That's a good first iteration. But I believe the whole section about persistent caches needs a rewrite to focus on the 3.1 constructor contract. And then leave a section indicating that the 3.0 contract remains valid for disk caches.

albinsuresh:
Refactored the content in such a way that the main content is about the latest contracts and there is a backward reference to the older contract.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -145,6 +151,14 @@ will be sanctioned with an exception upon cache initialization.
 
 Registering a serializer that lacks such constructor at the cache manager level will prevent it from being chosen for persistent caches.
 
+NOTE: The constructor with the signature _(ClassLoader classLoader, FileBasedPersistenceContext persistenceContext)_
```
ljacomet: Code extract - use back ticks

albinsuresh: Addressed.

albinsuresh:
test this please

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -145,6 +151,14 @@ will be sanctioned with an exception upon cache initialization.
 
 Registering a serializer that lacks such constructor at the cache manager level will prevent it from being chosen for persistent caches.
 
+NOTE: The constructor with the signature `(ClassLoader classLoader, FileBasedPersistenceContext persistenceContext)`
```
ljacomet: I would make the note much shorter sa we do not want too much focus on this.

Something like:
> The constructor with the signature `(ClassLoader classLoader, FileBasedPersistenceContext persistenceContext)` that existed in v3.0 is still supported to respect backward compatibility but its usage is limited to disk based caches.

albinsuresh: Shortened as suggested.

ljacomet:
Good once rebased

PR #1285, opened 2016-07-04
Closes #1003 - Expose a way to fetch/create entities re-using the same connetion

### LINKED ISSUE #1003: Being able to fetch other entities from Ehcache services by using the same opened cluster connection
The goal is that the (sooner added) `clustered-management` module of Ehcache could fetch the `management-entity` (from the `tc-management` project) using the same voltron connection (=> _SAME CLIENT_) in order to expose through it the metadata from the Ehcache `management` module.

So perhaps a solution I see would be a sort of `ClusteredConnectionService` that would be responsible for providing a voltron connection given a URI, and this service could be used for the clustered stuff, M&M and eventually other user-land ecache services.

The goal is to have the SAME connection (==same CLIENT for this product instance) re-used across the services that would need to fetch entities.

mathieucarbou:
This is a proposal for #1003.

The goal is to be able from other Ehcache services (such as management) to fetch other entities through the same cluster connection. This will come in another PR later (that will depend on this one).

I didn't want to reuse the existing `EhcacheClientEntityFactory` which does really special things concerning Ehcache config validation. 

So I instead created an isolated `ClientEntityFactory`, accessible from the `ClusteringService`, responsible to create, retrieve and destroy entities... This only depends on spi stuff and does not expose internal stuff.

Note: I moved `EhcacheEntityBusyException ` to `EntityBusyException `, into the non internal package since this exception is also used and not really related to a special Ehcache op, because it is used the same way for `tryDestroy()` (in case a destroy is not possible).

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
@@ -18,6 +18,7 @@ package org.ehcache.clustered.client.internal;
 
 import org.ehcache.clustered.client.internal.service.ClusteredTierManagerConfigurationException;
 import org.ehcache.clustered.client.internal.service.ClusteredTierManagerValidationException;
+import org.ehcache.clustered.client.service.EntityBusyException;
 import org.ehcache.clustered.common.ServerSideConfiguration;
 import org.ehcache.clustered.client.internal.lock.VoltronReadWriteLock;
 import org.ehcache.clustered.client.internal.lock.VoltronReadWriteLock.Hold;
@@ -76,18 +77,18 @@ public class EhcacheClientEntityFactory {
    *
    * @throws EntityAlreadyExistsException if the {@code EhcacheActiveEntity} for {@code identifier} already exists
    * @throws EhcacheEntityCreationException if an error preventing {@code EhcacheActiveEntity} creation was raised
-   * @throws EhcacheEntityBusyException if another client holding operational leadership prevented this client
+   * @throws EntityBusyException if another client holding operational leadership prevented this client
    *        from becoming leader and creating the {@code EhcacheActiveEntity} instance
    */
   public void create(final String identifier, final ServerSideConfiguration config)
-      throws EntityAlreadyExistsException, EhcacheEntityCreationException, EhcacheEntityBusyException {
+      throws EntityAlreadyExistsException, EhcacheEntityCreationException, EntityBusyException {
     Hold existingMaintenance = maintenanceHolds.get(identifier);
     Hold localMaintenance = null;
     if (existingMaintenance == null) {
       localMaintenance = createAccessLockFor(identifier).tryWriteLock();
     }
     if (existingMaintenance == null && localMaintenance == null) {
-      throw new EhcacheEntityBusyException("Unable to create clustered tier manager for id "
+      throw new EntityBusyException("Unable to create clustered tier manager for id "
               + identifier + ": another client owns the maintenance lease");
     } else {
       try {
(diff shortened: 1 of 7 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -157,28 +158,26 @@ public EhcacheClientEntity retrieve(String identifier, ServerSideConfiguration c
...
           if (!ref.tryDestroy()) {
-            throw new EhcacheEntityBusyException("Destroy operation failed; " + identifier + " clustered tier in use by other clients");
+            throw new EntityBusyException("Destroy operation failed; " + identifier + " clustered tier in use by other clients");
           }
         } catch (EntityNotProvidedException e) {
           LOGGER.error("Unable to delete clustered tier manager for id {}", identifier, e);
           throw new AssertionError(e);
         } catch (EntityNotFoundException e) {
           throw new EhcacheEntityNotFoundException(e);
-        } catch (EhcacheEntityBusyException e) {
```
mathieucarbou: Note: @chrisdennis : I removed this block:

```
        } catch (EhcacheEntityBusyException e) {
          throw new EhcacheEntityBusyException(e);
        }
```

anthonydahanne:
LGTM, but I too suggest @ljacomet , @chrisdennis , @lorban or @cljohnso give their input

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/AbstractClientEntityFactory.java
```diff
@@ -0,0 +1,126 @@
...
+        ref.create(configuration);
+        try {
+          ref.fetchEntity().close();
+          return;
+        } catch (EntityNotFoundException e) {
+          //continue;
+        }
+      }
+    } catch (EntityNotProvidedException e) {
+      LOGGER.error("Unable to create clustered tier manager for id {}", entityIdentifier, e);
```
ljacomet: Copy pasted exception message need to be adapted

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClientEntityFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package org.ehcache.clustered.client.service;
+
+import org.terracotta.connection.entity.Entity;
+import org.terracotta.exception.EntityAlreadyExistsException;
+import org.terracotta.exception.EntityNotFoundException;
+
+/**
+ * @author Mathieu Carbou
```
ljacomet: Please get rid of `@author` tags, we do not use them

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClientEntityFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface ClientEntityFactory<E extends Entity, C> {
+
+  String getEntityIdentifier();
```
ljacomet: This class, being SPI, needs to be javadoced

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClusteringService.java
```diff
@@ -31,6 +31,8 @@
 
   ClusteringServiceConfiguration getConfiguration();
 
+  <E extends Entity, C> ClientEntityFactory<E, C> newClientEntityFactory(String entityIdentifier, Class<E> entityType, long entityVersion, C configuration);
```
ljacomet: javadoc missing

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/AbstractClientEntityFactory.java
mathieucarbou: will fix!

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClientEntityFactory.java
mathieucarbou: will fix!

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClientEntityFactory.java
mathieucarbou: will fix!

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClusteringService.java
mathieucarbou: will fix!

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClientEntityFactory.java
```diff
@@ -0,0 +1,70 @@
...
+   */
+  String getEntityIdentifier();
+
+  /**
+   * @return The entity type created by this factory
+   */
+  Class<E> getEntityType();
+
+  /**
+   * @return The entoty version
```
ljacomet: typo

mathieucarbou:
@ljacomet : PR updated according to what we discussed :-)

Issue #1282, opened 2016-07-01
Creation/Destruction of Cache Managers does not free offheap memory in clustered case.


cschanck:
I've a test case where the server has a single offheap resource of 1024M.

The test repeatedly:

1. creates a cache manager with 512M of clustered resource pool
2. creates 2 caches using this default resource pool
3. does stuff...
4. destroys the caches
5. destroys the cache manager

On destroy, the OffheapServerStore gets detached from memory, but not reclaimed by gc. I have a heapdump to this effect (showing a number of OffheapServerStore instances with no root; they each eventually keep Page references which have Direct byte buffer references. 

Eventually, the server grows large enough that the allocation of a new cache manager fails. This is not how the allocation of byte buffers is supposed to work, but it is a loooong chain from OffheapServerStore tot he actual buffers -- perhaps some links in that chain need to be broken to make it easier for the VM to see that the buffers can be reclaimed? (speculation).

Alternatively, we might need to call gc() when cache managers (or even caches) are destroyed. Or something.

cschanck:
As a test, I added a destroy() method to ServerStore. In ServerStoreImpl this calls an OffheapServerStore.destroy().

OffheapServerStore.destroy() 1) calls clear() 2) clears the segment list (wrapped in a lock/unlock all). 

Calling this destroy in EhcacheActiveEntity destroy...() then the memory consumption problem goes away, as the offheap resources seem to be shallow enough to be reclaimed.

cschanck:
PR: https://github.com/ehcache/ehcache3/pull/1288

PR #1288, opened 2016-07-05
#1282 Add close() method down ServerStoreImpl chain.

cschanck:
During rapid creation/destruction of caches/cache managers,
I observed that GC did not cleanup the dangling offheap resources
leading to extremely high memory consumption and an eventual
failure to allocate more buffers.

The theory was that the object was to complex for easy GC'ing in the
direct buffer allocation path. By closing the server store, we
eventually remove the reference to each of the ChainMaps, which then
makes it easy for the GC to clean up the memory.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
@@ -182,4 +182,15 @@ public class OffHeapServerStore implements ServerStore {
 
     return evicted;
   }
+
+  public void close() {
+    writeLockAll();
+    try {
+      clear();
+      segments.clear();
+    } finally {
+      writeUnlockAll();
+    }
+  }
+
 }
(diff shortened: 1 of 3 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -15,16 +15,16 @@
...
 import org.ehcache.clustered.common.internal.store.Chain;
 import org.ehcache.clustered.common.internal.store.ServerStore;
 import org.ehcache.clustered.server.ServerStoreEvictionListener;
 import org.terracotta.offheapstore.exceptions.OversizeMappingException;
 import org.terracotta.offheapstore.paging.PageSource;
 
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
```
GaryWKeim: Unnecessary import shuffling.

mathieucarbou:
And commit message should include ref to #1282, such as "Close #1282" ;-)

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
cschanck: Stupid auto format in IntelliJ.

cljohnso:
The "closes", "fixes", etc does not have to be in the top-line commit comment -- anywhere in the initial PR description is good enough.

Issue #1277, opened 2016-07-01
Clustered buffer underflow seen when under pressure


cschanck:
o.e.c.i.r.LoggingRobustResilienceStrategy - Ehcache key 45099 in possible inconsistent state due to 
org.ehcache.core.spi.store.StoreAccessException: java.nio.BufferUnderflowException
	at org.ehcache.core.exceptions.StorePassThroughException.handleRuntimeException(StorePassThroughException.java:72) ~[ehcache-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT dbaf9901e0e2786eef9e107b174c3f9a869650cf]
	at org.ehcache.clustered.client.internal.store.ClusteredStore.silentPut(ClusteredStore.java:189) ~[ehcache-clustered-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT dbaf9901e0e2786eef9e107b174c3f9a869650cf]
	at org.ehcache.clustered.client.internal.store.ClusteredStore.put(ClusteredStore.java:160) ~[ehcache-clustered-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT dbaf9901e0e2786eef9e107b174c3f9a869650cf]
	at org.ehcache.impl.internal.store.tiering.TieredStore.put(TieredStore.java:149) ~[ehcache-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT dbaf9901e0e2786eef9e107b174c3f9a869650cf]
	at org.ehcache.core.Ehcache.put(Ehcache.java:196) ~[ehcache-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT dbaf9901e0e2786eef9e107b174c3f9a869650cf]
	at com.terracotta.qa.ehcache.validation.ValidatingPutOperation.exec(ValidatingPutOperation.java:56) [classes/:na]
	at io.rainfall.execution.RunsDuring$1.call(RunsDuring.java:79) [rainfall-core-1.1.3-20160624.105609-14.jar:na]
	at io.rainfall.execution.RunsDuring$1.call(RunsDuring.java:71) [rainfall-core-1.1.3-20160624.105609-14.jar:na]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [na:1.8.0_66]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_66]
[...]

PR #1294, opened 2016-07-06
 #1277 ensure the limit and position of the byte buffers passed as arguments aren't modified

lorban:
This is now also specified by the `ServerStore`'s javadoc.


cschanck:
Looks good to me.

PR #1300, opened 2016-07-07
Some management module cleanups

mathieucarbou:


management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatistics.java
```diff
@@ -221,17 +220,21 @@ public void dispose() {
 
 
   @SuppressWarnings({"unchecked", "rawtypes"})
-  private ConcurrentMap<String, OperationStatistic<?>> discoverCountStatistics() {
-    ConcurrentHashMap<String, OperationStatistic<?>> result = new ConcurrentHashMap<String, OperationStatistic<?>>();
```
anthonydahanne: no more putIfAbsent ?

mathieucarbou: not in a normal map in jdk6

anthonydahanne: ok, what I meant is : why the switch from Concurrent to normal ?

mathieucarbou: Why would you use a ConcurrentMap here, that, besides, depend on the internal impl of Ehcache ? It's just a local map which is created sequentially, then read-only.

anthonydahanne: haha, true, put I liked the conciseness of putIfAbsent :-)

PR #1305, opened 2016-07-11
Bad message reported in status transitioner (branch 3.0)

mathieucarbou:


core/src/main/java/org/ehcache/core/StatusTransitioner.java
```diff
@@ -220,7 +220,7 @@ public void succeeded() {
     public StateTransitionException failed(Throwable t) {
       if (st.done()) {
         if (t != null) {
-          throw new AssertionError("Throwable cannot be null if Transition is done.");
+          throw new AssertionError("Throwable cannot be thrown if Transition is done: " + t.getMessage());
```
chrisdennis: We should probably init the cause here too.

mathieucarbou: I tried, but we cannot in jdk6 :-(
But we could call AssertionError.setStatckTrace(), but I am not sure it's right ?

chrisdennis:
```java
throw (AssertionError) new AssertionError("Throwable cannot be thrown if Transition is done.").initCause(t)
```

Issue #1004, opened 2016-04-27
Being able to expose Ehcache client management metadata (clustered-management module)


mathieucarbou:
The goal of this new module would be to expose Ehcache management metadata from an Ehcache client (management module) to the server for M&M, through the management-entity.

This module would depend on already existing ehcache management module, already existing management-entity in tc-management, and on a clustering service providing a voltron connection, so that this module can fetch the management-entity and expose the management metadata as needed.

- [x] Requires Terracotta-OSS/terracotta-management#25
- [x] Requires Terracotta-OSS/terracotta-platform#106

PR #1306, opened 2016-07-11
Improving collector speed and auto-start at CM init (#1004)

mathieucarbou:
Prepares #1004, do not include any clustering work.

Collector service, if setup, is automatically started once CM is initialized.


anthonydahanne:
LGTM and in sync with master

PR #1307, opened 2016-07-11
[DONT MERGE] Issue #1004 - Clustered management support

### LINKED ISSUE #1004: Being able to expose Ehcache client management metadata (clustered-management module)

mathieucarbou:
# DO NOT MERGE

Closes #1004. 

Adds support for clustering in management module.

Several notes:

 __1.__ I moved several versions from `clustered` to the top build file because some versions are common from management and clustered and management depends on clustered (provided)

=> Confirmed with @chrisdennis and @cljohnso : OK (only impact on clustered is the dependency on tc-platform, which still compile fine and pass the IT tests)

 __2.__ To avoid creating another module depending on both management AND clustered, I put the clustering work in an isolated package which won't impact the standalone usage. Thus, the clustering management service is started by the Management registry service ONLY if the `EntityService` is found in the classpath AND in the service locator.

=> @ljacomet : the little trick is 3 lines added to the `DefaultManagementRegistryService` to load the clustered management service if needed. This avoids creating the new module.

 __3.__ The `passthrough` scheme needs to be supported in Ehcache so that we can Ehcache test with the OSS passthrough system from another module.

=> @ljacomet : this is something I discussed with @chrisdennis on Friday. I think it's a valid use case to be able to test Ehcache. And I cannot access the `UnitTestConnectionService` from management tests.

__DO NOT MERGE BEFORE @ljacomet's APPROVAL__

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,11 +18,14 @@ import scripts.*
 ext {
   baseVersion = '3.1.1-SNAPSHOT'
   offheapVersion = '2.2.2'
-  managementVersion = '5.0.3.beta'
   statisticVersion = '1.1.0'
   jcacheVersion = '1.0.0'
   slf4jVersion = '1.7.7'
   sizeofVersion = '0.3.0'
+  terracottaPlatformVersion = '5.0.3.beta'
+  terracottaApisVersion = '1.0.3.beta'
+  entityTestLibVersion = '1.0.3.beta'
+  terracottaCoreVersion = '5.0.0-beta2'
 
   utils = new Utils(baseVersion, logger)
   isReleaseVersion = !baseVersion.endsWith('SNAPSHOT')
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ import scripts.*
 ext {
   baseVersion = '3.1.1-SNAPSHOT'
   offheapVersion = '2.2.2'
-  managementVersion = '5.0.0.beta'
+  managementVersion = '5.0.3.beta'
   statisticVersion = '1.1.0'
   jcacheVersion = '1.0.0'
   slf4jVersion = '1.7.7'
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
@@ -24,6 +24,9 @@ import org.ehcache.config.builders.CacheManagerConfiguration;
 import org.ehcache.spi.service.ServiceCreationConfiguration;
 
 import java.net.URI;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
 
 import org.ehcache.clustered.common.ServerSideConfiguration;
 
@@ -36,7 +39,7 @@ public class ClusteringServiceConfiguration
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<PersistentCacheManager> {
 
-  private static final String CLUSTER_SCHEME = "terracotta";
+  private static final Collection<String> CLUSTER_SCHEMES = new HashSet<String>(Arrays.asList("terracotta", "passthrough"));
 
   private final URI clusterUri;
   private final boolean autoCreate;
@@ -159,7 +162,7 @@ public class ClusteringServiceConfiguration
       throw new NullPointerException("Cluster URI cannot be null.");
     }
 
-    if (!CLUSTER_SCHEME.equals(clusterUri.getScheme())) {
+    if (!CLUSTER_SCHEMES.contains(clusterUri.getScheme())) {
       throw new IllegalArgumentException("Cluster Uri is not valid, clusterUri : " + clusterUri.toString());
     }
   }
--- a/clustered/client/src/main/resources/ehcache-clustered-ext.xsd
+++ b/clustered/client/src/main/resources/ehcache-clustered-ext.xsd
@@ -77,7 +77,7 @@
 
   <xs:simpleType name="connectionUrl">
     <xs:restriction base="xs:anyURI">
-      <xs:pattern value="terracotta://([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(/[^\?#]*)?(\?[^#]*)?(#.*)?"/>
+      <xs:pattern value="(terracotta|passthrough)://([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(/[^\?#]*)?(\?[^#]*)?(#.*)?"/>
     </xs:restriction>
   </xs:simpleType>
 
--- a/management/src/main/java/org/ehcache/management/cluster/Clustering.java
+++ b/management/src/main/java/org/ehcache/management/cluster/Clustering.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.management.cluster;
+
+import org.ehcache.spi.service.Service;
+import org.ehcache.spi.service.ServiceProvider;
+
+/**
+ * Feature-check class that is there only so that the management module can contain some specific classes
+ * depending on clustering, and activated only when clustering is available
+ */
+public class Clustering {
+
+  private static final String ENTITY_SERVICE_FQCN = "org.ehcache.clustered.client.service.EntityService";
+  private static final Class<? extends Service> ENTITY_SERVICE_CLASS;
+
+  static {
+    Class<? extends Service> serviceClass = null;
+    try {
+      serviceClass = Clustering.class.getClassLoader().loadClass(ENTITY_SERVICE_FQCN).asSubclass(Service.class);
+    } catch (ClassNotFoundException ignored) {
+    }
+    ENTITY_SERVICE_CLASS = serviceClass;
+  }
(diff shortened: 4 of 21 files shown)


management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -0,0 +1,276 @@
...
+
+    runTest();
+  }
+
+  private void runTest() throws Exception {
+    // get the queues where notifs and stats are put
+    BlockingQueue<Object[]> statistics = consumer.getValueForNode(new String[]{"management", "statistics"}, BlockingQueue.class).get();
+    BlockingQueue<Object[]> notifications = consumer.getValueForNode(new String[]{"management", "notifications"}, BlockingQueue.class).get();
+
+    // Issue a management call to configure the stats to collect
```
anthonydahanne: it's not a "management call", but rather a local (ie on the client) management registry configuration change (or update)

management/src/main/java/org/ehcache/management/cluster/Clustering.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Check if clustering is active for this cache manager
+   */
+  public static boolean isAvailable(ServiceProvider<Service> serviceProvider) {
+    return ENTITY_SERVICE_CLASS != null && serviceProvider.getService(ENTITY_SERVICE_CLASS) != null;
+  }
+
+  /**
+   * Creates a new ${@link ClusteringManagementService} to handle the management integration with teh cluster
```
anthonydahanne: typo : the cluster

build.gradle
```diff
@@ -18,11 +18,14 @@ import scripts.*
 ext {
   baseVersion = '3.1.1-SNAPSHOT'
   offheapVersion = '2.2.2'
-  managementVersion = '5.0.0.beta'
   statisticVersion = '1.1.0'
   jcacheVersion = '1.0.0'
   slf4jVersion = '1.7.7'
   sizeofVersion = '0.3.0'
+  terracottaPlatformVersion = '5.0-SNAPSHOT'
```
ljacomet: yuck - can't we stick with released versions?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -77,7 +77,7 @@
 
   <xs:simpleType name="connectionUrl">
     <xs:restriction base="xs:anyURI">
-      <xs:pattern value="terracotta://([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(/[^\?#]*)?(\?[^#]*)?(#.*)?"/>
+      <xs:pattern value="(terracotta|passthrough)://([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(/[^\?#]*)?(\?[^#]*)?(#.*)?"/>
```
ljacomet: ouch ... that I really do not like

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -36,7 +39,7 @@
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<PersistentCacheManager> {
 
-  private static final String CLUSTER_SCHEME = "terracotta";
+  private static final Collection<String> CLUSTER_SCHEMES = new HashSet<String>(Arrays.asList("terracotta", "passthrough"));
```
ljacomet: This is bad, but I find XSD even worse. Can't we handle that elsewhere?

management/src/main/java/org/ehcache/management/cluster/Clustering.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class Clustering {
+
+  private static final String ENTITY_SERVICE_FQCN = "org.ehcache.clustered.client.service.EntityService";
+  private static final Class<? extends Service> ENTITY_SERVICE_CLASS;
+
+  static {
+    Class<? extends Service> serviceClass = null;
+    try {
+      serviceClass = Clustering.class.getClassLoader().loadClass(ENTITY_SERVICE_FQCN).asSubclass(Service.class);
```
ljacomet: That's going to be an OSGi nightmare ... but I have no ready answer

lorban: I agree with that, but the only way to avoid this problem would be to avoid having to load the service interface itself.

The only fully safe way I can think of would be to add a method like the following to the `ServiceProvider`: 
`ServiceProvider.containsImplFor(String intfName)`

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
lorban: I agree with you, @ljacomet but on the other hand I find it a bit extreme to have the URI scheme hardcoded in the XSD validation rule. IMHO the validation done in `ClusteringServiceConfiguration` is enough.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
lorban: Wouldn't a test-module subclass of `ClusteringServiceConfiguration` that allows the `passthrough` scheme would do the job?

management/src/main/java/org/ehcache/management/cluster/Clustering.java
lorban: Another thing you could do since you have declared `:clustered:client` as `compileOnly` is since you do have access to the `EntityService` class at compile time, you may want to directly use the class in your `Clustering` class without reflection and load the `Clustering` class via reflection. That way you'd get a CNFE while trying to load `Clustering` if the `EntityService` isn't available.

management/src/main/java/org/ehcache/management/cluster/DefaultClusteringManagementService.java
```diff
@@ -0,0 +1,178 @@
...
+          managementAgentEntity = managementAgentEntityFactory.retrieve();
+        } catch (EntityNotFoundException e) {
+          try {
+            managementAgentEntityFactory.create();
+          } catch (EntityAlreadyExistsException ignored) {
+          }
+          try {
+            managementAgentEntity = managementAgentEntityFactory.retrieve();
+          } catch (EntityNotFoundException bigFailure) {
+            throw new AssertionError(bigFailure);
```
lorban: `AssertionError` means bug, hence I'd include a human-readable error message so that one would have a rough idea of why the exception was thrown in case we ever run into that bug.

management/src/main/java/org/ehcache/management/cluster/DefaultClusteringManagementService.java
```diff
@@ -0,0 +1,178 @@
...
+        this.cacheManager.deregisterListener(this);
+        break;
+      }
+
+      case MAINTENANCE:
+        // in case we need management capabilities in maintenance mode
+        break;
+
+      default:
+        throw new AssertionError(to);
```
lorban: same as above: `AssertionError` should always have a human-readable message.

build.gradle
mathieucarbou: Yes now that it's tested, Anthony could release and we can change that.
@anthonydahanne : can you release please ?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
ljacomet: Having it in the XSD provides contextual error for users who do you either an IDE or an XML editor.

mathieucarbou: As long as we have a way to override the scheme, both in XML and programmatic API, OR as long as Ehcache provides a "testing feature" (like a module we could use to test Ehcache, that will contain basically the specific connection service implementation), so that eternal service implementors can use it, then it's fine.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
mathieucarbou: See my other comment. I think either Ehcache miss a more loose validation, either it misses a test module so that other module, like clustered and management, could depend on it...

management/src/main/java/org/ehcache/management/cluster/Clustering.java
mathieucarbou: @ljacomet @lorban : I do not understand the OSGI issue ? I mean... I just need to check whether "org.ehcache.clustered.client.service.EntityService" is on the classpath. Even in OSGI, since it's a service management would optionally need, it should be in the classpath right ? And OSGI supports optional dependencies ? Do you mean I should use another classloader for the check ?

@lorban : if the `Clustering` class has some imports onto `EntityService`, and Clustering class is used in management registry standalone, and is using the `EntityService` directly, where I should catch the CNFE ? In management registry, when using `Clustering` class ?

management/src/main/java/org/ehcache/management/cluster/DefaultClusteringManagementService.java
mathieucarbou: Fixed at 4 places

mathieucarbou:
UPDATED PR to simplify and prepare for management calls.

DEPENDS ON Terracotta-OSS/terracotta-platform#113

mathieucarbou:
test this please

mathieucarbou:
UPDATED PR to support management calls!

DEPENDS ON Terracotta-OSS/terracotta-platform#116

mathieucarbou:
test this please

mathieucarbou:
@anthonydahanne : I'll let you trigger again the build when the snapshot of tc-platform will be deployed.
Then you'll be able to release tc-platform :-)

mathieucarbou:
test this please

mathieucarbou:
Note: I rebased this PR on PR #1316, because it is a dependency.

anthonydahanne:
replaced by PR #1329

Issue #1207, opened 2016-06-23
Move entity state storage in to a specialized Voltron service


chrisdennis:
In order to support active passive failover we will have to move all state that must survive a promotion event in to a Voltron service that is external to the Ehcache entity itself.  The passive entity can then, on promotion, reconnect to this service to gain access to it's state.

chrisdennis:
Blocked by #1206

PR #1310, opened 2016-07-13
Issue 1206

AbfrmBlr:


AbfrmBlr:
test this please

AbfrmBlr:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -204,7 +168,7 @@
    */
   // This method is intended for unit test use; modifications are likely needed for other (monitoring) purposes
   Set<String> getStores() {
-    return Collections.unmodifiableSet(new HashSet<String>(stores.keySet()));
+    return Collections.unmodifiableSet(new HashSet<String>(serverStoreManager.getStores().keySet()));
```
ljacomet: With the changes, does it still make sense to keep these "unit test" methods here?

AbfrmBlr: You are right they don't belong here. Actually they will be moved to actual test cases of the service emerging out of #1207, and then these methods will be gone from entity.

PR #1313, opened 2016-07-13
[master] Fixes #1156 - JSR-107 integration support

### LINKED ISSUE #1156: JCache MutableConfiguration setWriteThrough(true) enables CacheLoader
Taking the code:

```
MutableConfiguration<String, String> configuration = new MutableConfiguration<>();
    configuration.setTypes(String.class, String.class);
    configuration.setCacheLoaderFactory(new FactoryBuilder.ClassFactory<>("org.jsoftbiz.service.SomeCacheLoader"));
    configuration.setReadThrough(false);
    configuration.setWriteThrough(true);
```

will enable the cacheloader because the setWriteThrough was set to true, eventhough the setReadThrough was set to false. (Since it is the cacheloader and not the cachewriter, it should not be enabled)

Additional note, in the spec, it is written:
```

  /**
   * Set if read-through caching should be used.
   * <p>
   * It is an invalid configuration to set this to true without specifying a
   * {@link CacheLoader} {@link Factory}.
   *
   * @param isReadThrough <code>true</code> if read-through is required
   * @return the {@link MutableConfiguration} to permit fluent-style method calls
   */
  public MutableConfiguration<K, V> setReadThrough(boolean isReadThrough) {
```

However, if calling  setReadThrough(true) but not specifying the cacheloader, we do not get an error, this is not taken as an invalid configuration in Ehcache.

ljacomet:


107/src/main/java/org/ehcache/jsr107/internal/JsrCacheLoaderWriter.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.ehcache.spi.loaderwriter.BulkCacheLoadingException;
+import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
+
+import java.util.Map;
+
+/**
+ * Extension of the CacheLoaderWriter interface for doing tricks in 107 world.
+ */
+public interface JsrCacheLoaderWriter<K, V> extends CacheLoaderWriter<K, V> {
```
chrisdennis: Typical prefixes here seem to be either `Jsr107` or `Eh107` - any reason for picking a new one?

107/src/main/java/org/ehcache/jsr107/MultiCacheException.java
```diff
@@ -74,7 +74,22 @@ public String getMessage() {
 
   @Override
   public Throwable initCause(Throwable cause) {
-    throw new UnsupportedOperationException();
```
chrisdennis: Isn't this the same as `addThrowable(Throwable t)`?

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -228,10 +237,17 @@ private static void addIdentityCopierIfNoneRegistered(Map<Class<?>, ClassInstanc
     return new ExpiryPolicyToEhcacheExpiry<K, V>(config.getExpiryPolicyFactory().create());
   }
 
-  private <K, V> CacheLoaderWriter<K, V> initCacheLoaderWriter(CompleteConfiguration<K, V> config, MultiCacheException mce) {
+  private <K, V> JsrCacheLoaderWriter<K, V> initCacheLoaderWriter(CompleteConfiguration<K, V> config, MultiCacheException mce) {
     Factory<CacheLoader<K, V>> cacheLoaderFactory = config.getCacheLoaderFactory();
     Factory<CacheWriter<K, V>> cacheWriterFactory = getCacheWriterFactory(config);
```
chrisdennis: You strayed too close to this line... could we inline this method and annotate the declaration - rather than having a static method just for this?

107/src/main/java/org/ehcache/jsr107/Eh107CacheLoaderWriter.java
```diff
@@ -28,26 +28,31 @@
...
 /**
  * @author teck
  */
-class Eh107CacheLoaderWriter<K, V> implements CacheLoaderWriter<K, V>, Closeable {
+class Eh107CacheLoaderWriter<K, V> implements JsrCacheLoaderWriter<K, V>, Closeable {
 
   private final CacheLoader<K, V> cacheLoader;
+  private final boolean skipLoader;
   private final CacheWriter<K, V> cacheWriter;
+  private final boolean writeThrough;
```
chrisdennis: This is still redundant with `cacheWriter != null` right?

107/src/main/java/org/ehcache/jsr107/Eh107CacheLoaderWriter.java
```diff
@@ -28,26 +28,31 @@
...
+import org.ehcache.spi.loaderwriter.BulkCacheWritingException;
 
 /**
  * @author teck
  */
-class Eh107CacheLoaderWriter<K, V> implements CacheLoaderWriter<K, V>, Closeable {
+class Eh107CacheLoaderWriter<K, V> implements JsrCacheLoaderWriter<K, V>, Closeable {
 
   private final CacheLoader<K, V> cacheLoader;
+  private final boolean skipLoader;
```
chrisdennis: I think the logic in this class would read clearer is we kept this as `readThrough`

107/src/main/java/org/ehcache/jsr107/internal/JsrCacheLoaderWriter.java
ljacomet: `Eh107CacheLoaderWriter` already exists (and is probably badly named ...), but you are right, should go with `Jsr107CacheLoaderWriter`.

107/src/main/java/org/ehcache/jsr107/MultiCacheException.java
ljacomet: Adds the throwable as the _first_ one

ljacomet:
Still need to address the last comment

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
```diff
@@ -247,18 +264,10 @@ private static void addIdentityCopierIfNoneRegistered(Map<Class<?>, ClassInstanc
...
     } else {
-      return new Eh107CacheLoaderWriter<K, V>(cacheLoader, cacheWriter);
+      return new Eh107CacheLoaderWriter<K, V>(cacheLoader, config.isReadThrough(), cacheWriter, config.isWriteThrough());
     }
   }
 
-  @SuppressWarnings("unchecked")
-  private static <K, V> Factory<CacheWriter<K, V>> getCacheWriterFactory(CompleteConfiguration<K, V> config) {
-    // I could be wrong, but I don't think this factory should be typed the way it is. The factory
-    // should be parameterized with (K, V) and it's methods take <? extend K>, etc
```
chrisdennis: Should we preserve this comment on the relocated cast?

107/src/main/java/org/ehcache/jsr107/MultiCacheException.java
```diff
@@ -72,6 +73,27 @@ public String getMessage() {
...
+      throw new NullPointerException();
+    }
+
+    if (t == this) {
+      throw new IllegalArgumentException("cannot add to self");
+    }
+
+    if (t instanceof MultiCacheException) {
+      MultiCacheException multiCacheException = (MultiCacheException) t;
+      List<Throwable> throwables = multiCacheException.getThrowables();
```
chrisdennis: You're hiding the instance variable here... which means none of the code below does the right thing... right?

107/src/main/java/org/ehcache/jsr107/MultiCacheException.java
```diff
@@ -72,6 +73,27 @@ public String getMessage() {
...
+      ListIterator<Throwable> listIterator = throwables.listIterator(throwables.size());
+      while (listIterator.hasPrevious()) {
+        throwables.add(0, listIterator.previous());
```
chrisdennis: You could do an `addAll(...)` for this.

107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
ljacomet: Not much we can do about it as it talks about a JSR-107 type thus external to us.

107/src/main/java/org/ehcache/jsr107/MultiCacheException.java
ljacomet: Duh - fixing

107/src/main/java/org/ehcache/jsr107/MultiCacheException.java
ljacomet: Why do in one line what can be done in five? 😞

Issue #951, opened 2016-04-07
API to get complete Ehcache configuration in text format


skbansal:
It will be helpful if we can get an API which will give us the complete Ehcache Configuration in text format (Probably xml).

This will be helpful for debugging purpose and sharing the configuration among Ehcache users.

chrisdennis:
I think this is a good idea... but I would vote for not XML.

skbansal:
XMLs are easy to parse (using some standard parsers) and help in understanding the tree structure.

Moreover, for debugging, I could just use the same XML and can create original like configuration to debug/reproduce the issue with 0% error rate.

chrisdennis:
The set of all programatic configurations is larger than the universe of all XML configurations... not every one is mappable.

> Moreover, for debugging, I could just use the same XML and can create original like configuration to debug/reproduce the issue with 0% error rate.

Do we really want to make that a goal/feature?  It's a pretty lofty one, and one that caused us pain in the past.  I'd want to see a strong use case for this.

ljacomet:
After discussion with @mathieucarbou, there is some need to be able to visualise configuration, without ever thinking about XML format.
In this context, the work for this enhancement is to created a _formatted_ `String` out of the `Configuration` object and its graph.

This should give us a valid `String` representation from the `Cache` level as well.

PR #1314, opened 2016-07-13
Provide a humanReadable representation of a Cache Manager configuration

anthonydahanne:
To provide a human readable representation of a CacheManager

This PR is here to try and solve #951 

Thank you for your initial feedback, just checking if I'm down the right path before going through all the classes


GaryWKeim:
No comment.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1079,4 +1079,25 @@ public AtomicInteger getActualRemoveCount() {
     }
   }
 
+  @Override
+  public String toString() {
+    return "statusTransitioner:\n    "
```
ljacomet: This (and the CM level) is way more than configuration - it even gives information on the current state. Is that really necessary?

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1079,4 +1079,25 @@ public AtomicInteger getActualRemoveCount() {
...
+        + statusTransitioner.toString().replace("\n","\n    ") + "\n"
```
ljacomet: What is the expected use of that information?

ljacomet:
My general fear with that global approach instead of limiting it to configuration is the amount of internal state that gets exposed. How much of this will then be considered _contractual_ or _api_ by users of said information?

core/src/main/java/org/ehcache/core/Ehcache.java
mathieucarbou: I think `statusTransitioner.toString()` is wrong: what we need i something like:

`"status: " + getStatus()`

core/src/main/java/org/ehcache/core/Ehcache.java
mathieucarbou: More generally, it is for #951, and to have a "dump" state in M&M.
I agree we should not expose internal stuff but only settings configured and states.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -1079,4 +1079,25 @@ public AtomicInteger getActualRemoveCount() {
...
+        + "store:\n    "
+        + store.toString().replace("\n","\n    ") ;
+//        ", resilienceStrategy=" + resilienceStrategy +
+//        ", runtimeConfiguration=" + runtimeConfiguration +
+//        ", jsr107Cache=" + jsr107Cache +
+//        ", getObserver=" + getObserver +
+//        ", getAllObserver=" + getAllObserver +
+//        ", putObserver=" + putObserver +
+//        ", putAllObserver=" + putAllObserver +
+//        ", removeObserver=" + removeObserver +
```
mathieucarbou: For observers and listeners, I wouldn't `toString()` them. But just knowing some have been setup is helpful for troubleshooting.  So perhaps something like `", getAllObserver=" + getAllObserver.size() +` and `cacheManagerListeners=` + cacheManagerListeners.size()` would be enough ?

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -728,6 +728,31 @@ void destroyInternal() throws CachePersistenceException {
...
+
+    if(!caches.isEmpty()) {
+      cachesToStringBuilder.deleteCharAt(cachesToStringBuilder.length() -1);
+    }
+
+    return "configuration:\n    "
+        + configuration.toString().replace("\n","\n    ") + "\n"
+        + "caches:\n    "
+        + cachesToStringBuilder.toString().replace("\n","\n    ");
+//        ", listeners=" + listeners +
```
mathieucarbou: Just `", listeners=" + listeners.size()` I think

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -728,6 +728,31 @@ void destroyInternal() throws CachePersistenceException {
...
+  @Override
+  public String toString() {
+    StringBuilder cachesToStringBuilder = new StringBuilder();
+    for (Map.Entry<String, CacheHolder> cacheHolderEntry : caches.entrySet()) {
+      cachesToStringBuilder
+          .append(cacheHolderEntry.getKey())
+          .append(":\n    ")
+          .append(cacheHolderEntry.getValue().toString().replace("\n","\n    "))
+          .append("\n");
+    }
```
mathieucarbou: I wouldn't dump the whole content of a cache: what if a cache has 1GB of data ?

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -777,5 +802,14 @@ public synchronized void setCache(final InternalCache<?, ?> cache) {
       this.isValueSet = true;
       notifyAll();
     }
+
+    @Override
+    public String toString() {
+      return "keyType: " + keyType + "\n"
```
mathieucarbou: `keyType.getName()` instead. Otherwize you'll print the `@hash` at the end.

core/src/main/java/org/ehcache/core/EhcacheRuntimeConfiguration.java
```diff
@@ -152,4 +152,18 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
     }
   }
+
+  @Override
+  public String toString() {
+    return
+//        "serviceConfigurations=" + serviceConfigurations +
+//        ", config=" + config +
+        "keyType: " + keyType + "\n    " +
+        "valueType: " + valueType ;
```
mathieucarbou: same here

core/src/main/java/org/ehcache/core/InternalStatus.java
```diff
@@ -140,6 +140,13 @@ public InternalStatus to() {
     public boolean done() {
       return done != null;
     }
+
+    @Override
+    public String toString() {
+      return "to: " + to + "\n" +
```
mathieucarbou: We shouldn't add `toString()` in such internal classes. Instead, in the CM, just output `status: getStatus()`

core/src/main/java/org/ehcache/core/StatusTransitioner.java
```diff
@@ -235,4 +235,13 @@ public StateTransitionException failed(Throwable t) {
...
     }
   }
+
+  @Override
+  public String toString() {
+    return "currentState: \n    "
+        + currentState.get().toString().replace("\n","\n    ") + "\n"
+        + "maintenanceLease: " + maintenanceLease + "\n"
+        + "hooks: " + hooks + "\n"
+        + "listeners: " + listeners ;
```
mathieucarbou: same here: `StatusTransitioner ` is really internal stuff

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1747,4 +1747,38 @@ public void initHigherCachingTier(HigherCachingTier<?, ?> resource) {
     }
   }
 
+  @Override
+  public String toString() {
+    return "map: \n    " + map.toString().replace("\n","\n    ") + "\n" +
+        "keyType: " + keyType + "\n" +
+        "valueType: " + valueType;
+//        "valueCopier: " + valueCopier + "\n"+
```
mathieucarbou: same comment for keyType and valueType

management/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
 
   testCompile project(':xml')
   testCompile "com.fasterxml.jackson.core:jackson-databind:2.7.5"
+  testCompile 'org.yaml:snakeyaml:1.17'
```
mathieucarbou: We shouldn't test it: it's not official ;-)

management/src/test/java/org/ehcache/management/providers/settings/EhcacheManagerToStringTest.java
```diff
@@ -0,0 +1,42 @@
...
+            .build())
+        .build(true);
+
+    String toString = cacheManager.toString();
+
+    Yaml yaml = new Yaml();
+    Object actual = yaml.load(toString);
+    Object expected = yaml.load(this.getClass().getResourceAsStream("/onHeap.yml"));
+
+    Assert.assertThat(actual, equalTo(expected));
```
mathieucarbou: same comment - not official support of yaml

chrisdennis:
Now that I think about it I'm not a huge fan of using `toString()` to do this.  I'm thinking we should really constrain it to be a 'concise but informative representation' - i.e. we should try to keep it to a single line.  This also means we can keep the more information rich method hidden from the API by keeping it out of the interfaces entirely.

anthonydahanne:
@ljacomet @GaryWKeim @lorban @chrisdennis hello I've updated this PR with some of your feedback; in a nutshell, only the configuration is concerned AND a new interface is being implemented by configuration implementors
Let me know what you think !

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -80,7 +80,7 @@
  *   {@code org.ehcache}.
  * </P>
  */
-public class EhcacheManager implements PersistentCacheManager, InternalCacheManager {
+public class EhcacheManager implements PersistentCacheManager, InternalCacheManager, HumanReadable {
```
ljacomet: Is it desired at this level? I feel the gain from implementing the interface here is really minimal.

core/src/main/java/org/ehcache/core/EhcacheRuntimeConfiguration.java
```diff
@@ -152,4 +152,48 @@ public void registerCacheEventListener(CacheEventListener<? super K, ? super V>
...
+      }
+    }
+
+    if(serviceConfigurationsToStringBuilder.length() > 0) {
+      serviceConfigurationsToStringBuilder.deleteCharAt(serviceConfigurationsToStringBuilder.length() -1);
+    }
+
+
+    StringBuilder cacheConfigurationListenerToStringBuilder = new StringBuilder();
+    for (CacheConfigurationChangeListener cacheConfigurationChangeListener : cacheConfigurationListenerList) {
```
ljacomet: These are internal things that get notified when configuration gets updated - pretty sure it makes no sense to have them in the human readable format.

ljacomet: Actually, reading the code below, looks like you are not even using the result of this.

core/src/main/java/org/ehcache/core/HumanReadable.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Classes implementing this interface provide a human readable
+ * string to describe themselves.
+ *
+ * Should only be used in configuration classes
+ */
+public interface HumanReadable {
+
+  /**
+   * @return the human readable representation of the object
```
ljacomet: _A_ instead of _the_ ?

core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
```diff
@@ -123,4 +124,24 @@ public static void validateResourcePools(Collection<? extends ResourcePool> pool
       }
     }
   }
+
+  @Override
+  public String readableString() {
+
+    StringBuilder poolsToStringBuilder = new StringBuilder();
+    for (Map.Entry<ResourceType<?>, ResourcePool> poolEntry : pools.entrySet()) {
```
ljacomet: I would sort the pools

core/src/main/java/org/ehcache/core/EhcacheManager.java
anthonydahanne: ok, I could call it directly from the DefaultConfiguration

core/src/main/java/org/ehcache/core/EhcacheRuntimeConfiguration.java
anthonydahanne: I removed it, true
not entirely though, fixing that

core/src/main/java/org/ehcache/core/HumanReadable.java
anthonydahanne: ok

core/src/main/java/org/ehcache/core/config/ResourcePoolsImpl.java
anthonydahanne: yeah, that could be better, looking into it

management/src/test/java/org/ehcache/core/EhcacheManagerToStringTest.java
```diff
@@ -0,0 +1,152 @@
...
+              }
+            })
+            .build())
+        .build(true);
+
+    String readableString = ((HumanReadable) cacheManager.getRuntimeConfiguration()).readableString();
+
+    System.out.println(readableString);
+
+    Yaml yaml = new Yaml();
```
lorban: Nothing in the `HumanReadable.readableString()` says that the string has to be yaml-parseable.

PR #1316, opened 2016-07-14
Upgraded dependencies

mathieucarbou:
Upgrade to:

- tc platform 5.0.3.beta
- tc apis 1.0.3.beta
- testing libs 1.0.3.beta

Required for Issue #1004, PR #1307

### DIFF
--- a/clustered/build.gradle
+++ b/clustered/build.gradle
@@ -15,14 +15,13 @@
  */
 
 ext {
-    terracottaPlatformVersion = '5.0.0.beta'
-    terracottaApisVersion = '1.0.0.beta'
+    terracottaPlatformVersion = '5.0.3.beta'
+    terracottaApisVersion = '1.0.3.beta'
     terracottaCoreVersion = '5.0.0-beta2'
-    coordinatorVersion = terracottaPlatformVersion
     clusteredMapVersion = terracottaPlatformVersion
     offheapResourceVersion = terracottaPlatformVersion
     entityApiVersion = terracottaApisVersion
-    entityTestLibVersion = terracottaApisVersion
+    entityTestLibVersion = '1.0.3.beta'
 }
 
 subprojects {
--- a/clustered/client/build.gradle
+++ b/clustered/client/build.gradle
@@ -31,4 +31,5 @@ dependencies {
   }
   testCompile "org.terracotta.entities:clustered-map-server:$parent.clusteredMapVersion"
   testCompile "org.terracotta:entity-test-lib:$parent.entityTestLibVersion"
+  testCompile "org.terracotta:passthrough-server:$parent.entityTestLibVersion"
 }
--- a/clustered/client/src/test/java/org/ehcache/clustered/server/ObservableEhcacheServerEntityService.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/server/ObservableEhcacheServerEntityService.java
@@ -21,9 +21,9 @@ import org.ehcache.clustered.common.internal.messages.EhcacheEntityResponse;
 import org.terracotta.entity.ActiveServerEntity;
 import org.terracotta.entity.ClientDescriptor;
 import org.terracotta.entity.ConcurrencyStrategy;
+import org.terracotta.entity.EntityServerService;
 import org.terracotta.entity.MessageCodec;
 import org.terracotta.entity.PassiveServerEntity;
-import org.terracotta.entity.ServerEntityService;
 import org.terracotta.entity.ServiceRegistry;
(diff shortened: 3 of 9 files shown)


mathieucarbou:
tc-platform is currently being deployed, so keep retrying ;-)

clustered/client/build.gradle
```diff
@@ -31,4 +31,5 @@ dependencies {
   }
   testCompile "org.terracotta.entities:clustered-map-server:$parent.clusteredMapVersion"
   testCompile "org.terracotta:entity-test-lib:$parent.entityTestLibVersion"
+  testCompile "org.terracotta:passthrough-server:$parent.entityTestLibVersion"
```
chrisdennis: Why does the passthrough-server dependency use the entity-test-lib version property?

clustered/client/src/test/java/org/ehcache/clustered/server/ObservableEhcacheServerEntityService.java
```diff
@@ -35,17 +35,20 @@
 
 /**
  * Provides an alternative to {@link EhcacheServerEntityService} for unit tests to enable observing
+ * Provides an alternative to {@link EhcacheServerEntityService} for unit tests to enable observing
+ * Provides an alternative to {@link EhcacheServerEntityService} for unit tests to enable observing
+ * Provides an alternative to {@link EhcacheServerEntityService} for unit tests to enable observing
```
chrisdennis: IDE screwed up here?

clustered/client/src/test/java/org/ehcache/clustered/server/ObservableEhcacheServerEntityService.java
```diff
@@ -35,17 +35,20 @@
...
-    implements ServerEntityService<EhcacheEntityMessage, EhcacheEntityResponse> {
+    implements EntityServerService<EhcacheEntityMessage, EhcacheEntityResponse> {
   private final EhcacheServerEntityService delegate = new EhcacheServerEntityService();
 
   private final List<EhcacheActiveEntity> servedActiveEntities = new ArrayList<EhcacheActiveEntity>();
 
   /**
    * Gets a list of {@link ObservableEhcacheActiveEntity} instances wrapping the
-   * {@link EhcacheActiveEntity} instances served by this {@link ServerEntityService}.
+   * {@link EhcacheActiveEntity} instances served by this {@link org.terracotta.entity.EntityServerService}.
```
chrisdennis: To FQCN or not to FQCN, that is the question.

clustered/build.gradle
```diff
@@ -15,14 +15,13 @@
...
-    terracottaApisVersion = '1.0.0.beta'
+    terracottaPlatformVersion = '5.0.3.beta'
+    terracottaApisVersion = '1.0.3.beta'
     terracottaCoreVersion = '5.0.0-beta2'
-    coordinatorVersion = terracottaPlatformVersion
     clusteredMapVersion = terracottaPlatformVersion
     offheapResourceVersion = terracottaPlatformVersion
     entityApiVersion = terracottaApisVersion
-    entityTestLibVersion = terracottaApisVersion
+    entityTestLibVersion = '1.0.3.beta'
```
chrisdennis: Are all these changes intentional... and can anyone explain to me what the system is here?

clustered/client/build.gradle
mathieucarbou: they are in the same repo (https://github.com/Terracotta-OSS/tc-passthrough-testing). I re-used the same property name

clustered/build.gradle
anthonydahanne: @chrisdennis are you talking about the versioning ? 
.z means api changes, while .beta, .beta2, etc. means fixes releases
all tc components are .0.3.beta currently

mathieucarbou: And entityTestLibVersion != terracottaApisVersion now.

chrisdennis: I meant the rerouting of how versions are defined.  I think I understand now that this is make it so that each 'repository' in which the source for these projects is defined has an equivalent property in the build file here.  I think that makes sense but the naming is a bit off (`entityTestLibVersion` instead of `passthroughTestingVersion`), and there is no convention on whether each module should have it's own version variable that is defined to be equal to the "repo version variable" (the things in platform for example) or if there should only be variables for the 'repo', and the module should use that one directly (entityTestLibVersion).  Would be nice to make all this consistent if we're going to change stuff.

ljacomet:
Working on doing some cleanup in the properties in a separate PR, so closing this one.

PR #1317, opened 2016-07-15
Close #778 release 3.0 Improve Ehcache JSR-107 documentation

### LINKED ISSUE #778: Improve Ehcache JSR-107 documentation
The documentation here: http://www.ehcache.org/documentation/3.0/107.html probably needs some rework.

If the the user reading that doc is expected to have some prior knowledge about what a JSR-107 cache is, the advantages of using a JSR cache and how a JSR cache can be configured/created then this issue can be closed. If that is not the case, then we need to start with some brief material on what JSR-107 is and the advantages of using it. Then we probably need to start with a very basic sample of cache creation/configuration. Using the JSR MutableConfiguration or other configuation types etc. Currently from a user's perspective it looks like some advanced material, straight away talking about retrieving RuntimeConfigurations and other n number of configurations.

albinsuresh:


docs/src/docs/asciidoc/user/107.adoc
```diff
@@ -9,25 +9,62 @@ ifdef::notBuildingForSite[]
...
+== JCache overview
 
-There is nothing special for you to do to use Ehcache 3.0 as the caching provider for your application: add the
-`ehcache-3.0.0.jar` to your application's classpath (possibly removing the previous provider's jar) and you are ready to
-go:
+JCache (aka JSR-107) specification defines the standard caching API for Java.
+The specification was developed under the Java Community Process v2.9 by an expert group including members from the Ehcache developer community.
+Hence being one of the pioneers in the Java caching domain, Ehcache 3 also offers an implementation that is fully compliant with the JCache specification.
+JCache provides a very simple API set that is easy to use and vendor neutral.
+But it is not as rich and robust as the native Ehcache APIs.
```
ljacomet: I would scratch that sentence. What did you have in mind when writing it?

albinsuresh: Intent was to promote the usage of the native API set as it offers better configurability over the JCache ones. But yeah in the context of this document which is about JSR-107 I guess that's not really needed.

docs/src/docs/asciidoc/user/107.adoc
```diff
@@ -9,25 +9,61 @@ ifdef::notBuildingForSite[]
...
-`ehcache-3.0.0.jar` to your application's classpath (possibly removing the previous provider's jar) and you are ready to
-go:
+JCache (aka JSR-107) specification defines the standard caching API for Java.
+The specification was developed under the Java Community Process v2.9 by an expert group including members from the Ehcache developer community.
+Hence being one of the pioneers in the Java caching domain, Ehcache 3 also offers an implementation that is fully compliant with the JCache specification.
+JCache provides a very simple API set that is easy to use and vendor neutral.
+
+For years the biggest problems that application developers have faced while wanting to try cache implementations by different vendors is the stark contrast in the APIs offered by these vendors.
+Developers were forced to rewrite a whole lot of their application code just to try out out a new caching solution.
+This lead to a lot of developers remain stuck with what they had even when they knew that there is a better product out there.
```
ljacomet: This lead to developers sticking with what they had as the bar to investigating other products was too high.

docs/src/docs/asciidoc/user/107.adoc
```diff
@@ -9,25 +9,61 @@ ifdef::notBuildingForSite[]
...
-There is nothing special for you to do to use Ehcache 3.0 as the caching provider for your application: add the
-`ehcache-3.0.0.jar` to your application's classpath (possibly removing the previous provider's jar) and you are ready to
-go:
+JCache (aka JSR-107) specification defines the standard caching API for Java.
+The specification was developed under the Java Community Process v2.9 by an expert group including members from the Ehcache developer community.
+Hence being one of the pioneers in the Java caching domain, Ehcache 3 also offers an implementation that is fully compliant with the JCache specification.
+JCache provides a very simple API set that is easy to use and vendor neutral.
+
+For years the biggest problems that application developers have faced while wanting to try cache implementations by different vendors is the stark contrast in the APIs offered by these vendors.
+Developers were forced to rewrite a whole lot of their application code just to try out out a new caching solution.
```
ljacomet: Developers were forced to rewrite a whole lot of their caching related code in an application just to try out a new caching solution.

docs/src/docs/asciidoc/user/107.adoc
```diff
@@ -9,25 +9,61 @@ ifdef::notBuildingForSite[]
...
+This lead to a lot of developers remain stuck with what they had even when they knew that there is a better product out there.
+JCache specification was the biggest boon for those developers as now there is a Java standard caching API that they can use.
```
ljacomet: Would have a line in-between this and the above to start a new paragraph.
And rephrase to:
The availability of the JCache specification gives real added value for developers as there is now a standard caching API they can use.

docs/src/docs/asciidoc/user/107.adoc
```diff
@@ -9,25 +9,61 @@ ifdef::notBuildingForSite[]
...
+So it is fairly easier for an application developer to switch between the products by different vendors and choose the one that suits them best without changing a single line of their application code.
```
ljacomet: drop _fairly_, "between products by different vendors", "of their application code interacting with caches."

docs/src/docs/asciidoc/user/107.adoc
```diff
@@ -9,25 +9,61 @@ ifdef::notBuildingForSite[]
...
+All they have to do is swap the caching library from one vendor with another.
```
ljacomet: I would add a line stating that most likely there will be configuration impacts.

ljacomet:
This and #1318 do break what we want to achieve with daggy fixes - the fix commit is no longer the same.
I believe having a single fix commit and opening two PRs should be enough, the merge is not needed.

PR #1318, opened 2016-07-15
Closes #778 master Improve Ehcache JSR-107 documentation

### LINKED ISSUE #778: Improve Ehcache JSR-107 documentation

albinsuresh:


ljacomet:
This and #1317 do break what we want to achieve with daggy fixes - the fix commit is no longer the same.
I believe having a single fix commit and opening two PRs should be enough, the merge is not needed.

PR #1329, opened 2016-07-18
Issue #1004 - Clustered management support 

### LINKED ISSUE #1004: Being able to expose Ehcache client management metadata (clustered-management module)

anthonydahanne:
# rebases and replaces PR #1307 
In a nutshell, we "just" need to figure out what to do with the "passthrough" scheme

Closes #1004. 

Adds support for clustering in management module.

Several notes:

 __1.__ I moved several versions from `clustered` to the top build file because some versions are common from management and clustered and management depends on clustered (provided)

=> Confirmed with @chrisdennis and @cljohnso : OK (only impact on clustered is the dependency on tc-platform, which still compile fine and pass the IT tests)

 __2.__ To avoid creating another module depending on both management AND clustered, I put the clustering work in an isolated package which won't impact the standalone usage. Thus, the clustering management service is started by the Management registry service ONLY if the `EntityService` is found in the classpath AND in the service locator.

=> @ljacomet : the little trick is 3 lines added to the `DefaultManagementRegistryService` to load the clustered management service if needed. This avoids creating the new module.

 __3.__ The `passthrough` scheme needs to be supported in Ehcache so that we can Ehcache test with the OSS passthrough system from another module.

=> @ljacomet : this is something I discussed with @chrisdennis on Friday. I think it's a valid use case to be able to test Ehcache. And I cannot access the `UnitTestConnectionService` from management tests.

__DO NOT MERGE BEFORE @ljacomet's APPROVAL__

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
@@ -36,8 +36,6 @@ public class ClusteringServiceConfiguration
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<PersistentCacheManager> {
 
-  private static final String CLUSTER_SCHEME = "terracotta";
-
   private final URI clusterUri;
   private final boolean autoCreate;
   private final ServerSideConfiguration serverConfiguration;
@@ -157,10 +155,6 @@ public class ClusteringServiceConfiguration
   private static void validateClusterUri(URI clusterUri) {
     if (clusterUri == null) {
       throw new NullPointerException("Cluster URI cannot be null.");
-    }
-
-    if (!CLUSTER_SCHEME.equals(clusterUri.getScheme())) {
-      throw new IllegalArgumentException("Cluster Uri is not valid, clusterUri : " + clusterUri.toString());
     }
   }
 
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/config/ClusteringServiceConfigurationTest.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/config/ClusteringServiceConfigurationTest.java
@@ -76,18 +76,6 @@ public class ClusteringServiceConfigurationTest {
   }
 
   @Test
-  public void testInvalidURI() {
-
-    URI uri = URI.create("http://localhost:9540");
-    try {
-      new ClusteringServiceConfiguration(uri);
-      fail();
-    } catch (IllegalArgumentException e) {
-      assertThat(e.getMessage(), is("Cluster Uri is not valid, clusterUri : http://localhost:9540"));
-    }
-  }
-
-  @Test
   public void testValidURI() {
     URI uri = URI.create("terracotta://localhost:9540");
     ClusteringServiceConfiguration serviceConfiguration = new ClusteringServiceConfiguration(uri);
--- a/management/build.gradle
+++ b/management/build.gradle
@@ -23,26 +23,17 @@ dependencies {
   compile project(':api')
   compile project(':core')
   compile project(':impl')
-  compile "org.terracotta.management:management-registry:$parent.terracottaPlatformVersion"
-  compile "org.terracotta.management:management-entity-client:$parent.terracottaPlatformVersion"
+  compile "org.terracotta.management:management-entity-client:$parent.managementVersion"
 
   testCompile project(':xml')
   testCompile project(':clustered:client')
   testCompile project(':clustered:server')
   testCompile "org.terracotta:entity-test-lib:$parent.entityTestLibVersion"
   testCompile "org.terracotta:passthrough-server:$parent.entityTestLibVersion"
-  testCompile "org.terracotta.management:monitoring-service:$parent.terracottaPlatformVersion"
-  testCompile "org.terracotta.management:management-entity-server:$parent.terracottaPlatformVersion"
+  testCompile "org.terracotta.management:monitoring-service:$parent.managementVersion"
+  testCompile "org.terracotta.management:management-entity-server:$parent.managementVersion"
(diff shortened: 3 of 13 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -36,7 +39,7 @@
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<PersistentCacheManager> {
 
-  private static final String CLUSTER_SCHEME = "terracotta";
+  private static final Collection<String> CLUSTER_SCHEMES = new HashSet<String>(Arrays.asList("terracotta", "passthrough"));
```
chrisdennis: I would like to see this relaxed to allow any URI here.  Imho it's up to Voltron to make a statement on what a legal identifier is or isn't.  To that end I'd even contemplate advocating only stripping the last element of the path, and using the remainder as part of the cluster URI.

management/build.gradle
```diff
@@ -18,16 +18,41 @@ apply plugin: EhDeploy
 
 dependencies {
   compileOnly project(':xml')
+  compileOnly project(':clustered:client')
```
chrisdennis: This bothers me... but at least it's compileOnly,

management/build.gradle
```diff
@@ -18,16 +18,41 @@ apply plugin: EhDeploy
...
 
   compile project(':api')
   compile project(':core')
   compile project(':impl')
-  compile "org.terracotta.management:management-registry:$parent.managementVersion"
+  compile "org.terracotta.management:management-registry:$parent.terracottaPlatformVersion"
+  compile "org.terracotta.management:management-entity-client:$parent.terracottaPlatformVersion"
```
chrisdennis: How much do we transitively pull through here?

management/build.gradle
```diff
@@ -18,16 +18,41 @@ apply plugin: EhDeploy
...
+  testCompile "org.terracotta:entity-test-lib:$parent.entityTestLibVersion"
+  testCompile "org.terracotta:passthrough-server:$parent.entityTestLibVersion"
+  testCompile "org.terracotta.management:monitoring-service:$parent.terracottaPlatformVersion"
+  testCompile "org.terracotta.management:management-entity-server:$parent.terracottaPlatformVersion"
+  testCompile "org.terracotta.entities:clustered-map-server:$parent.terracottaPlatformVersion"
   testCompile "com.fasterxml.jackson.core:jackson-databind:2.7.5"
 }
 
 repositories {
   maven { url "http://snapshots.terracotta.org/" }
```
chrisdennis: This should be able to be removed.

management/build.gradle
```diff
@@ -18,16 +18,41 @@ apply plugin: EhDeploy
...
 }
+
+def java8 = {
```
chrisdennis: What is this for?

management/build.gradle
anthonydahanne: oh, right, management-entity-client depends on management-registry, will update

management/build.gradle
anthonydahanne: right

management/build.gradle
anthonydahanne: checking

anthonydahanne:
test this please

anthonydahanne:
@ljacomet please review, I have removed the scheme validation

clustered/client/src/test/java/org/ehcache/clustered/client/config/ClusteringServiceConfigurationTest.java
```diff
@@ -76,20 +76,8 @@ public void testBuilder() throws Exception {
...
-      fail();
-    } catch (IllegalArgumentException e) {
-      assertThat(e.getMessage(), is("Cluster Uri is not valid, clusterUri : http://localhost:9540"));
-    }
-  }
-
-  @Test
   public void testValidURI() {
-    URI uri = URI.create("terracotta://localhost:9540");
+    URI uri = URI.create("hello://localhost:9540");
```
ljacomet: I would not do that change. Leave `terracotta` there.

anthonydahanne:
cloudbees, test this please

management/build.gradle
```diff
@@ -18,16 +18,32 @@ apply plugin: EhDeploy
 
 dependencies {
   compileOnly project(':xml')
+  compileOnly project(':clustered:client')
 
   compile project(':api')
   compile project(':core')
   compile project(':impl')
-  compile "org.terracotta.management:management-registry:$parent.managementVersion"
+  compile "org.terracotta.management:management-entity-client:$parent.terracottaPlatformVersion"
```
ljacomet: There is a managementVersion property - let's keep using that one

management/build.gradle
```diff
@@ -18,16 +18,32 @@ apply plugin: EhDeploy
...
 
   testCompile project(':xml')
+  testCompile project(':clustered:client')
+  testCompile project(':clustered:server')
+  testCompile "org.terracotta:entity-test-lib:$parent.entityTestLibVersion"
+  testCompile "org.terracotta:passthrough-server:$parent.entityTestLibVersion"
+  testCompile "org.terracotta.management:monitoring-service:$parent.terracottaPlatformVersion"
+  testCompile "org.terracotta.management:management-entity-server:$parent.terracottaPlatformVersion"
```
ljacomet: On the two lines above, would use the `managementVersion` property

PR #1333, opened 2016-07-19
[master] Fix #1331 Invert invalid eviction prioritizer

### LINKED ISSUE #1331: On heap eviction prioritizer is inverted
As [reported on the users group](https://groups.google.com/forum/#!category-topic/ehcache-users/ehcache-core/OJdaVpQw6PU), we incorrectly evict the most recent element instead of the least recent.

A fix for this + basic tests making sure it never gets reintroduced are needed.

ljacomet:


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -128,7 +132,7 @@ public int compare(ValueHolder<?> t, ValueHolder<?> u) {
       } else if (u instanceof Fault) {
         return 1;
       } else {
-        return Long.signum(t.lastAccessTime(TimeUnit.NANOSECONDS) - u.lastAccessTime(TimeUnit.NANOSECONDS));
+        return Long.signum(u.lastAccessTime(TimeUnit.NANOSECONDS) - t.lastAccessTime(TimeUnit.NANOSECONDS));
```
chrisdennis: We should probably do:
```java
return -Long.compare(t.lastAccessTime(TimeUnit.NANOSECONDS), u.lastAccessTime(TimeUnit.NANOSECONDS));
```

ljacomet: Indeed - will adapt

PR #1339, opened 2016-07-25
Issue #951 : adding clustered configuration

anthonydahanne:


management/src/test/resources/clusteredConfiguration.txt
```diff
@@ -0,0 +1,24 @@
...
+                offheap:
+                    size: 1 MB
+                    tierHeight: 1000
+                clustered-dedicated:
+                    size: 1 MB (persistent)
+                    tierHeight: 10
+services:
+    - org.ehcache.clustered.client.config.ClusteringServiceConfiguration:
+        clusterUri: passthrough://server-1:9510/my-server-entity-1
+        readOperationTimeout: null
```
lorban: Wouldn't that make sense to exclude `readOperationTimeout` from the output when it's null?

management/src/test/resources/clusteredConfiguration.txt
```diff
@@ -0,0 +1,24 @@
+caches:
+    cache-1:
+        keyType: java.lang.String
+        valueType: java.lang.String
+        serviceConfigurations:
+        evictionAdvisor:
```
lorban: Why is `evictionAdvisor` empty? Does that mean it's null? In that case, shouldn't it be left out?

anthonydahanne: hum, interesting thoughts... I'm not sure it makes sense to display only non null values.
Anybody else cares to chime in ?
If that would be the case, then I guess I'd need to make sure all null values keys are not displayed

lorban: purely from a human reading perspective, I find those nulls to just add extra noise. But maybe you have a good reason to include them?

ljacomet: I am surprised this can be empty - it should indicate `None`

anthonydahanne: well I thought it made it obvious that, for example, no eviction advisor, etc. was configured

management/src/test/resources/clusteredConfiguration.txt
ljacomet: I believe the decision on including `null` is going to be a case by case basis. But honestly, there should be no `null` in any configuration element as we have defaults in place normally - but for example, in the case here, I would indicate `MAX_LONG` when that is the value rather than the numerical value of that.

ljacomet: Ignore the `null` here, that is a bug - see #1344 
However, I stand by what I said earlier ... as it will end up being MAX_LONG <unit>, this should be reflected instead of having the explicit long printed there.

anthonydahanne:
@lorban @ljacomet so I added None, but not on read timeout, since #1344 will take care of that.
thanks for reviewing and merging

PR #1345, opened 2016-07-26
Adding nexus staging plugin

akomakom:
Allows jobs to add "closeRepository" or "closeAndPromoteRepository" to the task list.

https://github.com/Codearte/gradle-nexus-staging-plugin/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,24 @@
  */
 import scripts.*
 
+buildscript {
+  repositories {
+    mavenCentral()
+  }
+  dependencies {
+    classpath "io.codearte.gradle.nexus:gradle-nexus-staging-plugin:0.5.3"
+  }
+}
+
+// This adds tasks to auto close or release nexus staging repos
+// see https://github.com/Codearte/gradle-nexus-staging-plugin/
+project.plugins.apply 'io.codearte.nexus-staging'
+project.nexusStaging {
+  username = project.sonatypeUser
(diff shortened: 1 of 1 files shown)


akomakom:
@ljacomet ping

ljacomet:
@akomakom before merging this, could we make sure `closeAndPromoteRepository` is disabled? I do not want a release build just yet to be synced to maven central without some checks before.

akomakom:
@ljacomet I could add some code that will raise an error if closeAndPromoteRepository is on the task list - aside from that there is nothing in the plugin config to disable it. 

Either way, the ony time closeAndPromoteRepository would be used is when it is explicitly passed on the command line, and only on hosts that have the credentials configured.

ljacomet:
Some mess up with the rebase here - there is a not needed merge commit

akomakom:
Yup fighting with git on this one

akomakom:
... I guess there is always --force.

Anyway, I think that this plugin doesn't work with JDK1.6 anymore, so this may be throwaway.

albinsuresh:
Can't this be closed then?

akomakom:
Closing, I'll revisit this down the line.

akomakom:
@ljacomet do we still want to do this?  Now that ehcache3 requires JDK8, this plugin would work.  I also disabled auto promote on my branch.  I'll open a new one if this is useful.

PR #1350, opened 2016-07-28
Close #1349 : clear stats for given capability when selected set is empty

### LINKED ISSUE #1349: Do not collect stats for capability when list of stats is empty
In the DefaultCollectorService, when a user tries to update the set of stats for a given capability with an empty selection, we should assume he/she is trying to stop collecting stats for this capability

anthonydahanne:
see #1349

### DIFF
--- a/management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
+++ b/management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
@@ -40,6 +40,7 @@ import org.terracotta.management.registry.StatisticQuery;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


ljacomet:
Comment added on #1349 as I am not too fond of the solution proposed.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -217,8 +217,23 @@ public synchronized void stopStatisticCollector() {
...
+      selectedStatsPerCapability.put(capabilityName, builder);
+    } else {
+      // we clear the stats set
+      selectedStatsPerCapability.remove(capabilityName);
+    }
+  }
+
+  // for test purposes
+  ConcurrentMap<String, StatisticQuery.Builder> getSelectedStatsPerCapability() {
+    return selectedStatsPerCapability;
```
ljacomet: Even for tests, I would make sure this is not modifiable. And thus only the Map interface is needed.

PR #1353, opened 2016-07-29
Close #1352 : make sure the request returned before handling the response

### LINKED ISSUE #1352: Fix test issue on our infra
There's this test, ClusteringManagementServiceTest that seems to be failing from time to time on our infra.
Looks like a race issue, occurring when we receive the response whereas the request still has not returned

anthonydahanne:


management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -234,15 +235,22 @@ private void remotelyUpdateCollectedStatistics() throws Exception {
...
       AtomicReference<String> managementCallId = new AtomicReference<>();
       BlockingQueue<ContextualReturn<?>> returns = new LinkedBlockingQueue<>();
 
       agent.setContextualReturnListener(new ContextualReturnListener() {
         @Override
         public void onContextualReturn(ClientIdentifier from, String id, ContextualReturn<?> aReturn) {
-          assertEquals(ehcacheClientIdentifier, from);
-          assertEquals(managementCallId.get(), id);
-          returns.offer(aReturn);
+          try {
```
GaryWKeim: Does this really need to catch InterruptedException? I don't recall seeing this type of construct in 4.x tests that use a latch.

anthonydahanne: I can't say that for this test I expect the exception to be thrown, and if it does, something went really wrong with this test

I'm using the latch here just to make sure the call returned, before doing any kind of assertion in the response (in some executions, the response was received before the request "call" returned)

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -234,15 +235,22 @@ private void remotelyUpdateCollectedStatistics() throws Exception {
...
+            assertEquals(ehcacheClientIdentifier, from);
+            // make sure the call completed
+            callCompleted.await();
```
ljacomet: Please do not have unlimited awaits - instead have a timeout just in case something really bad happens.

anthonydahanne:
PR fixed and rebased

PR #1357, opened 2016-08-02
Fix #1344 Default timeout in configuration

### LINKED ISSUE #1344: ClusteringServiceConfiguration#readOperationTimeout defaults to null
This means that the default is actually hidden deeper in the code. The design decision a while ago was that these defaults would live in the configuration classes as opposed to builders.

So `ClusteringServiceConfiguration` needs to be updated for that and the existence of the default removed from the rest of the code.

ljacomet:
Previously the ClusteringServiceConfiguration could have a null field
. This commit changes that by making sure to use the defined default.
 This simplifies configuration handling by not having to perform null
  checks.


ljacomet:
Whoops - fixing the tests

PR #1373, opened 2016-08-08
Fix #1342 replaceAtHead only when chain modified

### LINKED ISSUE #1342: ClusteredStore.get() : calling replaceAtHead() every time looks suboptimal
On average, `ServerStoreProxy.replaceAtHead()` is twice slower than `ServerStoreProxy.get()`, excluding chain resolving time.

Hence, it looks suboptimal to call `replaceAtHead()` on every `get()`. We should find a way to either make `replaceAtHead()` faster, or call it less often, or both.

ljacomet:
Still misses some `ClusteredStore` tests but want to get feedback on the general idea already.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
@@ -36,6 +36,14 @@ public interface ResolvedChain<K, V> {
   Result<V> getResolvedResult(K key);
 
   /**
+   * Indicates whether the {@link #getCompactedChain()} is effectively compacted
+   * compared to the original chain it was built from.
+   *
+   * @return {@code true} if the chain has been compacted during resolution, {@code false} otherwise
+   */
+  boolean isCompacted();
+
+  /**
    * Represents the {@link ResolvedChain} result of a resolver that resolves
    * all the keys in a {@link Chain}
    */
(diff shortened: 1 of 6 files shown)


albinsuresh:
Wouldn't this task be simpler if we have a _size()_ method in `Chain` interface? From what I see, implementing the _size()_ method in the current implementations of `Chain` is a trivial task as most of those are backed by an internal `List`. With that method in place this task would be just a size comparison between the original and compacted chains before performing the replace.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
```diff
@@ -36,21 +36,31 @@
   Result<V> getResolvedResult(K key);
 
   /**
+   * Indicates whether the {@link #getCompactedChain()} is effectively compacted
+   * compared to the original chain is was built from.
```
albinsuresh: minor typo: "chain is was built" -> "chain it was built"

ljacomet: Thanks for the catch

ljacomet:
@albinsuresh Having size would not cover 100% of the use cases. We still need to account for a chain having the same size but timestamps replaced by expiry times. And as soon as you account for that, it makes more sense to simply handle it inside the resolver. The only thing is that `isCompacted` may not be the best term. Maybe `isReplaceNeeded` or a close variation would be better.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -116,11 +120,14 @@ public V value() {
         chainBuilder = chainBuilder.add(payload);
       }
     }
-    if(result != null) {
+    compacted = compacted || keyMatch > 1;
+    if(result == null) {
+      compacted = compacted || keyMatch > 0;
```
albinsuresh: I burned my brain for nearly an hour and I still don't understand why this line is 
`compacted = compacted || keyMatch > 0;`
and not just
`compacted = keyMatch > 0;`

ljacomet: you are right, let me simplify this.

ljacomet:
PR now contains all tests - `ChainResolver` ones and `ClusteredStore.get` ones.

albinsuresh:
Looks like 2 test cases are missing or I failed to find those from the existing ones:

- Getting a single entry chain from the server that is already expired, which is the
```java
if (result == null) {
  compacted = keyMatch > 0;
}
```
case.

- Getting a single entry chain from the server and just the timeStamp of the entry gets updated because of the expiry calculation, which is the else case of 
```java
if (result == null) {
  compacted = keyMatch > 0;
} else {
  compacted = compacted || keyMatch > 1;
}
```
where `compacted` is _true_ and `keyMatch` is just 1.

albinsuresh:
Forget my previous comment. `ChainResolverTest.testResolveSingleRemove()` covers the first case(not really using an expired entry, which is fine I guess) and `ChainResolverTest.testResolveSinglePut()` covers the latter case. So LGTM. I'll merge if no one else has any comments.

PR #1374, opened 2016-08-08
clustered management - fix missing thread delegation when receiving management call requests

mathieucarbou:
When using the client communicator, the execution should be delegated to another thread.

So this PR is to use the ehcache execution service to create an executor specialized in executing the management call requests in another thread.

### DIFF
--- a/management/src/main/java/org/ehcache/management/cluster/DefaultClusteringManagementService.java
+++ b/management/src/main/java/org/ehcache/management/cluster/DefaultClusteringManagementService.java
@@ -30,6 +30,7 @@ import org.ehcache.management.registry.DefaultCollectorService;
 import org.ehcache.spi.service.Service;
 import org.ehcache.spi.service.ServiceDependencies;
 import org.ehcache.spi.service.ServiceProvider;
+import org.slf4j.LoggerFactory;
 import org.terracotta.exception.EntityAlreadyExistsException;
 import org.terracotta.exception.EntityNotFoundException;
 import org.terracotta.management.entity.ManagementAgentConfig;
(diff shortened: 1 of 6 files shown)


anthonydahanne:
ok, I think I got it : 
1) when a new DefaultClusteringManagementService is constructed, pass it a configuration
2) when the DefaultClusteringManagementService starts, we retrieve the executor to be used, using the alias from the configuration (if null, it will use the default pool alias)
3) when the DefaultClusteringManagementService switches to AVAILABLE, wrap the executor retrieved previously in a LoggingExecutor that will log RuntimeExceptions

LGTM, nice enhancement to the threads use in ehcache3 management calls

@lorban @ljacomet feel free to add your comments, otherwise I'll merge it tomorrow (if still based on current master)

management/src/main/java/org/ehcache/management/cluster/DefaultClusteringManagementService.java
```diff
@@ -40,20 +41,37 @@
 import org.terracotta.management.model.stats.ContextualStatistics;
 
 import java.util.Collection;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ExecutorService;
 
-@ServiceDependencies({CacheManagerProviderService.class, ExecutionService.class, TimeSourceService.class, ManagementRegistryService.class, EntityService.class})
+@ServiceDependencies({CacheManagerProviderService.class, ExecutionService.class, TimeSourceService.class, ManagementRegistryService.class, EntityService.class, ExecutionService.class})
```
ljacomet: That's twice the same dependency here - I know this was merged but please clean it up.

mathieucarbou: oh!! zut!

PR #1375, opened 2016-08-09
Closes #1355 Move state repository handling to Ehcache entity

### LINKED ISSUE #1355: Move handling of StateRepository inside Ehcache entity
Having it separated complicates a whole lot the create and destroy stories. It will just be easier if that is managed by a single entity.

albinsuresh:
Some more tests being added.


albinsuresh:
Added some more tests and did some more refactoring.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ClusteredStateRepository.java
```diff
@@ -28,17 +30,17 @@
 class ClusteredStateRepository implements StateRepository {
 
   private final ClusteringService.ClusteredCacheIdentifier clusterCacheIdentifier;
-  private final String composedId;
-  private final DefaultClusteringService defaultClusteringService;
+  private final EhcacheClientEntity clientEntity;
+  private final ConcurrentMap<String, ConcurrentClusteredMap> knownMaps = new ConcurrentHashMap<String, ConcurrentClusteredMap>();
 
-  ClusteredStateRepository(ClusteringService.ClusteredCacheIdentifier clusterCacheIdentifier, String id, DefaultClusteringService defaultClusteringService) {
+  ClusteredStateRepository(ClusteringService.ClusteredCacheIdentifier clusterCacheIdentifier, String id, EhcacheClientEntity clientEntity) {
```
AbfrmBlr: Do we use `id` any more ?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheEntityMessage.java
```diff
@@ -23,13 +23,43 @@
...
+      return this.code;
+    }
+
+    public static Type toType(byte code) {
+      for (Type type: Type.values()) {
+        if(type.getCode() == code) {
+          return type;
+        }
+      }
+      throw new UnsupportedOperationException("Invalid message type code: " + code);
```
lorban: Shouldn't that be a `IllegalArgumentException`?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ServerStoreOpMessage.java
```diff
@@ -23,12 +23,12 @@
...
-    APPEND((byte) 12),
-    REPLACE((byte) 13),
-    CLIENT_INVALIDATION_ACK((byte) 14),
-    CLEAR((byte) 15),
+    GET((byte) 11),
+    GET_AND_APPEND((byte) 12),
+    APPEND((byte) 13),
+    REPLACE((byte) 14),
+    CLIENT_INVALIDATION_ACK((byte) 15),
+    CLEAR((byte) 16),
```
lorban: Why renumbering the op codes? This is okay as we said we wouldn't guarantee compatibility across versions for now but I fail to see the benefit here.

albinsuresh: With the introduction of the state repo operations, one more top level message got introduced, in addition to life-cycle messages and server-store messages. So I thought I'll categorize them into 3 buckets:

01<value<= 10 : life cycle messages
11<value<=20: server store messages
21<value<=30: state repo messages

This was based on the assumption that the sub-level message won't go beyond 10.
So basically I marked the upper bounds: 10, 20 and 30 and used those upper bound values as the enum values for the top-level message types defined in `EhcacheEntityMessage.Type`
And then there was a clash: `EhcacheEntityMessage.Type.LIFECYCLE_OP` and `ServerStoreOpMessage.ServerStoreOp.GET` had the same value: 10. So to avoid that clash I advanced all the `ServerStoreOpMessage.ServerStoreOp` types starting from 11.

I could have made the upper bound of life cycle message type as 9. But then the sequence of upper bounds looked weird: 9, 20 and 30. That's why I didn't do that. But I could revert to that if that's fine.

Or there is something even better that I could do. Just change the value of `ServerStoreOpMessage.ServerStoreOp.GET` to 16 from 10 instead of changing the value of all enums. So I think that's the best thing to do here.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheEntityMessage.java
albinsuresh: Yup

AbfrmBlr:
LGTM

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -285,24 +310,24 @@ private EhcacheEntityResponse invokeLifeCycleOperation(ClientDescriptor clientDe
       case DESTROY_SERVER_STORE:
         destroyServerStore(clientDescriptor, (DestroyServerStore) message);
         break;
+      case CREATE_CLUSTERED_MAP:
+        createConcurrentMap(clientDescriptor, (CreateClusteredMap) message);
```
ljacomet: I am wondering if this explicit operation is needed. It creates IMO unnecessary message exchanges with the server. The equivalent operation could be done lazily on the first map interaction.
@chrisdennis @AbfrmBlr wdyt?

chrisdennis: I don't know that seems like a bit of a premature optimization... it's only 1 extra message right?

ljacomet: Well, it is two messages _per client_ because the way it is implemented is based on ignoring result of `putIfAbsent` then doing `get`. And it creates added complexity it seems to me.

AbfrmBlr: I understand creation of map is happening in single round trip to server.  Only once the create map is called per map.The default serializer uses only one map. Right @albinsuresh ?

albinsuresh: There won't be two server round trips for a single map creation/retrieval. There'll only be one. I believe @ljacomet is referring to the _putIfAbsent_ and subsequent _get_ calls in `ClusteredStateRepository.getPersistentConcurrentMap()` method. Here the _putIfAbsent_ and _get_ happens on a local map *knownMaps* that keeps track of the client-side end points of clustered maps created. It is the `ConcurrentClusteredMap` constructor that is responsible for the creation/retrieval of the clustered map on the server side. The failure of _putIfAbsent_ would mean that there was a clustered map endpoint available already on the client-side which is attached to the server-side map. The success of _putIfAbsent_ implies that a clustered map has been created on *or attached to* the server side and the client-side endpoint to that map is added to the *knownMaps*. In all cases there is only one round trip.

I think the operation name is a bit confusing here. It should have been `CREATE_OR_ATTACH_CLUSTERED_MAP` and not just `CREATE_CLUSTERED_MAP` because that's what actually happens on the server side. There is no 2-step creation and retrieval. Both happens with that single message.

ljacomet: Indeed, I was wrong on the number of messages. The `get` does not cause a message. But since there is nothing that indicates if the map is in use or not, I still think creating it implicitly would just be easier. Of course, the only reason I am thinking about allowing the _implicit_ here is that we are protected by the lifecycle of the clustered tier and the entity itself above it.

albinsuresh: Yeah it could be done. Since every clustered map operation message contains the cache name and the map name, we can delay the creation of server side state repository and the map itself till the first operation on the map is performed.

So just to reiterate the approach, when a client performs a map operation and that message reaches the server side, the cache name is extracted and a state repository with that name will be created if one is not found, and then within the retrieved/created state repository the map is created using the extracted map name if one is not found and then the actual operation is performed on that created/retrieved map.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ClusteredStateRepository.java
```diff
@@ -28,17 +30,19 @@
...
     this.composedId = clusterCacheIdentifier.getId() + "-" + id;
-    this.defaultClusteringService = defaultClusteringService;
+    this.clientEntity = clientEntity;
   }
 
   @Override
   public <K extends Serializable, V extends Serializable> ConcurrentMap<K, V> getPersistentConcurrentMap(String name, Class<K> keyClass, Class<V> valueClass) {
-    return defaultClusteringService.getConcurrentMap(clusterCacheIdentifier, composedId + "-" + name, keyClass, valueClass);
+    knownMaps.putIfAbsent(name, new ConcurrentClusteredMap<K, V>(clusterCacheIdentifier.getId(), composedId + "-" + name, clientEntity));
+    return knownMaps.get(name);
```
ljacomet: That pattern of doing `putIfAbsent` followed by a blind `get` is broken, does not give any of the guarantees we expect from `putIfAbsent`. Also if we decide to forget about the create message, I believe this client state becomes obsolete. You can create a `Map` each time it is requested.

albinsuresh: With the delayed creation logic as in the following commit, this should be fine, right?

ljacomet: With that latest change, the `ConcurrentClusteredMap` is effectively stateless and so you can drop the `knownMaps` field in `ClusteredStateRepository`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -272,35 +267,23 @@ public boolean handlesResourceType(ResourceType<?> resourceType) {
...
 
-  private ConcurrentHashMap<EntityRef<ConcurrentClusteredMap, Object>, ConcurrentClusteredMap<?, ?>> getNewMapEntityMap() {
-    return new ConcurrentHashMap<EntityRef<ConcurrentClusteredMap, Object>, ConcurrentClusteredMap<?, ?>>();
-  }
-
   @Override
   public void releasePersistenceSpaceIdentifier(PersistenceSpaceIdentifier<?> identifier) throws CachePersistenceException {
-    if (!isKnownIdentifier(identifier)) {
+    ClusteredCacheIdentifier clusterCacheIdentifier = (ClusteredCacheIdentifier) identifier;
+    if (!knownPersistenceSpaces.containsKey(clusterCacheIdentifier)) {
```
ljacomet: This can be replaced by the remove call directly:
`if (knownPersistenceSpaces.remove(clusterCacheIdentifier) == null)`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -272,35 +267,23 @@ public boolean handlesResourceType(ResourceType<?> resourceType) {
...
 
   @Override
   public StateRepository getStateRepositoryWithin(PersistenceSpaceIdentifier<?> identifier, String name) throws CachePersistenceException {
-    if (!isKnownIdentifier(identifier)) {
-      throw new CachePersistenceException("Unknown identifier: " + identifier);
-    }
-    DefaultClusterCacheIdentifier clusterCacheIdentifier = (DefaultClusterCacheIdentifier) identifier;
-    return new ClusteredStateRepository(clusterCacheIdentifier, name, this);
+    ClusteredCacheIdentifier clusterCacheIdentifier = (ClusteredCacheIdentifier) identifier;
+    knownPersistenceSpaces.putIfAbsent(clusterCacheIdentifier, new ClusteredStateRepository(clusterCacheIdentifier, name, this.entity));
```
ljacomet: This use of `putIfAbsent` is wrong - see [there](https://github.com/ehcache/ehcache3/blob/080be1433862d32fcaf485436a04f61f17417546/impl/src/main/java/org/ehcache/impl/serialization/TransientStateRepository.java#L38) for the proper pattern

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -481,21 +418,25 @@ public String getId() {
     }
 
     @Override
-    public String toString() {
-      return getClass().getSimpleName() + "@" + id;
+    public boolean equals(final Object o) {
```
ljacomet: Why is this required now? It kind of breaks the idea of opaque identifiers.

albinsuresh:
Will squash all the commits if these changes are good to go.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -486,16 +444,15 @@ public String toString() {
...
-    final V second;
+  private static class ClusteredSpace {
+
+    private final ClusteredCacheIdentifier identifier;
+    private final ConcurrentMap<String, ClusteredStateRepository> stateRepositories;
 
-    Tuple(K first, V second) {
-      this.first = first;
-      this.second = second;
+    public ClusteredSpace(final ClusteredCacheIdentifier identifier, final EhcacheClientEntity clientEntity) {
```
ljacomet: `clientEntity` is unused. Also this constructor does not need to be public.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/DefaultClusteringServiceTest.java
```diff
@@ -1933,6 +1938,63 @@ public void testGetServerStoreProxyReturnsStrongStore() throws Exception {
...
+        new ClusteringServiceConfiguration(URI.create(CLUSTER_URI_BASE), true, new ServerSideConfiguration(Collections.<String, Pool>emptyMap()));
+    DefaultClusteringService service = new DefaultClusteringService(configuration);
+    PersistableResourceService.PersistenceSpaceIdentifier cacheIdentifier1 = service.getPersistenceSpaceIdentifier("myCache1", null);
+    PersistableResourceService.PersistenceSpaceIdentifier cacheIdentifier2 = service.getPersistenceSpaceIdentifier("myCache2", null);
+    StateRepository repository1 = service.getStateRepositoryWithin(cacheIdentifier1, "myRepo");
+    StateRepository repository2 = service.getStateRepositoryWithin(cacheIdentifier2, "myRepo");
+    assertThat(repository1, not(sameInstance(repository2)));
+  }
+
+  @Test(expected = CachePersistenceException.class)
```
ljacomet: I really dislike using this construct when the test is more than a single line. Got bitten in the past by a test that was reporting the proper exception type but absolutely not the expected one.
Will merge the PR anyway but would appreciate a follow up PR fixing this and other similar tests in here.

PR #1379, opened 2016-08-09
do not call cacheManager.getStatus() into a scheduled task because it would block the shutdown of the executor

mathieucarbou:
do not call cacheManager.getStatus() into a scheduled task
because it would block the shutdown of the executor


chrisdennis:
Simple fix, waiting on CI.

mathieucarbou:
@chrisdennis @anthonydahanne :

I added another commit, related to a missing exposed configuration for the management call timeout. This was there, but not configurable. It is now.

PR #1381, opened 2016-08-11
Closes #1343 On-demand encode/decode of Operation values during chain resolution

### LINKED ISSUE #1343: Suboptimal ChainResolver
`ChainResolver.resolve()` spends a lot of time decoding and encoding operations.

There certainly are optimizations that could be done here. For instance, the key and value of the payload always are decoded even if the key doesn't match -which in that case makes the decoded value useless-.

albinsuresh:
The chain resolution now handles the following cases:

1. Decode value of target operations only when required
2. Avoid decoding values of non-target operations
3. No expiry calculations and expiryTime updates when expiry is not configured
4. Avoid rebuilding the chain after resolution if no compaction has happened

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
@@ -23,6 +23,7 @@ import org.ehcache.clustered.client.internal.store.operations.codecs.OperationsC
 import org.ehcache.clustered.common.internal.store.Chain;
 import org.ehcache.clustered.common.internal.store.Element;
 import org.ehcache.expiry.Duration;
+import org.ehcache.expiry.Expirations;
 import org.ehcache.expiry.Expiry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -39,6 +40,9 @@ public class ChainResolver<K, V> {
   private final Expiry<? super K, ? super V> expiry;
 
   public ChainResolver(final OperationsCodec<K, V> codec, Expiry<? super K, ? super V> expiry) {
+    if(expiry == null) {
+      throw new NullPointerException("Expiry can not be null");
+    }
     this.codec = codec;
     this.expiry = expiry;
   }
@@ -72,47 +76,49 @@ public class ChainResolver<K, V> {
         if(result == null) {
           continue;
         }
-        if(operation.isExpiryAvailable()) {
-          expirationTime = operation.expirationTime();
-          if (expirationTime == Long.MIN_VALUE) {
-            continue;
-          }
-          if (now >= expirationTime) {
-            result = null;
-          }
-        } else {
-          Duration duration;
-          try {
-            if(previousResult == null) {
-              duration = expiry.getExpiryForCreation(key, result.getValue());
-              if (duration == null) {
-                result = null;
-                continue;
-              }
-            } else {
-              duration = expiry.getExpiryForUpdate(key, new ValueSupplier<V>() {
-                @Override
-                public V value() {
-                  return previousResult.getValue();
-                }
-              }, result.getValue());
-              if (duration == null) {
-                continue;
-              }
+        if (expiry != Expirations.noExpiration()) {
+          if(operation.isExpiryAvailable()) {
+            expirationTime = operation.expirationTime();
+            if (expirationTime == Long.MIN_VALUE) {
+              continue;
             }
-          } catch (Exception ex) {
-            LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", ex);
-            duration = Duration.ZERO;
-          }
-          compacted = true;
-          if(duration.isInfinite()) {
-            expirationTime = Long.MIN_VALUE;
-            continue;
-          }
-          long time = TIME_UNIT.convert(duration.getLength(), duration.getTimeUnit());
-          expirationTime = time + operation.timeStamp();
-          if(now >= expirationTime) {
-            result = null;
+            if (now >= expirationTime) {
+              result = null;
+            }
+          } else {
+            Duration duration;
+            try {
+              if(previousResult == null) {
+                duration = expiry.getExpiryForCreation(key, result.getValue());
+                if (duration == null) {
+                  result = null;
+                  continue;
+                }
+              } else {
+                duration = expiry.getExpiryForUpdate(key, new ValueSupplier<V>() {
+                  @Override
+                  public V value() {
+                    return previousResult.getValue();
+                  }
+                }, result.getValue());
+                if (duration == null) {
+                  continue;
+                }
+              }
+            } catch (Exception ex) {
+              LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", ex);
+              duration = Duration.ZERO;
+            }
+            compacted = true;
+            if(duration.isInfinite()) {
+              expirationTime = Long.MIN_VALUE;
+              continue;
+            }
+            long time = TIME_UNIT.convert(duration.getLength(), duration.getTimeUnit());
+            expirationTime = time + operation.timeStamp();
+            if(now >= expirationTime) {
+              result = null;
+            }
           }
         }
       } else {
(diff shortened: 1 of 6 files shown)


lorban:
Maybe a few JHM tests would be handy to figure out how much extra performance this change gives us. And a Flight Recorder log would also help making sure this code is squeezing as much as it can from the hardware.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
```diff
@@ -73,7 +71,7 @@ public Chain getCompactedChain() {
 
     @Override
     public boolean isCompacted() {
-      return compacted;
+      return compactedChain != null;
```
ljacomet: Please no, let's not have compactedChain being `null` legal. What is it buying us exactly?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -72,62 +76,67 @@ public ChainResolver(final OperationsCodec<K, V> codec, Expiry<? super K, ? supe
...
+        ByteBuffer payload = codec.encode(resolvedOperation);
+        chainBuilder = chainBuilder.add(payload);
+      }
+      return new ResolvedChain.Impl<K, V>(chainBuilder.build(), key, result);
     } else {
-      compacted = compacted || keyMatch > 1;
-      Operation<K, V> resolvedOperation = new PutOperation<K, V>(key, result.getValue(), -expirationTime);
-      ByteBuffer payload = codec.encode(resolvedOperation);
-      chainBuilder = chainBuilder.add(payload);
+      return new ResolvedChain.Impl<K, V>(null, key, result);
```
ljacomet: I really do not like using null there.

ljacomet:
What about the test failure?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
albinsuresh: Was just trying to avoid the `compacted` _boolean_ flag.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
albinsuresh: I just wanted to avoid rebuilding a non-compacted chain to create the `ResolvedChain.Impl` instance. Passing _null_ was my way of avoiding the rebuilding and the _boolean_ flag indicating whether the `ResolvedChain` has a compacted chain. But with the _boolean_ flag reverted as mentioned in the previous comment I should be able to get rid of this null. Instead of rebuilding the chain I could just pass the origial chain itself here.

albinsuresh:
The test failures are due to the platform version updates. I never updated this PR after the version changes were reverted. Thought of doing it after the JMH test additions are also done.

albinsuresh:
The following branch has the JMH test added for the resolve operation: https://github.com/albinsuresh/ehcache3/tree/jmh

The average time for the _resolve_ operation was measured with the following configuration:
```
# JMH 1.12 (released 139 days ago, please consider updating!)
# VM version: JDK 1.8.0_51, VM 25.51-b03
# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk/Contents/Home/jre/bin/java
# VM options: -Dfile.encoding=UTF-8 -Duser.country=US -Duser.language=en -Duser.variant
# Warmup: 10 iterations, 1 s each
# Measurement: 10 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.ehcache.clustered.client.internal.store.operations.ChainResolverJMHTest.measureResolve
```

And the results were as follows:

Before the optimization:
```
# Run progress: 0.00% complete, ETA 00:00:20
# Fork: 1 of 1
# Warmup Iteration   1: 2.349 us/op
# Warmup Iteration   2: 1.941 us/op
# Warmup Iteration   3: 1.600 us/op
# Warmup Iteration   4: 1.660 us/op
# Warmup Iteration   5: 1.671 us/op
# Warmup Iteration   6: 1.620 us/op
# Warmup Iteration   7: 2.027 us/op
# Warmup Iteration   8: 1.917 us/op
# Warmup Iteration   9: 1.766 us/op
# Warmup Iteration  10: 1.723 us/op
Iteration   1: 1.712 us/op
Iteration   2: 1.690 us/op
Iteration   3: 1.677 us/op
Iteration   4: 1.666 us/op
Iteration   5: 1.666 us/op
Iteration   6: 1.714 us/op
Iteration   7: 1.686 us/op
Iteration   8: 1.678 us/op
Iteration   9: 1.688 us/op
Iteration  10: 1.675 us/op


Result "measureResolve":
  1.685 ±(99.9%) 0.025 us/op [Average]
  (min, avg, max) = (1.666, 1.685, 1.714), stdev = 0.017
  CI (99.9%): [1.660, 1.711] (assumes normal distribution)


# Run complete. Total time: 00:00:20

Benchmark                            Mode  Cnt  Score   Error  Units
ChainResolverJMHTest.measureResolve  avgt   10  1.685 ± 0.025  us/op
```

After optimization:
```
# Run progress: 0.00% complete, ETA 00:00:20
# Fork: 1 of 1
# Warmup Iteration   1: 1.136 us/op
# Warmup Iteration   2: 1.194 us/op
# Warmup Iteration   3: 0.888 us/op
# Warmup Iteration   4: 0.852 us/op
# Warmup Iteration   5: 0.851 us/op
# Warmup Iteration   6: 0.827 us/op
# Warmup Iteration   7: 0.855 us/op
# Warmup Iteration   8: 0.870 us/op
# Warmup Iteration   9: 0.837 us/op
# Warmup Iteration  10: 0.864 us/op
Iteration   1: 0.878 us/op
Iteration   2: 0.874 us/op
Iteration   3: 0.847 us/op
Iteration   4: 0.844 us/op
Iteration   5: 0.859 us/op
Iteration   6: 0.827 us/op
Iteration   7: 0.829 us/op
Iteration   8: 0.824 us/op
Iteration   9: 0.840 us/op
Iteration  10: 0.830 us/op


Result "measureResolve":
  0.845 ±(99.9%) 0.030 us/op [Average]
  (min, avg, max) = (0.824, 0.845, 0.878), stdev = 0.020
  CI (99.9%): [0.816, 0.875] (assumes normal distribution)


# Run complete. Total time: 00:00:20

Benchmark                            Mode  Cnt  Score   Error  Units
ChainResolverJMHTest.measureResolve  avgt   10  0.845 ± 0.030  us/op
```

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/ChainResolver.java
```diff
@@ -72,62 +76,67 @@ public ChainResolver(final OperationsCodec<K, V> codec, Expiry<? super K, ? supe
...
         }
       } else {
         payload.flip();
         chainBuilder = chainBuilder.add(payload);
       }
     }
-    if(result == null) {
-      compacted = keyMatch > 0;
+
+    compacted = (result == null) ? (keyMatch > 0) : (compacted || keyMatch > 1);
```
ljacomet: I really find this harder to read than the if/else blocks that were there before. Is it really needed in order to only have if/else without else if?

albinsuresh: There was no particular reason behind that. There are too many if-else blocks already in that method. Including the ones following this line. Since this was the simplest if-else block that could easily be replaced with a ternary operator I went for it. There is no other reason. I can revert it, if is affecting readability.

ljacomet: Let's leave it like that.

ljacomet:
Good to merge once rebased

PR #1387, opened 2016-08-16
Fix #1354 JCache compatible ClusteredStore

### LINKED ISSUE #1354: Unable to use clustered tier with JSR-107 support
As with #1146 there needs to be fix applied to the clustered tier implementation so that authoritative tier stats can be found.

ljacomet:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -666,4 +679,14 @@ public Consistency getConsistency() {
       return consistency;
     }
   }
+
+  private static final class ClusteredStoreStatsSettings {
+    @ContextAttribute("tags") private final Set<String> tags = new HashSet<String>(Arrays.asList("store"));
```
chrisdennis: Collections.singleton("store")?

clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
```diff
@@ -124,14 +125,15 @@ public void testDestroyCacheManagerDoesNotAffectsExistingCacheWithExistingClient
 
     assertThat(cache.get(1L), is("One"));
 
+    persistentCacheManager2.close();
```
chrisdennis: Finally block? And if we're doing this should we also destroy?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
ljacomet: Is "I copy pasted your code!" an excuse? 😜

clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
ljacomet: I have been wondering that ... but then I'd rather have it handled by passthrough server. Something like `destroyAllAndStop()`. Makes sense?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
ljacomet: Fixed

clustered/client/src/test/java/org/ehcache/clustered/client/CacheManagerDestroyTest.java
chrisdennis: Okay... what about the finally blocks?

ljacomet: I tried it and I find it makes the code more complex to read for little added value. I want the `close()` to happen when all goes well and pretty much do not care when stuff goes wrong ... because that should only happen locally for a dev. Makes sense?

ljacomet:
Added a test about the associate / dissociate. @chrisdennis does that look good? @AbfrmBlr does that fill the gap you highlighted in the other related PRs?

chrisdennis:
This looks fine to me.

PR #1388, opened 2016-08-17
[master] Fix #1328 Dissociates stats context

### LINKED ISSUE #1328: StatisticsManager associate in AbstractOffheap store causes leaks
The statistics manager associate call of a store causes leaks in tests as the equivalent dissassociate call is only available if and only if the management service is started...

The following line in AbstractOffHeapStore does not seem to have a corresponding disassociate call when the store is released.

`StatisticsManager.associate(offHeapStoreStatsSettings).withParent(this);`

This causes a reference to the store to remain. So when a cache with the same alias is created and removed a 100 times, 100 instances of the store remains in the heap. This is only a problem for stores that inherit from AbstractOffHeapStore.

ljacomet:


ljacomet:
test this please

AbfrmBlr:
Is it possible to add tests around the clustered store to test whether context objects are associated/dissociated properly?

Issue #1286, opened 2016-07-05
Statistics work


lorban:
Define a set of standard statistics and implement them.

ljacomet:
I certainly hope so. I believe starting with the set of stats exposed by JCache / JSR-107 is an idea. Then we can discuss the Java based API to expose them. And maybe later discuss about exposing more.

lorban:
In my mind, this task was more about first deciding what stats we want to expose. At the moment, there are dozens of counters all around but that's pretty much it. We don't calculate rates, ratios, latencies... and the store counters that should be used to calculate useful stats change depending on the cache topologies.

I think we should first work on the statistics themselves before defining an API to expose them.

lorban:
Here is for instance a set of stats that would be very handy to have when performing perf testing:

```
tier hit/miss ratio
 tier hit count
 tier hit latency
 tier miss/fault count
 tier miss/fault latency
 tier eviction count
 tier eviction latency
 tier resource usage in %

cache hit/miss ratio
 cache hit count
 cache hit latency
 cache miss count
 cache miss latency
 cache eviction count
 cache eviction latency
 cache resource usage in %
 
special cases:
 clustered tier's resource usage is not available on the client
 clustered tier should also report invalidation count and latencies
```

aurbroszniowski:
What would be useful is

- cache tiers hit distribution (e.g. if we have a heap+offheap+clustered cache, reporting where the hit came from, e.g. 10% from heap, 30% from offheap, 50% from clustered)
- for cache read (hit) and cache write, the latencies running medians, min, max and probably moving averages
- nb of elements in each tier (even approximative)

those 3 types of numbers would help to decide of the caches sizing, by giving clear feedback on how performance is updated when changing sizes of the cache tiers.

for the statistics you listed by @lorban, it would help to decide which one are really useful if you could explain for each one of them, how they can be used and for which purpose.

lorban:
Tier hit/miss ratio will tell you how useful a tier is.
Tier hit count and latencies will tell you how fast a tier serves its hits.
Tier miss/fault latencies will tell you how much overhead a tier adds to the lower tiers' hit (in case of higher tiers) or how fast it misses.
Tier eviction count and latencies will tell you how often the tier needs to make room to fault in lower tier's mappings and how fast it does that.
Tier resource usage will tell you how much of its allocated resources a tier consumes. I start to believe reporting count as well as bytes would be useful.

[...]

PR #1390, opened 2016-08-17
Issue 1286

gibsong:
This PR is for https://github.com/ehcache/ehcache3/issues/1286.  
I took over Ludovic's branch and am now submitting all the bug and unit test fixes for it.  I still need to add in some stats and finish a few tests in CacheStatisticsMXBeanTest.  But in the meantime I was hoping to get this PR moving forward.


build.gradle
```diff
@@ -20,7 +20,8 @@ ext {
 
   // Third parties
   offheapVersion = '2.2.2'
-  statisticVersion = '1.1.0'
+  managementVersion = '5.0.0.beta'
```
ljacomet: That needs to go away.

build.gradle
```diff
@@ -20,7 +20,8 @@ ext {
...
+  statisticVersion = '1.1-SNAPSHOT'
```
ljacomet: This seems to indicate a new version of stats is required. With a SNAPSHOT here PR will never build.
What are the required statistics changes? Is there an issue and a PR opened already?

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -239,32 +245,68 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
...
+    properties.put("discriminator", STATISTICS_TAG);
+    StatisticsManager.createPassThroughStatistic(this, "mappings", tags, properties, new Callable<Number>() {
       @Override
       public Number call() throws Exception {
         return map.mappingCount();
       }
     });
+    StatisticsManager.createPassThroughStatistic(this, "maxMappings", tags, properties, new Callable<Number>() {
+      @Override
+      public Number call() throws Exception {
```
mathieucarbou: Does it mean that we are allowed to throw in this code, when reading the current state ? If yes, how the exception is handled, and what output it will create if we query the statistics from M&M ?

build.gradle
gibsong: you mean remove line: managementVersion = '5.0.0.beta'?  This was in the code I inherited from Ludovic.

management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatisticsProvider.java
```diff
@@ -65,10 +65,10 @@ public Capability getCapability() {
   @Override
   public Map<String, Statistic<?, ?>> collectStatistics(Context context, Collection<String> statisticNames, long since) {
     Map<String, Statistic<?, ?>> statistics = new HashMap<String, Statistic<?, ?>>(statisticNames.size());
-    EhcacheStatistics ehcacheStatistics = (EhcacheStatistics) findExposedObject(context);
+    StandardEhcacheStatistics ehcacheStatistics = (StandardEhcacheStatistics) findExposedObject(context);
     if (ehcacheStatistics != null) {
       for (String statisticName : statisticNames) {
-        statistics.putAll(ehcacheStatistics.queryStatistic(statisticName, since));
+        statistics.put(statisticName, ehcacheStatistics.queryStatistic(statisticName, since));
```
mathieucarbou: Does it mean that there is not anymore the concept of "groups" of statistics ? 
Before, we had some group name, and when we asked for i.e. `AllCacheGet`, then it would return all associated statistics (AllCacheGetCount, AllCacheGetRate, AllCacheGetLatencyMinimum, ...).
Hence the `.putAll()` ;-)

gibsong: Let me look into re-implementing those stats.  That is my next step.  I still have to add some stats that were removed and some new stats that Anthony requested.  Unfortunately this PR only includes all the changes that were required to fix the broken unit tests and bugs.

build.gradle
ljacomet: Yes it was, and it was in code earlier. It changed at some point and you probably kept it around while resolving a conflict. Like I said, needs to go. The same prop is defined correctly a few lines below.

build.gradle
gibsong: Yes there are some other terracotta stat changes.  I had a chat with Anthony and he helped me sort out the details.  I have submitted PR: 
https://github.com/Terracotta-OSS/statistics/pull/10
These changes need to be merged and released.  Once this is done I will update the gradle.build with the new version.  I believe it will be 1.2.0

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
gibsong: unfortunately i am not sure on this one as well.  Ludovic also wrote this code and I haven't really looked at it yet.  I can take a look when i add in the stats.

mathieucarbou:
@gibsong : can we rename the title of this PR to "Issue 1286" ? So that it matches the issue it closes in git ?

gibsong:
I just pushed new code.  This code contains the following statistics:

working stats:
Cache:HitCount
Cache:HitRatio
Cache:MissCount
Cache:MissRatio
Cache:ClearCount
OnHeap:EvictionCount
OffHeap:EvictionCount

done but need better testing to verify correctness:
OnHeap:OccupiedBytesCount
OffHeap:OccupiedBytesCount
Disk:OccupiedBytesCount
Disk:EvictionCount

management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatisticsProvider.java
gibsong: yes the group stats are no longer available.

build.gradle
```diff
@@ -59,6 +59,7 @@ subprojects {
   targetCompatibility = 1.6
 
   repositories {
+    mavenLocal()
```
anthonydahanne: please revert

build.gradle
```diff
@@ -88,7 +89,7 @@ subprojects {
   }
 
   test {
-    maxHeapSize = "512m"
+    maxHeapSize = "1024m"
```
anthonydahanne: is it absolutely necessary ? can you please revert if not ?

gibsong:
test please

build.gradle
gibsong: i will remove it.  I assume that allowed the build to grab the statisticVersion = '1.1-SNAPSHOT'?

build.gradle
gibsong: i will change it back to 512m, but on my machine it won't work unless I increase it to 1024m.

gibsong: i reverted it back to 512m and checked in the changes.  However this causes 16 tests to fail with the following error:

Caused by: java.lang.IllegalArgumentException: An attempt was made to allocate more off-heap memory than the JVM can allow. The limit on off-heap memory size is given by the -XX:MaxDirectMemorySize command (or equivalent).
	at org.terracotta.offheapstore.paging.UpfrontAllocatingPageSource.allocateBackingBuffers(UpfrontAllocatingPageSource.java:468)
	at org.terracotta.offheapstore.paging.UpfrontAllocatingPageSource.<init>(UpfrontAllocatingPageSource.java:136)
	at org.terracotta.offheapstore.paging.UpfrontAllocatingPageSource.<init>(UpfrontAllocatingPageSource.java:119)
	at org.ehcache.clustered.server.EhcacheStateServiceImpl$ResourcePageSource.<init>(EhcacheStateServiceImpl.java:368)
	at org.ehcache.clustered.server.EhcacheStateServiceImpl$ResourcePageSource.<init>(EhcacheStateServiceImpl.java:359)
	at org.ehcache.clustered.server.EhcacheStateServiceImpl.createPageSource(EhcacheStateServiceImpl.java:222)

anthonydahanne: ok - that's interesting, maybe that's a windows thing (our CI on cloudbees only tests on Linux)
That maybe worth asking @ljacomet about the need to have windows slaves.
Anyway, I'm ok with increasing to 1024m as long as there's a reason, and you just gave it - thank you !

anthonydahanne:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -677,15 +779,24 @@ public ClusteredCacheIdentifier getCacheIdentifier() {
...
     }
+
+    public int getConcurrency() {
+      return concurrency;
+    }
   }
 
+  //added @SuppressFBWarnings to get rid of findBugs error which recommended making instance variables static
+  @SuppressFBWarnings
   private static final class ClusteredStoreStatsSettings {
```
lorban: I believe we should get rid of this class as I fail to see its use after the current stats refactoring.

clustered/integration-test/src/test/java/org/ehcache/Clust.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.terracotta.connection.Connection;
+import org.terracotta.testing.rules.BasicExternalCluster;
+import org.terracotta.testing.rules.Cluster;
+
+/**
+ * @author Ludovic Orban
+ */
+public class Clust {
```
lorban: This class should go away. It's a quick and dirty manual test I wrote to help me during development.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -29,16 +29,14 @@
 @CommonComponent
 public class ServerStoreImpl implements ServerStore {
 
-  private static final int OFFHEAP_CHAIN_SEGMENTS = 16;
```
lorban: This change has to do with a completely unrelated experiment of mine I accidentally committed together with the stats work. It should be reverted.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -293,10 +295,17 @@ private boolean removeInternal(final K key) {
    */
   @Override
   public void clear() {
+    this.clearObserver.begin();
     statusTransitioner.checkAvailable();
     try {
       store.clear();
+      //if(store.iterator().hasNext()) {  //TODO the clusteredStore Iterator does not exist!
```
lorban: Don't leave those comments in.

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -92,7 +92,7 @@
   private final Jsr107CacheImpl jsr107Cache;
   protected final Logger logger;
 
-  private final OperationObserver<GetOutcome> getObserver = operation(GetOutcome.class).named("get").of(this).tag("cache").build();
+  private final OperationObserver<org.ehcache.core.statistics.CacheOperationOutcomes.GetOutcome> getObserver = operation(org.ehcache.core.statistics.CacheOperationOutcomes.GetOutcome.class).named("get").of(this).tag("cache").build();
```
lorban: Why getting rid of the `GetOutcome` import?

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
lorban: IIRC when a stat collection throws, it is ignored and no sample ends up being collected. That basically leaves you with a hole in your history.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1740,15 +1837,11 @@ public void initHigherCachingTier(HigherCachingTier<?, ?> resource) {
     }
   }
 
+  //added @SuppressFBWarnings to get rid of findBugs error which recommended making instance variables static
+  @SuppressFBWarnings
   private static final class OnHeapStoreStatsSettings {
```
lorban: That class should also go away.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -239,32 +245,68 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
...
+    Set<String> tags = new HashSet<String>(Arrays.asList(STATISTICS_TAG, "tier"));
+    Map<String, Object> properties = new HashMap<String, Object>();
+    properties.put("discriminator", STATISTICS_TAG);
+    StatisticsManager.createPassThroughStatistic(this, "mappings", tags, properties, new Callable<Number>() {
       @Override
       public Number call() throws Exception {
         return map.mappingCount();
       }
     });
+    StatisticsManager.createPassThroughStatistic(this, "maxMappings", tags, properties, new Callable<Number>() {
```
lorban: `maxMappings` is probably not needed as this can be accessed from the config. Tip for the M&M folks: the configs *should* be historized too.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -239,32 +245,68 @@ public OnHeapStore(final Configuration<K, V> config, final TimeSource timeSource
...
+      @Override
+      public Number call() throws Exception {
+        if (byteSized) {
+          return -1L;
+        } else {
+          return capacity;
+        }
+      }
+    });
+    StatisticsManager.createPassThroughStatistic(this, "allocatedMemory", tags, properties, new Callable<Number>() {
```
lorban: Same as above: `allocatedMemory` actually is a config setting that shouldn't be exposed via the stats mechanism.

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -87,6 +92,9 @@ public void cachemanagerExample() {
     myCache.put(1L, "da one!"); // <7>
     String value = myCache.get(1L); // <8>
 
+    System.out.println(StatisticsManager.nodeFor(myCache).toTreeString());
```
lorban: No `System.out.println` please.

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -120,6 +128,10 @@ public void offheapCacheManager() {
             )
         .build(true);
 
+    Cache<Long, String> tieredCache = cacheManager.getCache("tieredCache", Long.class, String.class);
+
+    System.out.println(StatisticsManager.nodeFor(tieredCache).toTreeString());
```
lorban: No `System.out.println` please.

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -138,6 +150,11 @@ public void threeTiersCacheManager() throws Exception {
                 )
         ).build(true);
 
+    Cache<Long, String> threeTieredCache = persistentCacheManager.getCache("threeTieredCache", Long.class, String.class);
+
+    System.out.println(StatisticsManager.nodeFor(threeTieredCache).toTreeString());
```
lorban: No `System.out.println` please.

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -98,6 +102,38 @@
...
+        .withCache("cache-1", CacheConfigurationBuilder.newCacheConfigurationBuilder(
+            String.class, String.class,
+            newResourcePoolsBuilder()
+                .heap(10, EntryUnit.ENTRIES)
+                .with(ClusteredResourcePoolBuilder.clusteredDedicated("primary-server-resource", 1, MemoryUnit.MB)))
+            .build())
+        .build(true);
+
+    assertThat(managementRegistry.getCapabilities(), hasSize(5));
+    assertThat(new ArrayList<>(managementRegistry.getCapabilities()).get(0).getName(), equalTo("ActionsCapability"));
```
lorban: `new ArrayList<>` is JDK 8+ only and this source code is supposed to be 1.6 compatible.

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
```diff
@@ -0,0 +1,73 @@
...
+import org.terracotta.management.model.stats.AbstractStatisticHistory;
+
+/**
+ *
+ *
+ */
+public class StatsUtil {
+
+  public static boolean isHistoryReady(AbstractStatisticHistory counterHistory, Double defaultValue)
+  {
```
lorban: Watch out with those curly braces, they should be on the same line as the method declaration.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteredStoreConfiguration.java
```diff
@@ -25,13 +25,17 @@
...
+
   private final Consistency consistency;
+  private final int concurrency;
 
   /**
    * Creates a new configuration with consistency set to {@link Consistency#EVENTUAL EVENTUAL}.
    */
   public ClusteredStoreConfiguration() {
-    this(Consistency.EVENTUAL);
+    this(DEFAULT_CONSISTENCY, DEFAULT_CONCURRENCY);
```
ljacomet: This change and all related ones are irrelevant for this PR and should not be part of it. Can they be removed?

core/src/main/java/org/ehcache/core/Ehcache.java
gibsong: sorry that is a reminder for me.  I will clean up all //TODO's comments

core/src/main/java/org/ehcache/core/Ehcache.java
gibsong: no reason, I will add back the import

impl/src/test/java/org/ehcache/docs/GettingStarted.java
gibsong: just deleted it

impl/src/test/java/org/ehcache/docs/GettingStarted.java
gibsong: just deleted it

impl/src/test/java/org/ehcache/docs/GettingStarted.java
gibsong: just deleted it

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
gibsong: i just updated and put in the appropriate class

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
gibsong: shoot i thought I had fixed all those.  I just updated it to the correct format.

clustered/integration-test/src/test/java/org/ehcache/Clust.java
gibsong: ok, just deleted it.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
gibsong: just reverted to:

  private static final int OFFHEAP_CHAIN_SEGMENTS = 16;

  private final OffHeapServerStore store;

  public ServerStoreImpl(ServerStoreConfiguration storeConfiguration, PageSource pageSource) {
    this.storeConfiguration = storeConfiguration;
    this.pageSource = pageSource;
    this.store = new OffHeapServerStore(pageSource, OFFHEAP_CHAIN_SEGMENTS);
  }

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
gibsong: i removed this stat

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
gibsong: i removed this stat

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
gibsong: if I remove the ClusteredStoreStatsSettings class what is the best way to call associate()?
line 119:
StatisticsManager.associate(clusteredStoreStatsSettings).withParent(this);

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteredStoreConfiguration.java
gibsong: let me look into pulling these changes out.

AbfrmBlr:
I think couple of commits can be squashed as they have little meaning like cleanup, q&d and compilation related.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
gibsong: i removed this inner class

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
gibsong: i removed this inner class

mathieucarbou:
please do not merge yet: there's a test that has been `@Ignore`d: I'll investigate first the reason because it only fails with this PR...

107/src/main/java/org/ehcache/jsr107/Eh107CacheStatisticsMXBean.java
```diff
@@ -204,42 +202,39 @@ private static float normalize(float value) {
...
+        .descendants()
         .filter(context(attributes(Matchers.<Map<String, Object>>allOf(hasAttribute("name", statName), hasAttribute("type", type)))))
         .build();
 
-    Set<TreeNode> statResult = statQuery.execute(Collections.singleton(StatisticsManager.nodeFor(authoritativeTier)));
-    if (statResult.size() > 1) {
-      throw new RuntimeException("stat result must be unique");
+    Set<TreeNode> statResult = statQuery.execute(Collections.singleton(ContextManager.nodeFor(cache)));
+
+    if(statResult.size() < 1 || statResult.size() > 3) {
```
ljacomet: Why the upper bound? I am not sure it brings us anything, as long as we have one eviction stat, we can find the proper one.
And not having the upper bound means less code to change in case we end up allowing more than 3 tiers.

build.gradle
ljacomet: Now that you have cleaned up associations that were made in the different stores, I would be interested in what happens if you revert back to 512m

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -552,8 +621,9 @@ public void setInvalidationValve(InvalidationValve valve) {
 
 
       ClusteredStore<K, V> store = new ClusteredStore<K, V>(codec, resolver, timeSource);
-      StatisticsManager.associate(store.clusteredStoreStatsSettings).withParent(store);
+
       createdStores.put(store, new StoreConfig(cacheId, storeConfig, clusteredStoreConfiguration.getConsistency()));
+
```
ljacomet: These blank lines do not add anything, let's get rid of them

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -582,15 +651,19 @@ public void initStore(final Store<?, ?> resource) {
       clusteredStore.storeProxy.addInvalidationListener(new ServerStoreProxy.InvalidationListener() {
         @Override
         public void onInvalidateHash(long hash) {
+          Enum result = StoreOperationOutcomes.EvictionOutcome.SUCCESS;
+          clusteredStore.evictionObserver.begin();
```
ljacomet: That feels _wrong_. There is technically no local eviction for the clustered tier. What happens with the invalidation valve will trigger eviction stats in the upper tier. So by accounting them as eviction in this tier as well we are double counting. I would remove the stats interactions here.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/ServerStoreConfiguration.java
```diff
@@ -125,7 +125,7 @@ public boolean isCompatible(ServerStoreConfiguration otherConfiguration, StringB
     isCompatible &= compareField(sb, "actualValueType", actualValueType, otherConfiguration.getActualValueType());
     isCompatible &= compareField(sb, "keySerializerType", keySerializerType, otherConfiguration.getKeySerializerType());
     isCompatible &= compareField(sb, "valueSerializerType", valueSerializerType, otherConfiguration.getValueSerializerType());
-    isCompatible &= compareConsistencyField(sb, consistency, otherConfiguration.getConsistency());
+    isCompatible &= compareField(sb, "consistencyType", consistency, otherConfiguration.getConsistency());
```
ljacomet: Unrelated change - revert please

clustered/common/src/main/java/org/ehcache/clustered/common/internal/ServerStoreConfiguration.java
```diff
@@ -158,7 +158,7 @@ private boolean compareConsistencyField(StringBuilder sb, Consistency serverCons
     return false;
   }
 
-  private boolean compareField(StringBuilder sb, String fieldName, String serverConfigValue, String clientConfigValue) {
+  private boolean compareField(StringBuilder sb, String fieldName, Object serverConfigValue, Object clientConfigValue) {
```
ljacomet: Unrelated change - revert please

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -168,17 +171,17 @@ public V get(final K key) {
 
       // Check for expiry first
       if (valueHolder == null) {
-        getObserver.end(GetOutcome.MISS_NO_LOADER);
+        getObserver.end(org.ehcache.core.statistics.CacheOperationOutcomes.GetOutcome.MISS_NO_LOADER);
```
ljacomet: Revert this change - having the FQN here degrades readability

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
```diff
@@ -0,0 +1,231 @@
...
+import static org.terracotta.context.query.Matchers.identifier;
+import static org.terracotta.context.query.Matchers.subclassOf;
+import static org.terracotta.context.query.Queries.self;
+import static org.terracotta.context.query.QueryBuilder.queryBuilder;
+
+/**
+ * @author Ludovic Orban
+ */
+@ContextAttribute("this")
+public class TierOperationStatistic<S extends Enum<S>, D extends Enum<D>> implements OperationStatistic<D> {
```
ljacomet: Is this really a type that a user will have to interact with? Or should it leave inside an `internal` package?
To be clear, if only the `management` and `107` models are going to need it, that is `internal`.

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
```diff
@@ -0,0 +1,231 @@
...
+import static org.terracotta.context.query.Matchers.attributes;
+import static org.terracotta.context.query.Matchers.context;
+import static org.terracotta.context.query.Matchers.hasAttribute;
+import static org.terracotta.context.query.Matchers.identifier;
+import static org.terracotta.context.query.Matchers.subclassOf;
+import static org.terracotta.context.query.Queries.self;
+import static org.terracotta.context.query.QueryBuilder.queryBuilder;
+
+/**
+ * @author Ludovic Orban
```
ljacomet: We agreed to not have `@author` tags - remove please

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
```diff
@@ -0,0 +1,231 @@
...
+    return (OperationStatistic) node.getContext().attributes().get("this");
+  }
+
+  public static <X> Set<X> set(X... xs) {
+    return new HashSet<X>(Arrays.asList(xs));
+  }
+
+  public static class TierOperationOutcomes {
+
+    public enum GetOutcome {
```
ljacomet: `StoreOperationOutcomes.GetOutcome` has a `TIMEOUT` value - is that not needed here then?

impl/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -138,6 +145,9 @@ public void threeTiersCacheManager() throws Exception {
                 )
         ).build(true);
 
+    Cache<Long, String> threeTieredCache = persistentCacheManager.getCache("threeTieredCache", Long.class, String.class);
```
ljacomet: Why this change and the one above?

management/src/test/java/org/ehcache/management/registry/DefaultManagementRegistryServiceTest.java
```diff
@@ -74,6 +90,88 @@ public void testCanGetContext() {
   }
 
   @Test
+  public void descriptorOnHeapTest()
+  {
```
ljacomet: Code formatting

107/src/main/java/org/ehcache/jsr107/Eh107CacheStatisticsMXBean.java
gibsong: i can take this out then.  I didn't realize we might allow more than 3 tiers in the future.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
gibsong: will delete

clustered/common/src/main/java/org/ehcache/clustered/common/internal/ServerStoreConfiguration.java
gibsong: i will revert the change

clustered/common/src/main/java/org/ehcache/clustered/common/internal/ServerStoreConfiguration.java
gibsong: i will revert the change

core/src/main/java/org/ehcache/core/Ehcache.java
gibsong: yes i agree, sorry must have missed these ones.  I will fix.

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
gibsong: will remove

impl/src/test/java/org/ehcache/docs/GettingStarted.java
gibsong: I removed 3 System.out calls in this class.  Each System.out was right after the getCache() call.  In addition, the getCache() call was already in Ludovics code so I assumed it was there as an example of how to create a cache(its in the GettingStarted class).

management/src/test/java/org/ehcache/management/registry/DefaultManagementRegistryServiceTest.java
gibsong: will fix all of them.  Saw a few more in this class

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
gibsong: This class is heavily used by :impl and  :clustered:client.  So I assume it shouldn't be moved?

mathieucarbou:
@gibsong : I opened a PR to rework the clustered management test first so that you could rebase onto after: #1414.

Basically, the test __IS__ reliable as long as you are using counters for stats, which is the thing that changed in your PR.

So I propose that we merge #1414 first so that it is cleaner, then you'll be able to rebase on master and just change some little parts of the test, more specifically only the stat names if needed and the descriptor count. The idea is to not change the behavior of the test ;-)

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
gibsong: hmm... let me double check that in a test.  I actually didn't add that code.  Not that I don't trust you but I also need to do some clustered eviction testing so this will force me to do it now.  Where can I create a clustered test that uses management stats?  I tried in :clustered:client but it doesn't have a management dependency.

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
gibsong: i need to double check this one.  I didn't think it was actually needed but let me make sure.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
gibsong: I talked to Mathieu about creating some clustered eviction tests in :clustered:integration-test but at the moment there is an issue (Terracotta-OSS/terracotta-core#297) preventing use of the management module.  Once this is fixed I can proceed.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1624,6 +1642,7 @@ private static boolean eq(Object o1, Object o2) {
 
     private volatile ServiceProvider<Service> serviceProvider;
     private final Map<Store<?, ?>, List<Copier>> createdStores = new ConcurrentWeakIdentityHashMap<Store<?, ?>, List<Copier>>();
+    private final Map<OnHeapStore<?, ?>, Collection<TierOperationStatistic<?, ?>>> tierOperationStatistics = new ConcurrentWeakIdentityHashMap<OnHeapStore<?, ?>, Collection<TierOperationStatistic<?, ?>>>();
```
ljacomet: Whatever is being put in this map is never removed. And from the previous consequence of registering associations with the `StatisticsManager` I wonder if the operations contained in the value do not need to be de-associated.
This comment applies to all `Store.Provider` implementations that now have a `tierOperationStatistics` field.

mathieucarbou:
Note: will need to be rebased also.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
gibsong: i am looking into it.  I think you may be right.

ljacomet: Pretty sure that code was needed - see #1328 - it was just missing the `dissociate` calls.

core/src/main/java/org/ehcache/core/statistics/TierOperationStatistic.java
gibsong: I did some digging and also chatted with Ludovic to confirm that the TierOperationOutcomes.GetOutcome.TIMEOUT value isn't actually needed.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
gibsong: I discussed more in detail with Louis.  I will remove this code.

mathieucarbou:
@gibsong : I think you can talk to @ljacomet about ehcache rules to be sure, but I think you would like, instead of having 3 commits with weird messages, to have instead just 1 commit saying for example:
`Closes #1286 : New Ehcache3 statistic support` (or something similar)
(Commit messages must refer the issue the PR closes and had better to be short: that helps when looking back at the history)
I am not sure `q&d` and "combinations" would be accepted ;-)

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1640,7 +1658,21 @@ public int rankCachingTier(Set<ResourceType<?>> resourceTypes, Collection<Servic
...
+      OnHeapStore<K, V> store = createStoreInternal(storeConfig, new ScopedStoreEventDispatcher<K, V>(storeConfig.getDispatcherConcurrency()), serviceConfigs);
+
+      TierOperationStatistic<StoreOperationOutcomes.GetOutcome, TierOperationStatistic.TierOperationOutcomes.GetOutcome> get = new TierOperationStatistic<StoreOperationOutcomes.GetOutcome, TierOperationStatistic.TierOperationOutcomes.GetOutcome>(TierOperationStatistic.TierOperationOutcomes.GetOutcome.class, StoreOperationOutcomes.GetOutcome.class, store, new HashMap<TierOperationStatistic.TierOperationOutcomes.GetOutcome, Set<StoreOperationOutcomes.GetOutcome>>() {{
+        put(TierOperationStatistic.TierOperationOutcomes.GetOutcome.HIT, set(StoreOperationOutcomes.GetOutcome.HIT));
+        put(TierOperationStatistic.TierOperationOutcomes.GetOutcome.MISS, set(StoreOperationOutcomes.GetOutcome.MISS, StoreOperationOutcomes.GetOutcome.TIMEOUT));
+      }}, "get", ResourceType.Core.HEAP.getTierHeight(), "get", STATISTICS_TAG);
+      StatisticsManager.associate(get).withParent(store);
```
ljacomet: The remaining tests problems are caused by what happens here. The `TierOperationStatistic` no longer has any strong reference to it after this line.
And thus it will end up being cleaned up from the statistics tree.

The only way out that I can see it to re-introduce a reference to it. Either by expanding the value type in `createdStores` or by having a dedicated map for that - like it was done before.

I would not change the clean up though - keep it general instead of focused on these stats.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -448,7 +471,21 @@ public void stop() {
...
+
+      TierOperationStatistic<AuthoritativeTierOperationOutcomes.GetAndFaultOutcome, TierOperationStatistic.TierOperationOutcomes.GetOutcome> get = new TierOperationStatistic<AuthoritativeTierOperationOutcomes.GetAndFaultOutcome, TierOperationStatistic.TierOperationOutcomes.GetOutcome>(TierOperationStatistic.TierOperationOutcomes.GetOutcome.class, AuthoritativeTierOperationOutcomes.GetAndFaultOutcome.class, authoritativeTier, new HashMap<TierOperationStatistic.TierOperationOutcomes.GetOutcome, Set<AuthoritativeTierOperationOutcomes.GetAndFaultOutcome>>() {{
+        put(TierOperationStatistic.TierOperationOutcomes.GetOutcome.HIT, set(AuthoritativeTierOperationOutcomes.GetAndFaultOutcome.HIT));
+        put(TierOperationStatistic.TierOperationOutcomes.GetOutcome.MISS, set(AuthoritativeTierOperationOutcomes.GetAndFaultOutcome.MISS, AuthoritativeTierOperationOutcomes.GetAndFaultOutcome.TIMEOUT));
+      }}, "get", ResourceType.Core.DISK.getTierHeight(), "getAndFault", STATISTICS_TAG);
+      StatisticsManager.associate(get).withParent(authoritativeTier);
```
ljacomet: See comment on `OnHeapStore`

impl/src/main/java/org/ehcache/impl/internal/store/offheap/OffHeapStore.java
```diff
@@ -141,7 +151,21 @@ public int rankAuthority(ResourceType<?> authorityResource, Collection<ServiceCo
...
+      OffHeapStore<K, V> store = createStoreInternal(storeConfig, new ThreadLocalStoreEventDispatcher<K, V>(storeConfig.getDispatcherConcurrency()), serviceConfigs);
+
+      TierOperationStatistic<StoreOperationOutcomes.GetOutcome, TierOperationStatistic.TierOperationOutcomes.GetOutcome> get = new TierOperationStatistic<StoreOperationOutcomes.GetOutcome, TierOperationStatistic.TierOperationOutcomes.GetOutcome>(TierOperationStatistic.TierOperationOutcomes.GetOutcome.class, StoreOperationOutcomes.GetOutcome.class, store, new HashMap<TierOperationStatistic.TierOperationOutcomes.GetOutcome, Set<StoreOperationOutcomes.GetOutcome>>() {{
+        put(TierOperationStatistic.TierOperationOutcomes.GetOutcome.HIT, set(StoreOperationOutcomes.GetOutcome.HIT));
+        put(TierOperationStatistic.TierOperationOutcomes.GetOutcome.MISS, set(StoreOperationOutcomes.GetOutcome.MISS, StoreOperationOutcomes.GetOutcome.TIMEOUT));
+      }}, "get", ResourceType.Core.OFFHEAP.getTierHeight(), "get", STATISTICS_TAG);
+      StatisticsManager.associate(get).withParent(store);
```
ljacomet: See comment on `OnHeapStore`

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1678,6 +1715,8 @@ public void releaseStore(Store<?, ?> resource) {
           throw new IllegalStateException("Exception while releasing Copier instance.", e);
         }
       }
+      StatisticsManager.nodeFor(onHeapStore).clean();
```
ljacomet: because the code above can throw, this line and the following should be before the block releasing `Copier`s

ljacomet:
According to my reading - the remaining failing tests could be racy or caused by side effects.
test this please

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
gibsong: nice catch.  I moved the stats clean up to before the releasing of Copier s

Issue #1208, opened 2016-06-23
Implement skeletal active-passive replication


chrisdennis:
First stage in a functional active passive system is getting the lifecycle events (cache creation/configuration and deletion) correctly replicated to the passive.  This can probably be done by direct execution of the logical lifecycle operations on the passive entity.

chrisdennis:
Blockedby #1207

PR #1392, opened 2016-08-18
Issue 1208

AbfrmBlr:
Not to be merged yet


clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -247,7 +249,31 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
 
   @Override
   public void handleReconnect(ClientDescriptor clientDescriptor, byte[] extendedReconnectData) {
-    //nothing to do
+    ClientState clientState = this.clientStateMap.get(clientDescriptor);
+    if (clientState == null) {
+      LOGGER.error("Client {} trying to reconnect is not connected to entity.", clientDescriptor);
+      return;
```
ljacomet: I would throw an AssertionError here - this must be impossible from the platform's perspective!

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -247,7 +249,31 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
...
+    }
+    clientState.attach();
+    Set<String> cacheIds = reconnectDataCodec.decode(extendedReconnectData);
+    for (final String cacheId : cacheIds) {
+      ServerStoreImpl serverStore = ehcacheStateService.getStore(cacheId);
+      if (serverStore == null) {
+        //Client removes the cache's reference only when destroy has successfully completed
+        //This happens only when client thinks destroy is still not complete
```
ljacomet: I do not understand these comments

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
```diff
@@ -0,0 +1,172 @@
...
+        default:
+          throw new IllegalMessageException("Unknown message : " + message);
+      }
+    } catch (Exception e) {
+      LOGGER.error("Unexpected exception raised during operation: " + message, e);
+    }
+
+  }
+
+  private void invokeLifeCycleOperation(LifecycleMessage message) throws ClusterException{
```
ljacomet: I have the impression that `invoke` and this method could be pulled into a super abstract type that passive and active version of the entity would implement.

ljacomet: Forget what I said ... unless we hide the difference in parameters between a message with or without a `ClientDescriptor`

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
ljacomet: Got precision from @AbfrmBlr - given that, the following log statement should be `warn` level.

AbfrmBlr:
It will good if this can be merged. This does not close #1208. This is a partial PR, following PRs shall contain message de-duplication and refactoring around passthrough replication testing in `UnitTestConnectionService`.

Issue #1396, opened 2016-08-22
Eh107MBeanServerBuilder uses internal JDK types to create an MBean server instance


chrisdennis:
Eh107MBeanServerBuilder currently uses a static method on `com.sun.jmx.mbeanserver.JmxMBeanServer` to creater an MBean server.  Instead we should use an instance of `MBeanServerBuilder`.

PR #1400, opened 2016-08-22
Issue #1396 [master] Use public API for creating MBean servers.

chrisdennis:


107/src/main/java/org/ehcache/jsr107/internal/tck/Eh107MBeanServerBuilder.java
```diff
@@ -42,7 +40,7 @@ public Eh107MBeanServerBuilder() {
   @Override
   public MBeanServer newMBeanServer(String defaultDomain, MBeanServer outer, MBeanServerDelegate delegate) {
     MBeanServerDelegate decoratingDelegate = new Eh107MBeanServerDelegate(delegate);
-    return JmxMBeanServer.newMBeanServer(defaultDomain, outer, decoratingDelegate, false);
```
henri-tremblay: Why not `return super.newMBeanServer(defaultDomain, outer, decoratingDelegate)` since it extends MBeanServerBuilder already?

chrisdennis: No reason beyond not paying attention... fixing now.

Issue #1397, opened 2016-08-22
The serializer tests incorrectly assume a classloader is a URLClassLoader


chrisdennis:
Some of the serializer tests hard cast the classloader of the test classes to a URLClassLoader.  This isn't guaranteed to work (and doesn't in JDK9).  We should replace this with something that doesn't make this assumption.

PR #1401, opened 2016-08-22
Issue #1397 [master] Use duplicating classloader

chrisdennis:


henri-tremblay:
I'm missing something. Why do you need a new class loader instead of using the URLClassLoader like before?

chrisdennis:
What I was doing before was taking advantage of the fact that the application classloader was always a subclass of UrlClassLoader and using this fact to easily create a clone of it and therefore create duplicate copies of the regular classes that could be GCed in order to check that the serializer wouldn't block class unloading.  In JDK 9 the application classloader isn't a subclass of URLClassLoader any more (because of Jigsaw) so I have to create the 'cloning' classloader in a more generic way.

impl/src/test/java/org/ehcache/impl/serialization/DuplicateClassLoader.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override
+  protected Class<?> findClass(String name) throws ClassNotFoundException {
+    InputStream resource = getResourceAsStream(name.replace('.', '/') + ".class");
+    if (resource == null) {
+      throw new ClassNotFoundException(name);
+    }
+    try {
+      ByteArrayOutputStream bout = new ByteArrayOutputStream();
+      while (true) {
```
henri-tremblay: `bout` is never closed here (I know it doesn't really matter but things like findbugs don't like that)

Issue #1399, opened 2016-08-22
ByteAccountingTest assumes that simple ascii characters use 2 bytes in a String


chrisdennis:
ByteAccountingTest uses simple ascii strings and assumes they occupy 2n bytes.  In JDK9 there are string storage optimizations that defeat this and break the test.  We should consider moving to using extended characters to defeat the optimization and prevent the test from being JDK version sensitive.

PR #1403, opened 2016-08-22
Issue #1399 [master] use uncompactable non-ascii character to defeat JDK 9 string optimizations

chrisdennis:


henri-tremblay:
Again, I'm curious. Why do you need to fill it with non-ascii chars?

chrisdennis:
JDK 9 optimizes string storage such that ascii strings occupy one byte per character and not two.  This test however is trying to assert things about byte sized caches and their eviction behavior.  This means I need a string that is the same size "on-heap" in all JDK versions.

PR #1406, opened 2016-08-25
Add editorconfig (http://editorconfig.org) to define a standard indentation

henri-tremblay:


chrisdennis:
This looks fine and the settings are all inline with what checkstyle currently enforces.  It bothers me however that the same set of information is now captured incompletely in two separate places.  Is there not a tool out there that can both drive IDE configurations and also provide build time enforcement?

henri-tremblay:
None that I'm aware of. Sadly. Editorconfig is supported by IntelliJ, Eclipse (with a plugin I think) and most text editor (like sublime). But it's not a full-fledged formater. So, in fact, we would need a complete ehcache formater (Eclipse file is good since it can be imported in IntelliJ). But editorconfig is a good start to be able to work on the project.

chrisdennis:
Seems like this can't hurt, so lets merge it.

PR #1412, opened 2016-08-30
Clean some "unchecked" warnings

henri-tremblay:


core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -426,7 +427,7 @@ public void loadDependenciesOf(Class<?> clazz) {
             throw new IllegalStateException("Service dependency declared by " + clazz.getName() +
                 " is not a Service: " + dependency.getName());
           }
-          dependencies.add(dependency);
+          dependencies.add((Class<? extends Service>) dependency);
```
chrisdennis: Had a quick run through the product code (don't really care about the tests as much) - only problem I could see was that this cast could be eliminated if we type the for loop variable properly.

ljacomet:
Unhappy compiler 😉

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
henri-tremblay: There is a check to see if the class is actually an instance of `Service`. So it doesn't make sense to type it as a `(Class<? extends Service>)`.

However, since the annotation is now typed, we probably can remove all these checks. But I've left that to another time.

henri-tremblay:
Ok, everything is passing now. Should be good to go

PR #1414, opened 2016-09-01
Rewriting ClusteredManagementServiceTest

mathieucarbou:
This PR is just a rewriting of a big integration test in the management module.

It will help also #1390.

The idea is:

1. split the test in several smaller ones (which is done in this PR)
2. Separate the passthrough part from the tests (see below for reason)

Then, after:

1. Rebase #1390 on top of this PR so that #1390 uses this new test: #1390 will just need to adapt the statistic part, IF needed.
2. Move this test in the clusterd-it module (which will be done after because it requires to get rid of the passthrough things, use galvan instead, use the new release of tc-platform, and also requires that an issue in platform is fixed first: Terracotta-OSS/terracotta-core#297)

### DIFF
--- a/management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatistics.java
+++ b/management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatistics.java
@@ -74,7 +74,6 @@ class EhcacheStatistics extends ExposedCacheBinding {
   private static final Set<CacheOperationOutcomes.RemoveOutcome> ALL_CACHE_REMOVE_OUTCOMES = EnumSet.allOf(CacheOperationOutcomes.RemoveOutcome.class);
   private static final Set<CacheOperationOutcomes.GetOutcome> GET_WITH_LOADER_OUTCOMES = EnumSet.of(CacheOperationOutcomes.GetOutcome.HIT_WITH_LOADER, CacheOperationOutcomes.GetOutcome.MISS_WITH_LOADER);
   private static final Set<CacheOperationOutcomes.GetOutcome> GET_NO_LOADER_OUTCOMES = EnumSet.of(CacheOperationOutcomes.GetOutcome.HIT_NO_LOADER, CacheOperationOutcomes.GetOutcome.MISS_NO_LOADER);
-  private static final Set<CacheOperationOutcomes.CacheLoadingOutcome> ALL_CACHE_LOADER_OUTCOMES = EnumSet.allOf(CacheOperationOutcomes.CacheLoadingOutcome.class);
 
   private final StatisticsRegistry statisticsRegistry;
   private final Map<String, OperationStatistic<?>> countStatistics;
--- a/management/src/test/java/org/ehcache/management/cluster/AbstractClusteringManagementTest.java
+++ b/management/src/test/java/org/ehcache/management/cluster/AbstractClusteringManagementTest.java
@@ -1,0 +1,189 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.management.cluster;
+
+import org.ehcache.clustered.client.internal.EhcacheClientEntityService;
+import org.ehcache.clustered.client.internal.lock.VoltronReadWriteLockEntityClientService;
+import org.ehcache.clustered.lock.server.VoltronReadWriteLockServerEntityService;
+import org.ehcache.clustered.server.EhcacheServerEntityService;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.terracotta.connection.Connection;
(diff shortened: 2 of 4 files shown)


mathieucarbou:
test this please

gibsong:
Mathieu, thanks for putting this together so quickly.  I am pretty sure Ludovic's changes removed the ability to query by Counter(primitive counter), and the only option left now is to use CounterHistory(samples).  So unfortunately that is why I had to change all the tests that used primitive counters to samples.  I just sent Ludovic an email to double check my understanding.

management/src/test/java/org/ehcache/management/cluster/AbstractClusteringManagementTest.java
```diff
@@ -0,0 +1,189 @@
...
+      // now we're sure the call completed
+      callCompleted.countDown();
+
+      // ensure the call is made
+      returns.take();
+    }
+  }
+
+  protected static ContextualStatistics[] waitForNextStats() {
+    // uses the monitoring consuemr entity to get the content of the stat buffer when some stats are collected
```
anthonydahanne: typo : consumer

anthonydahanne:
I really like the smaller test methods that each test a management feature, that's real nice to read and use.
But I regret the simplicity of the former test - I'm not sure why now some methods are in the abstract some others in the concrete class; the use of the Parameters makes the code harder to read I think too.
I'm more used to have, for unit tests to be run under different conditions, 1 abstract class that defines all the small test methods and several concrete classes that define a different behavior (programmatic vs xml) to be tested.
But anyway, maybe that's just me - the most important thing is the test being split in small increments, so LGTM

mathieucarbou:
@anthonydahanne : for the abstract class, I explained in the PR comment: See:  _Separate the passthrough part from the tests (see below for reason)_ This is to prepare to migrate this test to galvan. Only the abstract part should change.

The parameters are there to keep the same behavior as before, which is that we applies the same test methods against both CM, creates from programmatic api and XML. This ensures all the required services are initialized the same way, and, in the case of XML, makes sur the teh clustered management service gets __auto initialized__. This is the important part.

management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatistics.java
```diff
@@ -74,7 +74,6 @@
   private static final Set<CacheOperationOutcomes.RemoveOutcome> ALL_CACHE_REMOVE_OUTCOMES = EnumSet.allOf(CacheOperationOutcomes.RemoveOutcome.class);
   private static final Set<CacheOperationOutcomes.GetOutcome> GET_WITH_LOADER_OUTCOMES = EnumSet.of(CacheOperationOutcomes.GetOutcome.HIT_WITH_LOADER, CacheOperationOutcomes.GetOutcome.MISS_WITH_LOADER);
   private static final Set<CacheOperationOutcomes.GetOutcome> GET_NO_LOADER_OUTCOMES = EnumSet.of(CacheOperationOutcomes.GetOutcome.HIT_NO_LOADER, CacheOperationOutcomes.GetOutcome.MISS_NO_LOADER);
-  private static final Set<CacheOperationOutcomes.CacheLoadingOutcome> ALL_CACHE_LOADER_OUTCOMES = EnumSet.allOf(CacheOperationOutcomes.CacheLoadingOutcome.class);
```
mathieucarbou: this was unused code.

management/src/test/java/org/ehcache/management/cluster/AbstractClusteringManagementTest.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import static java.util.Arrays.asList;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+public abstract class AbstractClusteringManagementTest {
```
mathieucarbou: FYI: handles the creation of a cluster with management entities / services.

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -32,299 +29,224 @@
...
-          new Parameter(asList("PutCounter", "InexistingRate"), Collection.class.getName())));
-
-      // now we're sure the call completed
-      callCompleted.countDown();
-
-      // ensure the call is made
-      returns.take();
-    }
-  }
+    System.out.println(Arrays.toString(stats));
```
AbfrmBlr: sysout ??

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -32,299 +29,224 @@
...
-    if (cacheManager != null && cacheManager.getStatus() == Status.AVAILABLE) {
-      cacheManager.close();
-    }
+    // do some other operations
+    cache1.get("key1");
+    cache1.get("key2");
+
+    stats = waitForNextStats();
+
+    System.out.println(Arrays.toString(stats));
```
AbfrmBlr: sysout??

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
mathieucarbou: oh sorry! i forgot them thanks!

management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
mathieucarbou: again :P

ljacomet:
Looks better - let's merge once  rebased 😉

PR #1419, opened 2016-09-06
Fix unreliable test : there can be several samples

anthonydahanne:


management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java
```diff
@@ -241,7 +242,7 @@ public void test_stats_collection() throws Exception {
     assertThat(stats[0].getContext().get("cacheName"), equalTo("cache-1"));
     assertThat(stats[0].getStatistic(Counter.class, "GetCounter").getValue(), equalTo(4L));
     assertThat(stats[0].getContext().get("cacheName"), equalTo("cache-1"));
-    assertThat(stats[0].getStatistic(CounterHistory.class, "AllCacheGetCount").getValue().length, equalTo(1));
+    assertThat(stats[0].getStatistic(CounterHistory.class, "AllCacheGetCount").getValue().length, greaterThanOrEqualTo(1));
```
chrisdennis: If there's more than one sample, why are you asserting that the first sample is 4, as opposed to the last sample, or any random sample?

anthonydahanne: because it's not the same stat : GetCounter is a counter (not even sampled), and either you call it once or several times, it should always return the same value.

mathieucarbou: Also it is because we are just interested in the exact next stat that has been computed. Not the eventual extra one. The other one can or cannot be there, depending if the scheduler had some time to compute it. Anyway, we do not care about this other one. Only the first next one.

chrisdennis: If there is more than one history sample what guarantees that the *first* sample contains the right value?  What prevents the scheduling from taking the first sample before the counter hits 4?

gibsong: I have actually seen this scenario happen with history samples.  You should make sure you grab the highest index from the array to ensure you get the most recent value.  You also need to wait for the value to initialized.  For instance, if the default value is 0, wait till its not 0 then grab it.  U need a check like belos (which is in a utility in my branch):

public static boolean isHistoryReady(AbstractStatisticHistory counterHistory, Long defaultValue) {
    if(counterHistory.getValue().length > 0) {
      int mostRecentIndex = counterHistory.getValue().length - 1;
      if((Long)counterHistory.getValue()[mostRecentIndex].getValue() > defaultValue) {
        return true;
      }
    }
    return false;
  }

mathieucarbou: It's ok ;-) we are still in master ;-) It's true for @gibsong's PR. 

But do not forget that in our test we only want to be sure stats goes through the whole voltron loop. We do not care about testing stats.

Also, do not forget that:

- computation frequency is at each 1 sec
- stat collect frequency is at each 1.5 sec

So first time we wait and get the stats from voltron (line 233) we have no samples in AllCacheGetCount, the stat has been collected, started, thus 0 sample.

1.5 sec later, line 239, we collect the stats again. So it means we are >= 3 sec since the beginning. So we have at least one other stat that has been computed. Anyway we WAIT and read next collected stats.

Line 245, we make sure that we have at least 1 value in the sample. We could have 2, if the stat scheduler had time to compute for the 3rd time. But anyway, the value will be also 4 because there has been no  other "get" operations.

So line 246, whether we take the first value or the last one, we do not care: it will be the same value anyway.

mathieucarbou: In @gibsong's PR #1390, the behavior changes: when we first collect the a sampled stats, instead of having an empty sample that started the computation, then we have one stats.

So in @gibsong's PR #1390, we have, at the end, in the sample list, not 1 value but 2! And in this case, yes we need to take the most recent one!

ljacomet: @mathieucarbou I guess if it is always the same value, then the assertion could make sure all values are 4 - or at least one value is 4. The comment from @chrisdennis was that the fixed line felt weird compared to the check following it as the race could happen in different ways.

gibsong: sorry I didn't see waitForNextStats() in the github review pane.  I just read your previous comment and looked at the entire file, and it makes sense.  Also thanks for putting in those comments.

mathieucarbou: In master:

- __T__ : start collector scheduler
- __get operations__
- __T + 1 sec__ : collector scheduler:
  - collects stats (__=> sample computation starts now__)
  - send them to voltron
- __T + 1.5 sec__ : [waitForNextStats](https://github.com/anthonydahanne/ehcache3/blob/master/management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java#L224)
  - we have 1 stat and __no sample__
- __get operations__
- __T + 2 sec__ : collector scheduler runs
  - collects stats
  - send them to voltron
- __T + 3 sec__ : collector scheduler runs
  - collects stats
  - send them to voltron
- __T + 3 sec__ : [waitForNextStats] (https://github.com/anthonydahanne/ehcache3/blob/master/management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java#L239)
  - we have 1 stat and __1 or 2 sample values, depending whether the collector ran after or before this__ Values will all be 4.

In PR #1390:

- __T__ : start collector scheduler __AND sample computation starts now__
- __get operations__
- __T + 1 sec__ : collector scheduler:
  - collects stats
  - send them to voltron
- __T + 1.5 sec__ : [waitForNextStats](https://github.com/anthonydahanne/ehcache3/blob/master/management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java#L224)
  - we have 1 stat and __1 sample value that is 2__
- __get operations__
- __T + 2 sec__ : collector scheduler runs
  - collects stats
  - send them to voltron
- __T + 3 sec__ : collector scheduler runs
  - collects stats
  - send them to voltron
- __T + 3 sec__ : [waitForNextStats] (https://github.com/anthonydahanne/ehcache3/blob/master/management/src/test/java/org/ehcache/management/cluster/ClusteringManagementServiceTest.java#L239)
  - we have 1 stat and __2 or 3 sample values, depending whether the collector ran after or before this__ Values will be 2, 4 OR 2, 4, 4

ljacomet: @mathieucarbou Given the explanation above, and the original question, if the fix had contained a modification on line 246 (new version) so that the assertion verified that the _last value_ was `4`, it would have made sense.

mathieucarbou: yes it would have also made sense ;-) Like checking the first value makes also sense ;-) This particular case anyway is not anymore there is @gibsong's branch (that we need in M&M to complete our user story) so I think it's not worth changing it, and much more wait for the merge :-)

PR #1420, opened 2016-09-06
Correctly wrap the exception and close the file on error in DefaultLocalPersistenceService

henri-tremblay:


impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -17,21 +17,22 @@
 package org.ehcache.impl.persistence;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
+import org.ehcache.CachePersistenceException;
```
mathieucarbou: I think Ehcache people don't like import reorganization ;-)

ljacomet: Indeed, please configure the IDE so that it does not reorganise stuff you did not change - or adapt the rules so they mostly match what exists.

ljacomet: They were broken ... let's have the fix since this is a small change anyway

Issue #1410, opened 2016-08-30
PersistentCacheManager.destroyCache: unclear lifecycle


ljacomet:
In its current implementation `PersistentCacheManager.destroyCache ` does not check any lifecycle state and simply removes the cache if it exists and then invokes `PersistableResourceService.destroy` on all instances.

This works for disk based caches, as starting and stopping the file persistence service is irrelevant for destroy. Although there is a bug hiding there already since you could destroy data from a folder that is in use by another `CacheManager`.
But it is broken for clustered cache manages as the `ClusteringService` cannot do anything unless started.

So the implementation of `PersistentCacheManager.destroyCache ` needs to be adapted to fix this problem.

ljacomet:
My proposal to fix this:
Make the implementation of `destroyCache` state dependent:
* If the `CacheManager` is available, simply do what's done currently.
* If the `CacheManager` is uninitialised, start it for maintenance, destroy, exit maintenance.

Looking for other proposals or flaws in this one.

PR #1421, opened 2016-09-07
Issue 1410

henri-tremblay:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -171,6 +163,20 @@ public void start(final ServiceProvider<Service> serviceProvider) {
     }
   }
 
+  private void createEntityFactory() {entityFactory = new EhcacheClientEntityFactory(clusterConnection, operationTimeouts);}
```
ljacomet: I would rather have this on 3 lines

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -248,6 +248,8 @@ private PersistenceSpace createSpace(String name, boolean persistent) throws Cac
    */
   @Override
   public void destroy(String name) throws CachePersistenceException {
+    if(!started)
```
ljacomet: `{` and `}` please

impl/src/test/java/org/ehcache/impl/persistence/DefaultLocalPersistenceServiceTest.java
```diff
@@ -34,25 +35,34 @@
 public class DefaultLocalPersistenceServiceTest {
 
   @Rule
+  public final ExpectedException expectedException = ExpectedException.none();
```
ljacomet: I see a duplicate definition of this `expectedException` in the PR ... GitHub rendering issue or real mistake?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -248,6 +248,8 @@ private PersistenceSpace createSpace(String name, boolean persistent) throws Cac
...
+      internalStart();
```
ljacomet: This is missing the `stop` in case the service had to be started.

ljacomet:
Could you edit commit message so that they contain a reference to the issue. Format is #1410 (just as here).
You can also prefix that with _Fix_ or _Close_ to automatically close the issue when it gets merged to master.

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
GaryWKeim: And also "if ("

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -320,6 +339,15 @@ public void destroy(String name) throws CachePersistenceException {
       throw new CachePersistenceException("Could not destroy clustered tier '" + name + "' on " + clusterUri
           + "; destroy operation timed out" + clusterUri, e);
     }
+    finally {
```
ljacomet: needs to be on same line as above `}`

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -654,9 +654,27 @@ public void destroyCache(final String alias) throws CachePersistenceException {
...
+    try {
+      removeAndCloseWithoutNotice(alias);
+      destroyPersistenceSpace(alias);
+    } finally {
+      if(maintenance != null) {
+        statusTransitioner.exitMaintenance().succeeded();
+      }
+    }
+
+    // if was started, stop it
```
ljacomet: That looks like a copy pasted commented unrelated to the lines around it

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -654,9 +654,27 @@ public void destroyCache(final String alias) throws CachePersistenceException {
...
     LOGGER.debug("Destroying Cache '{}' in {}.", alias, simpleName);
-    removeAndCloseWithoutNotice(alias);
-    destroyPersistenceSpace(alias);
+
+    StatusTransitioner.Transition maintenance = null;
+    try {
+      maintenance = statusTransitioner.maintenance();
+      maintenance.succeeded();
+    } catch(IllegalStateException e) {
+      // the cache manager is already started, no need to put it in maintenance
```
ljacomet: Just realised this could mean it is in maintenance for another thread. What about having a `checkAvailable` here?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -248,6 +248,10 @@ private PersistenceSpace createSpace(String name, boolean persistent) throws Cac
    */
   @Override
   public void destroy(String name) throws CachePersistenceException {
+    if (!started) {
+      internalStart();
```
ljacomet: Still misses the corresponding stop if it had to be started.

ljacomet: And the corresponding test(s)

core/src/main/java/org/ehcache/core/EhcacheManager.java
henri-tremblay: But it still means that someone else might put it in maintenance right after the check. Right now, there is no way to prevent that.

ljacomet: Correct - it just makes sure that if we failed to put in maintenance because it is already in maintenance in another thread we do not blindly proceed. Makes sense?

henri-tremblay: Agreed

mathieucarbou:
it's cute now 😍

PR #1422, opened 2016-09-07
Closes #1411 Stateful contract for serializers

### LINKED ISSUE #1411: Lifecycle issue with Serializer and StateRepository
For clustering support, we changed the contract of `Serializer` that need to support persistent state by adding the `StateRepository` API.

However, we kept injecting this at `Serializer` construction time, which while working in practice, has several downsides:
* The `StateRepository` really belongs to the `AuthoritativeTier.Provider` when thinking about it, which is what gives guarantees to have the same persistence level.
* The `Serializer` really benefits from being created once and passed in to all `Store` so that it can be final.

From these two competing requirements, the current scheme is not working. We need to find another way of wiring the `StateRepository` inside the `Serializer`.

Sadly this means breaking the SPI so this can only be done for 3.2. And since we are breaking the SPI, should we consider moving away from exposing a `ConcurrentMap` to a custom type in the `StateRepository`?

albinsuresh:


api/src/main/java/org/ehcache/spi/serialization/Stateful.java
```diff
@@ -0,0 +1,34 @@
...
+package org.ehcache.spi.serialization;
+
+import org.ehcache.spi.persistence.StateRepository;
+
+/**
+ * Serializers implementing this interface can have persistent state.
+ * The state will be persisted in the persistent tier of the cache.
+ *
+ */
+public interface Stateful {
```
ljacomet: Let's have this named `StatefulSerializer` and `extends `Serializer` because there is no plan of having that contract exposed / honoured anywhere else.

api/src/main/java/org/ehcache/spi/serialization/StatefulSerializer.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.spi.serialization;
+
+import org.ehcache.spi.persistence.StateRepository;
+
+/**
+ * Serializers implementing this interface can have persistent state.
```
ljacomet: Javadoc should be adapted following the name and hierarchy change

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -125,40 +125,37 @@ of the serialized types as they might not be available in the current class load
...
+`CacheConfigurationBuilder.withValueSerializer(Serializer valueSerializer)`
+will work since the instantiation is done by the user code itself.
+
+Registering a serializer that lacks such constructor at the cache manager level will prevent it from being chosen for caches.
 
-On caches that have no persistent capable store, serializers must have a constructor:
+Custom serializer implementations could have some state that is used in the serialization/deserialization process.
+When configured on a persistent cache, such serializers may need to persist and restore their state across restarts.
+For clustered caches there might be an additional requirement that the state of the serializer must be visible to all clients using the same cache(configured with the same serializer).
+To address these requirements you can mark your serializer `Stateful` by implementing the _init_ method in it:
```
ljacomet: That's incomplete. It is not duck typing - you actually have to extend the more specific interface.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -125,40 +125,37 @@ of the serialized types as they might not be available in the current class load
...
 ```java
-public MySerializer(ClassLoader classLoader) {
+public init(StateRepository repository) {
 }
 ```
-Attempting to configure a serializer that lacks such constructor on a transient cache using either of
-`CacheConfigurationBuilder.withKeySerializer(Class<? extends Serializer<K>> keySerializerClass)` or
-`CacheConfigurationBuilder.withValueSerializer(Class<? extends Serializer<V>> valueSerializerClass)`
-will be sanctioned with an exception upon cache initialization.
+The `StateRepository` is provided by the persistent tier of the cache and hence will have the same persistence properties of that tier.
```
ljacomet: replace persistent with authority

ljacomet:
From looking at the code changes and what was done for the `CompactJavaSerializer` I am wondering about making sure that non persistent authority do honour the `StatefulSerializer` contract by providing a `TransientStateRepository` when being configured with such a `Serializer`.
Makes sense?

albinsuresh:
Why do we want non-persistent authority to honor this contract? The stateful contract itself should be something for persistent serializers, right? Because the persistence property of the persistent tiers is what any stateful serilalizer implementor would be interested in when he uses the state repository. I don't see a reason why an implementor would want to keep the state in a non-persistent state repository. So I was trying to associate this new contract with persistent caches alone. That is exactly why I used the word "persistent tier" in documentation and javadoc. So may be the name `StatefulSerializer` is not the right one and it should rather be `PersistentSerializer` or something like that.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
albinsuresh: Reply in the comment below.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
albinsuresh: Yeah missed that after the updates.

albinsuresh:
Following the discussion with @ljacomet  it has been decided that the stateful contract would NOT be limited to persistent caches alone. The `StateRepository` injection will be associated with the authoritative tier of the cache which could be either persistent or non-persistent.

ljacomet:
Marked as blocked since this requires master to move to version 3.2.0-SNAPSHOT before merge.

api/src/main/java/org/ehcache/spi/serialization/StatefulSerializer.java
```diff
@@ -0,0 +1,39 @@
...
+package org.ehcache.spi.serialization;
+
+import org.ehcache.spi.persistence.StateRepository;
+
+/**
+ * Implementations of this interface can have their state maintained in a {@code StateRepository}.
+ * The state will be maintained by the authoritative tier of the cache for which this is configured.
+ * When used within the default serialization provider, there is an additional constructor requirement.
+ * The implementations must define a constructor that takes in a {@code ClassLoader}.
+ * Post instantiation, the state repository will be injected with the {@code init} method invocation.
```
ljacomet: I would add that this call will happen before any interaction with the `Serializer` interface.

ljacomet: And use `<P></P>` to separate paragraphs in the output Javadoc.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -125,40 +125,37 @@ of the serialized types as they might not be available in the current class load
...
+`CacheConfigurationBuilder.withValueSerializer(Serializer valueSerializer)`
+it will work since the instantiation is done by the user code itself.
 
-On caches that have no persistent capable store, serializers must have a constructor:
+Registering a serializer that lacks such constructor at the cache manager level will prevent it from being chosen for caches.
+
+Custom serializer implementations could have some state that is used in the serialization/deserialization process.
+When configured on a persistent cache, such serializers may need to persist and restore their state across restarts.
+For clustered caches there might be an additional requirement that the state of the serializer must be visible to all clients using the same cache(configured with the same serializer).
+To address these requirements you can have a `StatefulSerializer` implementation.
```
ljacomet: Blank line above to start a new paragraph.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -125,40 +125,37 @@ of the serialized types as they might not be available in the current class load
...
+public init(StateRepository repository) {
 }
 ```
-Attempting to configure a serializer that lacks such constructor on a transient cache using either of
-`CacheConfigurationBuilder.withKeySerializer(Class<? extends Serializer<K>> keySerializerClass)` or
-`CacheConfigurationBuilder.withValueSerializer(Class<? extends Serializer<V>> valueSerializerClass)`
-will be sanctioned with an exception upon cache initialization.
-
-Registering a serializer that lacks such constructor at the cache manager level will prevent it from being chosen for persistent caches.
+The `StateRepository.getPersistentConcurrentMap()` provides a `ConcurrentMap` that you can use to store any relevant state.
```
ljacomet: Add `String, Class<K>, Class<V>` to indicate that `getPersistentConcurrentMap` takes parameters.

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -125,40 +125,37 @@ of the serialized types as they might not be available in the current class load
...
+`CacheConfigurationBuilder.withKeySerializer(Serializer keySerializer)` or
+`CacheConfigurationBuilder.withValueSerializer(Serializer valueSerializer)`
+it will work since the instantiation is done by the user code itself.
 
-On caches that have no persistent capable store, serializers must have a constructor:
+Registering a serializer that lacks such constructor at the cache manager level will prevent it from being chosen for caches.
+
+Custom serializer implementations could have some state that is used in the serialization/deserialization process.
+When configured on a persistent cache, such serializers may need to persist and restore their state across restarts.
+For clustered caches there might be an additional requirement that the state of the serializer must be visible to all clients using the same cache(configured with the same serializer).
```
ljacomet: I would drop that line from here - maybe add it below where you mention that the `StateRepository` has the same persistence properties than the authority.

ljacomet:
Aside from the documentation remarks - LGTM

docs/src/docs/asciidoc/user/serializers-copiers.adoc
ljacomet: Sorry if my strike through made it unclear - I would still drop the sentence above. Precision on clustering is provided below when discussing `StateRepository`

albinsuresh: Okay I get it now. But with the structuring of those paragraphs I kinda feel that it is still relevant. In this paragraph (lines 145-147) we are listing the requirements of having a state repositiry: 1) persistence for availability across restarts 2) the visibility aspect which is specific for clustered caches. These are the requirements.

In the following paragraph we describe the state repository and the advantages of using it. There in we reiterate on how we solved the 2 points described above. We mention that state repositories solve the persistence problem by persisting the state also to the persistent tier of the cache and the availability problem is also solved automatically by keeping the data in the cluster.

So in a way, the structuring of both the paragraphs are consistent. Two problems are described in paragraph 1. Solutions to both are described in paragraph 2. Dropping this line alone would break that consistency, I feel. Otherwise we should drop both the requirements in line 146 and 147, right?. WDYT?

docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -125,40 +125,38 @@ of the serialized types as they might not be available in the current class load
...
 
-On caches that have no persistent capable store, serializers must have a constructor:
+To address these requirements you can have a `StatefulSerializer` implementation.
+`StatefulSerializer` is a specialized `Serializer` with an additional _init_ method with the following signature:
 ```java
-public MySerializer(ClassLoader classLoader) {
+public init(StateRepository repository) {
```
ljacomet: Misses the `void` return type

PR #1423, opened 2016-09-07
Moved ClusteringManagementServiceTest (and ClusterService.isConnected())

mathieucarbou:
This PR is to move `ClusteringManagementServiceTest` from management module to clustered/it-tests module. This move required:

- To switch to Galvan instead of passthrough
- To make clustered/it-tests depend on the management libraries
- To add additional server libs when starting galvan through sysprops (see gradle files)
- To add the ability for services to know when the cluster service is (not) connected (see below...)

I run after some issues: several times, at cache manager close, if the cluster service stops BEFORE the management services, then the management scheduled threads are not aware of the closing of the underlying connection, and a lot of error stacktraces are printed each second because of a failure in sending notifications or stats to voltron.

So I added `ClusteringService.isConnected() ` so that dependant services can check if there exists a connection to a cluster.

For @gibsong : if you need to test some clustered stats, you can make your tests extends `AbstractClusteringManagementTest`. This class starts a cluster with all the management services wired, and also provides access to `sendManagementCallToCollectStats()` if you need to start statistic collector, plus a `consumer` if you need to access the stats sent to voltron by the Ehcache collector service.


mathieucarbou:
test this please

mathieucarbou:
@ljacomet @chrisdennis : are you ok with the first commit (`ClusteringService.isConnected()`) ?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -162,8 +167,11 @@ public void start(final ServiceProvider<Service> serviceProvider) {
     } catch (RuntimeException e) {
       entityFactory = null;
       try {
-        clusterConnection.close();
+        Connection conn = clusterConnection;
```
ljacomet: What triggered these connection closing changes?

And by looking at them ... and their duplication ... I can tell you a method wants to be extracted 😉

ljacomet:
Makes sense, is SPI so not too critical ... although that technically means this PR needs to wait for Ehcache 3.2.0-SNAPSHOT on master.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
mathieucarbou: Yes I can extract then. The idea is to nullify as soon as possible the connection so that isConnected() returns false, because we can have some scheduled threads that will call `isConnected()`. So if the `close()` takes some time for any reason, this is to avoid a scheduled thread to still see that we are connected and thus try to send some messages.

mathieucarbou:
Do you know when 3.2.0-SNAPSHOT will be set on master ?

It is not critical for M&M byt @gibsong was waiting for this refactor to be able to write some tests regarding clustered stats (which needs also to have the management layer setup for galvan).

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -233,14 +228,7 @@ public void stop() {
...
 
-    try {
-      if (clusterConnection != null) {
-        clusterConnection.close();
-        clusterConnection = null;
-      }
-    } catch (IOException ex) {
-      throw new RuntimeException(ex);
-    }
+    closeConnection();
```
mathieucarbou: Note: change of behavior: before, the IOException was rethrown. I don't know the reason because all other places just logs it, and if it is rethrown, it will be rethrown at CM.close() and this prevents the transition listeners to be called, whereas all the services are closed.

mathieucarbou:
@anthonydahanne @ljacomet : The build should pass when Terracotta-OSS/terracotta-platform#147 will be merged and tc-plaform 5.0.6.beta7 will be released

mathieucarbou:
tc-platform 5.0.6.beta7 released.

test this please

build.gradle
```diff
@@ -26,7 +26,7 @@ ext {
   sizeofVersion = '0.3.0'
 
   // Clustered
-  terracottaPlatformVersion = '5.0.6.beta6'
+  terracottaPlatformVersion = '5.0.6.beta7'
```
anthonydahanne: you could have squashed this commit into the previous

mathieucarbou: no because the previous commit is not compatible with 5.0.6.beta7.
The last single commit is to make ehcache compatible with 5.0.6.beta6

anthonydahanne: well, actually I was not talking about just the build.gradle but all changes you made for the compatibility as well
it just feels weird to see 2 upgrades for the same lib in the same PR...

mathieucarbou: - the first upgrade was not necessary: I did it only to match the latest state as yesterday.
- I could split in 2 PR. But what would be the benefits ? It would take extra time to build, you'll need to review / merge 2 consecutive PRs...

anthonydahanne:
LGTM

PR #1425, opened 2016-09-08
Insure correct synchronization between start, stop and destroy

henri-tremblay:


impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -126,6 +126,7 @@ public synchronized void startForMaintenance(ServiceProvider<MaintainableService
   /**
    * Default scope for testing
    */
+  @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "If started, the lock is already done")
```
ljacomet: I am pretty sure that warning is genuine.Two ways to get out of it:
* Brute force: make `destroy` and `destroyAll` synchronized
* More subtle: make `started` volatile but enter a `synchronized(this)` before invoking `internalStart`

Issue #1430, opened 2016-09-12
Remove generics warnings


ljacomet:
The build now prints all warnings. And there is a lot of them.

They should get fixed, in order to flush out any bugs they could hide but also to make the developer's life easier when chasing genuine javac failures. For the moment finding the failure amongst the warnings is a real challenge.

Note that some changes concern API classes and thus must be considered with care.

PR #1431, opened 2016-09-12
Issue #1430 First batch of changes to clear warnings

ljacomet:
Probably easier to review commit by commit as they are each addressing a single concern.

### DIFF
--- a/api/src/main/java/org/ehcache/config/Eviction.java
+++ b/api/src/main/java/org/ehcache/config/Eviction.java
@@ -21,7 +21,7 @@ package org.ehcache.config;
  */
 public final class Eviction {
 
-  private static final EvictionAdvisor<?, ?> NO_ADVICE = new EvictionAdvisor<Object, Object>() {
+  private static final EvictionAdvisor<Object, Object> NO_ADVICE = new EvictionAdvisor<Object, Object>() {
     @Override
     public boolean adviseAgainstEviction(Object key, Object value) {
       return false;
@@ -31,12 +31,10 @@ public final class Eviction {
   /**
    * Returns an {@link EvictionAdvisor} where no mappings are advised against eviction.
    *
-   * @param <K> the key type for the advisor
-   * @param <V> the value type for the advisor
    * @return an advisor where no mappings are advised against eviction
    */
-  public static <K, V> EvictionAdvisor<K, V> noAdvice() {
-    return (EvictionAdvisor<K, V>) NO_ADVICE;
+  public static EvictionAdvisor<Object, Object> noAdvice() {
+    return NO_ADVICE;
   }
 
 }
--- a/impl/src/main/java/org/ehcache/impl/copy/SerializingCopier.java
+++ b/impl/src/main/java/org/ehcache/impl/copy/SerializingCopier.java
@@ -16,6 +16,7 @@
 
 package org.ehcache.impl.copy;
 
+import org.ehcache.spi.copy.Copier;
 import org.ehcache.spi.serialization.SerializerException;
 import org.ehcache.spi.serialization.Serializer;
 
@@ -25,6 +26,17 @@ import org.ehcache.spi.serialization.Serializer;
 public final class SerializingCopier<T> extends ReadWriteCopier<T> {
 
   private final Serializer<T> serializer;
+
+  /**
+   * Convenience method allowing to represent this copier's class as the expected type in configuration.
+   *
+   * @param <T> The type to work on
+   * @return the proper type
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> Class<? extends Copier<T>> asCopierClass() {
+    return (Class) SerializingCopier.class;
+  }
 
   /**
    * Creates a new copier that will using the provided {@link Serializer}.
--- a/impl/src/main/java/org/ehcache/impl/events/CacheEventAdapter.java
+++ b/impl/src/main/java/org/ehcache/impl/events/CacheEventAdapter.java
@@ -32,7 +32,7 @@ public abstract class CacheEventAdapter<K, V> implements CacheEventListener<K, V
    * {@inheritDoc}
    */
   @Override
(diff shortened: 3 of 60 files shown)


api/src/main/java/org/ehcache/config/Eviction.java
```diff
@@ -35,6 +35,7 @@ public boolean adviseAgainstEviction(Object key, Object value) {
    * @param <V> the value type for the advisor
    * @return an advisor where no mappings are advised against eviction
    */
+  @SuppressWarnings("unchecked")
```
chrisdennis: Couldn't this be fixed to return EvictionAdvisor<Object, Object> - shouldn't that be able to be consumed by all caches?

impl/src/main/java/org/ehcache/impl/events/CacheEventAdapter.java
```diff
@@ -19,6 +19,8 @@
 import org.ehcache.event.CacheEvent;
 import org.ehcache.event.CacheEventListener;
 
+import static org.ehcache.event.EventType.CREATED;
```
chrisdennis: Why is this suddenly necessary?

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -466,6 +463,7 @@ public void initAuthoritativeTier(AuthoritativeTier<?, ?> resource) {
    * This is kind of a hack, but it's safe to use this if the regular portability
    * is stateless.
    */
+  @SuppressWarnings("unchecked")
```
chrisdennis: Can this be better supressed with PersistentPortability.class.cast(...)?

impl/src/main/java/org/ehcache/impl/events/CacheEventAdapter.java
ljacomet: Weird - no idea, will check

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
henri-tremblay: No. We nee a `PersistentPortability<T>` so cast won't work

impl/src/main/java/org/ehcache/impl/events/CacheEventAdapter.java
henri-tremblay: It seems updating the method made the javadoc auto-import the class.

api/src/main/java/org/ehcache/config/Eviction.java
henri-tremblay: No.  It needs to be typed. However, typing K, V will allow noAdvice to be any wanted type

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
henri-tremblay: But that makes me think I've never advocate my `Util.uncheckedCast` to resolve these issues without a broad `@SuppressWarnings` method annotation. :-)

api/src/main/java/org/ehcache/config/Eviction.java
chrisdennis: Sorry, stupid github formatting, that should have been: `EvictionAdvisor<Object, Object>`

ljacomet: Tried it - does not work. We rely on specific generics bounds in some places. Think Store have `K,V` generics but have internal maps for `K, ValueHolder<V>` and in that context the generic bounds still apply only on `K` and `V`.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
ljacomet: I prefer having the annotation exist where it is really needed. I hate the fact that the annotation does not have an attribute to explain why you are using it btw.

henri-tremblay: The problem with the annotation is that you cannot always be as
fine-grained at wanted. And it can be `Util.uncheckedCast(Object object,
String why)` :-)

ljacomet: That's a good point! You are tempting me 😉

ljacomet:
Added a whole lot of other clean ups.
`api` and `core` are now warnings free for main sources.
`impl` still has some but they are in code @albinsuresh is about to clean up with #1432

impl/src/main/java/org/ehcache/impl/copy/SerializingCopier.java
```diff
@@ -27,6 +28,17 @@
...
 
   /**
+   * Convenience method allowing to represent this copier's class as the expected type in configuration.
+   *
+   * @param <T> The type to work on
+   * @return the proper type
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> Class<? extends Copier<T>> asCopierClass() {
+    return (Class) SerializingCopier.class;
```
henri-tremblay: Shouldn't be cast to `(Class)`. Probably `(Class<? extends Copier<T>>)`

ljacomet: Right - I did back and forth on this change and left it with the wrong one at the end. Fixing.

henri-tremblay: Weird. What was the error?

henri-tremblay: Ok. Let's drop it for now

henri-tremblay:
All changes look good. Merging

Issue #1211, opened 2016-06-23
Implement Passive Key/Value Replication


chrisdennis:
This is the meat of the passive-sync solution that has been discussed in detail in various mail threads and meetings.

chrisdennis:
Blocked by #1208 (and likely by #1209 too)

ljacomet:
Additional outcome: some documentation on the design

AbfrmBlr:
Handling Inflight Client Invalidations during failover :-

Right now we send InvalidationIds(server generated) to all the clients waiting to be invalidated. As and when we get the ack back the bookkeeping is updated and once all the clients have acked backed, the originating client is told that invalidation is complete. Now the problem is what happens if the failover happens during the invalidations are still in progress.
The way it can be solved is to add more states at client sides, The first thing is to maintain a higher watermark of invalidation ids at all the clients and the other thing is to  maintain pair of hash - invalidationId at all the originating clients until the invalidation is complete. 
Both of the these information will help newly promoted Active to regain the (nearly)same state as of earlier active. This helps entity to know about all the waiting originating client and inflight invalidations. 

Now when the first invoke comes to entity after all the handleReconnects are completed, entity sends all the invalidations again(which may end up having duplicate ones at clients, but that's fine higher watermark is at help there).

Handling Passive eviction followed by a failover :- 

This can be handled based on the fact that every eviction at passive can be tracked until it is followed by a replaceAtHead, thus during failover newly promoted Active can invalidate all the clients having that hash which is still not replaced.

AbfrmBlr:
As of now the scheme for replication of append/getAndAppend has to use `IEntityMessenger` api from active entity to send the updated chain(after invoke) to the passive entity so that passive can attain the same state as the Active for that hash. This can lead to really large chains getting replicated to passive. Since our most used operation `get` mandates an asynchronous call to replaceAtHead, similarly other operations can also asynchronously call replaceAtHead. So as an optimization for later(iff found worthy enough), we can decide to invoke `IEntityMessenger` api for the replaceAtHead message from the client as the sub-operation for the corresponding append/getAndAppend operations. This will help us gain throughput for write operations as they can complete without waiting for Active-Passive replication of the original append/getAndAppend operation, as that can be taken care by asynchronous replaceAtHead calls done by clients. I think this should work without any major concerns.It will also depend on what other optimization is enabled at the server side to resolve the chain.

PR #1434, opened 2016-09-12
Issue 1208

AbfrmBlr:


clustered/client/build.gradle
```diff
@@ -31,3 +31,17 @@ dependencies {
   testCompile "org.terracotta:entity-test-lib:$parent.entityTestLibVersion"
   testCompile "org.terracotta:passthrough-server:$parent.terracottaPassthroughTestingVersion"
 }
+
+def java8 = {
+  JavaVersion.current().isJava8Compatible()
```
ljacomet: What is this used for?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -90,7 +92,7 @@ public void abandonLeadership(String entityIdentifier) {
    * @throws TimeoutException if the creation and configuration of the {@code EhcacheActiveEntity} exceed the
    *        lifecycle operation timeout
    */
-  public void create(final String identifier, final ServerSideConfiguration config)
+  public void create(final String identifier, final ServerSideConfiguration config, final UUID clientId)
```
ljacomet: I know we had a chat about this, but now that I see it, I am bothered by having the clientId be injected here instead of generated here. And not having duplications handled as well here.
I believe that either the entity.configure or entity.validate message are a great place to do this validation and thus it could be handled here.
My concern is that with a higher level handling, we kind of tie this id to an object higher in the responsibility chain and may thus have to rework it the moment we need multiple ids for multiple entities.
@AbfrmBlr makes sense?

clustered/client/src/test/java/org/ehcache/clustered/server/ObservableEhcacheServerEntityService.java
```diff
@@ -136,5 +137,6 @@ public String getDefaultServerResource() {
     public Set<String> getDedicatedResourcePoolIds() {
       return ehcacheStateService.getDedicatedResourcePoolIds();
     }
+
```
ljacomet: Changes in this file should be reverted - bring nothing

clustered/server/build.gradle
```diff
@@ -29,6 +29,32 @@ dependencies {
   provided "org.terracotta:standard-cluster-services:$parent.terracottaApisVersion"
 }
 
+def java8 = {
+  JavaVersion.current().isJava8Compatible()
```
ljacomet: What is this for?

clustered/server/src/main/java/org/ehcache/clustered/server/state/ClientMessageTracker.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  public boolean shouldConfigure(UUID clientId, long timestamp) {
+    if (entityConfiguredStamp == null) {
+      return true;
+    }
+    if (clientId.equals(entityConfiguredStamp) && configuredTimestamp == timestamp) {
+      return false;
+    }
+    return true;
```
ljacomet: If I reads this correctly, the entity will be configured (an operation supposedly happening only at creation time) anytime a different client requests to do it, correct? And then why?

clustered/server/build.gradle
```diff
@@ -29,6 +29,27 @@ dependencies {
...
+  options.forkOptions.executable = MavenToolchain.javaExecutable(JavaVersion.VERSION_1_8, 'javac')
+}
+
+compileTestJava {
+  options.fork = true;
+  options.forkOptions.executable = MavenToolchain.javaExecutable(JavaVersion.VERSION_1_8, 'javac')
+}
+
+javadoc {
+  options.addStringOption('Xdoclint:none', '-quiet')
```
ljacomet: This will fail javadoc build when running with a JDK different than 1.8.
Either make it conditional or add the `executable` attribute to mandate usage of 1.8

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -80,12 +80,15 @@
...
 
+  private final AtomicLong sequenceGenerator = new AtomicLong(0L);
+
   private final EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint;
   private final LifeCycleMessageFactory messageFactory;
   private final Map<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>> responseListeners = new ConcurrentHashMap<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>>();
   private final List<DisconnectionListener> disconnectionListeners = new CopyOnWriteArrayList<DisconnectionListener>();
   private final ReconnectDataCodec reconnectDataCodec = new ReconnectDataCodec();
   private volatile boolean connected = true;
+  private volatile UUID clientId = UUID.randomUUID();
```
ljacomet: Must not initialise, this renders the check in `getClientId()` useless and this value will always be overwritten by the `validate` invocation.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -166,14 +176,26 @@ public void close() {
...
-      invokeInternal(timeouts.getLifecycleOperationTimeout(), messageFactory.validateStoreManager(config), false);
+      while (true) {
+        try {
+          clientId = UUID.randomUUID();
+          this.messageFactory.setClientId(clientId);
+          this.reconnectData.setClientId(clientId);
+          invokeInternal(timeouts.getLifecycleOperationTimeout(), messageFactory.validateStoreManager(config), false);
+          break;
+        } catch (InvalidClientIdException e) {
+          //nothing to do - loop again
```
ljacomet: Comment should be more specific - this loops again because of id conflict

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
ljacomet: Missed the fact that it needs generated in `configure` context - please do the generation there instead of during class initialisation.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -69,8 +70,7 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheClientEntity.class);
 
-  private Set<String> reconnectData = new HashSet<String>();
-  private int reconnectDatalen = 0;
+  private ReconnectData reconnectData = new ReconnectData();
```
ljacomet: Should be final and ideally moved next to other fields

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
```diff
@@ -143,6 +144,20 @@ private void invokeLifeCycleOperation(LifecycleMessage message) throws ClusterEx
...
   }
 
+  private void configure(ConfigureStoreManager message) throws ClusterException {
+    ehcacheStateService.configure(message);
+    ehcacheStateService.getClientMessageTracker().setEntityConfiguredStamp(message.getClientId(), message.getId());
+  }
+
+  private void trackAndApplyMessage(LifecycleMessage message) {
+    ClientMessageTracker clientMessageTracker = ehcacheStateService.getClientMessageTracker();
+    if (!clientMessageTracker.isAdded(message.getClientId())) {
```
ljacomet: This should be abstracted in the `ClientMessageTracker`

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -535,7 +547,9 @@ public void destroy() {
    */
   private void configure(ClientDescriptor clientDescriptor, ConfigureStoreManager message) throws ClusterException {
     validateClientConnected(clientDescriptor);
-    ehcacheStateService.configure(message);
+    if (ehcacheStateService.getClientMessageTracker().isConfigureDuplicate(message.getClientId(), message.getId())) {
```
ljacomet: Name is confusing - apply when message is a duplicate??

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -257,7 +266,10 @@ public void handleReconnect(ClientDescriptor clientDescriptor, byte[] extendedRe
       throw new AssertionError("Client "+ clientDescriptor +" trying to reconnect is not connected to entity");
     }
     clientState.attach();
-    Set<String> cacheIds = reconnectDataCodec.decode(extendedReconnectData);
+    ReconnectData reconnectData = reconnectDataCodec.decode(extendedReconnectData);
+    clientIdMap.put(clientDescriptor, reconnectData.getClientId());
+    invalidIds.add(reconnectData.getClientId());
```
ljacomet: This call and the one above make a logical group, duplicated elsewhere - extract method

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+  boolean shouldApply(long msgId) {
+    if (msgId < lowerWaterMark) {
+      return false;
+    }
+    if (msgId > higerWaterMark.get()) {
+      return true;
+    }
+    final AtomicBoolean shouldApply = new AtomicBoolean(false);
+    inProgressMessages.computeIfPresent(msgId, (id, state) -> {
+      if (state != true) {
```
ljacomet: `state` is a boolean - no need to _compare_

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+   * @param msgId
+   */
+  void applied(long msgId) {
+    inProgressMessages.computeIfPresent(msgId, ((id, state) -> state = true));
+    if (lwmLock.tryLock()) {
+      try {
+        for (long i = lowerWaterMark + 1; i<= higerWaterMark.get(); i++) {
+          final AtomicBoolean removed = new AtomicBoolean(false);
+          inProgressMessages.computeIfPresent(i, (id, state) -> {
+            if (state == true) {
```
ljacomet: `state` is a boolean - no need to _compare_

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+    return inProgressMessages.isEmpty();
+  }
+
+  private void updateHigherWaterMark(long msgId) {
+    if (msgId < higerWaterMark.get()) {
+      return;
+    }
+    while(true) {
+      long old = higerWaterMark.get();
+      if (higerWaterMark.compareAndSet(old, msgId)) {
```
ljacomet: not checking that `msgId` is lower than `old` risks making the higher watermark move back

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class MessageTracker {
+
+  private final ConcurrentHashMap<Long, Boolean> inProgressMessages = new ConcurrentHashMap<>();
+
+  private long lowerWaterMark = -1L;   //Always to be updated under lock below
```
ljacomet: but read outside of any lock - that's wrong

AbfrmBlr: The lwm is only read buy newly promoted active, and its written only by passive, so its never a problem. Rather the newly promoted active will eventually clear all the tracked clients in message tracker. Therefore reads and writes to lwm never criss cross.

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
AbfrmBlr: At line no 106  I am doing that I suppose. It checks if currentmsgId < currentHighermark then return

ljacomet: And what guarantees that never becomes true during the while loop?????

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
AbfrmBlr: gross blunder

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
AbfrmBlr: oops, you are correct.

AbfrmBlr:
Addressed all review comments

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
AbfrmBlr: Except this one as this will change in going forward for #1211. See https://github.com/AbfrmBlr/ehcache3/blob/issue-1208-bckup/clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java#L179

AbfrmBlr:
This PR has all the changes regarding #1208. Once this gets merged, this closes #1208 completely. Although it has some ongoing changes for #1211 also which might change going further.

clustered/server/build.gradle
```diff
@@ -29,6 +29,25 @@ dependencies {
...
+  options.fork = true;
+  options.forkOptions.executable = MavenToolchain.javaExecutable(JavaVersion.VERSION_1_8, 'javac')
+}
+
+compileTestJava {
+  options.fork = true;
+  options.forkOptions.executable = MavenToolchain.javaExecutable(JavaVersion.VERSION_1_8, 'javac')
+}
+
+javadoc.executable = MavenToolchain.javaExecutable(JavaVersion.VERSION_1_8, 'javadoc')
```
ljacomet: Can this respect the format?
```
javadoc {
  executable = ...
}
```

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
ljacomet: This is still incorrect per the Java memory model. Yes it will most likely work, not it is not correct.

AbfrmBlr:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
AbfrmBlr: Taken care of.

AbfrmBlr:
test this please

AbfrmBlr:
test this please . management tests are racy

Issue #1438, opened 2016-09-13
Our maven artifacts are Divergent!


chrisdennis:
![Divergent!](https://i.kinja-img.com/gawker-media/image/upload/s--A7P8WhsL--/c_scale,fl_progressive,q_80,w_800/rull9j7ubkw0tdvilh80.jpg)

We should add a Gradle version of Kate Winslet to crush the threat that divergent ~~teenagers~~ _transitive dependency sets_ pose to the stability of ~~our society~~ _downstream dependent projects_.

cschanck:
You understand, that is not Kate Winslet, right? It's Shailene Woodley who will save our builds.

PR #1439, opened 2016-09-13
Issue #1438 : use failOnVersionConflict dependency resolution strategy

chrisdennis:


chrisdennis:
I've fixed this in master at the moment.  If we want to do a daggy fix I guess we can modify the resolution strategy right at (or near) the initial commit and then merge downstream from there?

ljacomet:
Looks good - let's have this on master only.

PR #1435, opened 2016-09-13
Closes #1432 Remove Serializer's FileBasedPersistentContext constructor contract

albinsuresh:

PR #1440, opened 2016-09-14
Closes #1433 StateRepository provides StateHolder instead of ConcurrentMap

### LINKED ISSUE #1433: Reduce API of StateRepository
Following the work on #1411 and since we are reworking contracts, let's fix all mistakes that were introduced in 3.1.

The `StateRepository` currently does not need the complete `ConcurrentMap` API to work, only a reduced set of methods are used effectively by the current code.
Offering thus a custom API instead of a standard map based one will allow for more targeted evolutions in the future.

albinsuresh:
In a lot of places variables are still named "xxxMap" instead of "xxxStateHolder". Will do that if the name `StateHolder` gets finalized.

### DIFF
--- a/api/src/main/java/org/ehcache/spi/persistence/StateHolder.java
+++ b/api/src/main/java/org/ehcache/spi/persistence/StateHolder.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.spi.persistence;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A {@code Map} like structure that can hold key value mappings.
+ *
+ * @param <K> type of Keys
+ * @param <V> type of Values
+ */
+public interface StateHolder<K, V> {
+
+  /**
+   * If the specified key is not already associated with a value (or is mapped
+   * to {@code null}) associates it with the given value and returns
+   * {@code null}, else returns the current value.
+   *
+   * @param key a key
+   * @param value a value
+   * @return the previous value associated with the specified key, or
+   *         {@code null} if there was no mapping for the key.
+   */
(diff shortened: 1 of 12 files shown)


impl/src/test/java/org/ehcache/impl/serialization/CompactPersistentJavaSerializerTest.java
```diff
@@ -30,6 +31,7 @@
 /**
  * CompactPersistentJavaSerializerTest
  */
+@Ignore
```
albinsuresh: Ignore this change as this whole class has been removed in https://github.com/ehcache/ehcache3/pull/1435

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -92,12 +93,12 @@ public CompactJavaSerializer(ClassLoader loader) {
...
   @Override
   public void init(final StateRepository stateRepository) {
-    this.readLookup = stateRepository.getPersistentConcurrentMap("CompactJavaSerializer-ObjectStreamClassIndex", Integer.class, ObjectStreamClass.class);
+    this.readLookup = stateRepository.getPersistentStateHolder("CompactJavaSerializer-ObjectStreamClassIndex", Integer.class, ObjectStreamClass.class);
     loadMappingsInWriteContext(readLookup.entrySet(), true);
   }
 
   Map<Integer, ObjectStreamClass> getSerializationMappings() {
-    return Collections.unmodifiableMap(new HashMap<Integer, ObjectStreamClass>(readLookup));
+    return Collections.unmodifiableMap(new HashMap<Integer, ObjectStreamClass>(Collections.<Integer, ObjectStreamClass>emptyMap()));
```
albinsuresh: Ignore this change as this method has been removed in https://github.com/ehcache/ehcache3/pull/1435

api/src/main/java/org/ehcache/spi/persistence/StateHolder.java
```diff
@@ -0,0 +1,57 @@
...
+   * If the specified key is not already associated with a value (or is mapped
+   * to {@code null}) associates it with the given value and returns
+   * {@code null}, else returns the current value.
+   *
+   * @param key a key
+   * @param value a value
+   * @return the previous value associated with the specified key, or
+   *         {@code null} if there was no mapping for the key.
+   */
+  V add(K key, V value);
```
ljacomet: For evolution, I would stick with `Map` method names - so that would be `putIfAbsent`

albinsuresh: Having just `putIfAbsent` without `put` looked a bit weird to me. So it was deliberately changed to `add`. But yeah I agree with you. Since this could evolve in future, it makes sense to keep it as `putIfAbsent`.

albinsuresh:
Rebased. But management tests keep failing from time to time.

albinsuresh:
Rebased.

PR #1443, opened 2016-09-15
Closes #1441 Make the constructor for ehcache concurrent offheap clock cache public

### LINKED ISSUE #1441: Make the constructor for ehcache concurrent offheap clock cache public
The EhcacheConcurrentOffHeapClockCache is a public class, but its constructor is protected, This disallows   outside packages to construct this data structure. Making this constructor public allows other 'internal' components to construct this data structure as a backing map.

ljacomet:


ljacomet:
@rkavanap Can you rebase your branch?

PR #1447, opened 2016-09-16
Put operation for passive in serverstore impl

AbfrmBlr:
This PR introduces a put operations on `ServerStoreImpl` only to be used by `EhcachePassiveEntity`, i.e. why it is not part of the `ServerStore` api as it is not supposed to be used by clients.
Also the implementation is kind of a provisional thing for now, so that work for AP can move forward. The implementation for put needs to be rewritten in much proper way


AbfrmBlr:
See #1448

PR #1449, opened 2016-09-19
Version upgrade and ManagementTest fix

mathieucarbou:
needs tc-platform __5.0.7.beta__ to be released and deployed first

### DIFF
--- a/management/src/test/java/org/ehcache/docs/ManagementTest.java
+++ b/management/src/test/java/org/ehcache/docs/ManagementTest.java
@@ -86,6 +86,8 @@ public class ManagementTest {
           .build();
 
       long onHeapHitCount;
+      // it could be several seconds before the sampled stats could become available
+      // let's try until we find the correct value : 4
       do {
         ResultSet<ContextualStatistics> counters = query.execute();
 
@@ -95,6 +97,8 @@ public class ManagementTest {
 
         CounterHistory onHeapStore_Hit_Count = statisticsContext.getStatistic(CounterHistory.class, "Cache:HitCount");
 
+        // hit count is a sampled stat, for example its values could be [0,0,3,4].
+        // In the present case, only the last value is important to us , the cache was eventually hit 4 times
         int mostRecentIndex = onHeapStore_Hit_Count.getValue().length - 1;
         onHeapHitCount = onHeapStore_Hit_Count.getValue()[mostRecentIndex].getValue();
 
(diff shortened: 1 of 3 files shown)


mathieucarbou:
test this please

management/src/test/java/org/ehcache/docs/ManagementTest.java
```diff
@@ -78,26 +78,27 @@ public void usingManagementRegistry() throws Exception {
...
-      ResultSet<ContextualStatistics> counters = managementRegistry.withCapability("StatisticsCapability") // <6>
+      StatisticQuery query = managementRegistry.withCapability("StatisticsCapability") // <6>
           .queryStatistic("Cache:HitCount")
           .on(context)
-          .build()
-          .execute();
+          .build();
+
+      long onHeapHitCount;
+      do {
```
anthonydahanne: Could we add a comment explaining why we are looping ?
Something like : 
```
// it could be several seconds before the sampled stats could be available, let's try until we find the correct value : 4
```

management/src/test/java/org/ehcache/docs/ManagementTest.java
```diff
@@ -78,26 +78,27 @@ public void usingManagementRegistry() throws Exception {
       aCache.get(0L);
       aCache.get(0L);
 
-      Thread.sleep(1000);
```
anthonydahanne: 👍

management/src/test/java/org/ehcache/docs/ManagementTest.java
```diff
@@ -78,26 +78,27 @@ public void usingManagementRegistry() throws Exception {
...
+        Assert.assertThat(counters.size(), Matchers.is(1));
 
-      Assert.assertThat(counters.size(), Matchers.is(1));
+        CounterHistory onHeapStore_Hit_Count = statisticsContext.getStatistic(CounterHistory.class, "Cache:HitCount");
 
-      CounterHistory onHeapStore_Hit_Count = statisticsContext.getStatistic(CounterHistory.class, "Cache:HitCount");
-      while(!StatsUtil.isHistoryReady(onHeapStore_Hit_Count, 0L)) {}
-      int mostRecentIndex = onHeapStore_Hit_Count.getValue().length - 1;
-      long onHeapHitCount = onHeapStore_Hit_Count.getValue()[mostRecentIndex].getValue();
+        int mostRecentIndex = onHeapStore_Hit_Count.getValue().length - 1;
```
anthonydahanne: we could also add a comment like : 
```
// hit count is a sampled stat, for example its values could be [0,0,3,4]. In the present case, only the last value is important to us , the cache was eventually hit 4 times

```

PR #1453, opened 2016-09-20
Closes #1452 bump up offheapVersion

### LINKED ISSUE #1452: Bump up offheap-store version

rkavanap:


ljacomet:
@mathieucarbou Changing that assertion should not be required. off heap 2.2.2 and 2.3.0 are exactly the same code. Proof is that the PR #1457 breaks with the exact same failures without the library change. The test is unstable for some reason.

gibsong:
@ljacomet Chris D, Anthony and I are looking into the Expected: a collection with size <55> vs <50> 
Chris and I can't reproduce it locally but Anthony can, so we will meet tomorrow to discuss.

rkavanap:
test this please

PR #1457, opened 2016-09-20
Fix unstable stats tests

anthonydahanne:


clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -190,26 +190,43 @@ public void test_stats_collection() throws Exception {
...
+    do {
+
+      // get the stats (we are getting the primitive counter, not the sample history)
+      ContextualStatistics[] stats = waitForNextStats();
+      Sample<Long>[] samples = stats[0].getStatistic(CounterHistory.class, "Cache:HitCount").getValue();
+
+      if(stats.length == 1 && stats[0].getContext().get("cacheName").equals("cache-1")) {
+        val = samples[0].getValue();
+      }
+    } while(val != 2);
```
ljacomet: Can you add a timeout to the test then? Don't want to see it loop forever the day it is effectively broken by an invalid change.

ljacomet:
Would be great if this PR could address [this failure](https://ehcache.ci.cloudbees.com/job/ehcache3-pull-requests/3100/jdk=Oracle%20JDK%201.8(latest),label_exp=standard/testReport/junit/org.ehcache.docs/ManagementTest/usingManagementRegistry/) as well

gibsong:
Louis, I have a local fix for the java.lang.ArrayIndexOutOfBoundsException: -1.  I can check it in tomorrow morning.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
anthonydahanne: oh right, I'll make the timeout rule global

management/src/test/java/org/ehcache/docs/ManagementTest.java
```diff
@@ -99,8 +99,10 @@ public void usingManagementRegistry() throws Exception {
 
         // hit count is a sampled stat, for example its values could be [0,0,3,4].
         // In the present case, only the last value is important to us , the cache was eventually hit 4 times
-        int mostRecentIndex = onHeapStore_Hit_Count.getValue().length - 1;
-        onHeapHitCount = onHeapStore_Hit_Count.getValue()[mostRecentIndex].getValue();
+        if (onHeapStore_Hit_Count.getValue().length > 0) {
```
anthonydahanne: @ljacomet @gibsong this is the fix for the array out of bound

ljacomet: Well, might be one - but then a similar fix is missing for `ClusteringManagementServiceTest` - see https://ehcache.ci.cloudbees.com/job/ehcache3-pull-requests/lastCompletedBuild/jdk=Oracle%20JDK%201.8(latest),label_exp=standard/testReport/org.ehcache.clustered.management/ClusteringManagementServiceTest/test_stats_collection/

anthonydahanne: ah ! good catch

anthonydahanne: btw @ljacomet the other error (50 vs 55 for the size of descriptors : @gibsong and I looked into it, and we also took it to @chrisdennis too noticing some weird stats fwk 1.20 removal of stats registration (OnHeap:Eviction stats are sometimes missing , when using a disk or offheap store). No definitive answers yet, we could timebomb the incriminated tests if you want.
I'll also make sure the failure is more explicit (something like "OnHeap:Eviction stats are missing" instead of "got 50 was expecting 55")

gibsong: @anthonydahanne there is actually a better fix.   The test can still pass when the index is -1.  It just needs more time to gather the statistics.  So I wouldn't use an assert that will make the test fail.   Just wrap that check in an if statement.  I posted a similar fix below that I need to check in for the ManagementTest

gibsong: long val = -1;
      // it could be several seconds before the sampled stats could become available
      // let's try until we find the correct value : 2
      do {
        ResultSet<ContextualStatistics> counters = query.execute();

        ContextualStatistics statisticsContext1 = counters.getResult(context1);

        CounterHistory counterContext1 = statisticsContext1.getStatistic(CounterHistory.class, "Cache:MissCount");

        // miss count is a sampled stat, for example its values could be [0,1,2].
        // In the present case, only the last value is important to us , the cache was eventually missed 2 times
        int mostRecentSampleIndex = counterContext1.getValue().length - 1;
        if(mostRecentSampleIndex >= 0) {
          val = counterContext1.getValue()[mostRecentSampleIndex].getValue();
        }
      } while(val != 2);

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -190,26 +190,38 @@ public void test_stats_collection() throws Exception {
...
+    do {
+
+      // get the stats (we are getting the primitive counter, not the sample history)
+      ContextualStatistics[] stats = waitForNextStats();
+      Sample<Long>[] samples = stats[0].getStatistic(CounterHistory.class, "Cache:HitCount").getValue();
+
+      if(stats.length == 1 && stats[0].getContext().get("cacheName").equals("cache-1") && samples.length > 0) {
+        val = samples[samples.length - 1].getValue();
+      }
+    } while(val != 2);
```
chrisdennis: So the failure symptom is an infinite loop?

anthonydahanne: yeah, if the timeout exception is thrown,  we'll know from which test

PR #1459, opened 2016-09-22
Move to Gradle 3.1

ljacomet:
Some fundamental build changes in here:
* Requires to be built with Java 8
* Requires developer to setup `java6Home` property in `$HOME/.gradle/gradle.properties`

Will break the multi configuration builds because it effectively only requires to be run once ... though we will probably want to add the possibility of running tests with whatever JVM.

So before I go fix the jobs building this, I want feedback ... which most likely means pulling this locally and testing 😉 

skip ci


107/build.gradle
```diff
@@ -37,24 +39,25 @@ dependencies {
   }
 }
 
-javadoc {
-  exclude '**/tck/**'
+test {
+  if (JavaVersion.current().isJava9Compatible()) {
+    jvmArgs 'add-modules', 'java.xml.bind'
```
chrisdennis: I think you mean: `jvmArgs '--add-modules', 'java.xml.bind'`

xml/build.gradle
```diff
@@ -35,6 +35,16 @@ sourceSets.main.java {
   srcDir generatedSources
 }
 
+if (JavaVersion.current().isJava9Compatible()) {
+  test {
+    jvmArgs 'add-modules', 'java.xml.bind'
```
chrisdennis: `--add-modules`

xml/build.gradle
```diff
@@ -35,6 +35,16 @@ sourceSets.main.java {
...
+  }
+
+  javadoc {
+    options.addStringOption 'add-modules', 'java.xml.bind'
```
chrisdennis: `-add-modules`

107/build.gradle
ljacomet: Nope, the `--` is added by Gradle

ljacomet: To be clear, when you started this, you added `addmods`, the former version of this flag

chrisdennis: Not on my machine it doesn't.

xml/build.gradle
chrisdennis: For clarity the javadoc DSL here adds one dash for you, how helpful.

107/build.gradle
ljacomet: Indeed - fixing

ljacomet:
Well, my thinking was going the other direction. We want developers to be able to run with the minimum of setup and produce what we expect: class version 50 for most of the code, 52 for selected parts.

And I believe this PR achieves that. Now, we still need to be able to test behaviour of standalone in Java 7 and above.

So I wanted to add that as an option, on top of this changes.

In short, add a modifier so that tests are run either with Java 7 or with Java 8.

chrisdennis:
I'm confused - I think we might be talking past each other... let me try to clarify:

          | ./gradlew                                                      | ./gradlew -Pjava6 
----- | ----------------------------------------- | ------------------
javac | `java6Home`                                                | `java6Home`
java   | 'default' if sufficient else `java<X>Home` | `java6Home`

ljacomet:
Had a discussion with @chrisdennis and sticking to the proposed approach.

I pushed an updated version of the branch:
* No longer includes the Java 9 related preparations - will happen at a later time
* Adds a commit that enables a `testVM` property which can be 7 or 8 to enable running tests under a more recent VM than 6.

Given the changes, this will fail to build - waiting to get confirmation this looks good before updating job configurations.

build.gradle
```diff
@@ -72,11 +96,7 @@ subprojects {
   }
 
   dependencies {
-    if (JavaVersion.current().compareTo(JavaVersion.VERSION_1_7) >= 0) {
-      compileOnly 'com.google.code.findbugs:annotations:3.0.0'
-    } else {
-      compileOnly 'com.google.code.findbugs:annotations:2.0.3'
-    }
+    compileOnly 'com.google.code.findbugs:annotations:3.0.0'
```
henri-tremblay: findbugs-annotations should probably be pushed to 3.0.1

build.gradle
```diff
@@ -42,6 +43,29 @@ ext {
   logger.info("Is cloudbees? $isCloudbees")
 }
 
+// Java 6 build setup
+def java6Error = 'Set the poperty \'java6Home\' in your $HOME/.gradle/gradle.properties pointing to a Java 6 installation'
+assert (JavaVersion.current().isJava8Compatible()) : 'The Ehcache 3 build requires Java 8 to run and a configured Java 6 installation\n' + java6Error
+assert hasProperty('java6Home') : 'The Ehcache 3 build requires a configured Java 6 installation\n' + java6Error
+def java6ExecutablesPath = new File(java6Home)
+def testExectuablePath = java6ExecutablesPath
```
henri-tremblay: Why a s to Executable of `java6ExecutablesPath`? There's a typo in `testExecutablePath`

henri-tremblay:
Look good to me apart from some line comments I just added.

If I got the idea correctly:
* Gradle will run with Java 8
* Some projects that requires Java 8 will be compiled and have tests ran with it
* Otherwise, we compile with Java 6
* Tests are ran with the Gradle JVM unless a property is specified
* You can (and should) use the multiconfiguration build but it will play on the `testVM` property, not on the JDK versions

I agree with the idea.

build.gradle
ljacomet: Because `java6ExectuablesPath` points to the `bin` folder of the JDK - hence is the root for `java`, `javac` and `javadoc` for example..
Will add the `s` to `testExecutablePath` indeed.

ljacomet:
Indeed @henri-tremblay that's the idea (and implementation even)

build.gradle
henri-tremblay: Subtle. Ok. Don't forget the typo

henri-tremblay:
Perfect. Yes, I was rephrasing the implementation, not your discussion :-) LGTM (apart from the typo)

build.gradle
ljacomet: Funny - will rename them all - no longer point to `bin` but effectively to java home 😉
I pushed the `bin` concern down to the build utilities class.

ljacomet:
Given that I have @chrisdennis and @henri-tremblay agreement, I will do the following:
* Rebase this PR on the root of 3.1 / 3.2 branch so that the changes can be merged in both
* Open a matching PR for merge into 3.1
* Adapt cloudbees builds and validate PRs are building
* Merge without waiting ... cause the above point will cause all other PRs to fail anyway

ljacomet:
And forgot one point:
* drop an email on the ehcache-dev mailing list letting devs know about this change.

ljacomet:
test this please

PR #1462, opened 2016-09-23
Issue 1211

AbfrmBlr:


ljacomet:
Description? scope? all relevant info to help review this please ...

AbfrmBlr:
Scope - Handling independent evictions by Passive by keeping track of evictions by passive
Inflight invalidations during failover for strong caches.
Use of `EntityMessenger` for replicating proper chain.

Basically Key-value replication is done. some more tests will follow.

AbfrmBlr:
Not to be merged. This needs to remain opened for review/discussions on handling of in-flight in-validations during Active-Passive fail-over.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -104,6 +109,9 @@ public void handleMessage(EntityResponse messageFromServer) {
 
       @Override
       public byte[] createExtendedReconnectData() {
+        for (ReconnectListener reconnectListener : reconnectListeners) {
+          reconnectListener.onHandleReconnect(reconnectData);
```
AbfrmBlr: This part where reconnectListener updates reconnect data, (see `ReconnectData`) can also be refactored in a better way, I can use some suggestions there. IMO, should we have a seperate ReconnectMsg refactored and sent to server.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -241,13 +263,40 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
...
+            // is received which implies we can do some clean up in client ID states.
+            Set<String> caches = ehcacheStateService.getStores();
+            caches.stream()
+                .flatMap(cache -> ehcacheStateService.getEvictionTracker(cache)
+                    .getTrackedEvictedKeys()
+                    .stream()
+                    .map(key -> new Tuple<>(cache, key)))
+                .forEach(x -> {
+                  invalidateHashAfterEviction(x.getK(), x.getV());
+                  LOGGER.debug("Sending invalidation for key {} and cache {} since it was evicted by Passive", x.getV(), x.getK());
```
AbfrmBlr: these method names getK and getV does not make sense

AbfrmBlr:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -241,13 +262,40 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
                                                   " Check your server configuration and define at least one offheap resource.");
       }
 
+      //All the invokes are blocked till passive evicted keys are invalidated by server under lock
+      //Once invalidations are done, other threads will just move on by checking the status
+      if (!invalidationForPassiveEvictedKeysSent.get()) {
```
ljacomet: Let's ditch all that logic.

It is not wrong per caching contract to have clients return null while others return the valid mapping. In all likelihood, the first client seeing null will want to update and that will be correctly propagated to all clients.

And in addition, I am really not fond of that locking done here.

AbfrmBlr: Rather than ditching the logic, we can simply opt for sending invalidations without stopping other threads to make progress as we discussed last time. wdyt ?

AbfrmBlr:
Closing this, will raise another PR just for invalidations.

PR #1466, opened 2016-09-26
Closes #1465 Adds copying newCacheConfigurationBuilder(CacheConfigura…

### LINKED ISSUE #1465: Add 'copying' CacheConfigurationBuilder factory method to support inline modification of configuration
In order to support programatic tweaking of CacheConfiguration objects (and indirectly Configurations) we should add the ability to seed a CacheConfigurationBuilder with an existing Configuration object.  This would then allow us to build a cache configuration that is derived from a proposed configuration in some sensible way.

chrisdennis:
…tion) method


ljacomet:
Looks good to me - just need a rebase

PR #1478, opened 2016-10-04
Key value replication using IEntityMessenger #1211

AbfrmBlr:
This only tackles replication. The complete story can be understood from PR #1462.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
@@ -327,7 +327,7 @@ public class EhcacheClientEntity implements Entity {
    */
   public static final class Timeouts {
 
-    public static final TimeoutDuration DEFAULT_READ_OPERATION_TIMEOUT = TimeoutDuration.of(5, TimeUnit.SECONDS);
+    public static final TimeoutDuration DEFAULT_READ_OPERATION_TIMEOUT = TimeoutDuration.of(20, TimeUnit.SECONDS);
 
     private final TimeoutDuration readOperationTimeout;
     private final TimeoutDuration mutativeOperationTimeout;
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,7 @@ ext {
   terracottaCoreVersion = '5.0.7-beta'
   offheapResourceVersion = terracottaPlatformVersion
   entityApiVersion = terracottaApisVersion
-  terracottaPassthroughTestingVersion = '1.0.7.beta'
+  terracottaPassthroughTestingVersion = '1.0.7.beta2'
   entityTestLibVersion = terracottaPassthroughTestingVersion
   galvanVersion = '1.0.7-beta'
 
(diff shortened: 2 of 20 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -327,7 +327,7 @@ private EhcacheEntityResponse invokeInternal(TimeoutDuration timeLimit, EhcacheE
    */
   public static final class Timeouts {
 
-    public static final TimeoutDuration DEFAULT_READ_OPERATION_TIMEOUT = TimeoutDuration.of(5, TimeUnit.SECONDS);
+    public static final TimeoutDuration DEFAULT_READ_OPERATION_TIMEOUT = TimeoutDuration.of(20, TimeUnit.SECONDS);
```
ljacomet: Why is this change needed?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/RetirementMessage.java
```diff
@@ -0,0 +1,126 @@
...
+package org.ehcache.clustered.common.internal.messages;
+
+import org.ehcache.clustered.common.internal.store.Chain;
+
+import java.util.UUID;
+
+/**
+ * This message is sent by the Active Entity to Passive Entity.
+ */
+public class RetirementMessage extends EhcacheEntityMessage {
```
ljacomet: I do not understand why these messages are named `RetirementMessage`. That's how they are used not what they convey from the active to the passive.

ljacomet: ~~For example, something like `PassiveSyncMessage` might express intent better.~~ Scratch that, this one deals specifically with the registration of client ids - let's name it to reflect that!

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/RetirementMessage.java
```diff
@@ -0,0 +1,126 @@
...
+
+  public long getId() {
+    return msgId;
+  }
+
+  public UUID getClientId() {
+    return clientId;
+  }
+
+  public static class ServerStoreRetirementMessage extends RetirementMessage implements ConcurrentEntityMessage {
```
ljacomet: This is a message that syncs a cluster tier key to the passive - let's name it to reflect that

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
AbfrmBlr: Because the replication tests fail due to get time out. Since the fail-over takes more than 15 secs on my machine atleast.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/RetirementMessage.java
AbfrmBlr: `KeyValueReplicationMessage` ?

AbfrmBlr: `ChainReplicationMessage` ?

ljacomet:
Ok with the proposed name changes (that I have 👍 ). Do not forget to update codecs and other related utility classes.

AbfrmBlr:
Review comments addressed

AbfrmBlr:
test this please

AbfrmBlr:
rebased

PR #1480, opened 2016-10-05
Fix issue #1192

henri-tremblay:
It also adds AssertJ to our dependencies and I'm using `@Enclosed` in one test.

### DIFF
--- a/api/src/main/java/org/ehcache/spi/service/MaintainableService.java
+++ b/api/src/main/java/org/ehcache/spi/service/MaintainableService.java
@@ -22,11 +22,20 @@ package org.ehcache.spi.service;
  */
 @PluralService
 public interface MaintainableService extends Service {
+
+  enum MaintenanceScope {
+    /** Will impact the cache manager */
+    CACHE_MANAGER,
+    /** Will impact one or many caches */
+    CACHE
+  }
+
   /**
    * Start this service for maintenance, based on its default configuration.
-   *
    * @param serviceProvider enables to depend on other maintainable services
+   * @param maintenanceScope the scope of the maintenance
+   *
    */
-  void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider);
+  void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider, MaintenanceScope maintenanceScope);
 
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
@@ -163,6 +163,20 @@ class DefaultClusteringService implements ClusteringService, EntityService {
     }
   }
 
+  @Override
+  public void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider, MaintenanceScope maintenanceScope) {
+    initClusterConnection();
+    createEntityFactory();
+    if(maintenanceScope == MaintenanceScope.CACHE_MANAGER) {
+      if (!entityFactory.acquireLeadership(entityIdentifier)) {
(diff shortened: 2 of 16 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -163,6 +163,20 @@ public void start(final ServiceProvider<Service> serviceProvider) {
     }
   }
 
+  @Override
+  public void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider, MaintenanceScope maintenanceScope) {
```
chrisdennis: Could we leave this method where it was so the changes in the method contents are clearer?

henri-tremblay: I moved it because it was annoying to go back and forth between them. And the private methods used by them are now under them

api/src/main/java/org/ehcache/spi/service/MaintainableService.java
```diff
@@ -22,11 +22,20 @@
...
+
   /**
    * Start this service for maintenance, based on its default configuration.
-   *
    * @param serviceProvider enables to depend on other maintainable services
+   * @param maintenanceScope the scope of the maintenance
+   *
    */
-  void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider);
+  void startForMaintenance(ServiceProvider<MaintainableService> serviceProvider, MaintenanceScope maintenanceScope);
```
chrisdennis: I think the generic type here should be `ServiceProvider<? super MaintainableService>`, since it's fine to pass in a more generic provider here, just not okay for the consumer to use it as a wider type.  Also, isn't this the value that there is in the generic type on ServiceProvider?

impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
```diff
@@ -42,36 +42,44 @@
...
+  private boolean isStarted() {
+    return isStarted;
+  }
+
   /**
    * {@inheritDoc}
    */
   @Override
   public void start(final ServiceProvider<Service> serviceProvider) {
     persistenceService = serviceProvider.getService(LocalPersistenceService.class);
```
chrisdennis: Should one of these start methods call the other?

impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
```diff
@@ -113,7 +121,7 @@ public void releasePersistenceSpaceIdentifier(PersistenceSpaceIdentifier<?> iden
       }
     }
     if (name == null) {
-      throw new CachePersistenceException("Unknown space " + identifier);
+      throw newCachePersistenceException(identifier);
```
chrisdennis: The condition on which this exception is thrown is the same everywhere (no value matching the identifier is found in the map).  Would it be possible to push the throwing of this exception in a single method deeper in the call stack?

api/src/main/java/org/ehcache/spi/service/MaintainableService.java
henri-tremblay: You will like my other pull request (#1481) :-) I kept the typing like it was.

impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
henri-tremblay: Hum... You are right that some refactoring can be done. It is not related to my code so I prefer to merge this PR and do it after

PR #1481, opened 2016-10-05
Remove generic parameter from ServiceProvider

henri-tremblay:
I noticed `ServiceProvider` doesn't need to be generic. It only makes things more complicated. 

Only the methods returning a service need to return the appropriate service type. So removing the generic type helps simplifying the code.


ljacomet:
Well to me it needs to be - that's what allows the difference in return types of `getService` between `ServiceProvider<Service>` and `ServiceProvider<MaintainableService>`

ljacomet:
Closing, @henri-tremblay saw the light - see #1505 😉

PR #1486, opened 2016-10-05
Use org.junit AssumptionViolatedException since the org.junit.internal one is now deprecated

henri-tremblay:


ljacomet:
Good to go once rebased

henri-tremblay:
Done

PR #1505, opened 2016-10-06
startForMaintenance should take a more flexible serviceProvider

henri-tremblay:


ljacomet:
Good to go once rebased

PR #1507, opened 2016-10-06
Refactoring following discoveries during #1192

henri-tremblay:


ljacomet:
Good to go once rebased

PR #1512, opened 2016-10-07
Issue 1211 invalidations

AbfrmBlr:
This Pr covers the handling of inflight invalidations during failover.


clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -155,6 +163,10 @@ public void addDisconnectionListener(DisconnectionListener listener) {
     disconnectionListeners.add(listener);
   }
 
+  public void addReconnectListener(ReconnectListener listener) {
+    reconnectListeners.add(listener);
```
AbfrmBlr: This part where reconnectListener updates reconnect data, (see ReconnectData) can also be refactored in a better way, I can use some suggestions there. IMO, should we have a seperate ReconnectMsg refactored and sent to server

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -104,6 +109,9 @@ public void handleMessage(EntityResponse messageFromServer) {
 
       @Override
       public byte[] createExtendedReconnectData() {
+        for (ReconnectListener reconnectListener : reconnectListeners) {
```
henri-tremblay: Here, you might iterate on an old version of the list since it is a CopyOnWriteArrayList. Is it what we want?

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/BasicClusteredCacheOpsReplicationTest.java
```diff
@@ -62,6 +68,16 @@
   private static Cache<Long, String> CACHE1;
   private static Cache<Long, String> CACHE2;
 
+  @Parameters(name = "consistency={0}")
```
henri-tremblay: This can be simplified with

```java
  @Parameters
  public static Consistency[] data() {
    return Consistency.values();
  }

  @Parameter
  public Consistency cacheConsistency;
```

henri-tremblay:
I've commented the code but can't comment the algorithm since I don't know exactly what it is supposed to do. But we can chat about it if you want.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/EventualServerStoreProxy.java
```diff
@@ -111,11 +160,14 @@ public Chain get(long key) throws TimeoutException {
   @Override
   public void append(final long key, final ByteBuffer payLoad) throws TimeoutException {
     delegate.append(key, payLoad);
+    incrementInvalidationsCounter(key);
   }
 
   @Override
   public Chain getAndAppend(final long key, final ByteBuffer payLoad) throws TimeoutException {
-    return delegate.getAndAppend(key, payLoad);
+    Chain chain = delegate.getAndAppend(key, payLoad);
```
albinsuresh: Nitpicking here: Did you intend to do something with that returned chain? Otherwise I don't see why it's kept? Otherwise the change set could have been just:
```java
incrementInvalidationsCounter(key);
return delegate.getAndAppend(key, payLoad);
```
right?

albinsuresh:
I'm not quite fond of the idea of keeping the `reconnectDataLen` field in the `ReconnectData` class. That field is dynamically updated on every mutation of the `ReconnectData` instance and is finally used to allocate the buffer in the codec. I personally feel that the codec should be computing the size itself by reading the fields of the `ReconnectData` instance rather than delegating that computation responsibility to the `ReconnectData` class itself. If we do it that way, the `ReconnectData` class would be much simpler without the length computation logic spread across several different methods and all the computation logic can move to a single encode method in the `ReconnectDataCodec` class.

AbfrmBlr:
For ReconnectData see my comment on the approach with ReconnectMessage.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/EventualServerStoreProxy.java
AbfrmBlr: nope. we have wait for the call to return from server before that call be accounted for

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ReconnectDataCodec.java
```diff
@@ -20,16 +20,29 @@
...
+    Set<String> clearInProgress = reconnectData.getClearInProgressCaches();
     for (String cacheId : reconnectData.getAllCaches()) {
       encodedMsg.putInt(cacheId.length());
       CodecUtil.putStringAsCharArray(encodedMsg, cacheId);
+      if (clearInProgress.contains(cacheId)) {
+        encodedMsg.put((byte)1);
+      } else {
+        encodedMsg.put((byte)0);
+      }
+      Set<Long> hashToInvalidate = reconnectData.removeInvalidationsInProgress(cacheId);
```
AbfrmBlr: This will be taken care by a separate ReconnectMessage approach

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ReconnectData.java
```diff
@@ -44,22 +49,46 @@ public void setClientId(UUID clientId) {
...
+    return Collections.EMPTY_SET;
+  }
+
+  public void addClearInProgress(String cacheId) {
+    cachesWithClearInProgress.add(cacheId);
+  }
+
+  public Set<String> getClearInProgressCaches() {
+    Set<String> caches = new HashSet<String>(cachesWithClearInProgress);
+    cachesWithClearInProgress.clear();
```
AbfrmBlr: same as reconnect comment

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ReconnectData.java
```diff
@@ -44,22 +49,46 @@ public void setClientId(UUID clientId) {
 
   public void add(String name) {
     reconnectData.add(name);
-    reconnectDatalen.addAndGet(2 * name.length() + ENTRY_SIZE);
+    reconnectDatalen.addAndGet(2 * name.length() + 2 * ENTRY_SIZE + CLEAR_IN_PROGRESS_STATUS_SIZE);
   }
 
   public void remove(String name) {
     if (!reconnectData.contains(name)) {
```
AbfrmBlr: I think I missed there to throw exception & remove if its there only.

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/BasicClusteredCacheOpsReplicationTest.java
AbfrmBlr: cool

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/EventualServerStoreProxy.java
```diff
@@ -126,5 +178,13 @@ public void replaceAtHead(long key, Chain expect, Chain update) {
   @Override
   public void clear() throws TimeoutException {
     delegate.clear();
+    clearInProgressCounter.incrementAndGet();
+  }
+
+  private void incrementInvalidationsCounter(long key) {
+    AtomicInteger counter = new AtomicInteger(1);
+    if ((invalidationsInProgress.putIfAbsent(key, counter)) != null) {
```
AbfrmBlr: but the null check is there for the cases when key is was absent, so null will be returned leading to a nullpointer , right?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/EventualServerStoreProxy.java
```diff
@@ -17,28 +17,77 @@
...
   private final List<InvalidationListener> invalidationListeners = new CopyOnWriteArrayList<InvalidationListener>();
+  private final AtomicInteger clearInProgressCounter = new AtomicInteger(0);
+  private final ConcurrentMap<Long, AtomicInteger> invalidationsInProgress = new ConcurrentHashMap<Long, AtomicInteger>();
 
   public EventualServerStoreProxy(final ServerStoreMessageFactory messageFactory, final EhcacheClientEntity entity) {
     this.delegate = new NoInvalidationServerStoreProxy(messageFactory, entity);
+    entity.addReconnectListener(new EhcacheClientEntity.ReconnectListener() {
+      @Override
+      public void onHandleReconnect(ReconnectData reconnectData) {
+        Set<Long> inflightInvalidations = new HashSet<Long>(invalidationsInProgress.keySet());
```
AbfrmBlr: That's a mess, I will remove while refactoring.see my comment above

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
AbfrmBlr: Since multiple threads can create different caches concurrently, `CopyOnWriteArrayList` just helps there. Anyways its just the number of caches, so should not be a problem

henri-tremblay: If it doesn't matter to miss a listener because it will be processed later, it's fine. Otherwise, it should synchronize instead.

AbfrmBlr: You are right!. It does matter, I can't miss any listeners, that's a blunder.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/EventualServerStoreProxy.java
henri-tremblay: Silly me.... yes you are right. However, can you please change to
```java
    AtomicInteger counter = new AtomicInteger(1);
    counter = invalidationsInProgress.putIfAbsent(key, counter);
    if(counter != null) {
      counter.incrementAndGet();
    }
```

AbfrmBlr:
All review comments taken care of, but the ReconnectMessage refactoring needs to be done.If this looks good, this can be merged. Refactoring can be a follow-up PR.

AbfrmBlr:
There is some problem, build was aborted

AbfrmBlr:
test this please. problem resolved

AbfrmBlr:
test this please

clustered/integration-test/build.gradle
```diff
@@ -44,6 +44,13 @@ compileTestJava {
   options.forkOptions.executable = Jvm.current().javacExecutable
 }
 
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+checkstyle {
```
ljacomet: This change is not needed, it is already configured like that in the root build file.

clustered/integration-test/build.gradle
```diff
@@ -44,6 +44,13 @@ compileTestJava {
   options.forkOptions.executable = Jvm.current().javacExecutable
 }
 
+sourceCompatibility = 1.8
```
ljacomet: These changes should not be needed - can you drop them?

ljacomet:
@AbfrmBlr Need to have a chat around the algorithm used here.

AbfrmBlr:
test this please

AbfrmBlr:
test this please. cloudbees is no kicking the build

AbfrmBlr:
test this please build environment failure

AbfrmBlr:
Had a chat with @ljacomet , and as of now eventual invalidations is broken.

AbfrmBlr:
test this please build environment failure

AbfrmBlr:
Eventual invalidations fixed

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -481,13 +530,13 @@ private void invalidateHashForClient(ClientDescriptor originatingClientDescripto
...
+    if (originatingClientDescriptor != null) {
+      clientsToInvalidate.remove(originatingClientDescriptor);
+    }
 
     InvalidationHolder invalidationHolder = null;
-    if (ehcacheStateService.getStore(cacheId).getStoreConfiguration().getConsistency() == Consistency.STRONG) {
-      invalidationHolder = new InvalidationHolder(originatingClientDescriptor, clientsToInvalidate, cacheId, key);
-      clientsWaitingForInvalidation.put(invalidationId, invalidationHolder);
-    }
+    invalidationHolder = new InvalidationHolder(originatingClientDescriptor, clientsToInvalidate, cacheId, key);
```
ljacomet: This change means that the null check on line 552 is no longer relevant.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -536,23 +585,35 @@ private void invalidateAll(ClientDescriptor originatingClientDescriptor, String
...
 
-    if (ehcacheStateService.getStore(invalidationHolder.cacheId).getStoreConfiguration().getConsistency() == Consistency.STRONG) {
-      invalidationHolder.clientsHavingToInvalidate.remove(clientDescriptor);
-      if (invalidationHolder.clientsHavingToInvalidate.isEmpty()) {
-        if (clientsWaitingForInvalidation.remove(invalidationId) != null) {
-          try {
-            Long key = invalidationHolder.key;
-            if (key == null) {
+    if (invalidationHolder == null) { // Happens when client is re-sending/sending invalidations for which server has lost track since fail-over happened.
+      LOGGER.warn("Ignoring invalidation from client {} " + clientDescriptor);
```
ljacomet: let's keep this in debug as it is a _valid_ condition

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -536,23 +585,35 @@ private void invalidateAll(ClientDescriptor originatingClientDescriptor, String
...
+            } else {
+              entityMessenger.messageSelf(new InvalidationCompleteMessage(invalidationHolder.cacheId, key));
             }
-          } catch (MessageCodecException mce) {
-            //TODO: what should be done here?
-            LOGGER.error("Codec error", mce);
           }
+        } catch (MessageCodecException mce) {
+          //TODO: what should be done here?
+          LOGGER.error("Codec error", mce);
```
ljacomet: Throw `AssertionError` to crash server - this is a programming error, so we want it as visible as possible as early as possible.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -279,14 +287,18 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
...
       throw new AssertionError("Client "+ clientDescriptor +" trying to reconnect is not connected to entity");
     }
     clientState.attach();
-    ReconnectData reconnectData = reconnectDataCodec.decode(extendedReconnectData);
-    addClientId(clientDescriptor, reconnectData.getClientId());
-    Set<String> cacheIds = reconnectData.getAllCaches();
+    ReconnectMessage reconnectMessage = reconnectMessageCodec.decode(extendedReconnectData);
+    addClientId(clientDescriptor, reconnectMessage.getClientId());
+    Set<String> cacheIds = reconnectMessage.getAllCaches();
+//    Set<String> clearInProgressCaches = reconnectMessage.getClearInProgressCaches();
```
ljacomet: Is that commented line dead code or something missing that still needs done?

AbfrmBlr:
All review comments addressed.

AbfrmBlr:
Rebased

AbfrmBlr:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntity.java
```diff
@@ -104,7 +112,14 @@ public void handleMessage(EntityResponse messageFromServer) {
 
       @Override
       public byte[] createExtendedReconnectData() {
-        return reconnectDataCodec.encode(reconnectData);
+        synchronized (lock) {
+          ReconnectMessage reconnectMessage = new ReconnectMessage(clientId, caches);
+          for (ReconnectListener reconnectListener : reconnectListeners) {
+            reconnectListener.onHandleReconnect(reconnectMessage);
+          }
+//          return reconnectDataCodec.encode(reconnectData);
```
ljacomet: Can this be cleaned up please?

AbfrmBlr:
test this please.

PR #1519, opened 2016-10-07
Closes #1479 Active-Passive data sync

### LINKED ISSUE #1479: Implement active-passive data sync
In #1210 we decided that a passive joining a cluster must be fully synced with all the key-value pairs in the active. So this task is to implement the same in the least disruptive manner. The active should not be fully blocked while the sync is being performed. Instead, the sync should be performed per concurrency key so that there's only concurrency key level blocking on the active.

albinsuresh:
The most basic implementation of sync where all the entries are iterated one by one and synced as individual messages. Next step is to optmise this by minimizing the messages sent. As the first step, all entries in a single server store segment will be pushed in a single sync message. A better heuristic based division can be added later.


ljacomet:
I believe this PR is now obsolete as we have come up with a scheme. I would suggest keeping the same pattern of a message per key but move to the design suggested in #1479

clustered/client/src/test/java/org/ehcache/clustered/client/ActivePassiveSyncTest.java
```diff
@@ -1,158 +0,0 @@
...
-import java.util.concurrent.TimeoutException;
-
-import static org.ehcache.clustered.client.internal.UnitTestConnectionService.getOffheapResourcesType;
-import static org.ehcache.config.units.MemoryUnit.MB;
-import static org.hamcrest.Matchers.containsString;
-import static org.hamcrest.Matchers.instanceOf;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.fail;
-
-public class ActivePassiveSyncTest {
```
albinsuresh: Removed this pass-through test as it is redundant in the presence of a real integration test that tests the same.

clustered/integration-test/src/test/java/org/ehcache/clustered/sync/PassiveSyncTest.java
```diff
@@ -81,11 +84,71 @@ public void testSync() throws Exception {
...
+
+      for (long i = -5; i < 5; i++) {
+        assertThat(cache.get(i), equalTo("value" + i));
+      }
+    } finally {
+      cacheManager.close();
+    }
+  }
+
+  @Ignore
```
albinsuresh: This test will fail due to a platform bug. Kept it here so that we can verify if the bug is fixed with the next platform release.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ServerStoreOpMessage.java
```diff
@@ -131,7 +126,7 @@ public long getKey() {
 
     @Override
     public int concurrencyKey() {
-      return (int) (super.concurrencyKey() + key);
+      return (int) key;
```
ljacomet: This needs to be changed to return the `long` value.

clustered/server/src/main/java/org/ehcache/clustered/server/ConcurrencyStrategies.java
```diff
@@ -44,17 +46,20 @@ public DefaultConcurrencyStrategy(int bucketCount) {
 
     @Override
     public int concurrencyKey(EntityMessage entityMessage) {
-      if (entityMessage instanceof ConcurrentEntityMessage) {
+      if (entityMessage instanceof ServerStoreOpMessage.GetMessage) {
+        return UNIVERSAL_KEY;
+      } else if (entityMessage instanceof ConcurrentEntityMessage) {
         ConcurrentEntityMessage concurrentEntityMessage = (ConcurrentEntityMessage) entityMessage;
-        return DEFAULT_KEY + Math.abs(concurrentEntityMessage.concurrencyKey() % bucketCount);
+        return DATA_CONCURRENCY_KEY_OFFSET + Math.abs(concurrentEntityMessage.concurrencyKey() % bucketCount);
```
ljacomet: Here you must use the exact same logic as in `org.ehcache.clustered.server.offheap.OffHeapServerStore#segmentFor` and since both are inside the server code, I would extract a method that is responsible for doing it.

albinsuresh:
Addressed the review feedback on extracting the message to segment mapping from entities and concurrency strategy to a common place.

ljacomet:
@albinsuresh needs rebase

Issue #1508, opened 2016-10-06
Gradle Jetty plugin is deprecated


henri-tremblay:
It should be replaced by https://plugins.gradle.org/plugin/org.akhikhl.gretty

Jetty is used for the demos.

Also, it seems that for some reason, IntelliJ call this plugin and unlock the exploded files in a directory named `out` located at the root of ehcache3.

PR #1523, opened 2016-10-11
Set IntelliJ to put wars in the build directory (see https://youtrack.jetbrains.com/issue/IDEA-91952)

henri-tremblay:


ljacomet:
Not too fond of adding IDE specific stuff to the build files. In addition, this might be caused by the war / jetty stuff in the demo folder which needs an update anyway as filed in #1508 

Given that, I am recommending to close this and wait until #1508 is addressed.

henri-tremblay:
Sadly, #1533 is using gretty but it haven't fixed the problem. It is quite classic to have a little IDE configuration  in the build tool. I also try to avoid it as much as possible but sometime, there is not other choice. The `org.eclipse.m2e:ifecycle-mapping` plugin is a good example of that. The idea plugin for gradle seems to nicely fix my problem without causing any harm.

henri-tremblay:
The problem seems to have magically disappear. Might be from gretty that then has updated IntelliJ. So I will close that for now.

Issue #1468, opened 2016-09-28
EhcacheCachingProvider is loading the XmlConfiguration even if an existing CacheManager matches the requested URI


henri-tremblay:
See the code below with my comments

```java
  public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties properties) {
    uri = uri == null ? getDefaultURI() : uri;
    classLoader = classLoader == null ? getDefaultClassLoader() : classLoader;
    properties = properties == null ? new Properties() : cloneProperties(properties);

    if (URI_DEFAULT.equals(uri)) {
      URI override = DefaultConfigurationResolver.resolveConfigURI(properties);
      if (override != null) {
        uri = override;
      }
    }

    Configuration config;
    try {
      if (URI_DEFAULT.equals(uri)) {
        config = new DefaultConfiguration(classLoader);
      } else {
        config = new XmlConfiguration(uri.toURL(), classLoader); // configuration loaded...
      }
    } catch (Exception e) {
      throw new javax.cache.CacheException(e);
    }

    return getCacheManager(uri, config, properties); // ... whatever happen in there. Which is where the URI lookup in the cache manager map occurs
  }
```

Note that it still works. It's more a performance issue than anything else.

PR #1528, opened 2016-10-11
 Issue #1468 Do not create config for existing cachemanager

kcleereman:
Issue #1468 Do not create config for existing cachemanager

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
+++ b/107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
@@ -76,18 +76,7 @@ public class EhcacheCachingProvider implements CachingProvider {
       }
     }
 
-    Configuration config;
-    try {
-      if (URI_DEFAULT.equals(uri)) {
-        config = new DefaultConfiguration(classLoader);
-      } else {
-        config = new XmlConfiguration(uri.toURL(), classLoader);
-      }
-    } catch (Exception e) {
-      throw new javax.cache.CacheException(e);
-    }
-
-    return getCacheManager(uri, config, properties);
+    return getCacheManager(new ConfigSupplier(uri, classLoader), properties);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -113,9 +143,14 @@ public Eh107CacheManager getCacheManager(URI uri, Configuration config) {
    * @return a cache manager
    */
   public Eh107CacheManager getCacheManager(URI uri, Configuration config, Properties properties) {
+    return getCacheManager(new ConfigSupplier(uri, config), properties);
+  }
+
+  public Eh107CacheManager getCacheManager(ConfigSupplier configSupplier, Properties properties) {
```
ljacomet: This should not be public.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -52,6 +52,47 @@
 
   private final Map<ClassLoader, ConcurrentMap<URI, Eh107CacheManager>> cacheManagers = new WeakHashMap<ClassLoader, ConcurrentMap<URI, Eh107CacheManager>>();
 
+  public static class ConfigSupplier {
```
ljacomet: this should not be public either. And by usual code style in the project, should be at the end of the file.

ljacomet:
Looks good in principle - just visibility and code style remarks

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
kcleereman: I've made the visibility changes, not sure why github hasn't noticed but oh well

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -50,6 +51,8 @@
 
   private static final URI URI_DEFAULT;
 
+  private static final AtomicInteger CONFIG_COUNT = new AtomicInteger(0);
```
ljacomet: @kcleereman Your previous version was better - just make the inner class and added getter package private. Here this atomic is still public and just plain useless but for the test purpose.

ljacomet:
Good to go once rebased

PR #1533, opened 2016-10-12
Move to gretty plugin

henri-tremblay:


ljacomet:
Good to go once rebased

PR #1538, opened 2016-10-13
Issue #1537 : HitRatio not computed correctly

### LINKED ISSUE #1537: HitRatio not computed correctly
The HitRatio should be obtained by dividing the HitCount by (HitCount+ MissCount), over a certain window of time; so that its value should be included between 0 and 1 (with the exception of NaN when no hit nor miss)
Right now, it is just dividing the HitCount by MissCount, this is why we currently end up with values included between 0 and Infinity
This issue was initially created on the statistic repo (https://github.com/Terracotta-OSS/statistics/issues/11), but the issue is with ehcache3 stats.

anthonydahanne:
* fix hit ratio
* rewrite  some tests assertions
Fix #1537

### DIFF
--- a/management/src/main/java/org/ehcache/management/providers/statistics/StandardEhcacheStatistics.java
+++ b/management/src/main/java/org/ehcache/management/providers/statistics/StandardEhcacheStatistics.java
@@ -55,6 +55,11 @@ import java.util.concurrent.TimeUnit;
 import org.terracotta.management.model.capabilities.descriptors.StatisticDescriptor;
 import org.terracotta.management.model.stats.StatisticType;
 
+import static java.util.Collections.singleton;
+import static java.util.EnumSet.allOf;
+import static java.util.EnumSet.of;
+import static org.terracotta.context.extended.ValueStatisticDescriptor.descriptor;
+
 class StandardEhcacheStatistics extends ExposedCacheBinding {
 
   private final StatisticsRegistry statisticsRegistry;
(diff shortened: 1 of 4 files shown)


mathieucarbou:
@anthonydahanne : code change is good for me. There's just a remaining test to update (DefaultManagementRegistryServiceTest). it fails right now ;-)

management/src/main/java/org/ehcache/management/providers/statistics/StandardEhcacheStatistics.java
```diff
@@ -65,16 +65,22 @@
...
+    statisticsRegistry.registerRatios("Cache:HitRatio", getCacheStatisticDescriptor, hit, EnumSet.allOf(CacheOperationOutcomes.GetOutcome.class));
+    statisticsRegistry.registerRatios("Cache:MissRatio", getCacheStatisticDescriptor, miss, EnumSet.allOf(CacheOperationOutcomes.GetOutcome.class));
 
-    statisticsRegistry.registerCompoundOperations("Hit", OperationStatisticDescriptor.descriptor("get", Collections.singleton("tier"), TierOperationStatistic.TierOperationOutcomes.GetOutcome.class), EnumSet.of(TierOperationStatistic.TierOperationOutcomes.GetOutcome.HIT));
-    statisticsRegistry.registerCompoundOperations("Miss", OperationStatisticDescriptor.descriptor("get", Collections.singleton("tier"), TierOperationStatistic.TierOperationOutcomes.GetOutcome.class), EnumSet.of(TierOperationStatistic.TierOperationOutcomes.GetOutcome.MISS));
+    OperationStatisticDescriptor getTierStatisticDescriptor = OperationStatisticDescriptor.descriptor("get", Collections.singleton("tier"), TierOperationStatistic.TierOperationOutcomes.GetOutcome.class);
```
chrisdennis: Could we reference the tier types as `TierOperationOutcomes.GetOutcome` so that these lines are slightly less noisy.

chrisdennis: We could probably statically import `of`, `allOf`, `singleton` and `descriptor`, as well.

gibsong:
Are the DefaultManagementRegistryServiceTest's still failing?  It was due to the addition of MissRatio(thanks for adding that in Anthony, it was in a branch of mine but would have taken longer to get committed).  It looks like this may have already been fixed by updating the DefaultManagementRegistryServiceTest collections in loadStatsUtil().

anthonydahanne:
test this please
some unrelated tests failed on cloudbees/ jdk6 only

anthonydahanne:
grrr....
test this please
now it's jenkins crapping out

PR #1539, opened 2016-10-13
bump up terracotta-core & galvan version

AbfrmBlr:


AbfrmBlr:
This updates terracotta-core and galvan version to the latest release which resolves memory leak related issue to RetirementManager.

PR #1544, opened 2016-10-17
:bug: Fix #1543 : Use SIzeHistory for size in bytes related stats

### LINKED ISSUE #1543: Sizes in bytes should be stored as SizeHistory stats, not CounterHistory
Using SizeHistory, the unit can be expressed as Bytes, instead of number of entries.

anthonydahanne:
For example, OnHeap:OccupiedByteSize


mathieucarbou:
look at the PR, good for me, there's just some other tests that needs to be modified I think

anthonydahanne:
@mathieucarbou tests are passing now ! Feel free to merge ;-)

management/src/main/java/org/ehcache/management/providers/statistics/StandardEhcacheStatistics.java
```diff
@@ -145,6 +145,10 @@
 
         if (name.equals(statisticName)) {
           SampledStatistic<Long> count = (SampledStatistic<Long>) registeredValueStatistic.getSampledStatistic();
+          if(statisticName.endsWith("Size")) {
+            // a registeredValue size ? it should go into a SizeHistory !
+            return new SizeHistory(buildHistory(count, since), MemoryUnit.B);
```
ljacomet: Is there not a way to register this type information somewhere instead of having to rely on naming conventions? That seems really brittle.

management/src/main/java/org/ehcache/management/providers/statistics/StandardEhcacheStatistics.java
```diff
@@ -195,7 +200,9 @@
       } else if(registeredStatistic instanceof RegisteredRatioStatistic) {
         capabilities.add(new StatisticDescriptor(entry.getKey() + "Ratio", StatisticType.RATIO_HISTORY));
       } else if(registeredStatistic instanceof RegisteredValueStatistic) {
-        capabilities.add(new StatisticDescriptor(entry.getKey().toString(), StatisticType.COUNTER_HISTORY));
+        capabilities.add(new StatisticDescriptor(statisticName, statisticName.endsWith("Size") ?
```
ljacomet: Same question here - can't we find a way to not rely on naming convention?

ljacomet: How much code is it? I think that relying on `String` values to drive the difference is something that will cost in the future ...

anthonydahanne:
test this please
build failed because of environment

ljacomet:
test this please

anthonydahanne:
@ljacomet can you please merge ?
thanks

Issue #1522, opened 2016-10-11
get() sometimes returns null after putIfAbsent()


dmitrymurashenkov:
Basic disk-backed cached, no clustering, several threads accessing single Cache object. EhCache 3.1.3, Ubuntu 16.04, Oracle JDK 8 (1.8.0_101-b13)

The case is:

```
String value = cache.get(key);
if (value == null)
{
    cache.putIfAbsent(key, "val");
    value = cache.get(key);
}
```

Sometimes "value" is null after this block, but if we wait for some 100ms and retry get() - it returns correct value.

Error is reproduced nearly 100% in complex test over production code, but when I tried to isolate it into  unit test - reproduced very rarely, 1 error in 30-50 test runs. 

In production test usually several threads miss same value and manage to get() it only after small sleep. Error usually occurs in first 3-5 seconds of test.

Here is JUnit test for Java 8: we have small heap disk-backed cache and several threads trying to put a single key-value mapping to it. But, you have to run it numerous times to get a reproduce (test will fail and for each missed get - line will be printed into stdout).

```
import org.ehcache.Cache;
import org.ehcache.CacheManager;
import org.ehcache.config.builders.CacheConfigurationBuilder;
import org.ehcache.config.builders.CacheManagerBuilder;
import org.ehcache.config.builders.ResourcePoolsBuilder;
import org.ehcache.config.units.MemoryUnit;
import org.junit.Test;

import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.concurrent.*;
import java.util.concurrent.atomic.AtomicBoolean;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.concurrent.atomic.AtomicLong;

import static org.junit.Assert.assertFalse;

public class EhCacheTest
{
    @Test
    public void test() throws Exception
    {
        CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
                .with(CacheManagerBuilder.persistence("ehcache-data"))
                .build();
        cacheManager.init();

        Cache<String, String> cache = cacheManager.createCache("my-cache",
                CacheConfigurationBuilder.newCacheConfigurationBuilder(
                        String.class,
                        String.class,
                        ResourcePoolsBuilder.newResourcePoolsBuilder()
                                .heap(1, MemoryUnit.MB)
                                .disk(1, MemoryUnit.GB, false)
                )
        );

        int totalDataBytes = 20*1024*1024;
[...]

PR #1550, opened 2016-10-19
Fix for #1522

henri-tremblay:
This is the fix for #1522 at explained in the issue comment. However, it only contains the fix for putIfAbsent, not for all the other methods.

I just want you to have a look and give your comment on the test. Other tests will look the same and the fixes will all be as simple.


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -720,6 +728,10 @@ public K getKey() {
           }
         });
         cachedValue = backEnd.putIfAbsent(key, fault);
+
+        // Hook added for testing
+        beforeCompletingTheFault();
```
ljacomet: I really really want to work hard to make this not necessary

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -959,9 +984,9 @@ public void invalidateAllWithHash(long hash) throws StoreAccessException {
   private ValueHolder<V> getValue(final ValueHolder<V> cachedValue) {
     if (cachedValue instanceof Fault) {
       return ((Fault<V>)cachedValue).get();
-    } else {
-      return cachedValue;
     }
+
+    return cachedValue;
```
ljacomet: Code style preference ... but I'd rather stick with the explicit else. If someone was to screw up the if branch to no longer return, this would compile fine. The previous version would fail saying that there is a missing return statement.

ljacomet:
I really am not fond of the added abstract method, especially in such a contested code area as soon as a cache as more tier.

Will work from this PR in trying to get another form of test that does not require it.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
henri-tremblay: I tried but the `synchronized` in the fault is the only thing that happens right after. The only other solution would be to decorate the backEnd. But it means getting it by reflection, wrapping it and setting it back. This seemed the easiest and cleanest.

impl/src/test/java/org/ehcache/docs/UserManagedCaches.java
```diff
@@ -21,20 +21,14 @@
...
-import org.ehcache.event.EventType;
-import org.ehcache.impl.config.persistence.UserManagedPersistenceContext;
 import org.ehcache.config.units.EntryUnit;
 import org.ehcache.config.units.MemoryUnit;
+import org.ehcache.core.spi.service.LocalPersistenceService;
 import org.ehcache.docs.plugs.ListenerObject;
-import org.ehcache.docs.plugs.LongCopier;
+import org.ehcache.event.EventType;
+import org.ehcache.impl.config.persistence.DefaultPersistenceConfiguration;
+import org.ehcache.impl.config.persistence.UserManagedPersistenceContext;
```
henri-tremblay: It's related. A bit. Modifying StringCopier made me revise where it is used. And here was a unused import.

henri-tremblay:
Ok. I'm not a big fan either. But since it is empty, it will quickly be removed. And only called on cache miss.

A way out is to refactor into smaller methods. And then use a partial mock.

henri-tremblay:
Some notable changes:
* I'm using a more realistic Authoritative tier. Because it was really weird to mock it at some point. A more complete version of it could become a OnHeapAuthoritativeStore that could be useful for testing
* All methods are tested, they will fail on deadlock instead of invalid values

impl/src/test/java/org/ehcache/impl/internal/store/tiering/TieredStoreMutatorTest.java
```diff
@@ -0,0 +1,456 @@
...
+        removeKeyFromTieredStore();
+      }
+    });
+
+    progressLatch.await();
+
+    launchThread3();
+
+    // 4. Test Thread receives null from remove. We would expect the get to receive the same null after
+    //    a. Test Thread -> TieredStore.remove
```
ljacomet: Comments that mention remove seem off compared to the test.

henri-tremblay: Good catch. Done

henri-tremblay:
OK. Let me see. Also, I forgot the daggy fix. Still need it?

Le 26 oct. 2016 7:17 AM, "Louis Jacomet" <notifications@github.com> a
écrit :

*@ljacomet* requested changes on this pull request.

Some comments are not inline with the test - see above remark

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ehcache/ehcache3/pull/1550#pullrequestreview-5834481>,
or mute the thread

henri-tremblay:
Daggy fixed and rebased

ljacomet:
This is not right for a daggy fix - it can't be rebased, instead needs to be merged with master and then merged back.

ljacomet:
Good to merge once build passes

PR #1552, opened 2016-10-19
Issue 1530

gibsong:
This check in includes the server side allocatedMemory statistic.  Also I had to increase the build.gradle maxHeapSize to 2048m otherwise 48 of the :clustered:client tests failed because the JVM ran out of memory.  I did try to to avoid this by forcing the passthrough server to call the serverside destroy method but this didn't seem to work.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,7 +125,7 @@ subprojects {
   }
 
   test {
-    maxHeapSize = "1024m"
+    maxHeapSize = "1408m"
     systemProperty 'java.awt.headless', 'true'
     if (parent.isCloudbees) {
       systemProperty 'disable.concurrent.tests', 'true'
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
@@ -16,6 +16,7 @@
 
 package org.ehcache.clustered.server;
 
+import java.util.Arrays;
 import org.ehcache.clustered.common.PoolAllocation;
 import org.ehcache.clustered.common.ServerSideConfiguration;
 import org.ehcache.clustered.common.internal.ServerStoreConfiguration;
@@ -32,6 +33,7 @@ import org.ehcache.clustered.common.internal.exceptions.InvalidStoreException;
 import org.ehcache.clustered.common.internal.exceptions.InvalidStoreManagerException;
 import org.ehcache.clustered.common.internal.exceptions.LifecycleException;
 import org.ehcache.clustered.common.internal.exceptions.ResourceConfigurationException;
+import org.terracotta.context.TreeNode;
 import org.terracotta.entity.ServiceRegistry;
 import org.terracotta.offheapresource.OffHeapResource;
 import org.terracotta.offheapresource.OffHeapResourceIdentifier;
@@ -40,6 +42,7 @@ import org.terracotta.offheapstore.paging.OffHeapStorageArea;
 import org.terracotta.offheapstore.paging.Page;
 import org.terracotta.offheapstore.paging.PageSource;
 import org.terracotta.offheapstore.paging.UpfrontAllocatingPageSource;
+import org.terracotta.statistics.StatisticsManager;
 
 import java.util.Collections;
 import java.util.HashMap;
@@ -49,14 +52,42 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Callable;
+import java.util.function.Function;
 
 import static java.util.stream.Collectors.toMap;
 import static org.terracotta.offheapstore.util.MemoryUnit.GIGABYTES;
 import static org.terracotta.offheapstore.util.MemoryUnit.MEGABYTES;
 
+
 public class EhcacheStateServiceImpl implements EhcacheStateService {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheStateServiceImpl.class);
+
+  private static final String STATISTICS_STORE_TAG = "ServerStore";
+  private static final String STATISTICS_POOL_TAG = "Pool";
+  private static final String PROPERTY_STORE_KEY = "storeName";
+  private static final String PROPERTY_POOL_KEY = "poolName";
+
+  private static final Map<String, Function<ServerStoreImpl,Number>> STAT_STORE_METHOD_REFERENCES = new HashMap<>();
+  private static final Map<String, Function<ResourcePageSource,Number>> STAT_POOL_METHOD_REFERENCES = new HashMap<>();
+
+  static {
+    STAT_STORE_METHOD_REFERENCES.put("allocatedMemory", ServerStoreImpl::getAllocatedMemory);
+    STAT_STORE_METHOD_REFERENCES.put("dataOccupiedMemory", ServerStoreImpl::getDataOccupiedMemory);
+    STAT_STORE_METHOD_REFERENCES.put("occupiedMemory", ServerStoreImpl::getOccupiedMemory);
+    STAT_STORE_METHOD_REFERENCES.put("dataOccupiedMemory", ServerStoreImpl::getDataOccupiedMemory);
+    STAT_STORE_METHOD_REFERENCES.put("entries", ServerStoreImpl::getSize);
+    STAT_STORE_METHOD_REFERENCES.put("usedSlotCount", ServerStoreImpl::getUsedSlotCount);
+    STAT_STORE_METHOD_REFERENCES.put("dataVitalMemory", ServerStoreImpl::getDataVitalMemory);
+    STAT_STORE_METHOD_REFERENCES.put("vitalMemory", ServerStoreImpl::getVitalMemory);
+    STAT_STORE_METHOD_REFERENCES.put("reprobeLength", ServerStoreImpl::getReprobeLength);
+    STAT_STORE_METHOD_REFERENCES.put("removedSlotCount", ServerStoreImpl::getRemovedSlotCount);
+    STAT_STORE_METHOD_REFERENCES.put("dataSize", ServerStoreImpl::getDataSize);
+    STAT_STORE_METHOD_REFERENCES.put("tableCapacity", ServerStoreImpl::getTableCapacity);
+
+    STAT_POOL_METHOD_REFERENCES.put("allocatedSize", ResourcePageSource::getAllocatedSize);
+  }
 
   private final ServiceRegistry services;
   private final Set<String> offHeapResourceIdentifiers;
@@ -241,6 +272,7 @@ public class EhcacheStateServiceImpl implements EhcacheStateService {
     } else if (source.reserve(pool.getSize())) {
       try {
         pageSource = new ResourcePageSource(pool);
+        registerPoolStatistics(poolName, pageSource);
       } catch (RuntimeException t) {
         source.release(pool.getSize());
         throw new ResourceConfigurationException("Failure allocating pool " + pool, t);
@@ -250,6 +282,42 @@ public class EhcacheStateServiceImpl implements EhcacheStateService {
       throw new ResourceConfigurationException("Insufficient defined resources to allocate pool " + poolName + "=" + pool);
     }
     return pageSource;
+  }
+
+  private void registerStoreStatistics(ServerStoreImpl store, String storeName) throws InvalidStoreException {
+    STAT_STORE_METHOD_REFERENCES.entrySet().stream().forEach((entry)->
+      registerStatistic(store, storeName, entry.getKey(), STATISTICS_STORE_TAG, PROPERTY_STORE_KEY, () -> entry.getValue().apply(store) ));
+  }
+
+  private void registerPoolStatistics(String poolName, ResourcePageSource pageSource) {
+    STAT_POOL_METHOD_REFERENCES.entrySet().stream().forEach((entry)->
+      registerStatistic(pageSource, poolName, entry.getKey(), STATISTICS_POOL_TAG, PROPERTY_POOL_KEY, () -> entry.getValue().apply(pageSource))
+    );
+  }
+
+  private void unRegisterStoreStatistics(ServerStoreImpl store) {
+    TreeNode node = StatisticsManager.nodeFor(store);
+
+    if(node != null) {
+      node.clean();
+    }
+  }
+
+  private void unRegisterPoolStatistics(ResourcePageSource pageSource) {
+    TreeNode node = StatisticsManager.nodeFor(pageSource);
+
+    if(node != null) {
+      node.clean();
+    }
+  }
+
+  private void registerStatistic(Object context, String name, String observerName, String tag, String propertyKey, Callable<Number> callable) {
+    Set<String> tags = new HashSet<String>(Arrays.asList(tag,"tier"));
+    Map<String, Object> properties = new HashMap<String, Object>();
+    properties.put("discriminator", tag);
+    properties.put(propertyKey, name);
+
+    StatisticsManager.createPassThroughStatistic(context, observerName, tags, properties, callable);
   }
 
   private void releaseDedicatedPool(String name, PageSource pageSource) {
@@ -270,6 +338,7 @@ public class EhcacheStateServiceImpl implements EhcacheStateService {
 
   public void destroy() {
     for (Map.Entry<String, ServerStoreImpl> storeEntry: stores.entrySet()) {
+      unRegisterStoreStatistics(storeEntry.getValue());
       storeEntry.getValue().close();
     }
     stores.clear();
@@ -300,6 +369,7 @@ public class EhcacheStateServiceImpl implements EhcacheStateService {
     ServerSideConfiguration.Pool pool = resourcePageSource.getPool();
     OffHeapResource source = services.getService(OffHeapResourceIdentifier.identifier(pool.getServerResource()));
     if (source != null) {
+      unRegisterPoolStatistics(resourcePageSource);
       source.release(pool.getSize());
       LOGGER.info("Released {} bytes from resource '{}' for {} pool '{}'", pool.getSize(), pool.getServerResource(), poolType, poolName);
     }
@@ -313,11 +383,15 @@ public class EhcacheStateServiceImpl implements EhcacheStateService {
     PageSource resourcePageSource = getPageSource(name, serverStoreConfiguration.getPoolAllocation());
     ServerStoreImpl serverStore = new ServerStoreImpl(serverStoreConfiguration, resourcePageSource, mapper);
     stores.put(name, serverStore);
+
+    registerStoreStatistics(serverStore, name);
+
     return serverStore;
   }
 
   public void destroyServerStore(String name) throws ClusterException {
     final ServerStoreImpl store = stores.remove(name);
+    unRegisterStoreStatistics(store);
     if (store == null) {
       throw new InvalidStoreException("Clustered tier '" + name + "' does not exist");
     } else {
@@ -418,6 +492,10 @@ public class EhcacheStateServiceImpl implements EhcacheStateService {
 
     public ServerSideConfiguration.Pool getPool() {
       return pool;
+    }
+
+    public long getAllocatedSize() {
+      return delegatePageSource.getAllocatedSizeUnSync();
     }
 
     @Override
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
@@ -100,38 +100,68 @@ public class ServerStoreImpl implements ServerStore, MapInternals {
 
 
   @Override
-  public long getDataAllocatedMemory() {return store.getDataAllocatedMemory();}
+  public long getSize() {
+    return store.getSize();
+  }
 
   @Override
-  public long getAllocatedMemory() {return store.getAllocatedMemory();}
+  public long getTableCapacity() {
+    return store.getTableCapacity();
+  }
 
   @Override
-  public long getRemovedSlotCount() {return store.getRemovedSlotCount();}
+  public long getUsedSlotCount() {
+    return store.getUsedSlotCount();
+  }
 
   @Override
-  public long getDataVitalMemory() {return store.getDataVitalMemory();}
+  public long getRemovedSlotCount() {
+    return store.getRemovedSlotCount();
+  }
 
   @Override
-  public int getReprobeLength() {return store.getReprobeLength();}
+  public long getAllocatedMemory() {
+    return store.getAllocatedMemory();
+  }
 
   @Override
-  public long getDataSize() {return store.getDataSize();}
+  public long getOccupiedMemory() {
+    return store.getOccupiedMemory();
+  }
 
   @Override
-  public long getDataOccupiedMemory() {return store.getDataOccupiedMemory();}
+  public long getVitalMemory() {
+    return store.getVitalMemory();
+  }
 
   @Override
-  public long getUsedSlotCount() {return store.getUsedSlotCount();}
+  public long getDataAllocatedMemory() {
+    return store.getDataAllocatedMemory();
+  }
 
   @Override
-  public long getSize() {return store.getSize();}
+  public long getDataOccupiedMemory() {
+    return store.getDataOccupiedMemory();
+  }
 
   @Override
-  public long getVitalMemory() {return store.getVitalMemory();}
+  public long getDataVitalMemory() {
+    return store.getDataVitalMemory();
+  }
 
   @Override
-  public long getOccupiedMemory() {return store.getOccupiedMemory();}
+  public long getDataSize() {
+    return store.getDataSize();
+  }
 
   @Override
-  public long getTableCapacity() {return store.getTableCapacity();}
+  public int getReprobeLength() {
+    //TODO
+    //MapInternals Interface may need to change to implement this function correctly.
+    //Currently MapInternals Interface contains function: int getReprobeLength();
+    //however OffHeapServerStore.reprobeLength() returns a long
+    //Thus there could be data loss
+
+    throw new UnsupportedOperationException("Not supported yet.");
+  }
 }
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
@@ -32,6 +32,7 @@ import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
 
 import org.ehcache.clustered.client.internal.lock.VoltronReadWriteLockEntityClientService;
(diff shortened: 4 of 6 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -231,6 +232,14 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     } else if (source.reserve(pool.getSize())) {
       try {
         pageSource = new ResourcePageSource(pool);
+
+        registerStatistic(pageSource, poolName, "allocatedMemory" ,new Callable<Number>() {
+          @Override
+          public Number call() throws Exception {
+            return pageSource.delegatePageSource.getAllocatedSize();
```
mathieucarbou: you should use `getAllocatedSizeUnsync` to avoid blocking

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +251,15 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStatistic(Object context, String alias, String observerName, Callable callable) {
+    Set<String> tags = new HashSet<String>(Arrays.asList("ClusteredServerSide","tier"));
```
mathieucarbou: should replace ClusteredServerSide by `PageSource ` : and put a constant, like it is done for observers on client side.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -742,6 +743,11 @@ private void destroyServerStore(ClientDescriptor clientDescriptor, DestroyServer
     boolean isDuplicate = isLifeCycleMessageDuplicate(destroyServerStore);
 
     if (!isDuplicate) {
+      ServerStoreImpl store = ehcacheStateService.getStore(name);
+      if (store != null) {
+        StatisticsManager.nodeFor(store.getPageSource()).clean();
```
gibsong: I guess there is a small chance it could happen.  I added a null check here and also in EhcachePassiveEntity

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: I just updated it. I added a constant for "PageSource" and removed "ClusteredServerSide"

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -742,6 +744,14 @@ private void destroyServerStore(ClientDescriptor clientDescriptor, DestroyServer
     boolean isDuplicate = isLifeCycleMessageDuplicate(destroyServerStore);
 
     if (!isDuplicate) {
+      ServerStoreImpl store = ehcacheStateService.getStore(name);
```
mathieucarbou: I think this code should be moved into `ehcacheStateService.destroyServerStore(name);` 

It would be better to have all the code within the same class and classloader

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
```diff
@@ -222,6 +224,14 @@ private void destroyServerStore(DestroyServerStore destroyServerStore) throws Cl
 
     String name = destroyServerStore.getName();
 
+    ServerStoreImpl store = ehcacheStateService.getStore(name);
+    if (store != null) {
+      TreeNode node = StatisticsManager.nodeFor(store.getPageSource());
```
mathieucarbou: I think this code should be moved into ehcacheStateService.destroyServerStore(name);

It would be better to have all the code within the same class and classloader

gibsong:
Mathieu, yes i plan to add tests.  I can also do in a seperate PR if this needs to be merged ASAP.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
gibsong: So that is were I originally had the code, but the method never seemed to get called in my testing.  Although after talking with you yesterday now I know why it wasn't called.  I do agree the placement is better here.  I just made the change and tested it.  It seems to be working fine.

mathieucarbou:
no it's ok ;-) no urgency.

gibsong:
test please

gibsong:
please test

gibsong:
test this please

gibsong:
test this please

gibsong:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +247,105 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStatistics(String cacheAlias, PageSource pageSource) {
```
mathieucarbou: I think `cacheAlias ` should be renamed to `storeName` or `storeId`

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +247,105 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
...
+    });
+
+    registerStatistic(pageSource, cacheAlias, "dataOccupiedMemory" ,new Callable<Number>() {
+      @Override
+      public Number call() throws Exception {
+        return store.dataOccupiedMemory();
+      }
+    });
+
+    registerStatistic(pageSource, cacheAlias, "mappings" ,new Callable<Number>() {
```
mathieucarbou: perhaps much more `entries` or `size` instead of `mappings` ?

mathieucarbou:
You mixed up pool stats and store stats. pools and stores are 2 different concepts.

all the stats you registered are for stores, and they should be based on ServerStoreImpl 
And we should have 1 stat (allocatedMemory) for PageSource context (pool)

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -48,15 +47,21 @@
...
+import org.terracotta.context.TreeNode;
 import static org.terracotta.offheapstore.util.MemoryUnit.GIGABYTES;
 import static org.terracotta.offheapstore.util.MemoryUnit.MEGABYTES;
+import org.terracotta.statistics.StatisticsManager;
 
 public class EhcacheStateServiceImpl implements EhcacheStateService {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheStateServiceImpl.class);
 
+  private static final String STATISTICS_TAG = "PageSource";
```
mathieucarbou: should have 2 tags: one for stores and one for pools:
```
POOL_STATISTICS_TAG = "Pool"
STORE_STATISTICS_TAG = "ServerStore"
```

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
mathieucarbou: register__Store__Statistics instead and should take a ServerStoreImpl parameter, not a page source

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +247,105 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStatistics(String cacheAlias, PageSource pageSource) {
+
+    ServerStoreImpl store = stores.get(cacheAlias);
+
+    registerStatistic(pageSource, cacheAlias, "allocatedMemory" ,new Callable<Number>() {
```
mathieucarbou: store stats should be registered  with a ServerStoreImpl context and the storeName

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +247,105 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
...
+      }
+    });
+
+  }
+
+  private void registerStatistic(Object context, String alias, String observerName, Callable callable) {
+    Set<String> tags = new HashSet<String>(Arrays.asList(STATISTICS_TAG,"tier"));
+    Map<String, Object> properties = new HashMap<String, Object>();
+    properties.put("discriminator", STATISTICS_TAG);
+    properties.put("alias", alias);
```
mathieucarbou: change to 
`properties.put("storeName", alias);`
for stores, and
`properties.put("poolName", alias);`
for pools

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +247,105 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
...
+    registerStatistic(pageSource, cacheAlias, "tableCapacity" ,new Callable<Number>() {
+      @Override
+      public Number call() throws Exception {
+        return store.tableCapacity();
+      }
+    });
+
+  }
+
+  private void registerStatistic(Object context, String alias, String observerName, Callable callable) {
```
mathieucarbou: create one `registerStoreStatistic` (takes a SerevrStoreImpl as context) and one `registerPoolStatistic` (takes a ResourcePageSource as context). `alias` should be changed also to `poolname` and `storeName`

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i renamed to storeName

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i thought "mappings" was the keyword that was previously used?  I actually don't like it so I will change it to entries.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i added these 2 tags

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i made the change

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +247,105 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStatistics(String cacheAlias, PageSource pageSource) {
+
+    ServerStoreImpl store = stores.get(cacheAlias);
```
gibsong: i removed it

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i now register store stats with ServerStoreImpl and storeName

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i made this change

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i updated the code per your suggestions

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +250,139 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
...
+      @Override
+      public Number call() throws Exception {
+        return store.tableCapacity();
+      }
+    });
+
+  }
+
+  private void registerPoolStatistics(String poolName, ResourcePageSource pageSource) {
+    registerStatistic(pageSource, poolName, "allocatedMemory", STATISTICS_POOL_TAG, PROPERTY_POOL_KEY, new Callable<Number>() {
```
mathieucarbou: `allocatedMemory` is perhaps better to be renamed to `allocatedSize` because if we look at the store stats, `size` means something different than `allocated`... Could you check the code to be sure we can call it allocatedMemory ?

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +250,139 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
...
+        return pageSource.delegatePageSource.getAllocatedSizeUnSync();
+      }
+    });
+  }
+
+  private void unRegisterStoreStatistics(String storeName) throws InvalidStoreException {
+
+    ServerStoreImpl store = stores.get(storeName);
+    if (store == null) {
+      throw new InvalidStoreException("Clustered tier '" + storeName + "' does not exist");
```
mathieucarbou: not sure we should throw... perhaps just do nothing... @ljacomet ?

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: i actually think that allocatedSize is better because there is already a store stat called allocatedMemory.  This will avoid more confusion.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: in this case, i guess it doesn't really matter if we throw.  We can just let the code continue.  If we don't have access to the store instance then there is no way we can remove it from the shadow tree.  Since the store instance has been removed from the stores map it should get removed from the shadow tree when the store instance gets garbage collected.  i will remove this throw statement.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
```diff
@@ -43,11 +43,13 @@
 import org.terracotta.entity.PassiveServerEntity;
 import org.terracotta.entity.ServiceRegistry;
 import org.terracotta.offheapresource.OffHeapResources;
+import org.terracotta.statistics.StatisticsManager;
 
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import org.terracotta.context.TreeNode;
```
ljacomet: These strongly smell like dead import seeing there are no changes in the rest of the file ...

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +250,139 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStoreStatistics(String storeName) throws InvalidStoreException {
```
ljacomet: All of this should leverage Java 8 constructs and if still large be moved to a companion object.

For example, all the invocations of `registerStatistic` change by two params - the stat name and the method invoked in the callable. By having a map of these and writing some functional code, all this could be done in a more compact way.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -302,10 +444,16 @@ public ServerStoreImpl createStore(String name, ServerStoreConfiguration serverS
...
     stores.put(name, serverStore);
+
+    registerStoreStatistics(name);
+
     return serverStore;
   }
 
   public void destroyServerStore(String name) throws ClusterException {
+
+    unRegisterStoreStatistics(name);
```
ljacomet: You need to make sure the `EhcacheActiveEntity.destroy` method triggers this as well.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -289,6 +430,7 @@ private void releasePool(String poolType, String poolName, ResourcePageSource re
     ServerSideConfiguration.Pool pool = resourcePageSource.getPool();
     OffHeapResource source = services.getService(OffHeapResourceIdentifier.identifier(pool.getServerResource()));
     if (source != null) {
+      unRegisterPoolStatistics(resourcePageSource);
```
ljacomet: You need to make sure the `EhcacheActiveEntity.destroy` method triggers this as well.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -90,4 +90,30 @@ public void clear() {
   public void close() {
     store.close();
   }
+
+  // stats
+
+  public long allocatedMemory() {return store.allocatedMemory();}
```
ljacomet: Could these be added to a specific stats related interface that `ServerStoreImpl` would implement?

Also, changes to this file caused it to become executable - please revert that.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -222,4 +223,102 @@ public void close() {
     }
   }
 
+  // stats
+
+  public long allocatedMemory() {
```
ljacomet: Changes to this files caused it to become executable - please revert that.

build.gradle
```diff
@@ -125,7 +125,7 @@ subprojects {
   }
 
   test {
-    maxHeapSize = "1024m"
+    maxHeapSize = "2048m"
```
ljacomet: As discussed, let's try to not have to do that. Which means that most tests should properly destroy the entity when done.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
mathieucarbou: @ljacomet : I think the answer is yes because we could make `ServerStoreImpl` implements `MapInternals`, and then register stats over any `MapInternals` object.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
gibsong: I had made some changes here, but they have been reverted.  I guess I missed an import.  I will remove it.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: I actually call unRegisterPoolStatistics(resourcePageSource); in EhcacheStateServiceImpl.releasePool(...), which is called by EhcacheStateServiceImpl.destroy().  Is there a reason why I can't do it this way and need to do it from EhcacheActiveEntity.destroy()?

ljacomet: Scratch what I said ... again 😉 
Actually, this will be called when the entity is destroyed for shared pools, but not for dedicated pools AFAICT.
However, it will be properly called if a `ServerStore` using a dedicated pool is destroyed explicitly.
So there is still one path to cover - linked to my comment on `unregisterStoreStatistics`.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +250,139 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStoreStatistics(String storeName) throws InvalidStoreException {
+
+    ServerStoreImpl store = stores.get(storeName);
+    if (store == null) {
+      throw new InvalidStoreException("registerStoreStatistics, Clustered tier '" + storeName + "' does not exist");
```
ljacomet: I would not throw here - this would indicate a major bug.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
gibsong: @ljacomet :  yes Mathieu is right, I was able to make ServerStoreImpl implement MapInternals.  The only problem I encountered was that MapInternals.getReprobeLength() returns an int and OffHeapServerStore.reprobeLength() returns a long.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
ljacomet: Actually, do this differently. From the place where it is called, pass in the ServerStore directly, as retrieved from the `remove` call. That will resolve your problem of what to do if you cannot find it.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
ljacomet: Scratch what I said - just pass the `Store` instance into the method instead of a name that you have to look up. And then you do not even have to ask the question.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: I made the change

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
gibsong: fixed

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
gibsong: fixed file, removed exec permission

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: so unRegisterStoreStatistics(name); is called indirectly via EhcacheActiveEntity.destroy.  This is the call chain:
1. EhcacheActiveEntity.destroy()
2. ehcacheStateService.destroy()
3. releasePools(...)
4. releasePool(...)
5. unRegisterPoolStatistics(...);
Let me know if you want it done differently?

gibsong: I made sure it is called now.  It is called via the following call chain: 
1. EhcacheActiveEntity.destroy()
2. ehcacheStateService.destroy()
3. unRegisterStoreStatistics(...);

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: Just to clarify this is the call stack for calling unRegisterPoolStatistics(...);
1. EhcacheActiveEntity.destroy()
2. ehcacheStateService.destroy()
3. EhcacheStateServiceImpl.releasePools(...)
4. EhcacheStateServiceImpl.releasePool(...)
5. EhcacheStateServiceImpl.unRegisterPoolStatistics(...);

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
gibsong: I changed the permissions too

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: looking into this suggested change

build.gradle
gibsong: I was able to decrease the maxHeapSize to 1408m from 2048m.  This was due to the fix with the passthrough server.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: statistics are now registered via java 8 streams.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
ljacomet: Still is executable

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -242,6 +273,53 @@ private ResourcePageSource createPageSource(String poolName, ServerSideConfigura
     return pageSource;
   }
 
+  private void registerStoreStatistics(ServerStoreImpl store, String storeName) throws InvalidStoreException {
+    STAT_STORE_METHOD_REFERENCES.entrySet().stream().forEach((entry)->
+      registerStatistic(store, storeName, entry.getKey(), STATISTICS_STORE_TAG, PROPERTY_STORE_KEY, new Callable<Number>() {
```
ljacomet: That's much better - but can be even better: no need to explicitly create the `Callable`. Just invoking `entry.getValue().apply(store)` should allow it to appear as a `Callable`.

ljacomet:
Much better, including not having to bump test VMs to 2g. Still have two remarks though.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
gibsong: hmm... so weird.  I guess i missed it.  i thought I had changed it.  Anyways i just ran chmod 644 on it.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
gibsong: I am actually passing that Callable to the registerStatistic(...) method which then uses it to call:
StatisticsManager.createPassThroughStatistic(context, observerName, tags, properties, callable);
So I am pretty sure I need it, unless there is some magic I am not understanding with streams.

gibsong:
@ljacomet 
This code was rebased onto #1512 and #1555.  Please merge if everything looks good.

PR #1556, opened 2016-10-21
Issue 1517

AbfrmBlr:


AbfrmBlr:
tests are failing because lifecycle ops return before retired messages are processed on passive.Waiting on Terracotta-OSS/terracotta-core#324.

AbfrmBlr:
I will revert the last commit, if the version upgrade happens before this PR is merged. Its there so that the build passes.

AbfrmBlr:
test this please

ljacomet:
I would rather not have the upgrade commit included here.
The introduced method on `EntityServerService` really can't be left with their default impl for Ehcache.

AbfrmBlr:
The dependency needs to be there for PR build to pass.

AbfrmBlr:
Depends on #1561. The last needs to be removed before merge

ljacomet:
#1561 was merged - drop last commit and then rebase

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ReconnectData.java
```diff
@@ -48,7 +48,7 @@ public void add(String name) {
   }
 
   public void remove(String name) {
-    if (!reconnectData.contains(name)) {
+    if (reconnectData.contains(name)) {
       reconnectData.remove(name);
```
ljacomet: This can be simplified to:

```java
if (reconnectData.remove(name)) {
  reconnectDatalen.addAndGet(-(2 * name.length() + ENTRY_SIZE));
}
```

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -750,6 +750,11 @@ private void createServerStore(ClientDescriptor clientDescriptor, CreateServerSt
 
     serverStore.setEvictionListener(key -> invalidateHashAfterEviction(name, key));
     attachStore(clientDescriptor, name);
+    try {
+      entityMessenger.messageSelfAndDeferRetirement(createServerStore, new ClientIDTrackerMessage.ServerStoreLifeCycleReplicationMessage(createServerStore));
+    } catch (MessageCodecException e) {
+      LOGGER.error("Codec Exception", e);
```
ljacomet: Let's throw `AssertionError` as this should only be a programming error, nothing else. Same for the similar block below.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
```diff
@@ -186,4 +189,23 @@ public long getKey() {
       return key;
     }
   }
+
+  public static class ServerStoreLifeCycleReplicationMessage extends ClientIDTrackerMessage {
+
+    private final LifecycleMessage message;
```
ljacomet: Really not too fond of the wrap a message in a message approach. It seems lazy 😉

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
```diff
@@ -229,10 +250,8 @@ private void invokeLifeCycleOperation(LifecycleMessage message) throws ClusterEx
         trackAndApplyMessage(message);
         break;
       case CREATE_SERVER_STORE:
-        createServerStore((CreateServerStore) message);
-        break;
       case DESTROY_SERVER_STORE:
-        destroyServerStore((DestroyServerStore) message);
+        ehcacheStateService.getClientMessageTracker().track(message.getId(), message.getClientId());
```
ljacomet: Can you explain why tracking is needed here but not above?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
AbfrmBlr: Yes, I thought the one doing runnel integration will do this :smile:

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
ljacomet: Duh, I missed the fact that you removed the break ... ok, makes sense.

ljacomet:
Good to go once rebased

PR #1560, opened 2016-10-25
Closes #1525 Closes #1418 Making server side data strutures thread-safe

### LINKED ISSUE #1418: EhcacheActiveEntity.clientStateMap looks fishy for concurrent accesses
`EhcacheActiveEntity.clientStateMap` is a plain `Map` and thus may suffer from concurrency artefacts when used in the EhcacheActiveEntity.
Some of this may be answered by terracotta-oss/terracotta-apis#144.

Also, we seem to lookup the `ClientDescriptor` multiple times _per message_ which sounds pretty inefficient.

### LINKED ISSUE #1525: EhcacheStateServiceImpl is not thread safe
While the entity usage guarantees some protection because of the concurrency key mechanism, there still need to be thread safe with regards to messages happening on different concurrency keys.

For example, calls to `org.ehcache.clustered.server.EhcacheStateServiceImpl#getStore` will happen concurrently with `org.ehcache.clustered.server.EhcacheStateServiceImpl#createStore` but that means writing and reading to a non thread safe data structure.

albinsuresh:
Since most of the *state* of the classes in the change-set have data structures holding life-cycle related state, I didn't have to do much to address concurrent mutations as that will never happen since all the life-cycle operations are run on the same concurrency key. The only thing that I had to ensure was the visibility of these mutating operations. I could address that mostly by replacing non-thread safe data structures with the ones that provide visibility guarantees.

Now am not sure if I'm supposed to write tests for these changes. If yes, I'm gonna need some guidance on writing tests for these kinda changes.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
@@ -107,7 +107,7 @@ class EhcacheActiveEntity implements ActiveServerEntity<EhcacheEntityMessage, Eh
    * {@link #connected(ClientDescriptor)} method is invoked for a client and removed when the
    * {@link #disconnected(ClientDescriptor)} method is invoked for the client.
    */
-  private final Map<ClientDescriptor, ClientState> clientStateMap = new HashMap<>();
+  private final Map<ClientDescriptor, ClientState> clientStateMap = new ConcurrentHashMap<>();
 
   private final ConcurrentHashMap<String, Set<ClientDescriptor>> storeClientMap =
       new ConcurrentHashMap<>();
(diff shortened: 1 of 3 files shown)


ljacomet:
I will have a deeper look soon, but I believe what you did is sufficient. And I do not think we should try adding tests proving correctness for this.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -401,8 +401,11 @@ private void validateClientConnected(ClientDescriptor clientDescriptor) throws C
   }
 
   private void validateClientAttached(ClientDescriptor clientDescriptor) throws ClusterException {
-    validateClientConnected(clientDescriptor);
-    if (!clientStateMap.get(clientDescriptor).isAttached()) {
+    ClientState clientState = this.clientStateMap.get(clientDescriptor);
```
ljacomet: I would rather have another internal method that does what's in lines 398 to 400 and back in 406 to 407 but indeed takes the `ClientState` instead of looking it up - and being racy indeed.

ljacomet:
Good to go once build passes

PR #1555, opened 2016-10-21
Close Issue #1008 - Expose management data from ehcache server entity

mathieucarbou:
Depends on Terracotta-OSS/terracotta-platform#165

Depends on tc-platform release __5.0.8.beta3__

Close #1008

New events:

- EHCACHE_CLIENT_RECONNECTED
- EHCACHE_RESOURCE_POOLS_CONFIGURED
- EHCACHE_CLIENT_VALIDATED
- EHCACHE_SERVER_STORE_CREATED
- EHCACHE_SERVER_STORE_ATTACHED
- EHCACHE_SERVER_STORE_RELEASED
- EHCACHE_SERVER_STORE_DESTROYED

Exposes management descriptors for stores, client states, pools and server resources.

PR #1561, opened 2016-10-25
Fix #1559 Implement ExecutionStrategy

### LINKED ISSUE #1559: Adapt latest Terracotta version
This requires work around the `EntityServerService` to adopt the two added methods as the default implementation does not match Ehcache needs.

ljacomet:
* Bump to latest terracotta .9-beta
* ExecutionStrategy first cut
* Make reconfigure invalid for Ehcache

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,13 +30,13 @@ ext {
   // Clustered
   terracottaPlatformVersion = '5.0.8.beta2'
   managementVersion = terracottaPlatformVersion
-  terracottaApisVersion = '1.0.8.beta'
-  terracottaCoreVersion = '5.0.8-beta'
+  terracottaApisVersion = '1.0.9.beta'
+  terracottaCoreVersion = '5.0.9-beta2'
   offheapResourceVersion = terracottaPlatformVersion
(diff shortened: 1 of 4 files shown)


ljacomet:
I would like this to go after #1555 so that MnM version bumps can be propagated to server without the changes from .9-beta initially.

build.gradle
```diff
@@ -212,7 +212,7 @@ subprojects {
 allprojects {
   tasks.withType(JavaCompile) {
     options.encoding = 'UTF-8'
-    options.compilerArgs += ['-Xlint:unchecked']
+//    options.compilerArgs += ['-Xlint:unchecked']
```
ljacomet: That's a mistake

ljacomet:
Can be merged since #1555 is currently blocked by different bugs.

AbfrmBlr:
Can this PR also accommodate following version updates -

•	tc-passthrough-testing 1.0.9.beta2
•	terracotta-core 5.0.9-beta2
•	galvan 1.0.9-beta2

ljacomet:
Updated PR with latest terracotta versions

PR #1570, opened 2016-10-28
[master] Fix #1569 Proper ByteBuffer usage

### LINKED ISSUE #1569: Incorrect usage of ByteBuffer in Operations and ChainResolver
* In Operations, we work on the limit based of the capacity and not based of the set limit when receiving the `ByteBuffer`.
This makes the behaviour incorrect if the received `ByteBuffer` is a view on a larger `byte[]`.
* In ChainResolver, we `flip()` the `ByteBuffer` after decoding when matching a wrong key, but we need instead to `rewind()` otherwise we artificially set the limit to the position after the decoding.

This needs to be daggy fixed between `release/3.1` and `master`.

ljacomet:
* Set limit based on known limit instead of capacity
* rewind instead of flip after decoding


lorban:
LGTM

AbfrmBlr:
test this please

Issue #1567, opened 2016-10-27
Being able to pass Stats config to the server entity at creation / configuration time


mathieucarbou:
I am starting a discussion thread to resolve this, following PR #1573

mathieucarbou:
probably we won't need that because the stat config and the scheduler will be into a voltron service (Terracotta-OSS/terracotta-platform#179) so this will probably be the TMS entity that will configure the stat manager and the stats.

PR #1573, opened 2016-10-28
Resource usage stats from Ehcache server entity (close #1551 and close #1530)

### LINKED ISSUE #1530: Resource usage statistics
Implement server side statistics.  Add stats for server side resource usage.

Update by @mathieucarbou:
- [x] Implement the observers (PR #1552 )
- [ ] Implement the stats (like StandardEhcacheStatistics) on the management providers for serevr stores and pools (depends on #1551)

### LINKED ISSUE #1551: Being able to send statistics from the Ehcache server entity (active and passive)
Depends on #1008

mathieucarbou:
Close #1551 and close #1530.

**Important notes for Ehcache / M&M people:**

__THESE ARE NOT BLOCKERS for M&M: JUST THINGS TO TAKE CARE OF__

1. **TimeSource**

We need in management to get the server time in a uniform way both from ehcache server entity and within the monitoring service. Currently, we are calling: `System.currentTimeMillis()` because there is no time service in voltron.

See `StatisticCollectorManagementProvider` in this PR.

I started a discussion for that: Terracotta-OSS/terracotta-apis#167

1. **Scheduling tasks**

There is no way to schedule a task in voltron yet, so I needed to start a scheduled executor service (the thread is set to daemon) used only by management to compute stats.

See `Management` in this PR

I started an issue for that: Terracotta-OSS/terracotta-apis#158

1. **Statistics configuration**

We have currently no way to pass the stats config object to use server-side. One idea is that it could be an optional XML config in ehcache xml, sent when the entity is configured with the server config. For the moment, the stat config is then fixed and cannot be changed.

See `Management` and `StatisticConfiguration` in this PR

I started an issue for that: #1567 

1. **Management calls to server entities**

~~We have no way yet to issue a management call to the server entity to configure which stat to collect depending on what we need. So for the moment, all stats are collected.~~

~~See `Management` in this PR~~

~~I started an issue for that: Terracotta-OSS/terracotta-apis#168~~

__UPDATE:__ After discussion with @anthonydahanne and @lorban, I figured out we can do management calls already to exposed objects into the voltron registry service if we extend it.

See: #1575 and Terracotta-OSS/terracotta-platform#178

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
@@ -18,7 +18,6 @@ package org.ehcache.clustered.management;
 import org.ehcache.Cache;
 import org.ehcache.config.units.EntryUnit;
 import org.ehcache.config.units.MemoryUnit;
-import org.hamcrest.CoreMatchers;
 import org.junit.BeforeClass;
 import org.junit.FixMethodOrder;
 import org.junit.Ignore;
@@ -40,8 +39,8 @@ import org.terracotta.management.model.stats.history.CounterHistory;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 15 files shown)


mathieucarbou:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/management/PoolSettingsManagementProvider.java
```diff
@@ -69,7 +76,7 @@ public Context getContext() {
         Collections.singleton(new Settings(getContext())
           .set("serverResource", getBinding().getValue().getServerResource())
           .set("size", getBinding().getValue().getSize())
-          .set("allocationType", getBinding().getAllocationType()));
+          .set("allocationType", getBinding().getAllocationType().name().toLowerCase()));
```
anthonydahanne: cool ! I already got prepared to that change in the TMS

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -74,7 +74,7 @@
 
   static {
     STAT_STORE_METHOD_REFERENCES.put("allocatedMemory", ServerStoreImpl::getAllocatedMemory);
-    STAT_STORE_METHOD_REFERENCES.put("dataOccupiedMemory", ServerStoreImpl::getDataOccupiedMemory);
+    STAT_STORE_METHOD_REFERENCES.put("dataAllocatedMemory", ServerStoreImpl::getDataAllocatedMemory);
```
anthonydahanne: shouldn't there be a test to track the presence / absence of stats observers ?
could be done in a separate PR though

anthonydahanne:
several "hacks" are in place in this PR to make server side stats collection possible.
Those hacks are well documented and tracked in different issues.
LGTM
PS : please try and bring OffHeapResource stats in a next PR

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
mathieucarbou: There's one in the the integration test.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -353,6 +412,20 @@ public static void initDescriptors() throws ClassNotFoundException {
     CACHE_DESCRIPTORS.add(new StatisticDescriptor("Cache:MissCount", StatisticType.COUNTER_HISTORY));
     CACHE_DESCRIPTORS.add(new StatisticDescriptor("Cache:MissRatioRatio", StatisticType.RATIO_HISTORY));
 
+    POOL_DESCRIPTORS.add(new StatisticDescriptor("Pool:AllocatedSize", StatisticType.SIZE_HISTORY));
+
+    SERVER_STORE_DESCRIPTORS.add(new StatisticDescriptor("Store:AllocatedMemory", StatisticType.SIZE_HISTORY));
+    SERVER_STORE_DESCRIPTORS.add(new StatisticDescriptor("Store:DataAllocatedMemory", StatisticType.SIZE_HISTORY));
+    SERVER_STORE_DESCRIPTORS.add(new StatisticDescriptor("Store:OccupiedMemory", StatisticType.SIZE_HISTORY));
+    SERVER_STORE_DESCRIPTORS.add(new StatisticDescriptor("Store:DataOccupiedMemory", StatisticType.SIZE_HISTORY));
```
mathieucarbou: Done: see `test_D_server_capabilities_exposed` and `test_G_stats_collection`

mathieucarbou:
Yes, I'll add an observer for `OffHeapResource.remaining()`. Just need to see first if it is ok to make offheap lib depends on stats, or if it is preferable to duplicate he observers per server entity.

PR #1578, opened 2016-11-01
Issue 1454 Statistics Count and Ratio testing

gibsong:
This PR includes Count(hit and miss) and Ratio(hit and miss) tests on every tier and at the cache level.


clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteredStatisticsRatioTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.terracotta.management.model.call.ContextualReturn;
+import org.terracotta.management.model.stats.ContextualStatistics;
+import org.terracotta.management.model.stats.Sample;
+import org.terracotta.management.model.stats.history.RatioHistory;
+
+
+/**
+ *
+ *
+ */
```
mathieucarbou: perhaps remove this empty javadoc section ?

management/src/test/java/org/ehcache/management/providers/statistics/HitCountTest.java
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.Timeout;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.terracotta.management.model.context.Context;
+
+/**
+ *
+ *
+ */
```
mathieucarbou: perhaps remove this empty javadoc section ?

management/src/test/java/org/ehcache/management/providers/statistics/HitCountTest.java
```diff
@@ -0,0 +1,150 @@
...
+    Configuration cacheConfiguration = new DefaultConfiguration(EvictionTest.class.getClassLoader(),
+            new DefaultPersistenceConfiguration(diskPath.newFolder()));
+
+    Collection<Service> services = new ArrayList<Service>();
+    services.add(managementRegistry);
+
+    CacheManager cacheManager = null;
+
+    try {
+      cacheManager = new EhcacheManager(cacheConfiguration, services);
```
mathieucarbou: why not using the traditional builder api to create the cache manager ?

management/src/test/java/org/ehcache/management/providers/statistics/HitRatioTest.java
```diff
@@ -0,0 +1,183 @@
...
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.Timeout;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.terracotta.management.model.context.Context;
+
+/**
+ *
+ *
+ */
```
mathieucarbou: perhaps remove this empty javadoc section ?

management/src/test/java/org/ehcache/management/providers/statistics/HitRatioTest.java
```diff
@@ -0,0 +1,183 @@
...
+    Configuration cacheConfiguration = new DefaultConfiguration(EvictionTest.class.getClassLoader(),
+            new DefaultPersistenceConfiguration(diskPath.newFolder()));
+
+    Collection<Service> services = new ArrayList<Service>();
+    services.add(managementRegistry);
+
+    CacheManager cacheManager = null;
+
+    try {
+      cacheManager = new EhcacheManager(cacheConfiguration, services);
```
mathieucarbou: why not using the traditional builder api to create the cache manager ?

management/src/test/java/org/ehcache/management/providers/statistics/MissCountTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.Timeout;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.terracotta.management.model.context.Context;
+
+/**
+ *
+ *
+ */
```
mathieucarbou: empty comment

management/src/test/java/org/ehcache/management/providers/statistics/MissRatioTest.java
```diff
@@ -0,0 +1,173 @@
...
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.Timeout;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.terracotta.management.model.context.Context;
+
+/**
+ *
+ *
+ */
```
mathieucarbou: empty comment

management/src/test/java/org/ehcache/management/providers/statistics/MissRatioTest.java
```diff
@@ -0,0 +1,173 @@
...
+    Configuration cacheConfiguration = new DefaultConfiguration(EvictionTest.class.getClassLoader(),
+            new DefaultPersistenceConfiguration(diskPath.newFolder()));
+
+    Collection<Service> services = new ArrayList<Service>();
+    services.add(managementRegistry);
+
+    CacheManager cacheManager = null;
+
+    try {
+      cacheManager = new EhcacheManager(cacheConfiguration, services);
```
mathieucarbou: same comment: why not using the builder api ?

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
```diff
@@ -68,4 +76,64 @@ public static Context createContext(ManagementRegistryService managementRegistry
...
+
+      RatioHistory ratioHistory = statisticsContext.getStatistic(RatioHistory.class, statName);
+
+      if (ratioHistory.getValue().length > 0) {
+        int mostRecentIndex = ratioHistory.getValue().length - 1;
+        value = ratioHistory.getValue()[mostRecentIndex].getValue();
+        System.out.println("statName: " + statName + " value: " + value + " expectedResult: " + expectedResult);
+      }
+
+    }while(value != expectedResult);
```
mathieucarbou: same remark:
If the expected value is not in the list of samples, at the latest position, this could lead to an infinite loop ?
should you instead do a sort of for loop to iterate over the sample values from the end to the start ?

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteredStatisticsCountTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Test;
+import org.terracotta.management.model.call.ContextualReturn;
+import org.terracotta.management.model.stats.ContextualStatistics;
+import org.terracotta.management.model.stats.Sample;
+import org.terracotta.management.model.stats.history.CounterHistory;
+
+/**
+ *
+ *
+ */
```
gibsong: i will remove empty javadoc section

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteredStatisticsRatioTest.java
gibsong: i will remove empty javadoc section

management/src/test/java/org/ehcache/management/providers/statistics/HitCountTest.java
gibsong: i will remove empty javadoc section

management/src/test/java/org/ehcache/management/providers/statistics/MissCountTest.java
gibsong: i will remove empty comment

management/src/test/java/org/ehcache/management/providers/statistics/MissRatioTest.java
gibsong: i will remove empty javadoc section

management/src/test/java/org/ehcache/management/providers/statistics/HitCountTest.java
gibsong: You are right, that would make more sense since the management tests should use the api.  I will change it.

management/src/test/java/org/ehcache/management/providers/statistics/MissCountTest.java
```diff
@@ -0,0 +1,143 @@
...
+    Configuration cacheConfiguration = new DefaultConfiguration(EvictionTest.class.getClassLoader(),
+            new DefaultPersistenceConfiguration(diskPath.newFolder()));
+
+    Collection<Service> services = new ArrayList<Service>();
+    services.add(managementRegistry);
+
+    CacheManager cacheManager = null;
+
+    try {
+      cacheManager = new EhcacheManager(cacheConfiguration, services);
```
gibsong: You are right, that would make more sense since the management tests should use the api. I will change it.

management/src/test/java/org/ehcache/management/providers/statistics/MissRatioTest.java
gibsong: You are right, that would make more sense since the management tests should use the api. I will change it.

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
```diff
@@ -68,4 +76,64 @@ public static Context createContext(ManagementRegistryService managementRegistry
...
+
+      CounterHistory counterHistory = statisticsContext.getStatistic(CounterHistory.class, statName);
+
+      if (counterHistory.getValue().length > 0) {
+        int mostRecentIndex = counterHistory.getValue().length - 1;
+        value = counterHistory.getValue()[mostRecentIndex].getValue();
+        System.out.println("statName: " + statName + " value: " + value + " expectedResult: " + expectedResult);
+      }
+
+    }while(value != expectedResult);
```
gibsong: I don't know if that would be the best idea,, that could cause a test to pass when it shouldn't.  For instance if the samples are 0,0,0,1,2,3 and I am expecting 2 then this would find 2 but 3 is the actual value, which is incorrect.  If the latest value isn't the expected value then the code should wait for the correct value, and if it is never the expected value it should eventually fail.  All the tests that use this method have timeouts.  Does that make sense?

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
gibsong: I don't know if that would be the best idea,, that could cause a test to pass when it shouldn't. For instance if the samples are 0,0,0,1,2,3 and I am expecting 2 then this would find 2 but 3 is the actual value, which is incorrect. If the latest value isn't the expected value then the code should wait for the correct value, and if it is never the expected value it should eventually fail. All the tests that use this method have timeouts. Does that make sense?

gibsong: I added a comment to let others know to add timeouts to their tests if using this method.

mathieucarbou:
Good to me! We wait for the build to pass and I'll merge!

gibsong:
test this please

Issue #1475, opened 2016-09-30
Missing package version in Export-Package header


nenko-tabakov:
The MANIFEST.MF in ehcache-3.1.2 and ehcache-3.1.3 does not provide package version in the Export-Package header. While this won't prevent the jar from resolving in OSGI environment it might cause confusion and/or bad linking if a consumer bundle has versions in their Import-Package header.

henri-tremblay:
Looking at the `MANIFEST.MF`, the exported packages should look like `org.ehcache;version="3.1.3"`

ljacomet:
Looking at OSGi documentation, we should try to leverage http://bnd.bndtools.org/chapters/170-versioning.html and see if it works that way.

Of course we may then have to do some filtering on the files to apply the version from the project build file ...

chrisdennis:
I think that the move to the `bnd` plugin under #2562 should handle this.  I've assigned it to myself and will close when that is confirmed.

chrisdennis:
Resolved by merging of #2562

PR #1587, opened 2016-11-04
Issue #1475: include package version in Export-Packages in the manif…

kcleereman:
…est for the client ehcache library


mathieucarbou:
test this please

ljacomet:
As discussed with @kcleereman there is no way we are going to do that kind of approach. Commented on the issue to try finding another way.

PR #1589, opened 2016-11-07
Closes #1577 Align the concurrency keys

### LINKED ISSUE #1577: Missing cache entries on mirror.
This happens when replicaion & sync operations are mixed up by bringing up mirror when series of mutations are progressing in active. It can be reproduced by the below test case. Here are the steps performed by test.

1)Start one server to become active
2)Perform put operations in multiple threads
3)Start mirror while put operations are progressing
4)Failover and verify the entries on new active which doesn't match with the entries on original active.

    //start one server first
    control.start(control.getTerracottaServer(0));
    //Create cache manager
    LOG.info("Started one server: " + control.getTerracottaServer(0).getServerSymbolicName());
    CacheManagerBuilder<PersistentCacheManager> builder = CacheManagerBuilder.newCacheManagerBuilder()
        .with(ClusteringServiceConfigurationBuilder.cluster(clusterURI)
            .autoCreate()
            .resourcePool("resource-pool-a", 128, MemoryUnit.MB)
            .resourcePool("resource-pool-b", 128, MemoryUnit.MB, "secondary-server-resource"));
    PersistentCacheManager cacheManager = builder.build(true);
    //Create cache
    final Cache<Long, String> cache = cacheManager.createCache("cache1", CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
            .with(ClusteredResourcePoolBuilder.clusteredDedicated("primary-server-resource", 128, MemoryUnit.MB))));

    //Spawn threads to perform put operations and start mirror in the middle of operations
    final int numEntriesActive = 3000;
    final int threads = 3;
    Callable<Void> task = new Callable<Void>() {
      @Override
      public Void call() throws Exception {
        for (; ; ) {
          int key = counter.incrementAndGet();
          if (key > numEntriesActive) {
            break;
          }
          cache.put((long)key, "value");
          Thread.sleep(5);
        }
        return null;
      }
    };
    ExecutorService executorService = Executors.newFixedThreadPool(threads);
    Collection<Future<Void>> futures = new ArrayList<Future<Void>>();
    for (int i = 0; i < threads; ++i) {
      futures.add(executorService.submit(task));
    }
    //start mirror while put operations are progressing on active`
    control.start(control.getTerracottaServer(1));

    for (Future<Void> f : futures) {
      f.get();
    }
    executorService.shutdown();

    //stop active so mirror takes over and compare the number of entries there on mirror
    control.stop(control.getTerracottaServer(0));
    int numEntriesOnMirror = 0;
    for (int i = 1; i <= numEntriesActive; ++i) {
      if (cache.get((long)i) != null) {
        ++numEntriesOnMirror;
      }
    }
    //this fails as number of entries in mirror less than expected.
    assertThat(numEntriesOnMirror).isEqualTo(numEntriesActive);

albinsuresh:
of server store operation messages and replicated messages

### DIFF
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
@@ -132,7 +132,7 @@ public abstract class PassiveReplicationMessage extends EhcacheEntityMessage {
 
     @Override
     public long concurrencyKey() {
-      return (this.cacheId.hashCode() + key);
+      return key;
     }
   }
 
--- a/clustered/server/src/test/java/org/ehcache/clustered/server/EhcacheActiveEntityTest.java
+++ b/clustered/server/src/test/java/org/ehcache/clustered/server/EhcacheActiveEntityTest.java
@@ -30,6 +30,7 @@ import org.ehcache.clustered.common.internal.exceptions.LifecycleException;
 import org.ehcache.clustered.common.internal.exceptions.ResourceBusyException;
 import org.ehcache.clustered.common.internal.exceptions.ResourceConfigurationException;
 import org.ehcache.clustered.common.internal.exceptions.ServerMisconfigurationException;
+import org.ehcache.clustered.common.internal.messages.ConcurrentEntityMessage;
 import org.ehcache.clustered.common.internal.messages.EhcacheEntityMessage;
 import org.ehcache.clustered.common.internal.messages.EhcacheEntityResponse;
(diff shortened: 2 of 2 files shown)


clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
```diff
@@ -132,7 +132,7 @@ public ReplicationOp operation() {
 
     @Override
     public long concurrencyKey() {
-      return (this.cacheId.hashCode() + key);
+      return (key);
```
ljacomet: Get rid of the not needed parenthesis

albinsuresh:
Addressed the review feedback.

PR #1599, opened 2016-11-08
Issue #1430: Clean up compiler warnings

ljacomet:
This PR takes care of all main and test source compiler warnings for the `api`, `core`, `core-spi-test` and `impl` modules.

I recommend focusing the review on the main source changes as they are the one that could have an impact I missed.

It also makes any new warning in any of these module an error thus preventing any to be introduced by accident.

### DIFF
--- a/core/src/test/java/org/ehcache/core/EhcacheBasicRemoveAllTest.java
+++ b/core/src/test/java/org/ehcache/core/EhcacheBasicRemoveAllTest.java
@@ -259,14 +259,15 @@ public class EhcacheBasicRemoveAllTest extends EhcacheBasicCrudBase {
   }
 
   @Test
+  @SuppressWarnings("unchecked")
   public void removeAllStoreCallsMethodTwice() throws Exception {
-    this.store = mock(Store.class);
-    CacheLoaderWriter cacheLoaderWriter = mock(CacheLoaderWriter.class);
+    CacheLoaderWriter<String, String> cacheLoaderWriter = mock(CacheLoaderWriter.class);
     final List<String> removed = new ArrayList<String>();
     doAnswer(new Answer() {
       @Override
       public Object answer(InvocationOnMock invocation) throws Throwable {
-        Iterable<String> i = (Iterable) invocation.getArguments()[0];
+        @SuppressWarnings("unchecked")
+        Iterable<String> i = (Iterable<String>) invocation.getArguments()[0];
         for (String key : i) {
           removed.add(key);
         }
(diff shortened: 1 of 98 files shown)


core/src/test/java/org/ehcache/core/EhcacheBasicRemoveAllTest.java
```diff
@@ -275,13 +276,13 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
-    final ArgumentCaptor<Function> functionArgumentCaptor = ArgumentCaptor.forClass(Function.class);
+    final ArgumentCaptor<Function<Iterable<? extends Map.Entry<? extends String, ? extends String>>, Iterable<? extends Map.Entry<? extends String, ? extends String>>>> functionArgumentCaptor = (ArgumentCaptor) ArgumentCaptor.forClass(Function.class);
 
     when(store.bulkCompute(anySet(), functionArgumentCaptor.capture())).then(new Answer<Object>() {
       @Override
       public Object answer(InvocationOnMock invocation) throws Throwable {
-        Function function = functionArgumentCaptor.getValue();
-        Iterable arg = new HashMap((Map) function.getClass().getDeclaredField("val$entriesToRemove").get(function)).entrySet();
+        Function<Iterable<? extends Map.Entry<? extends String, ? extends String>>, Iterable<? extends Map.Entry<? extends String, ? extends String>>> function = functionArgumentCaptor.getValue();
+        Iterable<? extends Map.Entry<? extends String, ? extends String>> arg = new HashMap<String, String>((Map) function.getClass().getDeclaredField("val$entriesToRemove").get(function)).entrySet();
```
henri-tremblay: Should be `Map<String, String>` I guess

henri-tremblay:
I have flag some cases (but there are more) where we cast to a rawtype. If we want to prevent to use any rawtype, you have to use the cast to wildcard - cast to new type trick. I can survive either way.

Otherwise, LGTM and awesome work!

ljacomet:
I believe that the cast is the real issue but sometimes we cannot work around it. I certainly do not mind the use of raw types in tests. Maybe I could track the places where they are used in source.

However, the build is failing for another reason. Looks like on the CI environment, we get warnings about use of internal APIs, `Unsafe` namely and that is something you can't suppress. So the sad bit is that I have no other option than to turn off the error mode from the `impl` module.

henri-tremblay:
It seems you can try this ` javac -XDignore.symbol.file`. Another solution would be to fail only in Jenkins using a plugin looking at the warnings. And keeping only the `-Xlint:unchecked`

ljacomet:
Not really fond of using `-XDignore.symbol.file`. So instead keeping the unchecked instruction, but not the error on warnings.

Pushed an update version that fixes some variable extraction I did which are not legal in Mockito.

ljacomet:
test this please

PR #1610, opened 2016-11-15
Issue #1482: Use of runnel for client -> server messages

ljacomet:
This PR introduces runnel in Ehcache and covers all client -> server messages and the replication messages from active -> passive.

Follow up PRs will take care or responses and passive sync messages.

### DIFF
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
@@ -19,93 +19,73 @@ package org.ehcache.clustered.common.internal.messages;
 import org.ehcache.clustered.common.internal.store.Chain;
 import org.ehcache.clustered.common.internal.store.Element;
 import org.ehcache.clustered.common.internal.store.SequencedElement;
+import org.ehcache.clustered.common.internal.store.Util;
+import org.terracotta.runnel.Struct;
+import org.terracotta.runnel.StructBuilder;
+import org.terracotta.runnel.decoding.StructArrayDecoder;
+import org.terracotta.runnel.decoding.StructDecoder;
+import org.terracotta.runnel.encoding.StructArrayEncoder;
+import org.terracotta.runnel.encoding.StructEncoder;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 
-import static org.ehcache.clustered.common.internal.store.Util.getElement;
-import static org.ehcache.clustered.common.internal.store.Util.getChain;
-
 public class ChainCodec {
 
-  private static final byte NON_SEQUENCED_CHAIN = 0;
-  private static final byte SEQUENCED_CHAIN = 1;
-  private static final byte SEQ_NUM_OFFSET = 8;
-  private static final byte ELEMENT_PAYLOAD_OFFSET = 4;
+  private static final Struct ELEMENT_STRUCT = StructBuilder.newStructBuilder()
+    .int64("sequence", 10)
+    .byteBuffer("payload", 20)
+    .build();
 
-  //TODO: optimize too many bytebuffer allocation
+  static final Struct CHAIN_STRUCT = StructBuilder.newStructBuilder()
+    .structs("elements", 10, ELEMENT_STRUCT)
+    .build();
+
   public byte[] encode(Chain chain) {
-    ByteBuffer msg = null;
-    boolean firstIteration = true ;
+    StructEncoder encoder = CHAIN_STRUCT.encoder();
+
+    encode(encoder, chain);
+
+    ByteBuffer byteBuffer = encoder.encode();
+    return byteBuffer.array();
+  }
+
+  public void encode(StructEncoder encoder, Chain chain) {
+    StructArrayEncoder elementsEncoder = encoder.structs("elements");
     for (Element element : chain) {
-      if (firstIteration) {
-        firstIteration = false;
-        ByteBuffer buffer = ByteBuffer.allocate(1);
-        if (element instanceof SequencedElement) {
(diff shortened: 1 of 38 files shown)


ljacomet:
Unrelated failures, unstable stats tests

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
```diff
@@ -16,14 +16,15 @@
...
+import org.ehcache.clustered.common.internal.ServerStoreConfiguration;
 import org.ehcache.clustered.common.internal.store.Chain;
 
 import java.util.UUID;
 
 /**
  * This message is sent by the Active Entity to Passive Entity.
  */
-public abstract class PassiveReplicationMessage extends EhcacheEntityMessage {
+public abstract class PassiveReplicationMessage extends EhcacheOperationMessage {
```
AbfrmBlr: This class has findbugs failures

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
```diff
@@ -156,6 +158,16 @@ public static Object unmarshall(byte[] payload) {
     }
   }
 
+  public static Object unmarshall(ByteBuffer payload) {
```
AbfrmBlr: This method is not being used anywhere. This class has findbugs failures as well

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.terracotta.runnel.EnumMappingBuilder.newEnumMappingBuilder;
+
+/**
+ * EhcacheMessageType
+ *
+ * Whenever you edit this, you must think about enum mapping and helper methods
+ */
+public enum EhcacheMessageType {
+  // Lifecycle messages
+  CONFIGURE,
```
AbfrmBlr: Would it be possible to reuse these enums as operation enums in messages ? Right now we end up maintaining same set of enums at multiple places.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessageCodec.java
```diff
@@ -16,111 +16,256 @@
...
+import static org.ehcache.clustered.common.internal.messages.MessageCodecUtils.STORE_CONFIG_KEY_TYPE_FIELD;
+import static org.ehcache.clustered.common.internal.messages.MessageCodecUtils.STORE_CONFIG_VALUE_SERIALIZER_TYPE_FIELD;
+import static org.ehcache.clustered.common.internal.messages.MessageCodecUtils.STORE_CONFIG_VALUE_TYPE_FIELD;
+import static org.terracotta.runnel.StructBuilder.newStructBuilder;
 
+
+// TODO move all this to server side - no use in common
 class PassiveReplicationMessageCodec {
 
   private static final byte OP_CODE_SIZE = 1;
```
AbfrmBlr: Unused static variables

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessage.java
ljacomet: Fixing

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
ljacomet: Fixing the windbags though

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
ljacomet: I am deprecating the other enums, just did not do the clean up yet.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/PassiveReplicationMessageCodec.java
ljacomet: Removing

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
ljacomet: The reason I can't do the clean up yet is that sync messages are also `EhcacheEntityMessage`. However I had missed some call sites that could be cleaned up already - added them to the PR.

ljacomet:
Looks like my latest changes have a bug ... will be looking into that.

ljacomet:
Fixed the test assertions following message `toString` changes

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
```diff
@@ -19,93 +19,96 @@
...
+import org.terracotta.runnel.encoding.StructEncoder;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 
-import static org.ehcache.clustered.common.internal.store.Util.getElement;
-import static org.ehcache.clustered.common.internal.store.Util.getChain;
-
 public class ChainCodec {
```
albinsuresh: There is some fair amount of duplicated code between both the encode methods and decode methods. Is it possible to avoid that?

ljacomet: Indeed - fixed

PR #1612, opened 2016-11-15
[master] Fix start of file inclusion

ljacomet:


docs/src/docs/asciidoc/user/serializers-copiers.adoc
```diff
@@ -109,7 +109,7 @@ Implement the following interface, from package `org.ehcache.spi.serialization`:
 
 [source,java,indent=0]
 ----
-include::{sourcedir32}/api/src/main/java/org/ehcache/spi/serialization/Serializer.java[lines=21..-1]
+include::{sourcedir32}/api/src/main/java/org/ehcache/spi/serialization/Serializer.java[lines=20..-1]
```
lorban: Wouldn't it be wiser to add some "asciidoc tag comments" in the Serializer.java class file and be done with updating this file forever?

ljacomet: For some reason, I do not like having these tags leak in API source files ... can other chime in? @AbfrmBlr @albinsuresh @henri-tremblay ?

chrisdennis: It's an API file... not going to change any time soon.  To me there are two better fixes:

1. Write a Java syntax aware IncludeProcessor for AsciidoctorJ
2. Get rid of the include entirely (I think it's overkill).

chrisdennis: Also, point of process... is the result of this vote binding?

ljacomet: Given the votes cast, leaving it like that and not using tags in API files.

PR #1613, opened 2016-11-15
Closes #1605 Eventual proxy to send acks to server

### LINKED ISSUE #1605: Eventual consistency memory leak
When performing a 90/10 get/put test using an eventually consistent clustered store with two or more clients, we have a memory leak in the "clientsWaitingForInvalidation" map in the "org.ehcache.clustered.server.EhcacheActiveEntity" object.  We do not run into this problem with only one client or when using a strongly consistent clustered store with any number of clients; it appears that the client may not be sending invalidation acks in an eventually consistent store which means that their entries are never removed from the "clientsWaitingForInvalidation" map.

The cache consisted of 16gb of clustered storage, Eventual consistency.  The server ran with 2gb min and max ram and the ConcurrentMarkSweep collector.  Contact me for test code / server config as well as a heap dump - github isn't allowing me to attach files.

AbfrmBlr:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/NoInvalidationServerStoreProxy.java
```diff
@@ -20,21 +20,92 @@
...
 
 import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeoutException;
 
 /**
  * Provides client-side access to the services of a {@code ServerStore}.
  */
 class NoInvalidationServerStoreProxy implements ServerStoreProxy {
```
ljacomet: Clearly the name of the class no longer fits. Can you update to something like `CommonServerStoreProxy` or something else describing what it does?

AbfrmBlr:
Addressed review comments

AbfrmBlr:
unrelated test failure

AbfrmBlr:
Rebased

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/CommonServerStoreProxy.java
```diff
@@ -0,0 +1,218 @@
...
+  public void addInvalidationListener(InvalidationListener listener) {
+    invalidationListeners.add(listener);
+  }
+
+  @Override
+  public boolean removeInvalidationListener(InvalidationListener listener) {
+    return invalidationListeners.remove(listener);
+  }
+
+  @SuppressWarnings("unchecked")
```
ljacomet: This suppress warnings can be removed by introducing a generic type on the method declaration:
`<T extends EhcacheEntityResponse>` which allows the type of both to be bound to one and thus no longer require the cast in line 139

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/CommonServerStoreProxy.java
```diff
@@ -0,0 +1,218 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(CommonServerStoreProxy.class);
+
+  private final ServerStoreMessageFactory messageFactory;
+  private final EhcacheClientEntity entity;
+
+  private final List<InvalidationListener> invalidationListeners = new CopyOnWriteArrayList<InvalidationListener>();
+  private final Map<Class<? extends EhcacheEntityResponse>, EhcacheClientEntity.ResponseListener<? extends EhcacheEntityResponse>> responseListeners
+      = new ConcurrentHashMap<Class<? extends EhcacheEntityResponse>, EhcacheClientEntity.ResponseListener<? extends EhcacheEntityResponse>>();
+
+  @SuppressWarnings("unchecked")
```
ljacomet: This seems to be needed just for the last lines of this method. Please extract the lines to have a much smaller scope for the suppression.

AbfrmBlr:
comments taken care of

kcleereman:
I've confirmed that this PR fixes the memory leak

PR #1616, opened 2016-11-15
:books: Replace non existing RuntimeCacheConfiguration with CacheRunt…

anthonydahanne:
…imeConfiguration


ljacomet:
and the convention is to use :memo: for doc updates

PR #1625, opened 2016-11-16
Close #1624: fixes randomly failing ratio unit tests

### LINKED ISSUE #1624: Ratio tests failing
The HitRatioTest and MissRatioTest tests fail randomly, and end up timing out while waiting for the right expected value.

gibsong:
I added a Thread.sleep after the get calls to allow the ratio stats to be computed.

This is the original task:
https://github.com/ehcache/ehcache3/issues/1624

### DIFF
--- a/management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
+++ b/management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
@@ -99,14 +99,14 @@ public class StatsUtil {
          This should only occur if the stats value is different from your expectedResult, which may happen if the stats calculations
          change, the stats value isn't accessible or if you enter the wrong expectedResult.
   */
-  public static double getExpectedValueFromRatioHistory(String statName, Context context, ManagementRegistryService managementRegistry, double expectedResult) {
+  public static Double getExpectedValueFromRatioHistory(String statName, Context context, ManagementRegistryService managementRegistry, Double expectedResult) {
 
     StatisticQuery query = managementRegistry.withCapability("StatisticsCapability")
       .queryStatistics(Arrays.asList(statName))
       .on(context)
       .build();
 
-    double value = 0;
+    Double value = 0d;
     do {
       ResultSet<ContextualStatistics> counters = query.execute();
 
@@ -120,7 +120,7 @@ public class StatsUtil {
         int mostRecentIndex = ratioHistory.getValue().length - 1;
         value = ratioHistory.getValue()[mostRecentIndex].getValue();
       }
-    } while (!Thread.currentThread().isInterrupted() && value != expectedResult);
+    } while (!Thread.currentThread().isInterrupted() && !value.equals(expectedResult));
 
     assertThat(value, Matchers.is(expectedResult));
 
@@ -149,7 +149,7 @@ public class StatsUtil {
         .getStatistics();
 
       for (Map.Entry<String, Statistic<?, ?>> entry : statistics.entrySet()) {
-        if (((StatisticHistory<?, ?>) entry.getValue()).getValue().length == 0) {
+        if (((StatisticHistory<?, ?>) entry.getValue()).getValue().length < 2) {
           noSample = true;
           break;
         }
--- a/management/src/test/java/org/ehcache/management/providers/statistics/HitRatioTest.java
+++ b/management/src/test/java/org/ehcache/management/providers/statistics/HitRatioTest.java
@@ -170,7 +170,7 @@ public class HitRatioTest {
         Assert.assertThat(tierHitRatio, is(tierExpectedValues.get(i)));
       }
 
-      double hitRatio = StatsUtil.getExpectedValueFromRatioHistory("Cache:HitRatio", context, managementRegistry, cacheExpectedValue);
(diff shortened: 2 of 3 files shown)


management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
```diff
@@ -120,7 +120,7 @@ public static double getExpectedValueFromRatioHistory(String statName, Context c
         int mostRecentIndex = ratioHistory.getValue().length - 1;
         value = ratioHistory.getValue()[mostRecentIndex].getValue();
       }
-    } while (!Thread.currentThread().isInterrupted() && value != expectedResult);
+    } while (!Thread.currentThread().isInterrupted() && !value.equals(expectedResult));
```
ljacomet: Instead of having a major single sleep in the test method, I would rather have here a back off strategy of sleeping more and more, from 10ms up to 500ms at a time to break the tight loop.

chrisdennis:
```java
public static <T> void assertBy(long time, TimeUnit unit, Callable<T> value, Matcher<? super T> matcher) {
    boolean interrupted = false;
    long end = System.nanoTime() + unit.toNanos(time);
    try {
        for (long sleep = 10; ; sleep <<= 1L) {
            try {
                Assert.assertThat(value.call(), matcher);
                return;
            } catch (Throwable t) {
                //ignore - wait for timeout
            }

            long remaining = end - System.nanoTime();
            if (remaining <= 0) {
                break;
            } else {
                try {
                    Thread.sleep(Math.min(sleep, TimeUnit.NANOSECONDS.toMillis(remaining) + 1));
                    interrupted = true;
                }
            }
        }
    } finally {
        if (interrupted) {
            Thread.currentThread().interrupt();
        }
    }
    try {
        Assert.assertThat(value.call(), matcher);
        throw e;
    } catch (Exception e) {
        throw new AssertionError(e);
    }
}
```

mathieucarbou:
@gibsong : please rename your commit to `Close #1624: ...`

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
mathieucarbou: If another wait time must be added it matches what we saw in debug with Ludovic: we needed to wait at least for 2 computations of samples to happen before doing any cache ops because the ratio depends on previous stats.

So I think we would only need to wait for the first 2 samples to be computed into `StatsUtil .triggerStatComputation()`

i.e.: replacing

`if (((StatisticHistory<?, ?>) entry.getValue()).getValue().length == 0) {`

by:

`if (((StatisticHistory<?, ?>) entry.getValue()).getValue().length < 2) {`

will perhaps do the trick.

It would wait _before_ starting the test and doing some cache ops, and then, we wouldn't need to sleep in the loop.

`StatsUtil .triggerStatComputation()` is supposed to be the place where we do the necessary stuff to make sure the stats are _ready_.

gibsong: @ljacomet the sleep actually isn't in the loop.  I call sleep right after the get calls(allowing the stats time to be calculated) and before I enter the looping to gather the stat values.  Please let me know if I missed something but I am not sure how to implement your suggestion.

gibsong: @mathieucarbou let me try your suggestion.  It would be nice to eliminate the sleep as this is causing the test to run for a really long time, and sometimes timeout.

mathieucarbou: @gibsong : you probably need to increase the global timeout (junit rule) for these ratio tests.

gibsong: @mathieucarbou the global timeout is currently set to 60 seconds.  I was hoping I could actually decrease it.  If your solution works, I can remove my sleep 1500ms(which is called 17x, once per dataprovider row) and then I can play around with the timeout value.

mathieucarbou: Oh yes! The timeout was already increased I remember! Good!

gibsong:
test this please

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
gibsong: @mathieucarbou looks like your suggestion may have worked.  I tested it locally and the build just passed.  Although the build took an hour!

mathieucarbou:
@gibsong : please rebase __and__ rename your commit to `Close #1624: ...`
 thanks :-)

management/src/test/java/org/ehcache/management/providers/statistics/StatsUtil.java
mathieucarbou: @gibsong : damn.. this does not seem to work: https://ehcache.ci.cloudbees.com/job/ehcache3-pull-requests/3424/

ljacomet:
The failures have nothing to with stats timing but with relying on deterministic eviction. See #1629

PR #1629, opened 2016-11-17
Fix stats tests relying on deterministic eviction

ljacomet:


management/src/test/java/org/ehcache/management/providers/statistics/HitCountTest.java
```diff
@@ -100,7 +101,14 @@ public void test() throws InterruptedException, IOException {
       registryConfiguration.addConfiguration(new EhcacheStatisticsProviderConfiguration(1,TimeUnit.MINUTES,100,1,TimeUnit.MILLISECONDS,10,TimeUnit.MINUTES));
       ManagementRegistryService managementRegistry = new DefaultManagementRegistryService(registryConfiguration);
 
-      CacheConfiguration<Long, String> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class, resources).build();
+      CacheConfiguration<Long, String> cacheConfiguration = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class, resources)
+        .withEvictionAdvisor(new EvictionAdvisor<Long, String>() {
+          @Override
+          public boolean adviseAgainstEviction(Long key, String value) {
+            return key.equals(2L);
```
mathieucarbou: @ljacomet : I would be interested in knowing why trying to prevent the eviction of only `key.equals(2L);` ? 
Perhaps we can add a little comment ? Just to be sure we follow the same pattern and understand the reason behind for future tests. Thanks :-)

CC @gibsong : FYI.

ljacomet: I leave that as an exercise for the reader

mathieucarbou: oh! is it because of the 7th use case that is configured with `.heap(1, ENTRIES)`, and thus can evicts subsequents values put ?

AbfrmBlr: Since the test at line 132 expects the hit to be at highest tier for key 2L, the advisor will prevent it from evicting.

mathieucarbou: @gibsong @anthonydahanne : FYI! Thanks @AbfrmBlr !

ljacomet: New exercise for the reader: all but one test would work just as well if the advised key was `1L`. Why?

mathieucarbou: good question... ;-)

ljacomet: > 3 tier test: heap(1, ENTRIES).offheap(2, MB).disk(3, MB)
> 
> cache.put(1L, "1");
> cache.put(2L, "2");
> cache.put(3L, "3");
> -now keys 1,2 and 3 are all in the disk tier.

Correct

> 
> get(1L) 
> -hit on disk tier
> -key 1 is pushed to heap and offheap, and also remains on disk

installed on heap, off heap only misses but does not get the value

> -if evictionAdvisor designated 1L then this key would remain in Heap tier

eviction does not play here, only a single entry exists in heap

> 
> get(2L)
> -hit on disk tier
> -key 2 is pushed to heap and offheap, and also remains on disk

again, only heap is affected, off heap only misses

> -key 1 is removed from heap tier bc only allows 1 entry

since key `2L` is advised against eviction, then key `1L` is the guaranteed eviction victim. Without the eviction advisor, it would be undetermined if both had the same `lastAccess` timestamp, which can happen as failures have shown. Also when they have the same, the eviction victim will really be what is sampled first (or last, would need to check code) and that's _random_.

> -since evictionAdvisor designates 2L it will remain in Heap tier

correct

> 
> get(2L)
> -hit on heap tier
> 
> ***I am assuming at this point key 2L sometimes expires and this was causing the problem? Then by adding the eviction advisor the key no longer can expire and be evicted?

No expiry is at play in this test.

> 
> get(1L)
> -hit on offheap tier
> -Does the evictionAdvisor prevent 1L from being pushed to the heap(heap only allows 1 entry)?

Indeed, it will once again loose the eviction sampling because `2L` is advised against eviction.

gibsong: @ljacomet thanks for the explanation.  A few more quick questions.  In the case of 3 tiers, when there is a hit on the disk tier, that value is only pushed to the heap tier?  It isn't also put in the offheap tier?  I believe that is what you said above, but I just want to double check.  If that is the case then how does an entry ever get into the middle tier(offheap)?  Is it when an entry is evicted from the heap?  I have to work on eviction tests soon, so I need to understand this.  I may have been thinking of it wrong.

ljacomet: I am not sure I want to answer the question even more than I did already. This is internal implementation that we may want to change at some point and thus having too strict tests on existing behaviour will make the potential change more complicated.

gibsong: ok, for now I won't get too picky then.  I will probably have some more questions though when I work on the eviction testing.

PR #1635, opened 2016-11-18
Move to latest Terracotta version

ljacomet:
* Clean up lifecycle of integration tests by no longer starting the
servers in tearDown.


ljacomet:
rebased

PR #1639, opened 2016-11-21
Issue 1591

AbfrmBlr:
Some tests to follow

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
@@ -51,7 +51,6 @@ import org.ehcache.clustered.common.internal.messages.EhcacheOperationMessage;
 import org.ehcache.clustered.common.internal.messages.LifecycleMessage;
 import org.ehcache.clustered.server.internal.messages.PassiveReplicationMessage;
 import org.ehcache.clustered.server.internal.messages.PassiveReplicationMessage.ClearInvalidationCompleteMessage;
-import org.ehcache.clustered.server.internal.messages.PassiveReplicationMessage.ClientIDTrackerMessage;
 import org.ehcache.clustered.server.internal.messages.PassiveReplicationMessage.InvalidationCompleteMessage;
 import org.ehcache.clustered.server.internal.messages.PassiveReplicationMessage.ChainReplicationMessage;
 import org.ehcache.clustered.common.internal.messages.ReconnectMessage;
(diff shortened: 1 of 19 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -297,6 +306,17 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
     }
   }
 
+  private void clearClientTrackedAtReconnectComplete() {
+
+    if (!reconnectComplete.get()) {
+      boolean success = reconnectComplete.compareAndSet(false, true);
+      if (success) {
```
ljacomet: Fold the CAS op in the if, no need to the local var here.

clustered/server/src/main/java/org/ehcache/clustered/server/state/ClientMessageTracker.java
```diff
@@ -73,4 +91,8 @@ public boolean isConfigureApplicable(UUID clientId, long timestamp) {
     return true;
   }
 
+  public void reconcileTrackedClients(Set<UUID> trackedClients) {
+    messageTrackers.entrySet().removeIf(x -> !trackedClients.contains(x));
```
ljacomet: This does not work - using incompatible types.
And thus also indicates there is missing testing ...

clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheSyncMessageCodec.java
```diff
@@ -67,7 +62,7 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheSyncMessageCodec.class);
 
   private static final String STORES_SUB_STRUCT = "stores";
-  private static final String CLIENTS_SUB_STRUCT = "clients";
+//  private static final String CLIENTS_SUB_STRUCT = "clients";
```
ljacomet: Please delete code instead of commenting out

clustered/server/src/main/java/org/ehcache/clustered/server/EhcachePassiveEntity.java
```diff
@@ -152,14 +148,16 @@ private void invokeRetirementMessages(PassiveReplicationMessage message) throws
...
+        break;
       default:
         throw new IllegalMessageException("Unknown Retirement Message : " + message);
     }
   }
 
   private void untrackHashInvalidationForEventualCache(InvalidationCompleteMessage message) {
-    InvalidationCompleteMessage invalidationCompleteMessage = message;
-    ehcacheStateService.getInvalidationTracker(invalidationCompleteMessage.getCacheId()).getInvalidationMap().computeIfPresent(invalidationCompleteMessage.getKey(), (key, count) -> {
+    ehcacheStateService.getInvalidationTracker(message.getCacheId()).getInvalidationMap().computeIfPresent(message.getKey(), (key, count) -> {
```
ljacomet: Slightly unrelated but the `count--` and `count++` a few lines down should be replaced by `count - 1` and `count + 1` as they can't really modify the `count` value.

AbfrmBlr:
Review comments addressed

clustered/server/src/main/java/org/ehcache/clustered/server/state/ClientMessageTracker.java
```diff
@@ -73,4 +91,8 @@ public boolean isConfigureApplicable(UUID clientId, long timestamp) {
     return true;
   }
 
+  public void reconcileTrackedClients(Set<UUID> trackedClients) {
+    messageTrackers.entrySet().removeIf(x -> !trackedClients.contains(x.getKey()));
```
ljacomet: Do it on the keySet, and you can even use `retainAll` then instead of a lambda.

AbfrmBlr: cool

PR #1646, opened 2016-11-23
Proper else clause invoke #1482

AbfrmBlr:


ljacomet:
Killed build given comment above

AbfrmBlr:
@ljacomet what else we can do? Can we throw exception to bring down the server. The only way on passive entity is to log the exception. The only other stuff you can do is to log at other places, but that does not change anything in design, other than refactoring the code.

AbfrmBlr:
@ljacomet , ca we discuss this ?

ljacomet:
Fix will be part of #1657

PR #1653, opened 2016-11-24
Make sure we don’t suppress exceptions. Close #1651

### LINKED ISSUE #1651: EhcacheClientEntityFactory.retrieve might suppress exceptions
When `entity.close()` or `fetchHold.unlock()` throw an exception, it suppresses the actual exception thrown by `entity.validate`.

That shouldn't happen. I think we should suppress the later exception (and log it in error).

henri-tremblay:
I'm logging exceptions that would suppress other exceptions. This will do until Java 7.

I've also changed some code that felt weird:
* No unlock on `EntityVersionMismatchException` but one on all other exceptions 
* try/catch on `EntityVersionMismatchException` was really broad

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
@@ -91,7 +91,7 @@ public class EhcacheClientEntityFactory {
    *        lifecycle operation timeout
    */
   public void create(final String identifier, final ServerSideConfiguration config)
-      throws EntityAlreadyExistsException, EhcacheEntityCreationException, EntityBusyException, TimeoutException {
+    throws EntityAlreadyExistsException, EhcacheEntityCreationException, EntityBusyException, TimeoutException {
     Hold existingMaintenance = maintenanceHolds.get(identifier);
     Hold localMaintenance = null;
     if (existingMaintenance == null) {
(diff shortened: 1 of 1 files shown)


henri-tremblay:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -191,25 +220,50 @@ public void destroy(final String identifier) throws EhcacheEntityNotFoundExcepti
...
     } else {
-      try {
         EntityRef<EhcacheClientEntity, UUID> ref = getEntityRef(identifier);
         try {
           if (!ref.destroy()) {
             throw new EntityBusyException("Destroy operation failed; " + identifier + " clustered tier in use by other clients");
           }
         } catch (EntityNotProvidedException e) {
-          LOGGER.error("Unable to delete clustered tier manager for id {}", identifier, e);
+          LOGGER.error("Unable to delete clustered tier manager for id " + identifier, e);
```
ljacomet: Why that change? The previous version is the standard pattern for slf4j.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -100,44 +100,60 @@ public void create(final String identifier, final ServerSideConfiguration config
...
+      silentlyUnlock(localMaintenance, identifier);
+      throw new AssertionError(e);
+    } catch (EntityVersionMismatchException e) {
+      LOGGER.error("Unable to create clustered tier manager for id {}", identifier, e);
+      silentlyUnlock(localMaintenance, identifier);
+      throw new AssertionError(e);
+    } catch(EntityAlreadyExistsException e) {
+      silentlyUnlock(localMaintenance, identifier);
+      throw e;
+    } catch(EhcacheEntityCreationException e) {
```
ljacomet: `EhcacheEntityCreationException` is a `RuntimeException` and since we do not have any specific handling, catching the higher type makes more sense and that is already done below.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
henri-tremblay: I forgot to remove it I think. One note on the standard pattern though. The curly braces are useful to prevent string concatenation upfront. So we should use them when we know that most of the time, the log will be ignored. This is the case for DEBUG for instance. However, for an ERROR, we could argue that it's not a good idea. Because looking for curly braces in a string have a huge performance impact. So String concatenation is more appropriate. That said, errors are not frequent. So it is really more dodgy for INFO.

The same applies when you call complicated methods to generate a log. This is be called upfront. So you better shield the call with a `isInfoEnabled` and then use String concatenation.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
henri-tremblay: You a right. I missed that.

henri-tremblay:
Both requested changes have been made and the PR is rebased

henri-tremblay:
I did something simpler. Because I wasn't happy with my own code. I should also pass the tests.

ljacomet:
Good to go once rebased.

henri-tremblay:
Rebased

AbfrmBlr:
needs a rebase

henri-tremblay:
rebased

PR #1658, opened 2016-11-28
Fixes #1657 Error on decoded but unsupported message

### LINKED ISSUE #1657: Fix systematic stack trace on passive when processing message
Through #1482 I introduced a bug that causes all message processing on the passive to print a stack trace in logs.

However, looking at the problem I realised we have a very poor error handling around _unknown_ messages. Messages that come out of the codec but are not recognised by the entity are programming errors and thus should be reported through throwing `AssertionError`.

ljacomet:
* Includes a fix for `PassiveSyncTest` instability

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/sync/PassiveSyncTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/sync/PassiveSyncTest.java
@@ -33,11 +33,16 @@ import org.junit.Test;
 import org.terracotta.testing.rules.BasicExternalCluster;
 import org.terracotta.testing.rules.Cluster;
 
+import com.google.code.tempusfugit.temporal.Timeout;
+
 import java.io.File;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


clustered/integration-test/src/test/java/org/ehcache/clustered/sync/PassiveSyncTest.java
```diff
@@ -33,11 +33,16 @@
...
+import com.google.code.tempusfugit.temporal.Timeout;
+
 import java.io.File;
 import java.util.Collections;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import static com.google.code.tempusfugit.temporal.Duration.seconds;
+import static com.google.code.tempusfugit.temporal.Timeout.timeout;
+import static com.google.code.tempusfugit.temporal.WaitFor.waitOrTimeout;
```
ljacomet: Was added by @chrisdennis a while ago, so decided to use it instead of rolling a similar logic.

PR #1672, opened 2016-12-02
Put should close non-null chain

AbfrmBlr:


lorban:
Were is the test?

AbfrmBlr:
Will try to figure out, if this can be really tested

AbfrmBlr:
Since the side effect is caused by the behavior of the storage engine, I wonder if this can be tested.

AbfrmBlr:
PLEASE DO NOT MERGE.

PR #1673, opened 2016-12-02
Fixes #1430 Final clean up of compilation warnings

### LINKED ISSUE #1430: Remove generics warnings

ljacomet:
This PR finalises the clean up of compilation warnings and turns warnings into errors in all module where it is possible.

### DIFF
--- a/107/build.gradle
+++ b/107/build.gradle
@@ -40,6 +40,10 @@ dependencies {
   }
 }
 
+tasks.withType(JavaCompile) {
+  options.compilerArgs += ['-Werror']
+}
+
 javadoc {
   exclude '**/tck/**'
 }
--- a/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
+++ b/107/src/main/java/org/ehcache/jsr107/ConfigurationMerger.java
@@ -187,6 +187,7 @@ class ConfigurationMerger {
     return builder;
   }
 
+  @SuppressWarnings("unchecked")
   private static <K, V> CacheConfigurationBuilder<K, V> addDefaultCopiers(CacheConfigurationBuilder<K, V> builder, Class keyType, Class valueType ) {
     Set<Class> immutableTypes = new HashSet<Class>();
     immutableTypes.add(String.class);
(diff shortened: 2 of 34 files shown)


chrisdennis:
Is it clear how many of the `@SuppressWarnings` annotations are "required" and therefore inherently unfixable, and how many are fixable with time?  Would be good to have some way to distinguish these two classes so that fixing the fixable can be a trackable task.

ljacomet:
I tried hard to fix all those I could. Example of unfixable: Mockito matchers `any(GenericClass.class)` which cannot be declared as a variable and thus force the suppress to happen at the method level.

Maybe such a categorisation should be made on prod source not test.

chrisdennis:
Definitely makes more sense for the production usages.  Maybe we could use `//TODO` comments or something similar?

henri-tremblay:
LGTM. I personally like to add a comment telling why a warning can't be fixed. Sometimes it's obvious though.

Then, I start to wonder if we shouldn't had a small mockito wrapper. Like (but I'm not happy with it yet)

```java
  public static class TT<T> {
  }

  public static <T> T typedMock(TT<T> typeWitness) {
    ParameterizedType type = (ParameterizedType) typeWitness.getClass().getGenericSuperclass();
    Type paramType = type.getActualTypeArguments()[0];
    if(paramType instanceof ParameterizedType) {
      paramType = ((ParameterizedType) paramType).getRawType();
    }
    return Mockito.mock((Class<T>) paramType);
  }
  
  @Test
  public void test() {
    List<String> l1 = typedMock(new TT<List<String>>() {});
  }
```

ljacomet:
@henri-tremblay Yes, need to do one more pass on the suppressed warnings in prod code, will get to it tomorrow.
Regarding mockito, whenever possible I replace the `mock(Class)` call with an annotated field. I have the impression that your solution is overkill for tests, where the `@SuppressWarnings` is clear enough.

henri-tremblay:
It's just that I'm afraid we will have a lot of `@SuppressWarning`. Both I agree the best way is to use `@Mock`.

A simpler way that adds a parameter but removes the anonymous inner class is:

```java
  public static class TT<T> {
  }

  public static <T> T typedMock(TT<T> typeWitness, Class<?> rawType) {
    return (T) Mockito.mock(rawType);
  }
  
  @Test
  public void test() {
    List<String> l1 = typedMock(new TT<List<String>>());
  }
```

PR #1680, opened 2016-12-06
A 10 seconds timeout is too tight for the CI

henri-tremblay:


anthonydahanne:
did not solve the issue : https://ehcache.ci.cloudbees.com/job/ehcache3-pull-requests/3557/jdk=Oracle%20JDK%201.8(latest),label_exp=standard,testVM=6/testReport/junit/org.ehcache.management.providers.statistics/HitLatencyTest/test_5_/

henri-tremblay:
Sad. I've noticed that sometimes the build run on a server out of RAM. So
it swaps and takes ages to allocate of heap.

So we can raise the timeout even more but I think cloubees should
investigate. We can also migrate some test on heap.

Le 6 déc. 2016 11:29 AM, "Anthony Dahanne" <notifications@github.com> a
écrit :

PR #1683, opened 2016-12-06
Upgrade to new tc-platform

mathieucarbou:
to fix management service lifecycle, closing entity requested services from the monitoring service when the entity is gone, to avoid having ehcache knowing it needs to close the management service

In 5.0.13.beta3, I got rid of all weak references and do the cleanup of client descriptors and consumer services thanks to some listeners on the topology changes (when client disconnects and entities are destroyed). This change allows also to get rid of the close() methods that were required before.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ ext {
   sizeofVersion = '0.3.0'
 
   // Clustered
-  terracottaPlatformVersion = '5.0.13.beta2'
+  terracottaPlatformVersion = '5.0.13.beta3'
   managementVersion = terracottaPlatformVersion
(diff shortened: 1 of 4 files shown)


mathieucarbou:
test this please

mathieucarbou:
@ljacomet @gibsong : when building this PR locally and on cloudbee, I get some errors in `HitLatencyTest`:

```
org.ehcache.management.providers.statistics.HitLatencyTest > test[5] FAILED
    org.junit.runners.model.TestTimedOutException at HitLatencyTest.java:142
```

We said yesterday that we finally do not care about latencies: can we remove this test ? If yes, I'll remove it in this PR

Issue #1684, opened 2016-12-06
Compute only compute basic Statistics in Ehcache (not rates, ratio, etc)


mathieucarbou:
Stats in `StandardEhcacheStatistics ` are not correctly computed (see `Eh107CacheStatisticsMXBean `) and only counters have to be logged and sent to TMS. Not latencies, rate or ratio. Rates and Ratio will be computed in TMS based on its own window and counters in its DB.

@gibsong @anthonydahanne  : here are the steps we discussed yesterday:

- [x] Disable latency stats (in tc-platform, ehcache, m&m) => @mathieucarbou (PR #1692)

- [x] expose stats such as Cache:Hits, OnHeapHits, etc. (work started Geoff’s PR : https://github.com/ehcache/ehcache3/pull/1693/files) - think about creating a class defining hitcounts, etc. that would be used by the JSR107 Mbean and the Collector. It could be an __Ehcache Statistic Service__ that would be used by both management and JSR107 and will contains all rules to compute stats, like it is done currently in JSR107 Mbean. @henri-tremblay 

- [ ] compute other stats in TMS (i.e. rates, ratios, across clients, etc) => @anthonydahanne / @GaryWKeim 

- [x] get rid of other older now unused stats (ratios / rates) in stat registry @henri-tremblay 

- [x] get rid of stats registry, scheduling and sampling and history stats. TMS : would do a call to start the collector - we would not choose metrics anymore - the collector will collect all ehcache raw counters (a counter = stat name + value + timestamp, so lighter than the stat history object we currently have, and moreover, timestamp is the same for all; collector will push all these counters to monitoring buffer => @mathieucarbou (https://github.com/Terracotta-OSS/terracotta-platform/pull/259)

mathieucarbou:
@ljacomet @chrisdennis @GaryWKeim : can you please review the description to make sure I didnt forget something ? Thanks!

mathieucarbou:
Just to add some notes:

Stat computation will happen in TMS now.

So The only goal of the `StandardEhcacheStatistics` now is to output all Ehcache counter values, in the form: `Tier:operation:outcome`, i.e. `Cache:get:HitNoLoader`, `Cache:get:HitWithLoader`, etc.
TMS will do the `Cache:Hit` computation based on the google doc and what is done in the jsr107 mbean as an example.

mathieucarbou:
[...]

PR #1692, opened 2016-12-07
#1684: Disabling Latency statistics

mathieucarbou:


management/src/main/java/org/ehcache/management/providers/statistics/EhcacheStatisticsProvider.java
```diff
@@ -55,6 +69,17 @@ protected void dispose(ExposedObject<CacheBinding> exposedObject) {
   }
 
   @Override
+  public final Collection<? extends Descriptor> getDescriptors() {
+    Collection<StatisticDescriptor> capabilities = new HashSet<StatisticDescriptor>();
+    for (ExposedObject o : getExposedObjects()) {
+      capabilities.addAll(((StandardEhcacheStatistics) o).getDescriptors());
+    }
+    List<StatisticDescriptor> list = new ArrayList<StatisticDescriptor>(capabilities);
+    Collections.sort(list, STATISTIC_DESCRIPTOR_COMPARATOR);
```
mathieucarbou: Note: I forgot to apply this code before. I also did this change for server entity stats. 
This really helps during tests or in TMS/TMC when generating a json view of the backend topology with management data. Having them unsorted makes it hard to find wrong stt items.

GaryWKeim: Well, you know in your heart that this should be in a different commit.

mathieucarbou: Yes, should have been! Now... Is it worth loosing another  1h for the new build to pass for this small thing and delaying more the PR for M&M ? :-S

gibsong: looks good to me, anything that helps with faster testing is always a plus.

gibsong:
looks good to merge

PR #1703, opened 2016-12-12
Bump to latest Terracotta version

ljacomet:
* Adapt to EntityRef.create and EntityRef.destroy API changes
* Adapt to the passthrough changes

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
@@ -27,9 +27,11 @@ import org.slf4j.LoggerFactory;
 import org.terracotta.connection.Connection;
 import org.terracotta.connection.entity.EntityRef;
 import org.terracotta.exception.EntityAlreadyExistsException;
+import org.terracotta.exception.EntityConfigurationException;
 import org.terracotta.exception.EntityNotFoundException;
 import org.terracotta.exception.EntityNotProvidedException;
 import org.terracotta.exception.EntityVersionMismatchException;
+import org.terracotta.exception.PermanentEntityException;
 
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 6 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -140,6 +142,12 @@ public void create(final String identifier, final ServerSideConfiguration config
       } catch (EntityVersionMismatchException e) {
         LOGGER.error("Unable to create clustered tier manager for id {}", identifier, e);
         throw new AssertionError(e);
+      } catch (PermanentEntityException e) {
+        LOGGER.error("Unable to create entity - server indicates it is permanent",e );
```
henri-tremblay: Bad spacing here `, e`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -227,6 +235,9 @@ public void destroy(final String identifier) throws EhcacheEntityNotFoundExcepti
         throw new AssertionError(e);
       } catch (EntityNotFoundException e) {
         throw new EhcacheEntityNotFoundException(e);
+      } catch (PermanentEntityException e) {
+        LOGGER.error("Unable to destroy entity - server says it is permanent");
```
henri-tremblay: Why isn't the exception logged here?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/UnitTestConnectionService.java
```diff
@@ -240,6 +246,8 @@ public static PassthroughServer remove(URI uri) {
           LOGGER.error("Entity destroy failed: ", ex);
         } catch (EntityNotFoundException ex) {
           LOGGER.error("Entity destroy failed: ", ex);
+        } catch (PermanentEntityException e) {
+          LOGGER.error("Entity destroy failed (permanent???): ", e);
```
henri-tremblay: The exception is called `e` while the others are called `ex`. The 3 question marks are to say it is weird to have a permanent destroy?

ljacomet: Yes, since that is in test code, attracting attention to that point. A permanent entity is a special kind of entity that has no configuration and the lifecycle of the server.

ljacomet:
Unstable test 😢

ljacomet:
test this please

PR #1713, opened 2016-12-13
Improve javadoc on CacheManagerBuilder.persistence

henri-tremblay:


impl/src/main/java/org/ehcache/config/builders/CacheManagerBuilder.java
```diff
@@ -375,11 +375,14 @@ T cast(EhcacheManager ehcacheManager) {
   }
 
   /**
-   * Convenience method to get a {@link CacheManagerConfiguration} for a {@link PersistentCacheManager}.
+   * Convenience method to get a {@link CacheManagerConfiguration} for a {@link PersistentCacheManager} stored on disk. The
+   * disk resource pool should be configured as persistent as well.
```
ljacomet: That is just wrong. You need to setup the `CacheManagerBuilder` so that it returns a properly configured `PersistentCacheManager` to be able to use a `disk` resource but that does not mean you want the _reload my data_ feature always.

henri-tremblay: Is the new version better? I will daggy fix afterwards

ljacomet:
Looks good - just needs the daggy fix approach now

PR #1724, opened 2016-12-14
Adding Nexus Staging Plugin - take 2

akomakom:
Allows jobs to add "closeRepository" or "closeAndPromoteRepository" to the task list.

https://github.com/Codearte/gradle-nexus-staging-plugin/

This is a copy of the closed PR #1345 since it is now possible to do this in the absence of JDK 1.6

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,28 @@
 import scripts.*
 import org.gradle.internal.jvm.Jvm
 
+buildscript {
+  repositories {
+    mavenCentral()
+  }
+  dependencies {
+    classpath "io.codearte.gradle.nexus:gradle-nexus-staging-plugin:0.5.3"  
+  }
+}
+
+// This adds tasks to auto close or release nexus staging repos
+// see https://github.com/Codearte/gradle-nexus-staging-plugin/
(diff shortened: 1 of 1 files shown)


akomakom:
So apparently I cannot test this plugin with our Nexus - only with oss.sonatype...  here is why:

1. The plugin retrieves the staging profile id's via an API call (this works)
1. The plugin then looks for a profile with the artifact group ("org.ehcache" in our case) as its exact name.   This fails since our profile name is "Ehcache OS"

The good news is that it works with Sonatype because there our profile is in fact named "org.ehcache".   The bad news is that this is probably a poor implementation (and also the only one out there).

I might fork his code to do this properly, assuming that there is an API for Repository Targets

akomakom:
.. I hacked it by setting profileGroup="Ehache OS", but the plugin disappointed once again, it couldn't figure out which of the 20 open staging repos in our Nexus to close.   The problem appears to be that it does not receive the staging repository info from gradle the way the maven plugin does, so it has to guess by interrogating nexus.   As long as only one repo is open at a time, it would work (which I guess is usually the case at oss.sonatype.org)

PR #1725, opened 2016-12-14
Fix #1710 Entity gets configured when created

### LINKED ISSUE #1710: Inject ServerSideConfiguration at entity construction time
now that we can fail entity construction, we can inject the `ServerSideConfiguration` during construction, getting rid of the create - configure chained calls.

This also supersedes #1675 and #1709 and is a move in the good direction for #1686.

However this change is breaking client - server contract and thus can only be done for the next minor version.

ljacomet:
* Entity receives configuration during create - much more than the previous `UUID`
  * An identifier - currently a `String` by default sourced from the entity URL in the connection
  * The `ServerSideConfiguration` that used to be passed during `configure` step
* Entity uses the `CreationException` in case of mismatch between server resources and specified ones, same for shared pools.
* State service destroy effectively removes the instance from the provider
* Clean up related to the above changes

Flagged as no merge for now since this means breaking compatibility of client -> server communication and thus requires bumping version to 3.3.0-SNAPSHOT on master.
(description copied in the issue for better reference)

### DIFF
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactoryTest.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactoryTest.java
@@ -16,10 +16,8 @@
 
 package org.ehcache.clustered.client.internal;
 
-import java.util.UUID;
-
-import org.ehcache.clustered.client.internal.service.ClusteredTierManagerConfigurationException;
 import org.ehcache.clustered.common.ServerSideConfiguration;
+import org.ehcache.clustered.common.internal.ClusteredTierManagerConfiguration;
 import org.ehcache.clustered.common.internal.lock.LockMessaging.HoldType;
 import org.ehcache.clustered.client.internal.lock.VoltronReadWriteLockClient;
 import org.junit.Before;
@@ -35,6 +33,7 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyInt;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.isA;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
@@ -43,6 +42,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import org.terracotta.connection.entity.EntityRef;
 import org.terracotta.exception.EntityAlreadyExistsException;
+import org.terracotta.exception.EntityConfigurationException;
 import org.terracotta.exception.EntityNotFoundException;
 
 public class EhcacheClientEntityFactoryTest {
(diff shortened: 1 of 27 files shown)


chrisdennis:
It kinda bothers me that the state service still has a separate configure phase even though it's only called from the entity constructors and it's idempotent.  Couldn't we just have it be done on construction of the service instances?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
```diff
@@ -61,7 +60,6 @@
   public static final String MESSAGE_TYPE_FIELD_NAME = "opCode";
   public static final int MESSAGE_TYPE_FIELD_INDEX = 10;
   public static final EnumMapping<EhcacheMessageType> EHCACHE_MESSAGE_TYPES_ENUM_MAPPING = newEnumMappingBuilder(EhcacheMessageType.class)
-    .mapping(CONFIGURE, 1)
```
chrisdennis: Since we'd be breaking the comms protocol anyway we could correct all these index values at the same time right?

clustered/server/src/main/java/org/ehcache/clustered/server/state/config/EhcacheStateServiceConfig.java
```diff
@@ -42,6 +43,10 @@ public EhcacheStateServiceConfig(ServiceRegistry serviceRegistry, final KeySegme
     return EhcacheStateService.class;
   }
 
+  public ClusteredTierManagerConfiguration getConfig() {
+    return config;
+  }
+
   public ServiceRegistry getServiceRegistry() {
     return this.serviceRegistry;
```
chrisdennis: This looks to be unused now.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
ljacomet: Good point - will change

clustered/server/src/main/java/org/ehcache/clustered/server/state/config/EhcacheStateServiceConfig.java
ljacomet: Will leave it in for now - multi entity probably will require it

ljacomet:
@chrisdennis Indeed the two part configuration of the service is a bummer. I will file a feature request on terracotta-api so that we have the ability to fail a service retrieval with the exception getting to the requesting entity.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactoryTest.java
```diff
@@ -69,14 +70,12 @@ public void testCreate() throws Exception {
     EhcacheClientEntityFactory factory = new EhcacheClientEntityFactory(connection);
     factory.create("test", null);
     verify(entityRef).create(any(UUID.class));
```
albinsuresh: Wondering how that verify worked. Shouldn't the `create` be verfied on `any(ClusteredTierManagerConfiguration.class)`?

ljacomet: Good point ... let me look further into that!

clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
```diff
@@ -179,7 +179,7 @@ public static void beforeClass() throws Exception {
     List<String> notificationTypes = notificationTypes(messages);
 
     Map<String, List<String>> counts = notificationTypes.stream().collect(Collectors.groupingBy(o -> o));
-    assertThat(counts.keySet(), hasSize(12));
+//    assertThat(counts.keySet(), hasSize(12));
```
ljacomet: left over when trying to find why this was failing - good catch

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/LifeCycleMessageCodec.java
```diff
@@ -142,10 +135,6 @@ public LifeCycleMessageCodec(ConfigCodec configCodec) {
     return encoder.encode().array();
   }
 
-  private byte[] encodeTierManagerConfigureMessage(LifecycleMessage.ConfigureStoreManager message) {
-    return encodeTierManagerCreateOrValidate(message, message.getConfiguration(), configureMessageStruct.encoder());
```
AbfrmBlr: I guess we can get rid of `configureMessageStruct`

ljacomet: Indeed, removing

PR #1744, opened 2016-12-20
Closes #1743 Extract ChainMap interface from OffHeapChainMap

### LINKED ISSUE #1743: Extract an interface out of OffheapChainMap
The extracted interface must be used in `ServerSideServerStore` in places where the concrete `OffheapChainMap` is used.

albinsuresh:


clustered/server/src/main/java/org/ehcache/clustered/server/ServerSideServerStore.java
```diff
@@ -29,6 +29,6 @@
 public interface ServerSideServerStore extends ServerStore, MapInternals {
   void setEvictionListener(ServerStoreEvictionListener listener);
   ServerStoreConfiguration getStoreConfiguration();
-  List<OffHeapChainMap<Long>> getSegments();
+  List<? extends ChainMap<Long>> getSegments();
```
ljacomet: So that's the method making the concrete implementation leak. Can't we instead really on a specific type for here?
From the usage, it seems we only need to get at a given `Set<K>` so a `List<Set<K>>` would be good enough.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -92,8 +92,9 @@ public void close() {
     store.close();
   }
 
-  public List<OffHeapChainMap<Long>> getSegments() {
-    return store.getSegments();
+  @Override
+  public List<Set<Long>> getSegmentKeySets() {
+    return store.getSegments().stream().map(OffHeapChainMap::keySet).collect(Collectors.toList());
```
ljacomet: Why not simply return an anonymous abstract list that delegates the two methods you have to implement.

albinsuresh:
test this please

PR #1757, opened 2017-01-02
Closes #1754 Release write locks on clear

### LINKED ISSUE #1754: write lock of all segments not release in server store during close()
In OffHeapServerStore, the following code snippet does a segment.clear() before doing writeUnlockAll(). This makes the writeUnlockAll() never release the held write locks. Probably this did not manifest as a big problem in OSS because a close() is only called during destroy and after destroy the same references are not used. But this may become a problem if there are other threads trying to hold a read or write lock to work on the cleared map.

  public void close() {
    writeLockAll();
    try {
      clear();
      segments.clear();
    } finally {
      writeUnlockAll();
    }
  }

here is the code for writeUnlockAll() which will never unlock due to offending code that clears the segments in close()

  private void writeUnlockAll() {
    for (OffHeapChainMap<Long> s : segments) {
      s.writeLock().unlock();
    }
  }

rkavanap:
OffheapServerStore was not releasing write locks on destroy.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
@@ -224,10 +224,10 @@ public class OffHeapServerStore implements ServerStore, MapInternals {
     writeLockAll();
     try {
       clear();
-      segments.clear();
     } finally {
       writeUnlockAll();
     }
+    segments.clear();
   }
 
   // stats


clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -221,12 +221,16 @@ boolean handleOversizeMappingException(long hash) {
   }
 
   public void close() {
+    boolean success = false;
     writeLockAll();
     try {
       clear();
-      segments.clear();
+      success = true;
```
ljacomet: What does this buys us? There is no _expected_ failure path once we managed to `writeLockAll()` so we should simply move the `segments.clear()` after the finally block.

PR #1758, opened 2017-01-02
[master] Fix #1619 ClassLoading documentation

### LINKED ISSUE #1619: Document class loading in Ehcache
Ehcache takes into account complex environment where class loading may not be straightforward. This includes OSGi and a number of APIs / SPIs taking `ClassLoader` as parameters.

The rationale behind it should be documented so that users know what to do when their environment requires it.

This should be applied to `release/3.1` and `master` branches.

ljacomet:
Single PR for reviews against master.

Will open other ones against `release/3.1` and `release/3.2` once content is agreed upon.

### DIFF
--- a/docs/src/docs/asciidoc/user/class-loading.adoc
+++ b/docs/src/docs/asciidoc/user/class-loading.adoc
@@ -1,0 +1,70 @@
+---
+---
+= Class loading
+ifndef::sourcedir32[]
+include::common.adoc[]
+endif::sourcedir32[]
+
+ifdef::notBuildingForSite[]
+include::menu.adoc[]
+endif::notBuildingForSite[]
+
+== Introduction
+
+Since Ehcache is a library and supports user types both in configuration and in mapping keys or values,
+it must offer flexibility around class loading.
+
+[[cl-default]]
+== Default `ClassLoader` in Ehcache
+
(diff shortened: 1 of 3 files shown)


docs/src/docs/asciidoc/user/class-loading.adoc
```diff
@@ -0,0 +1,65 @@
...
+* An optional one at the `CacheManager` level,
+* A map of `<String, ClassLoader>` which will be used to link a specific `ClassLoader` to a given `Cache`.
+
+As for the Java configuration, the `Cache` level configuration takes precendence of the `CacheManager` level one.
+If no class loaders are specified, the <<cl-default,default class loader>> from Ehcache will be used.
+
+[[cl-runtime]]
+=== At runtime
+
+As soon as serialization is involved with `Cache` keys or values, class loader play a role at runtime.
```
albinsuresh: class loader *plays* a role...

ljacomet:
Build failure caused by a checkout failure - no need to re test as long as content not approved

chrisdennis:
I think the essence of this is correct.  I'm not convinced however that we always use the correct classloader during object deserialization.  The specific case I'm concerned about is the deserialization in `FileBasedStateRepository`.  This looks to be using the default 'latestUserDefinedLoader' rather than the correct user specified loader at deserialization.

PR #1760, opened 2017-01-02
Fix #1752 Bump entity version

### LINKED ISSUE #1752: org.terracotta.runnel.utils.LimitReachedException when running 3.3 server and 3.2.1 client
When running the 3.3.0-SNAPSHOT server with a 3.2.1-SNAPSHOT client, the server throws a LimitReachedException when the client attempts to connect - instead we should see an exception that the entity versions do not match

[exception.txt](https://github.com/ehcache/ehcache3/files/668651/exception.txt)
[terracotta-server.txt](https://github.com/ehcache/ehcache3/files/668652/terracotta-server.txt)
[tc-config.txt](https://github.com/ehcache/ehcache3/files/668653/tc-config.txt)

ljacomet:
This represents incompatibility - should have been done when the
configuration on create happened ideally.

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/BasicEntityInteractionTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/BasicEntityInteractionTest.java
@@ -18,13 +18,16 @@ package org.ehcache.clustered;
 import java.io.File;
 import java.util.Collections;
 import org.ehcache.clustered.client.internal.EhcacheClientEntity;
+import org.ehcache.clustered.common.EhcacheEntityVersion;
 import org.ehcache.clustered.common.ServerSideConfiguration;
 import org.ehcache.clustered.common.internal.ClusteredTierManagerConfiguration;
 import org.hamcrest.Matchers;
 import org.junit.BeforeClass;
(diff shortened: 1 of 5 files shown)


clustered/integration-test/src/test/java/org/ehcache/clustered/BasicEntityInteractionTest.java
```diff
@@ -60,7 +61,7 @@ public static void waitForActive() throws Exception {
   public void testAbsentEntityRetrievalFails() throws Throwable {
     Connection client = CLUSTER.newConnection();
     try {
-      EntityRef<EhcacheClientEntity, ClusteredTierManagerConfiguration> ref = client.getEntityRef(EhcacheClientEntity.class, 1, "testAbsentEntityRetrievalFails");
+      EntityRef<EhcacheClientEntity, ClusteredTierManagerConfiguration> ref = client.getEntityRef(EhcacheClientEntity.class, EhcacheEntityVersion.ENTITY_VERSION, "testAbsentEntityRetrievalFails");
```
henri-tremblay: Can you please extract a method and retrieve the test name using a Rule?

PR #1763, opened 2017-01-04
Close #1762: Reactivate M&M after a failover.

### LINKED ISSUE #1762: Reactivate M&M on failover
Caused by https://github.com/Terracotta-OSS/terracotta-core/issues/426

After a failover, we should reactivate M&M after the entity is added to the monitoring tree.

mathieucarbou:
- [x] Applied workaround for Terracotta-OSS/terracotta-core#426 (to restart m&m after a failover)
- [x] Needs new tc-platform release __5.2.0-pre3__
- [x] Activated management server libs by default in integration tests
- [ ] @anthonydahanne : after this, you'll be able to add more ehcache failover tests with M&M :-)

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/management/EhcacheConfigWithManagementTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/management/EhcacheConfigWithManagementTest.java
@@ -50,12 +50,8 @@ public class EhcacheConfigWithManagementTest {
       + "</ohr:offheap-resources>" +
       "</config>\n";
 
-  private static final List<File> MANAGEMENT_PLUGINS = Stream.of(System.getProperty("managementPlugins", "").split(File.pathSeparator))
-    .map(File::new)
-    .collect(Collectors.toList());
-
   @ClassRule
-  public static Cluster CLUSTER = new BasicExternalCluster(new File("build/cluster"), 1, MANAGEMENT_PLUGINS, "", RESOURCE_CONFIG, "");
+  public static Cluster CLUSTER = new BasicExternalCluster(new File("build/cluster"), 1, Collections.emptyList(), "", RESOURCE_CONFIG, "");
 
   @BeforeClass
   public static void beforeClass() throws Exception {
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
@@ -407,8 +407,6 @@ public class EhcacheActiveEntity implements ActiveServerEntity<EhcacheEntityMess
 
   @Override
   public void createNew() {
-    management.init();
-    management.sharedPoolsConfigured();
   }
 
   @Override
--- a/build.gradle
+++ b/build.gradle
@@ -50,13 +50,13 @@ ext {
   sizeofVersion = '0.3.0'
 
   // Clustered
-  terracottaPlatformVersion = '5.2.0-pre1'
+  terracottaPlatformVersion = '5.2.0-pre3'
   managementVersion = terracottaPlatformVersion
   terracottaApisVersion = '1.2.0-pre1'
-  terracottaCoreVersion = '5.2.0-pre3'
+  terracottaCoreVersion = '5.2.0-pre5'
   offheapResourceVersion = terracottaPlatformVersion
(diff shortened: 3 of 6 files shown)


clustered/integration-test/src/test/java/org/ehcache/clustered/management/EhcacheConfigWithManagementTest.java
```diff
@@ -50,9 +50,11 @@
       + "</ohr:offheap-resources>" +
       "</config>\n";
 
-  private static final List<File> MANAGEMENT_PLUGINS = Stream.of(System.getProperty("managementPlugins", "").split(File.pathSeparator))
-    .map(File::new)
-    .collect(Collectors.toList());
+  private static final List<File> MANAGEMENT_PLUGINS = System.getProperty("managementPlugins") == null ?
+    Collections.emptyList() :
+    Stream.of(System.getProperty("managementPlugins").split(File.pathSeparator))
```
anthonydahanne: it seems that in the build.gradle, your removed : 

     systemProperty 'managementPlugins', project.configurations.serverLibs.join(File.pathSeparator)

How is that supposed to return something else than null now ?

More genrally, what is the goal of the new method "copyServerLibs()" compared to what we used to have ?

mathieucarbou: Goal now is to put the serevr libs directly into the kit after it has been unzipped, so that ALL integration tests are using now the management stack.

Before, we had to pass a system property and add the server libs in only specifically some integration tests (m&m ones). System property "managementPlugins" is not there anymore so `Collections.emptyList()` will be returned.

I could remove this dead code though...

anthonydahanne: ok , that sounds good (including mnm plugins directly in the unzipped kit

anthonydahanne:
build failed since tc platform 5.2.0-pre2 not yet synced to s3.
@mathieucarbou can you please re trigger the build tomorrow am ? thanks

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -407,8 +407,6 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
 
   @Override
   public void createNew() {
-    management.init();
```
ljacomet: Why is this change not also applied to the passive version of the entity?

ljacomet:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
mathieucarbou: because the passive does not have the loadExisting method. 
The way platform works now works fine for the createNew() so I didn't change the passive entities.
This workaround will be removed once I'll have done https://github.com/Terracotta-OSS/terracotta-platform/issues/256

ljacomet: But so a passive will do the wiring during `createNew` and the call on the registered listener will be a no-op because of the boolean flag?
If you were to clean up the passive entity version as well and always rely on the event, you do not need the flag. Correct?

mathieucarbou: No because the `EntityEventService` is really specific to an entity: listeners are removed:
- either when the paltform tells us an entity is destroyed
- either when a new `EntityEventService` is requested on the same consumerId, then previous ones are removed.
In the case of a failover, there were no listener registered, so the active  entity just add a new one for itself.

The boolean flag is redundant. I just did the same thing we did in our sample entity while testing. But before, we kept the `createNew()` in the active, plus this listener, so I had a "double" init. This is not the case now.

mathieucarbou:
test this please

mathieucarbou:
@anthonydahanne @ljacomet : I found a bug in tc-platform: which was causing management to be initialized twice (https://github.com/Terracotta-OSS/terracotta-platform/pull/258).
I'll release again a new tc-platform version once this will be merged (-pre3)

mathieucarbou:
Closing: replaced by https://github.com/ehcache/ehcache3/pull/1766 from @chrisdennis

PR #1756, opened 2016-12-28
Move JMX statistics to an ehcache service

henri-tremblay:
This is the first step of a series of refactoring. It moves the statistics into ehcache.

PR #1764, opened 2017-01-05
Clean ups following #1756

ljacomet:


107/src/main/java/org/ehcache/jsr107/Eh107InternalCacheManager.java
```diff
@@ -30,11 +30,11 @@
  */
 class Eh107InternalCacheManager extends EhcacheManager {
 
-  public Eh107InternalCacheManager(Configuration config, Collection<Service> services, boolean useLoaderInAtomics) {
+  Eh107InternalCacheManager(Configuration config, Collection<Service> services, boolean useLoaderInAtomics) {
     super(config, services, useLoaderInAtomics);
   }
 
-  public ServiceLocator getServiceLocator() {
+  ServiceProvider<Service> getServiceLocator() {
```
henri-tremblay: Maybe also rename the method so it fits the return type?

107/src/main/java/org/ehcache/jsr107/Eh107MXBean.java
```diff
@@ -28,17 +28,16 @@
...
 
   private String sanitize(String string) {
     return string == null ? "" : string.replaceAll(",|:|=|\n", ".");
   }
 
-  private ObjectName createObjectName(String cacheName, Eh107CacheManager cacheManager, String beanName) {
-    URI uri = cacheManager.getURI();
-    String cacheManagerName = sanitize(uri != null ? uri.toString() : "null");
+  private ObjectName createObjectName(String cacheName, URI cacheManagerURI, String beanName) {
+    String cacheManagerName = sanitize(cacheManagerURI != null ? cacheManagerURI.toString() : "null");
```
henri-tremblay: I just been wondering. Can the cache manager URI really be null? Or it should be a precondition that it isn't?

107/src/main/java/org/ehcache/jsr107/Eh107InternalCacheManager.java
ljacomet: Fixing

107/src/main/java/org/ehcache/jsr107/Eh107MXBean.java
ljacomet: Indeed, `URI` is used as key in a map, can't be null. Cleaning up code

PR #1766, opened 2017-01-05
Upgrade to latest terracotta-platform and terracotta-core versions.

chrisdennis:


clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
```diff
@@ -26,12 +26,24 @@
...
+import org.terracotta.runnel.encoding.StructEncoderFunction;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 
-public class ChainCodec {
+public final class ChainCodec {
+
+  private ChainCodec() {
```
chrisdennis: I made this class static... was there some reason we wanted instances of it?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ExceptionCodec.java
```diff
@@ -26,11 +26,23 @@
...
 import org.terracotta.runnel.encoding.StructEncoder;
+import org.terracotta.runnel.encoding.StructEncoderFunction;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 
-class ExceptionCodec {
+final class ExceptionCodec {
+
+  private ExceptionCodec() {
```
chrisdennis: I made this one static too... was that wrong?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ResponseCodec.java
```diff
@@ -91,24 +89,13 @@
     .byteBuffer(MAP_VALUE_FIELD, 20)
     .build();
 
-  private final ChainCodec chainCodec;
-
-  public ResponseCodec() {
```
chrisdennis: I might be able to make this one static now too... but I didn't.  Should I have done?

build.gradle
```diff
@@ -50,13 +50,13 @@ ext {
   sizeofVersion = '0.3.0'
 
   // Clustered
-  terracottaPlatformVersion = '5.2.0-pre1'
+  terracottaPlatformVersion = '5.2.0-pre2'
```
mathieucarbou: should be `5.2.0-pre3` (I've released again after a bugfix)

mathieucarbou:
@chrisdennis : I approve the PR, though tc-platform version should be changed to __5.2.0-pre3__. Otherwise there will be issue with M&M when integrating into "other" projects.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
ljacomet: None I can think of

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ExceptionCodec.java
ljacomet: Don't think so

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ResponseCodec.java
ljacomet: Maybe, can happen at a later stage though.

Issue #1723, opened 2016-12-14
ehcache-clustered OSGi bundle


lburgazzoli:
I'm testing Ehcache in an OSGi environment and in relation to ehcache-clustered module, I noticed the following issues:

* the bundle name is the same as ehcache, so it is not easy to distinguish: 
  ```
  62 | Installed |  80 | 3.2.0    | Ehcache 3
  63 | Active    |  80 | 3.2.0    | Ehcache 3
  ```
* the bundle does not start as it requires ```org.ehcache.core.internal.service``` from ehcache bundle which is not exported:
  ```
  Error executing command: Error executing command on bundles:
    Error starting bundle 62: Unable to resolve org.ehcache.ehcache-clustered [62](R 62.0): missing requirement [org.ehcache.ehcache-clustered [62](R 62.0)] osgi.wiring.package; (osgi.wiring.package=org.ehcache.core.internal.service) 
    Unresolved requirements: [[org.ehcache.ehcache-clustered [62](R 62.0)] osgi.wiring.package; (osgi.wiring.package=org.ehcache.core.internal.service)]
```

ljacomet:
Hi,

Thanks for the report. These are issues that need fixing on the release/3.1, release/3.2 branches and master.

ljacomet:
Needs #1475 to be fixed first

ljacomet:
Not sure why I said there is a dependency on #1475, both issue can be addressed independently.

However, I am really surprised `ehcache-clustered` depends on internal packages. Need to investigate why this is the case and potentially resolve the underlying issue to exclude that dependency.

ljacomet:
Internal packages are required because:
* Usage of `org.ehcache.core.internal.service.ServiceLocator#findSingletonAmongst` - these static methods need to be moved to another type outside of the `internal` packages.
* `org.ehcache.core.internal.util.ConcurrentWeakIdentityHashMap` is used in clustered. So this type needs to move out of the `internal` package.
* `org.ehcache.impl.internal.concurrent.ConcurrentHashMap` is used in clustered. We should simply use the JDK `ConcurrentHashMap` there.
* `org.ehcache.impl.internal.events.NullStoreEventDispatcher` is used, should probably simply be moved to `org.ehcache.core.events`
* `org.ehcache.impl.internal.store.AbstractValueHolder` is used. Investigate and consider moving out of `internal` package.
[...]

PR #1771, opened 2017-01-09
Issue #1723 - rename packages, export org.ehcache.core.internal.service

kcleereman:
Issue #1723 - rename packages, export org.ehcache.core.internal.service

### DIFF
--- a/buildSrc/src/main/groovy/EhOsgi.groovy
+++ b/buildSrc/src/main/groovy/EhOsgi.groovy
@@ -61,7 +61,7 @@ class EhOsgi implements Plugin<Project> {
         }
 
         // Metadata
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/EhOsgi.groovy
```diff
@@ -76,7 +76,7 @@ class EhOsgi implements Plugin<Project> {
           }
         }
 
-        instruction 'Export-Package', '*'
+        instruction 'Export-Package', 'org.ehcache.core.internal.service', '*'
```
ljacomet: That's too coarse grained, and will cause issues because multiple OSGi packages will end up exporting the same.
Check the `gradle.properties` found in specific modules - `core` here - and add the entry there.

PR #1776, opened 2017-01-11
[closes #1775] force creation of threads in the service creator's thread group

### LINKED ISSUE #1775: Thread pools create threads in random thread groups
Since the `DefaultClusteringService` starts the TC client that itself creates thread pools with its own thread group, `OnDemandExecutionService` and `PooledExecutionService` create threads that can mistakenly be assigned to the TC client's thread group.

When the latter then tries to clean after itself during `CacheManager.close()` it notices the rogue threads in its thread group and incorrectly tries to shut them down.

lorban:


lorban:
this leaves me wondering if we shouldn't create a thread group for our thread pools though.

PR #1786, opened 2017-01-13
Compat' with new tc-platform (get rid of statistic wrappers)

mathieucarbou:
See: https://github.com/Terracotta-OSS/terracotta-platform/issues/261

Depends on new tc-platform release __5.2.0-pre6__ (https://github.com/Terracotta-OSS/terracotta-platform/pull/264)

Note: All Ehcache tests passed locally with the tc-platform snapshot.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ ext {
   sizeofVersion = '0.3.0'
 
   // Clustered
-  terracottaPlatformVersion = '5.2.0-pre5'
+  terracottaPlatformVersion = '5.2.0-pre6'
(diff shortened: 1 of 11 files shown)


mathieucarbou:
@henri-tremblay : they were not managed at all. If you look at the dependen PR (https://github.com/Terracotta-OSS/terracotta-platform/pull/264), memory unit was fixed to byte and time unit to millisecond. So all stats either return bytes or ms when of type size or time.

mathieucarbou:
test this please

PR #1789, opened 2017-01-16
After a clean, compensating counters for bulk stats are removed twice

henri-tremblay:


anthonydahanne:
build failure looks like an infra temporary issue
test this please

PR #1793, opened 2017-01-18
Refactoring required for management and state service extensions

albinsuresh:


clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -106,31 +106,39 @@ public void test_C_client_capabilities_exposed() throws Exception {
...
     Settings settings = (Settings) capabilities[0].getDescriptors().iterator().next();
     assertThat((String[]) settings.get("attachedStores")).containsOnly("dedicated-cache-1", "shared-cache-2", "shared-cache-3");
 
-    // ServerStoreSettings
+    // ClusteredTierManagerStateSettings
+
+    assertThat(capabilities[1].getDescriptors()).hasSize(1);
+    settings = (Settings) capabilities[1].getDescriptors().iterator().next();
+    assertThat(settings.get("type")).isEqualTo("ClusteredTierManager");
+    assertThat(settings.get("defaultServerResource")).isEqualTo("primary-server-resource");
```
mathieucarbou: @anthonydahanne : FYI, a new provider is responsible to expose the cluster tier manager now.
So `defaultServerResource ` is not anymore under `PoolSettings ` capability. `PoolSettings` will only expose things about pools now.

mathieucarbou:
@anthonydahanne : FYI, a new provider is responsible to expose the cluster tier manager now.
So `defaultServerResource` is not anymore under `PoolSettings` capability but under `ClusteredTierManagerSettings `. 
`PoolSettings` will only expose things about pools now.

mathieucarbou:
@ljacomet : for the refactoring into multi entities, perhaps you might want to reuse what @albinsuresh did for his needs: he had to introduce a provider for the cluster manager, which is the right thing to do (and is extensible) and also is something that you'll need anyway i think to split the providers ?

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
anthonydahanne: okee

Issue #1686, opened 2016-12-06
Refactor Ehcache entity usage to create multiple entities


ljacomet:
The goal is to better fit with the platform abstractions.

We would have:
* ClusteredTierManagerEntity for top level configuration and some book keeping
* ClusteredStoreEntity for each clustered tier

ljacomet:
Unblocked as #1710 has been merged in

ljacomet:
Functional changes are in.

Still to do:
* EhcacheStateSyncMessage is not relevant anymore. Can be removed and EhcacheSyncMessageCodec should be cleaned up as well.
* DefaultConcurrencyStrategy can be simplified as the concurrency key offsetting is not required anymore.
* ... and some general unused code cleanup.

PR #1798, opened 2017-01-25
Issue #1686 Move to multi entity

ljacomet:
This PR covers all the required bits to make multi entity work.

What's left to be done is around clean up, like refactoring and renames.


clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -121,54 +93,21 @@
...
   private final Set<UUID> trackedClients = Collections.newSetFromMap(new ConcurrentHashMap<>());
   private final ReconnectMessageCodec reconnectMessageCodec = new ReconnectMessageCodec();
-  private final ServerStoreCompatibility storeCompatibility = new ServerStoreCompatibility();
   private final EhcacheEntityResponseFactory responseFactory;
-  private final ConcurrentMap<Integer, InvalidationHolder> clientsWaitingForInvalidation = new ConcurrentHashMap<>();
-  private final AtomicInteger invalidationIdGenerator = new AtomicInteger();
-  private final ClientCommunicator clientCommunicator;
   private final EhcacheStateService ehcacheStateService;
   private final IEntityMessenger entityMessenger;
   private volatile ConcurrentHashMap<String, List<InvalidationTuple>> inflightInvalidations;
```
AbfrmBlr: This should be dead code as there is no invalidations in this entity

AbfrmBlr: Same class `addInflightInvalidationsForEventualCaches` is also not required

clustered/server/src/main/java/org/ehcache/clustered/server/state/config/EhcacheStoreStateServiceConfig.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public Class<EhcacheStateService> getServiceType() {
+    return EhcacheStateService.class;
+  }
+
+  public String getManagerIdentifier() {
+    return managerIdentifier;
+  }
+
+  public KeySegmentMapper getMapper() {
```
AbfrmBlr: This is not used, or did we miss something

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
```diff
@@ -35,6 +35,7 @@
   VALIDATE_SERVER_STORE,
   RELEASE_SERVER_STORE,
   DESTROY_SERVER_STORE,
```
AbfrmBlr: `CREATE_SERVER_STORE` & `DESTROY_SERVER_STORE` are not relevant message types anymore, right ?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
```diff
@@ -85,7 +87,7 @@
     .mapping(DESTROY_SERVER_STORE_REPLICATION, 66)
```
AbfrmBlr: same for `CREATE_SERVER_STORE_REPLICATION` & `DESTROY_SERVER_STORE_REPLICATION`

clustered/server/src/main/java/org/ehcache/clustered/server/management/Notification.java
```diff
@@ -22,5 +22,6 @@
   EHCACHE_SERVER_STORE_CREATED,
   EHCACHE_SERVER_STORE_ATTACHED,
   EHCACHE_SERVER_STORE_RELEASED,
-  EHCACHE_SERVER_STORE_DESTROYED
+  EHCACHE_SERVER_STORE_DESTROYED,
+  EHCACHE_SERVER_STORE_CLIENT_RECONNECTED
```
mathieucarbou: @anthonydahanne @GaryWKeim  : FYI here are the notifications coming from Ehcache for multi-entity.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusteredTierActiveEntity.java
```diff
@@ -0,0 +1,683 @@
...
+  private final ConcurrentMap<ClientDescriptor, Boolean> connectedClients = new ConcurrentHashMap<>();
+  private final AtomicBoolean reconnectComplete = new AtomicBoolean(true);
+  private final AtomicInteger invalidationIdGenerator = new AtomicInteger();
+  private final ConcurrentMap<Integer, InvalidationHolder> clientsWaitingForInvalidation = new ConcurrentHashMap<>();
+  private final Set<UUID> trackedClients = Collections.newSetFromMap(new ConcurrentHashMap<>());
+  private final InvalidStoreException creationException;
+  private final ReconnectMessageCodec reconnectMessageCodec = new ReconnectMessageCodec();
+  private final ClusterTierManagement management;
+
+  private volatile ConcurrentHashMap<String, List<InvalidationTuple>> inflightInvalidations;
```
AbfrmBlr: `inflightInvalidations` should now just be `List<InvalidationTuple>` because this entity maps to only one store. This should even simply compute on this variable.

clustered/server/src/main/java/org/ehcache/clustered/server/management/ClusterTierManagement.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  public void clientValidated(ClientDescriptor clientDescriptor) {
+    if (managementRegistry != null) {
+      LOGGER.trace("clientValidated({})", clientDescriptor);
+      ClusterTierClientStateBinding clientStateBinding = new ClusterTierClientStateBinding(clientDescriptor, new ClusterTierClientState(storeIdentifier, true));
+      managementRegistry.unregister(clientStateBinding);
+      managementRegistry.registerAndRefresh(clientStateBinding);
+      managementRegistry.pushServerEntityNotification(clientStateBinding, EHCACHE_SERVER_STORE_ATTACHED.name());
```
mathieucarbou: same here:

```
managementRegistry.registerAndRefresh(clientStateBinding)
    .thenRun(() -> managementRegistry.pushServerEntityNotification(clientStateBinding, EHCACHE_SERVER_STORE_ATTACHED.name()));
}
```

clustered/server/src/main/java/org/ehcache/clustered/server/management/ClusterTierManagement.java
```diff
@@ -0,0 +1,132 @@
...
+      ServerSideServerStore serverStore = ehcacheStateService.getStore(storeIdentifier);
+      ServerStoreBinding serverStoreBinding = new ServerStoreBinding(storeIdentifier, serverStore);
+      managementRegistry.register(serverStoreBinding);
+      ServerSideConfiguration.Pool pool = ehcacheStateService.getDedicatedResourcePool(storeIdentifier);
+      if (pool != null) {
+        managementRegistry.register(new PoolBinding(storeIdentifier, pool, PoolBinding.AllocationType.DEDICATED));
+      }
+      managementRegistry.refresh();
+      managementRegistry.pushServerEntityNotification(serverStoreBinding, EHCACHE_SERVER_STORE_CREATED.name());
+    }
```
mathieucarbou: @ljacomet : I forgot to update this part in my previous PR. Now, since called can be delayed we need to use completable futures to delegate the refreshing and push of the notification, in case platform does not send us the entity creation event fast. Is it possible to change to something like this:

```
      LOGGER.trace("init({})", storeIdentifier);
      ServerSideServerStore serverStore = ehcacheStateService.getStore(storeIdentifier);
      ServerStoreBinding serverStoreBinding = new ServerStoreBinding(storeIdentifier, serverStore);
      CompletableFuture<?> r1 = managementRegistry.register(serverStoreBinding);
      ServerSideConfiguration.Pool pool = ehcacheStateService.getDedicatedResourcePool(storeIdentifier);
      CompletableFuture<?> r2;
      if (pool != null) {
        r2 = managementRegistry.register(new PoolBinding(storeIdentifier, pool, PoolBinding.AllocationType.DEDICATED));
      } else {
        r2 = CompletableFuture.completedFuture(null);
      }
      CompletableFuture.allOf(r1, r2)
          .thenRun(() -> managementRegistry.refresh()
              .thenRun(() -> managementRegistry.pushServerEntityNotification(serverStoreBinding, EHCACHE_SERVER_STORE_CREATED.name())));
    }
```

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -278,12 +183,8 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
...
         EhcacheOperationMessage operationMessage = (EhcacheOperationMessage) message;
         EhcacheMessageType messageType = operationMessage.getMessageType();
-        if (isStoreOperationMessage(messageType)) {
-          return invokeServerStoreOperation(clientDescriptor, (ServerStoreOpMessage) message);
-        } else if (isLifecycleMessage(messageType)) {
+        if (isLifecycleMessage(messageType)) {
           return invokeLifeCycleOperation(clientDescriptor, (LifecycleMessage) message);
-        } else if (isStateRepoOperationMessage(messageType)) {
-          return invokeStateRepositoryOperation(clientDescriptor, (StateRepositoryOpMessage) message);
         } else if (isPassiveReplicationMessage(messageType)) {
```
AbfrmBlr: I guess `EhcacheActiveEntity` should never be getting any `PassiveReplicationMessage` since execution strategy prohibits that, right ?

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
ljacomet: removing

clustered/server/src/main/java/org/ehcache/clustered/server/ConcurrencyStrategies.java
```diff
@@ -20,21 +20,38 @@
...
 
-  public static class DefaultConcurrencyStrategy<T extends EntityMessage> implements ConcurrencyStrategy<T> {
+  public static ConcurrencyStrategy<EhcacheEntityMessage> clusterTierManagerConcurrency() {
+    return CLUSTER_TIER_MANAGER_CONCURRENCY_STRATEGY;
+  }
+
+  private static final ConcurrencyStrategy<EhcacheEntityMessage> CLUSTER_TIER_MANAGER_CONCURRENCY_STRATEGY = new ConcurrencyStrategy<EhcacheEntityMessage>() {
+    @Override
+    public int concurrencyKey(EhcacheEntityMessage message) {
+      return 1;
```
AbfrmBlr: Can we use `DEFAULT_KEY ` instead?

clustered/server/src/main/java/org/ehcache/clustered/server/state/config/EhcacheStoreStateServiceConfig.java
ljacomet: Currently unused since we have a single mapping strategy. However, needed in order to move to more dynamic mapping based on cache size. So I suggest to leave it here for now.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
ljacomet: Indeed, removing

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/EhcacheMessageType.java
ljacomet: Indeed, removing

clustered/server/src/main/java/org/ehcache/clustered/server/management/Notification.java
ljacomet: hehe, forgot to delete the `EHCACHE_SERVER_STORE_DESTROYED` one, not used as discussed.

clustered/server/src/main/java/org/ehcache/clustered/server/ConcurrencyStrategies.java
ljacomet: Yes!

clustered/server/src/main/java/org/ehcache/clustered/server/management/ClusterTierManagement.java
mathieucarbou: Perhaps it's worse checking at other places too.... It's possible there's other places. The idea is that
- register must happen first before a refresh
- and pushNotification must happen after a register.
- refresh and push could happen at the same time

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusteredTierActiveEntity.java
ljacomet: Indeed, simplifying

clustered/server/src/main/java/org/ehcache/clustered/server/management/ClusterTierManagement.java
ljacomet: You did not forget that one, I added it. Will look and adapt.

ljacomet: As discussed, doing these changes break tests, so will leave them for a later PR

clustered/client/src/main/java/org/ehcache/clustered/client/internal/EhcacheClientEntityFactory.java
```diff
@@ -258,11 +297,72 @@ private VoltronReadWriteLock createAccessLockFor(String entityIdentifier) {
...
       throw new AssertionError(e);
     }
   }
 
+  public ClusteredTierClientEntity fetchOrCreateClusteredStoreEntity(UUID clientId, String clusterTierManagerIdentifier,
+                                                                     String storeIdentifier, ServerStoreConfiguration clientStoreConfiguration,
+                                                                     boolean autoCreate) throws EntityNotFoundException, CachePersistenceException {
+    EntityRef<ClusteredTierClientEntity, ClusteredTierEntityConfiguration> entityRef;
+    try {
+      entityRef = connection.getEntityRef(ClusteredTierClientEntity.class, ENTITY_VERSION, clusterTierManagerIdentifier + storeIdentifier);
```
AbfrmBlr: Why are we not using the same acquireleadership/lock concept for cache-entity lifecycle as it is used in cachemanager-entity lifecycle. IMHO the purpose served by acquireleadership/lock concept should apply to both caches and cachemanager as far as their respective entities are concerned. That way in ehcache we can have only one way. Just that in case of cache the lock will for that cache, wdyt ?

ljacomet: The entity API provides the guarantee that only one caller will succeed. We do not need anything more here. The locking at the `CacheManager` level is required for maintenance but that is not true for `Cache` level lifecycle.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusteredTierClientEntity.java
```diff
@@ -0,0 +1,268 @@
...
+
+  private final EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint;
+  private final LifeCycleMessageFactory messageFactory;
+  private final AtomicLong sequenceGenerator = new AtomicLong(0L);
+  private final Object lock = new Object();
+  private final ReconnectMessageCodec reconnectMessageCodec = new ReconnectMessageCodec();
+  private final Map<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>> responseListeners =
+    new ConcurrentHashMap<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>>();
+
+  private UUID clientId;
```
albinsuresh: Doesn't this need to be volatile? And the subsequent fields as well?

ljacomet: I do not think so because they are all set before the object is made visible to another thread.

ljacomet: Double checked, since the `SimpleClusteredTierClientEntity` is only ever used by other threads after being assigned to a final field in a `ServerStoreProxy` it is safe.

ljacomet:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
```diff
@@ -309,101 +188,21 @@ private void clearClientTrackedAtReconnectComplete() {
...
-            });
-          if (!mappingsToSend.get().isEmpty()) {
-            syncChannel.synchronizeToPassive(new EhcacheDataSyncMessage(name, mappingsToSend.get()));
-            mappingsToSend.set(new HashMap<>());
-            size.set(0);
-          }
-        });
-    }
-    LOGGER.info("Sync complete for concurrency key {}.", concurrencyKey);
+    // Nothing to sync
```
albinsuresh: Nothing to sync? What about the default server resource configuration and the pool definitions?

ljacomet: Part of the entity configuration

ljacomet:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheActiveEntity.java
albinsuresh: Aah the platform does it. Okay.

albinsuresh:
LGTM. Some cleanup can be done later, like

1. EhcacheStateSyncMessage is not relevant anymore. Can be removed and EhcacheSyncMessageCodec should be cleaned up as well.
2. DefaultConcurrencyStrategy can be simplified as the concurrency key offsetting is not required anymore.
3. ... and some general unused code cleanup.

ljacomet:
@albinsuresh thanks for that quick list - copying it on the issue

PR #1799, opened 2017-01-25
Closes #1800 Abstract out invalidation tracker responsibilities from EhcacheStateService

### LINKED ISSUE #1800: Abstract out invalidation tracker responsibilities from EhcacheStateService

albinsuresh:


albinsuresh:
test this please

ljacomet:
@albinsuresh Can you elaborate on why you think the changes you did make the test more stable?

albinsuresh:
Analyzing the logs of a failure scenario showed some weird ordering problem of server startups that caused the failure. The order of events is as follows:

1. The test class starts up with the class level rule starting up the cluster
2. `testMultipleClients` runs and on the failover step it brings down the active and starts up that node immediately.
3. `testMultipleClients` connects to the newly promoted server and succeeds in its verification.
4. `testMultipleClients` ends with one active and the restarted node still starting up.
5. `testSingleThreadSingleClientInteraction` starts and performs the initial locking part followed by the terminate active call. The active gets terminated when the old node startup is still in progress.
6. Terminate active call is followed by startup one node which will startup the terminate node again. The old node in startup phase still hasn't completed it. At this stage we do not have an active in the cluster but only 2 nodes in startup phase.
7. The old node completes the startup by winning the election and becomes the active.
8. The client in tries to connect to this fresh active and that reconnect request gets rejected as the server does not recognize this client. 
9. Test fails with a connection exception.

Since the cluster was defined as a class level rule, a test running after another could be affected by the side effects of the previous test. On top of that the entities remain created after each test since the tests aren't doing any cleanup.

Now the intent of the test as far as I understand is testing passive failover behavior. So ideally each test should start with a fresh cluster with both nodes up. To test failover, bringing down the active was enough.  I wasn't sure why the brought down server was being started up again before verifying the failover behavior.  I felt hat wasn't necessary. If that was done to ensure that both servers are up and available for the subsequent test then the tests should have at least waited for the started up node to become a passive. And one test relying on the successful execution of the previous test felt wrong. That's why I changed the class level rule a method level one.

PR #1801, opened 2017-01-25
Management registration call fixes

mathieucarbou:


clustered/server/src/main/java/org/ehcache/clustered/server/management/Management.java
```diff
@@ -127,26 +129,31 @@ public void clientDisconnected(ClientDescriptor clientDescriptor, ClientState cl
   public void clientReconnected(ClientDescriptor clientDescriptor, ClientState clientState) {
     if (managementRegistry != null) {
       LOGGER.trace("clientReconnected({})", clientDescriptor);
-      managementRegistry.refresh();
+      managementRegistry.refresh(); // required because ClientState fields have been modified
```
mathieucarbou: @ljacomet : I think this is why the build was failing... These refresh() calls are needed because `ClientState` can have changed

gradle.properties
```diff
@@ -6,4 +6,4 @@ deployUrl = https://oss.sonatype.org/service/local/staging/deploy/maven2/
 # Enable the daemon by adding org.gradle.daemon in USER_HOME/.gradle/gradle.properties
 org.gradle.parallel=true
 
-java6Home=/Library/Java/JavaVirtualMachines/1.6.0_65-b14-462.jdk/Contents/Home
+java6Home=/Library/Java/JavaVirtualMachines/1.6.0.jdk/Contents/Home
```
mathieucarbou: I've commited this line by mistake in a previous commit, but perhaps we can leave it there and set it to `1.6.0.jdk` so that devs just needs to set a symlink from `1.6.0.jdk` to their jdk6 version.
On mac, if installed from brew cask, default folder is `1.6.0.jdk` anyway.

ljacomet: Nope, needs to be removed, as it is a local only thing.

clustered/server/src/main/java/org/ehcache/clustered/server/management/Management.java
```diff
@@ -127,26 +129,31 @@ public void clientDisconnected(ClientDescriptor clientDescriptor, ClientState cl
...
-      ehcacheStateService.getSharedResourcePools()
+      CompletableFuture.allOf(ehcacheStateService.getSharedResourcePools()
         .entrySet()
-        .forEach(e -> managementRegistry.register(new PoolBinding(e.getKey(), e.getValue(), PoolBinding.AllocationType.SHARED)));
-      managementRegistry.refresh();
-      managementRegistry.pushServerEntityNotification(PoolBinding.ALL_SHARED, EHCACHE_RESOURCE_POOLS_CONFIGURED.name());
+        .stream()
+        .map(e -> managementRegistry.register(new PoolBinding(e.getKey(), e.getValue(), PoolBinding.AllocationType.SHARED)))
+        .collect(Collectors.toList())
+        .toArray(new CompletableFuture<?>[0]))
```
ljacomet: this line and the one above can be collapsed in
`.toArray(CompletableFuture[]::new)`

gradle.properties
ljacomet: However, that line is present in release branches, so will do a daggy fix to get rid of it.

mathieucarbou: ok, but what's the difference then with other properties in this file ? To me, if a file contains some local things, it should be ignored from git. Because otherwise, people will keep having untracked modifications locally and it's quite a burden.  Couldn't we get rid of this file then and put in the global README it's content, saying that devs needs to have this file with this content, modified as they need ?

ljacomet: None of the other properties are environment dependent, they are required by the build and stable across all use. The `SONATYPE*` ones are an exception but must exist for the build to work from what I can remember.

ljacomet:
Merging in the face of failures.

PR #1803, opened 2017-01-26
[release/3.2] Remove java6Home definition from gradle.properties

ljacomet:
Defining these props is part of the local project setup, not
something that makes sense to share.


ljacomet:
Merging as the test failures are unrelated

PR #1808, opened 2017-01-26
Checkstyle fix

mathieucarbou:


ljacomet:
Merging in the face of test failures cause by TC platform

PR #1811, opened 2017-01-27
Issue #1475 - add package versions to exported api packages

kcleereman:


ljacomet:
Actually the `;version=<version>` can be passed on the manifest `Export-Package` and thus that's most likely the easiest way - thus closing this

PR #1815, opened 2017-01-30
Issue #1686 For now, prepare for destroy only on active

ljacomet:
This is a temporary change, to last until the _prepare for destroy_ call effectively sets some state server side. Until then, since the passive cannot support the message, make it go to the active only.


ljacomet:
Merging in the face of test failures

PR #1819, opened 2017-02-01
Add tests for passthrough statistics

henri-tremblay:


gibsong:
Is it sufficient to only test eviction count is > 0? Is this check more a test to see if evictions are occurring? I recall testing the exact number of evictions to be a little troublesome.

integration-test/src/test/java/org/ehcache/integration/statistics/TierCalculationTest.java
```diff
@@ -267,4 +276,60 @@ private void innerClear() {
...
+  }
+
+  @Test
+  public void testOccupiedByteSize() {
+    assertThat(tierStatistics.getOccupiedByteSize()).isEqualTo(0);
+    cache.put(1, "a");
+    assertThat(tierStatistics.getOccupiedByteSize()).isGreaterThan(0);
+  }
+
+  @Test
```
gibsong: Is it sufficient to only test eviction count is > 0? Is this check more a test to see if evictions are occurring? I recall testing the exact number of evictions to be a little troublesome.

henri-tremblay: Because when eviction occurs, a "random" number of entries are evicted. So I just want to make sure the statistic works and returns me a number that is not 0 when an eviction has occurred.

integration-test/src/test/java/org/ehcache/integration/statistics/TierCalculationTest.java
```diff
@@ -267,4 +276,60 @@ private void innerClear() {
...
+    do {
+      cache.put(i++, payload);
+      evictions = tierStatistics.getEvictions();
+    }
+    while(evictions == 0 && i < 100_000); // The 100 000 threshold is to prevent an infinite loop in case of a bug
+
+    assertThat(evictions).isGreaterThan(0);
+  }
+
+  @Test
```
gibsong: Does this test pass?  At the moment I am not seeing any ExpirationCount values > 0 .  I am checking via the TMS REST api.

henri-tremblay: Yes, all tests pass :-) But you need to set an expiry policy. By default, nothing expires

gibsong: nice to hear!  I do see ExpirationCount values now.  I just had to swap the clustered tier for the disk tier.  Thanks for your help.

integration-test/src/test/java/org/ehcache/integration/statistics/TierCalculationTest.java
gibsong: ok i guess I will check the same then for the TMS system tests.  I was just curious as I had some issues testing eviction counts in tests, since they are "random".  Its just a mystery I guess.

henri-tremblay:
test this please

PR #1821, opened 2017-02-02
Fix TierCalculationTest that might fail on slow builds

henri-tremblay:


ljacomet:
test this please

For some reason, no build was triggered

ljacomet:
test this please

henri-tremblay:
Interesting. It failed but on a different thing. It really feels like the entry was evicted. I will dig a bit more.

henri-tremblay:
test this please

henri-tremblay:
test this please

integration-test/src/test/java/org/ehcache/integration/statistics/TierCalculationTest.java
```diff
@@ -328,7 +328,7 @@ public void testEviction() {
   @Test
   public void testExpiration() throws InterruptedException {
     cache.put(1, "a");
-    Thread.sleep(100);
+    Thread.sleep(300);
```
ljacomet: huh? please inject a `TimeSource` that enables you to simulate time - that `sleep` needs to go

henri-tremblay: Wasn't aware of the `TimeSource`. Much better solution. Will do.

impl/src/main/java/org/ehcache/impl/internal/DefaultTimeSourceService.java
```diff
@@ -29,6 +29,14 @@
 
   private final TimeSource timeSource;
 
+  public DefaultTimeSourceService() {
+    this.timeSource = SystemTimeSource.INSTANCE;
+  }
+
+  public DefaultTimeSourceService(TimeSource timeSource) {
```
ljacomet: Not needed, see following comment

integration-test/src/test/java/org/ehcache/integration/statistics/TierCalculationTest.java
```diff
@@ -68,6 +74,7 @@ public void before() throws Exception {
       .withCache("cache", cacheConfiguration)
       .using(new DefaultPersistenceConfiguration(diskPath.newFolder()))
       .using(statisticsService)
+      .using(new DefaultTimeSourceService(timeSource))
```
ljacomet: Just use an anonymous inner `TimeSourceService` to hand off the  `TestTimeSource` and then you do not need to add the injection constructor on the default one.

henri-tremblay: Will pass the `TimeSourceServiceConfig` instead

ljacomet:
Unstable test

PR #1828, opened 2017-02-04
:art: Adding license header

anthonydahanne:


henri-tremblay:
Wasn't the build failing when a header was missing previously?

anthonydahanne:
nope; the cloudbees build does not do checkstyle verification
so it bit me while doing a release; a job that _does_ checkstyle verification.

PR #1832, opened 2017-02-07
Issue #1686 Further clean ups following multi entity changes

ljacomet:
This PR contains some more code clean up. It might be easier to review per commit.

It also contains the rename of `EhcacheEntity` into `ClusterTierManagerEntity` ... except for messages which I decided to leave as `EhcacheEntityMessage` and `EhcacheEntityResponse`.


clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerActiveEntity.java
```diff
@@ -79,8 +79,8 @@
   private final AtomicBoolean reconnectComplete = new AtomicBoolean(true);
```
albinsuresh: Adding this comment here as I'm not able to comment on the truncated lines.

`storeClientMap` can be removed and `trackedClients` set is redundant. `clientIdMap` has that information already.

ljacomet: Fixing that

ljacomet:
For 1, aligning on `ClusterTier`, looking into 2

ljacomet:
For 2, that's done with the `connectedClients` field

PR #1835, opened 2017-02-08
Upgrade to new tc-platform v5.2.0-pre18

mathieucarbou:
Compat' with https://github.com/Terracotta-OSS/terracotta-platform/pull/285

### DIFF
--- a/clustered/integration-test/build.gradle
+++ b/clustered/integration-test/build.gradle
@@ -27,7 +27,8 @@ dependencies {
   testCompile project(':dist')
   testCompile project(':clustered:clustered-dist')
   testCompile project(':management')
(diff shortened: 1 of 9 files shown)


clustered/integration-test/build.gradle
```diff
@@ -27,7 +27,8 @@ dependencies {
   testCompile project(':dist')
   testCompile project(':clustered:clustered-dist')
   testCompile project(':management')
-  testCompile "org.terracotta.management.dist:management-client:$parent.managementVersion"
+  testCompile "org.terracotta.management.dist:mnm-nms:$parent.managementVersion"
```
anthonydahanne: much clearer to me 👍  we see here that mnm-nms is only for testing in the context of ehcache

Issue #1814, opened 2017-01-30
HA and stateful serializers: missing syncing of data


ljacomet:
Looks like we missed a feature combination in the 3.2.0 release of Ehcache. We do not sync to the passive server the content of the `StateRepository` we keep inside the `EhcacheStateService`.

This needs to be fixed in 3.2.1 and on master. Given the multi entity work that went into master, I believe this can't/shouldn't be fixed daggy.

ljacomet:
Upon further investigation, what gets sent to the server needs to change as well. That is specific types linked to a serialization library can only exist on the server in binary format.

PR #1838, opened 2017-02-09
[release/3.2] Issue #1814 Passive sync of state repository data

ljacomet:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConcurrentClusteredMap.java
```diff
@@ -106,12 +121,20 @@ public void clear() {
...
+      entries.add(new AbstractMap.SimpleEntry<K, V>(keyCodec.decode(objectEntry.getKey()),
+                                                    valueCodec.decode(objectEntry.getValue())));
+    }
+    return entries;
   }
 
   @Override
   public V putIfAbsent(final K key, final V value) {
-    return (V) getResponse(messageFactory.putIfAbsentMessage(key, value));
+    Object response = getResponse(messageFactory.putIfAbsentMessage(keyCodec.encode(key), valueCodec.encode(value)));
```
albinsuresh: Why isn't there an explicit key and value types check here as done for `get`?

PR #1840, opened 2017-02-09
Refactor InvalidationTrackerImplTest

henri-tremblay:


clustered/server/src/test/java/org/ehcache/clustered/server/state/InvalidationTrackerImplTest.java
```diff
@@ -18,44 +18,97 @@
...
-import java.util.concurrent.ConcurrentMap;
-
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.nullValue;
-import static org.junit.Assert.*;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class InvalidationTrackerImplTest {
 
+  private final InvalidationTrackerImpl tracker = new InvalidationTrackerImpl();
```
albinsuresh: Wouldn't this cause side effects when tests are run in unpredictable order? For example if `trackHashInvalidation_differentKeysCreateDifferentCounts()` is run after `trackHashInvalidation_incrementOnsameValue()` wouldn't it fail on the expected invalidation count of 5? Since the tracker isn't getting reset after every test?

henri-tremblay: No. That's not the way JUnit works. Each test is ran using a brand new test class instance.

PR #1842, opened 2017-02-10
Closes #1834 Client message tracking at cluster tier level

### LINKED ISSUE #1834: Client Message tracking should only be done at cache/server store level.
Client message tracking should only be done at cache/server store level. Its buggy now as each cache has its own sequence generator but EhcacheStateService has one ClientMessageTracker for all caches.

albinsuresh:
Client tracking need not happen at the cluster tier manager level as it adds an unnecessary contention point for different caches. So moving that to the cluster tier level.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
@@ -42,33 +42,6 @@ public class ClusterTierManagerPassiveEntity implements PassiveServerEntity<Ehca
   private final EhcacheStateService ehcacheStateService;
   private final Management management;
 
-  @Override
-  public void invoke(EhcacheEntityMessage message) {
-
-    try {
-      if (message instanceof EhcacheOperationMessage) {
-        EhcacheOperationMessage operationMessage = (EhcacheOperationMessage) message;
-        EhcacheMessageType messageType = operationMessage.getMessageType();
-        if (isLifecycleMessage(messageType)) {
-          invokeLifeCycleOperation((LifecycleMessage) message);
-        } else if (isPassiveReplicationMessage(messageType)) {
-          try {
-            invokeRetirementMessages((PassiveReplicationMessage)message);
-          } catch (ClusterException e) {
-            LOGGER.error("Unexpected exception raised during operation: " + message, e);
-          }
-        } else {
-          throw new AssertionError("Unsupported EhcacheOperationMessage: " + operationMessage.getMessageType());
-        }
-      } else {
-        throw new AssertionError("Unsupported EhcacheEntityMessage: " + message.getClass());
-      }
-    } catch (ClusterException e) {
-      // Reaching here means a lifecycle or sync operation failed
-      throw new IllegalStateException("A lifecycle or sync operation failed", e);
-    }
-  }
-
   public ClusterTierManagerPassiveEntity(ClusterTierManagerConfiguration config,
                                          EhcacheStateService ehcacheStateService, Management management) throws ConfigurationException {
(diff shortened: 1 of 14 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
```diff
@@ -42,34 +42,7 @@
...
-        throw new AssertionError("Unsupported EhcacheEntityMessage: " + message.getClass());
-      }
-    } catch (ClusterException e) {
-      // Reaching here means a lifecycle or sync operation failed
-      throw new IllegalStateException("A lifecycle or sync operation failed", e);
-    }
-  }
-
-  public ClusterTierManagerPassiveEntity(ClusterTierManagerConfiguration config,
+  ClusterTierManagerPassiveEntity(ClusterTierManagerConfiguration config,
```
ljacomet: Why the visibility change?

clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
```diff
@@ -87,31 +60,11 @@ public ClusterTierManagerPassiveEntity(ClusterTierManagerConfiguration config,
...
-    switch (message.getMessageType()) {
-      case VALIDATE:
-        applyMessage(message);
-        break;
-      default:
-        throw new AssertionError("Unsupported LifeCycle operation " + message.getMessageType());
-    }
+  @Override
+  public void invoke(EhcacheEntityMessage message) {
+    throw new AssertionError("No invokes supported by this entity");
```
ljacomet: I am more and more thinking this is a choice that prevents some evolutions. Not sure you need to do anything different in that PR but I will most likely raise an issue on this topic and request a conversation.

clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
albinsuresh: Not really something under the scope of this issue. The same is applicable for other entity classes as well. IDE prompted and since I was touching this file anyway I did it. Want me to revert?

clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
albinsuresh: I kept this here just to make sure that no surprise invocations are coming to this entity.

clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerPassiveEntity.java
ljacomet: Yes please

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -524,12 +553,20 @@ public void handleReconnect(ClientDescriptor clientDescriptor, byte[] extendedRe
     ServerSideServerStore serverStore = stateService.getStore(storeIdentifier);
     addInflightInvalidationsForStrongCache(clientDescriptor, reconnectMessage, serverStore);
 
-    attachStore(clientDescriptor, storeIdentifier);
+    attachStore(clientDescriptor, reconnectMessage.getClientId(), storeIdentifier);
```
AbfrmBlr: As discussed, little more cleanup has to be done in attach store

AbfrmBlr:
needs rebase

PR #1844, opened 2017-02-10
[release/3.2] Issue #1814 Remove findbugs warning

ljacomet:
Suppressing the warning as the type is used in a specific context
where exposing that byte[] is not an issue.


ljacomet:
Replaced by #1845

Issue #1847, opened 2017-02-10
put, update and remove counts/rates at the tier levels are not correct


gibsong:
put, update and remove counts/rates are working at the cache level but not for each tier: onheap, offheap, disk and clustered. The hit/miss counts are not what I am expecting. It appears that the put, update and remove counts are being set to either the hitCount or missCount(In my test these 2 are the same) This also causes the put, update and remove rate stats to be wrong as they are calculated using the corresponding count stat.

PR #1848, opened 2017-02-11
#1847 put, update and remove counts/rates at the tier level are incorrect

gibsong:
The issue is https://github.com/ehcache/ehcache3/issues/1847


henri-tremblay:
I will add some test.

PR #1849, opened 2017-02-13
[master] Closes #1637: Added migration guide for ehcache2 to ehcache3 custom expiry.

### LINKED ISSUE #1637: Document how to simulate Ehcache 2.x per mapping expiration
The question comes up often on StackOverflow, the mailing list or even when doing the shiro and play integrations.

We should have a document showing this as a possible custom expiry implementation.

ravichaturvedi:


impl/build.gradle
```diff
@@ -19,6 +19,7 @@ apply plugin: EhDeploy
 dependencies {
   compile project(':api'), project(':core')
   compile group: 'org.terracotta', name: 'offheap-store', version: parent.offheapVersion
+  compile group: 'net.sf.ehcache', name: 'ehcache', version: parent.ehcache2Version
```
albinsuresh: `testCompile` scope is enough I think.

ravichaturvedi:
Changed the scope of ehcache2 dependency to testCompile and converted sample code to Test.

impl/src/test/java/org/ehcache/docs/Ehcache2.java
```diff
@@ -0,0 +1,78 @@
...
+    CacheManager cacheManager = initCacheManager();
+    CacheConfiguration cacheConfiguration = new CacheConfiguration().name("cache")
+        .maxEntriesLocalHeap(100)
+        .timeToLiveSeconds(20);   // <1>
+    cacheManager.addCache(new Cache(cacheConfiguration));
+
+    Element element = new Element(10L, "Hello");
+
+    // Set the expiry of the element
+    element.setTimeToLive(getTimeToLiveInSeconds((Long)element.getObjectKey(), (String)element.getObjectValue()));  // <2>
```
ljacomet: I would make that example a bit more evolved, getting the TTL and only if different than default above set it.

docs/src/docs/asciidoc/user/migration-guide.adoc
```diff
@@ -0,0 +1,37 @@
...
+endif::notBuildingForSite[]
+
+== Introduction
+
+This guide provide sample code snippets to help migrate ehcache2 code to ehcache3 code.
+
+
+== Custom Expiry
+
+Ehcache2 Code snippet:
```
ljacomet: That's really terse. I would add introductory paragraph explaining the goal of this section and the describe a little bit the example in text before diving into code.

docs/src/docs/asciidoc/user/expiry.adoc
```diff
@@ -84,3 +84,6 @@ include::{sourcedir32}/xml/src/test/resources/configs/docs/getting-started.xml[t
 ----
 
 <1> Simply pass the fully qualified class name of your custom expiry.
+
+
+Ehcache2 to Ehcache3 Expiry migration: <<migration-guide.adoc#custom-expiry,_Custom Expiry_>>
```
ljacomet: The linking from here is good but I would preface this with a bit more text explaining it and link in text instead of having text then link.

ljacomet:
@albinsuresh is right about using tests, so that they get exercised during builds. And he is right as well about using `Thread.sleep()`, do not do it for Ehcache 3 for sure, we have ways to work around it. However that specific code should leave outside of the sample code itself to not be _advertised_. For Ehcache 2, I do not remember from the top of my head, but I believe there was something similar in place.

impl/src/test/java/org/ehcache/docs/Ehcache2.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.docs;
+
+import net.sf.ehcache.*;
```
ljacomet: code style - not `*` imports

AbfrmBlr:
Why this has to live in ehcache code base? Why can't we move it to ehcache3-samples and ask users to refer that. Having a ehcache2 dependency is what I don't like, even if it's for test. @ljacomet , @albinsuresh , @henri-tremblay wdyt?

ljacomet:
@AbfrmBlr having Ehcache in test scope is not an issue for me.

docs/src/docs/asciidoc/user/migration-guide.adoc
```diff
@@ -0,0 +1,48 @@
...
+This guide provides sample code snippets to help migrate Ehcache 2.x code to Ehcache 3.x code.
+
+[[per-mapping-expiry]]
+== Per Mapping Expiry
+
+Per mapping expiry indicates that the cache should expire a mapping (key/value pair) post the defined expiry duration.
+
+
+.Ehcache 2.x Code snippet:
+Here we are creating a cache manager with a cache having default expiry TTL (Time To Live). And before adding the element to the cache we are computing its expiry and only setting it if other than default expiry. +
```
ravichaturvedi: '+' is for adding new line, though at a wrong place here.

ravichaturvedi:
test this please

ravichaturvedi:
test this please

ravichaturvedi:
test this please

PR #1850, opened 2017-02-13
Fix #1814 Passive sync state repository data

### LINKED ISSUE #1814: HA and stateful serializers: missing syncing of data

ljacomet:


clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -584,32 +584,37 @@ private void addInflightInvalidationsForStrongCache(ClientDescriptor clientDescr
...
-        }
-      });
-    if (!mappingsToSend.get().isEmpty()) {
-      syncChannel.synchronizeToPassive(new EhcacheDataSyncMessage(mappingsToSend.get()));
-      mappingsToSend.set(new HashMap<>());
-      size.set(0);
+    if (concurrencyKey == DEFAULT_KEY) {
+      stateService.getStateRepositoryManager().syncMessageFor(storeIdentifier).forEach(syncChannel::synchronizeToPassive);
+    } else {
+      int segmentId = concurrencyKey - DEFAULT_KEY - 1;
```
albinsuresh: Since you are doing this additional offsetting of 1 here, the same needs to reflect in the `DefaultConcurrencyStrategy` as well don't u think? The concurrency key for a `ConcurrentEntityMessage` should be `DEFAULT_KEY + mapper.getSegmentForKey(concurrentEntityMessage.concurrencyKey()) + 1`, no?

ljacomet: Good catch! Fixed and modified a test to guard against future similar mistakes

PR #1851, opened 2017-02-14
Fixes #1834 Perform client ID validation at the clustered tier manager level

### LINKED ISSUE #1834: Client Message tracking should only be done at cache/server store level.

albinsuresh:
Client ID validation done at the clustered tier must be moved to the clustered tier manager level since all the duplicate client ID resolution logic is at the clustered tier manager level and it's earlier in the life cycle.


clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerActiveEntity.java
```diff
@@ -247,9 +248,23 @@ public void destroy() {
    */
   private void validate(ClientDescriptor clientDescriptor, ValidateStoreManager message) throws ClusterException {
     validateClientConnected(clientDescriptor);
+    UUID clientId = clientStateMap.get(clientDescriptor).getClientIdentifier();
```
ljacomet: when is that ever set outside of the reconnect path?

PR #1857, opened 2017-02-15
:snowflake: Close #1856 Create 2 subinterfaces 

### LINKED ISSUE #1856: Need for PoolAllocation sub interfaces
The interface PoolAllocation is directly implemented by Dedicated and Shared in the code base.

Having  2 new sub interfaces such as : DedicatedPoolAllocation and SharedPoolAllocation would allow the management and monitoring stack to provide sizes and resourcePoolNames for several implementations of Dedicated and Shared pools, via the ServerStoreSettingsManagementProvider

anthonydahanne:
to ease other Pool implementations use

### DIFF
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/PoolAllocation.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/PoolAllocation.java
@@ -23,11 +23,20 @@ import java.io.Serializable;
  */
 public interface PoolAllocation extends Serializable {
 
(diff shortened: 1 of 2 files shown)


mathieucarbou:
@ljacomet @anthonydahanne : would it be possible to merge https://github.com/Terracotta-OSS/terracotta-platform/pull/288 so that we can only do 1 ehcache release for both this PR and the -pre15 upgrade ?

PR #1858, opened 2017-02-16
Message tracker refactoring

albinsuresh:


clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -507,8 +517,19 @@ private void invalidateHashForClient(ClientDescriptor originatingClientDescripto
     }
   }
 
+  private void trackMessage(EhcacheOperationMessage message) {
+    ClientMessageTracker clientMessageTracker = stateService.getClientMessageTracker(storeIdentifier);
```
AbfrmBlr: This part of code implies that an Active will always be tracking clients, which is not required. I would suggest to not do that. Instead we can push a state in `DefaultClientMessageTracker` during `loadExisting` that its promoted to Active from passive and should not track now.

clustered/server/src/main/java/org/ehcache/clustered/server/EhcacheStateServiceImpl.java
```diff
@@ -499,6 +498,11 @@ public StateRepositoryManager getStateRepositoryManager() {
   }
 
   @Override
+  public void createClientMessageTracker(String name, boolean fromActive) {
+    this.messageTrackers.put(name, new DefaultClientMessageTracker(fromActive));
```
AbfrmBlr: if its from active, we can choose to not even create this object

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -244,7 +247,9 @@ public EhcacheEntityResponse invoke(ClientDescriptor clientDescriptor, EhcacheEn
   }
 
   private EhcacheEntityResponse invokeStateRepositoryOperation(StateRepositoryOpMessage message) throws ClusterException {
-    return stateService.getStateRepositoryManager().invoke(message);
+    EhcacheEntityResponse response = stateService.getStateRepositoryManager().invoke(message);
```
AbfrmBlr: cool, this one was missed earlier

AbfrmBlr:
test this please

PR #1860, opened 2017-02-16
Simplify life cycle of message and invalidation trackers

albinsuresh:
1. Got rid of InvalidationTrackerManager which was an unnecessary level of indirection.
2. Create/Destroy both invalidation and message trackers along with their respective stores.
3. Make both the tracking happen before the the invocation of the actual message. This helps in avoiding certain error scenarios associated with some racy failures even though it could cause some false invalidations in some rare with failure scenarios. But that doesn't break correctness. So it should be fine.


clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -138,26 +137,23 @@ public ClusterTierActiveEntity(ServiceRegistry registry, ClusterTierEntityConfig
...
   @Override
   public void loadExisting() {
-    stateService.loadStore(storeIdentifier, configuration).setEvictionListener(this::invalidateHashAfterEviction);
-    LOGGER.debug("Preparing for handling Inflight Invalidations and independent Passive Evictions in loadExisting");
     inflightInvalidations = synchronizedList(new ArrayList<>());
     if (!isStrong()) {
       addInflightInvalidationsForEventualCaches();
     }
+    stateService.loadStore(storeIdentifier, configuration).setEvictionListener(this::invalidateHashAfterEviction);
+    LOGGER.debug("Preparing for handling Inflight Invalidations and independent Passive Evictions in loadExisting");
```
AbfrmBlr: this debug statement will move as well. Change the message as well

PR #1862, opened 2017-02-17
Rename Clustered stats to Cluster stat

henri-tremblay:


mathieucarbou:
@ljacomet : I remember this is something we discussed in an M&M scrum with @anthonydahanne @henri-tremblay and @GaryWKeim .

We didn't like "Clustered" because "Clustered" is an adjective, And a HitCount is not Clustered.

So perhaps we should get rid of Cluster/Clustered and replace this with something like: ClusteredStore, or something else ?

ljacomet:
@mathieucarbou the problem is that the stats are actually about a _single_ client view of a given `ClusterTier`. The real `Cluster:HitCount` is the sum across clients. And I admit not having found the term that makes sense. But if indeed you qualify the stats by the type of tier, `Clustered` is the right one for me, not `Cluster`.

mathieucarbou:
@ljacomet other try: I understand your concern. So could we get rid of clustered/cluster and instead having a name much more representing the clustered part of the client ? `Entity:HitCount` perhaps ?

PR #1866, opened 2017-02-21
Fix #1755 ConcurrentSkipListMap for invalidation tracking

### LINKED ISSUE #1755: EhcacheActiveEntity#clientsWaitingForInvalidation is unbounded
The ConcurrentMap org.ehcache.clustered.server.EhcacheActiveEntity#clientsWaitingForInvalidation grows without bound, causing it to eventually exhaust the heap when given a workload with enough clients and threads per client (e.g. 4 clients, 128 test threads per client will quickly exhaust a 2gb heap, and will eventually exhaust a 32gb heap).  We need a way to bound this map so that it will e.g. consume at most Xmb of heap per client.

ljacomet:
This should be merged only if scale testing proves it solves #1755


ljacomet:
Does not help with the issue.

Issue #1772, opened 2017-01-10
Putting large values in cache throws after a while with storage engine failure


AbfrmBlr:
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 2017-01-10 18:20:05,657 [WorkerThread(request_processor_stage, 1, 4)] ERROR org.ehcache.clustered.server.EhcacheActiveEntity - Unexpected exception raised during operation: GET_AND_APPEND
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] org.terracotta.offheapstore.exceptions.OversizeMappingException: Storage Engine and Eviction Failed - Empty Map
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] Storage Engine : org.ehcache.clustered.server.offheap.OffHeapChainStorageEngine@649955de
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.terracotta.offheapstore.AbstractOffHeapClockCache.storageEngineFailure(AbstractOffHeapClockCache.java:84)
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.terracotta.offheapstore.OffHeapHashMap.writeEntry(OffHeapHashMap.java:689)
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.terracotta.offheapstore.OffHeapHashMap.put(OffHeapHashMap.java:471)
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.terracotta.offheapstore.AbstractLockedOffHeapHashMap.put(AbstractLockedOffHeapHashMap.java:145)
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.terracotta.offheapstore.OffHeapHashMap.put(OffHeapHashMap.java:461)
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.terracotta.offheapstore.AbstractLockedOffHeapHashMap.put(AbstractLockedOffHeapHashMap.java:134)
[org.ehcache.clustered.BasicClusteredCacheOpsTest][testServer0] 	at org.ehcache.clustered.server.offheap.OffHeapChainMap.getAndAppend(OffHeapChainMap.java:111)
[...]

PR #1872, opened 2017-02-22
[release/3.1] Change Max page size for offheap store #1772

AbfrmBlr:
For small caches the max page size should be as low as
256 KB so that defragmenetaion in offheap storage is less

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
@@ -21,22 +21,46 @@ import java.util.List;
 
 import org.ehcache.clustered.common.internal.store.Chain;
 import org.ehcache.clustered.common.internal.store.ServerStore;
+import org.ehcache.clustered.server.EhcacheStateServiceImpl.ResourcePageSource;
 import org.ehcache.clustered.server.ServerStoreEvictionListener;
 import org.terracotta.offheapstore.exceptions.OversizeMappingException;
 import org.terracotta.offheapstore.paging.PageSource;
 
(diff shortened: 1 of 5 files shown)


AbfrmBlr:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -21,24 +21,49 @@
...
+  public OffHeapServerStore(ResourcePageSource source, int concurrency) {
+    segments = new ArrayList<OffHeapChainMap<Long>>(concurrency);
+    long maxSize = getMaxSize(source);
+
+    for (int i = 0; i < concurrency; i++) {
+      segments.add(new OffHeapChainMap<Long>(source, LongPortability.INSTANCE, KILOBYTES.toBytes(4), (int) KILOBYTES.toBytes(maxSize), false));
+    }
+  }
+
   public OffHeapServerStore(PageSource source, int concurrency) {
```
ljacomet: Can this constructor used only in tests move to package private?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -21,24 +21,49 @@
...
     segments = new ArrayList<OffHeapChainMap<Long>>(concurrency);
     for (int i = 0; i < concurrency; i++) {
       segments.add(new OffHeapChainMap<Long>(source, LongPortability.INSTANCE, KILOBYTES.toBytes(4), MEGABYTES.toBytes(8), false));
     }
   }
 
+  private static long getMaxSize(ResourcePageSource source) {
```
ljacomet: Good candidate for package private and a few tests

PR #1877, opened 2017-02-22
Register the underlying store of the XAStore (close #1701)

### LINKED ISSUE #1701: java.lang.RuntimeException: store result must not be null with jsr107 and xa-store
It doesn't work to configure an ehcache with XA transactions and jcache.

Details are explained in the [ehcache-users forum](https://groups.google.com/forum/#!category-topic/ehcache-users/ehcache-core/IuZ-XUYGxM8).

Thanks for the support.

henri-tremblay:


ljacomet:
Replaced by daggy fixes PR #1879 #1881 #1882

PR #1880, opened 2017-02-22
Register the underlying store of the XAStore (release 3.2) (#1701)

henri-tremblay:


henri-tremblay:
Should not be on master

PR #1883, opened 2017-02-23
Replace should not get tracked

AbfrmBlr:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -250,10 +250,10 @@ private EhcacheEntityResponse invokeInternal(TimeoutDuration timeLimit, EhcacheE
     }
   }
 
-  private InvokeFuture<EhcacheEntityResponse> internalInvokeAsync(EhcacheEntityMessage message, boolean replicate)
+  private InvokeFuture<EhcacheEntityResponse> internalInvokeAsync(EhcacheEntityMessage message, boolean replicate, boolean track)
         throws MessageCodecException {
     getClientId();
-    if (replicate) {
+    if (track) {
```
ljacomet: This use of replicate was what made me cautious about removing it. Since you are changing to use `track` let's get rid of replicate calls. That has been replaced with the `ExecutionStrategy` concept on the server.

PR #1899, opened 2017-03-02
Still try to get the value when it is expired (close #1598)

### LINKED ISSUE #1598: CacheLoaderWriter and multiple tiers are not playing nice
As reported in the [linked StackOverflow question](http://stackoverflow.com/questions/40396653/ehcache-cacheloaderwriter-load-not-called-after-cache-entry-expired) we have an issue when using a cache loader in a multiple tiers cache.

What happens is that we end up checking for expiry in the caching tier but forget that upon expiration, we may want to enable the authority to effectively load a value instead of simply returning null.

This is again a slight difference in behaviour between a simple cache and a _cache-through_ configured cache.

henri-tremblay:
Seems to look perfectly. The only drawback is when there is no loader-writer and the value expires. The source will be called for nothing since it would have returned null.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
@@ -735,14 +735,27 @@ public class OnHeapStore<K, V> implements Store<K,V>, HigherCachingTier<K, V> {
       }
 
       // If we have a real value (not a fault), we make sure it is not expired
-      // If yes, we return null and remove it. If no, we return it (below)
+      // If yes, we remove it and ask the source just in case. If no, we return it (below)
       if (!(cachedValue instanceof Fault)) {
         if (cachedValue.isExpired(now, TimeUnit.MILLISECONDS)) {
           expireMappingUnderLock(key, cachedValue);
-          getOrComputeIfAbsentObserver.end(CachingTierOperationOutcomes.GetOrComputeIfAbsentOutcome.MISS);
-          return null;
+
+          // On expiration, we might still be able to get a value from the fault. For instance, when a load-writer is used
+          final Fault<V> fault = new Fault<V>(new NullaryFunction<ValueHolder<V>>() {
+            @Override
+            public ValueHolder<V> apply() {
+              return source.apply(key);
+            }
+          });
+          cachedValue = backEnd.putIfAbsent(key, fault);
+
+          if (cachedValue == null) {
+            return resolveFault(key, backEnd, now, fault);
+          }
         }
-        setAccessTimeAndExpiryThenReturnMappingOutsideLock(key, cachedValue, now);
+        else {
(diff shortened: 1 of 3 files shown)


henri-tremblay:
From a benchmark using permanent expiration (the value expires at each get), the solution is 7% slower (disk storage on SSD). Which is that the `get` discovering the expiration will be 7% slower. 

However, we expect this get to be rare and followed by a `put`. When benching for a `get` and `put`, the performance are identical which means that the `put` is so much slower (as expected) that the slower `get` gets erased.

Here is the benchmark if anyone wants to validate.

```java
@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
@Fork(4)
@Threads(Threads.MAX)
@Warmup(iterations = 10, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 5, time = 10, timeUnit = TimeUnit.SECONDS)
public class LoaderWriterBenchmark {

    private static final long SIZE = 100;

    private PersistentCacheManager cacheManager;

    private Cache<Long, String> cache;

    private volatile long key = 1L;

    private String payload;

    @Setup
    public void prepare() {
        Expiry<Long, String> expiry = new Expiry<Long, String>() {
            @Override
            public Duration getExpiryForCreation(Long key, String value) {
                return Duration.ZERO;
            }

            @Override
            public Duration getExpiryForAccess(Long key, ValueSupplier<? extends String> value) {
                return Duration.INFINITE;
            }

            @Override
            public Duration getExpiryForUpdate(Long key, ValueSupplier<? extends String> oldValue, String newValue) {
                return Duration.INFINITE;
            }
        };

        CacheConfigurationBuilder<Long, String> cacheConfiguration =
            CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
                    .heap(SIZE)
                    .disk(1, MemoryUnit.MB))
                .withExpiry(expiry);

        cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
            .with(new CacheManagerPersistenceConfiguration(Paths.get("target", "cache").toFile()))
            .withCache("cache", cacheConfiguration).build(true);

        cache = cacheManager.getCache("cache", Long.class, String.class);

        char[] array = new char[256];
        Arrays.fill(new char[256], 'a');
        payload = String.valueOf(array);
        for (long i = 0; i < SIZE; i++) {
            cache.put(i, payload);
        }
    }

    @TearDown
    public void after() throws CachePersistenceException {
        cacheManager.close();
        cacheManager.destroy();
    }

    @Benchmark
    public String get() {
        String value = cache.get(key);
        key = (key + 1) % SIZE; // yes, this is not atomic but I don't care
        return value;
    }

    @Benchmark
    public String getAndPut() {
        String value = cache.get(key);
        key = (key + 1) % SIZE; // yes, this is not atomic but I don't care
        return value;
    }
}
```

Seing the results, I will go on with the daggy fix.

henri-tremblay:
Replaced by #1902 and #1903

PR #1901, opened 2017-03-06
Delegate pool allocation compatibility checks

albinsuresh:
Delegate the checks from server store configuration level to individual pool allocation types

### DIFF
--- a/clustered/common/src/main/java/org/ehcache/clustered/common/PoolAllocation.java
+++ b/clustered/common/src/main/java/org/ehcache/clustered/common/PoolAllocation.java
@@ -23,6 +23,8 @@ import java.io.Serializable;
  */
 public interface PoolAllocation extends Serializable {
 
+  boolean isCompatible(PoolAllocation other);
+
   interface DedicatedPoolAllocation extends PoolAllocation {
     long getSize();
     String getResourceName();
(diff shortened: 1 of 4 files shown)


clustered/common/src/main/java/org/ehcache/clustered/common/PoolAllocation.java
```diff
@@ -72,6 +74,24 @@ public long getSize() {
     public String getResourceName() {
       return resourceName;
     }
+
+    @Override
+    public boolean isCompatible(PoolAllocation other) {
```
AbfrmBlr: IMHO compatibility check makes sense on `ServerStoreConfiguration`, rather than anywhere else. As I understand its the two configs which we want to see if they are compatible and based on that the pool allocations are compatible or not. I would rather have compatibility check somewhere else.

ljacomet: We have a root type for doing compatibility checks: `ServerStoreCompatibility`
So either we use that type only and move all compatibility checks in it _or_ we agree that we delegate the compatibility check to the graph composing a `ServerStoreConfiguration`. But I do not understand why we would do a bit of both like is the case now.

albinsuresh: The thing that I like about delegation is that it enforces the developer to add the compatibility check in every new `PoolAllocation` that gets added. If every check is in `ServerStoreCompatibility` things are distributed in multiple places and they someone might forget to keep the updates in sync(which is exactly what happened in our case). Of course through testing would expose that hole at some point. And personally I'm not a big fan of huge classes that handle a lot of logic in if-else blocks. But that's just personal opinion. If the rest of you are in favor of the "all in `ServerStoreCompatibility`" approach I'm not completely against that either.

PR #1902, opened 2017-03-06
[master] Still try to get the value when it is expired (close #1598)

### LINKED ISSUE #1598: CacheLoaderWriter and multiple tiers are not playing nice

henri-tremblay:


impl/src/main/java/org/ehcache/impl/internal/store/heap/holders/OnHeapValueHolder.java
```diff
@@ -23,7 +23,7 @@
 /**
  * @author Ludovic Orban
  */
-public abstract class OnHeapValueHolder<V> extends AbstractValueHolder<V> {
+public abstract class   OnHeapValueHolder<V> extends AbstractValueHolder<V> {
```
ljacomet: I do not get that change, it is isolated in a commit but does not get rid of spaces, instead adds them. Can you remove the commit from that branch?

henri-tremblay: Hish... Must have been in a strange mood when I did that. Good catch.

PR #1904, opened 2017-03-07
Add an ExpiryBuilder allowing to refactor redundant code

henri-tremblay:


api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -57,11 +57,24 @@
     return new TimeToIdleExpiry(timeToIdle);
   }
 
+  /**
+   * Builder allowing to specify TTL for creation, access and update. Unspecified values will be set to {@code Duration.INFINITE}
+   * meaning not expiration.
+   *
+   * @return an {@link Expiry} builder
+   */
+  public static <K, V> ExpiryBuilder<K, V> builder() {
```
ljacomet: I am not a 100% convinced about having this builder made public. I understand it cleans up a lot of code in SPI tests, so that's a good thing.
But given the defaults to `Duration.INFINITE` I can't say I want it suddenly widely used.

Also if we wanted it public, it should live in a `.builder.` package.

api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -74,16 +87,21 @@ private Expirations() {
...
 
     @Override
-    public Duration getExpiryForCreation(Object key, Object value) {
+    public Duration getExpiryForCreation(K key, V value) {
       return create;
     }
 
     @Override
-    public Duration getExpiryForAccess(Object key, ValueSupplier<?> value) {
+    public Duration getExpiryForAccess(K key, ValueSupplier<? extends V> value) {
```
ljacomet: Good catch!

api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -108,36 +126,83 @@ public int hashCode() {
     @Override
     public String toString() {
       return this.getClass().getSimpleName() + "{" +
-          "create=" + create +
```
ljacomet: This whitespace change and the method move make the diff harder to read.

api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -108,36 +126,83 @@ public int hashCode() {
...
+  /**
+   * Builder to create a simple {@link Expiry}.
+   *
+   * @param <K> Key type of the cache entries
+   * @param <V> Value type of the cache entries
+   */
+  public static final class ExpiryBuilder<K, V> {
+
+    private Duration create = Duration.INFINITE;
+    private Duration access = Duration.INFINITE;
```
ljacomet: This is the main gripe I have with defaults. This would mean that unless set, all access makes the entry never expire. Seems really wrong as a default.
From the `Expiry` API, a sensible default is `null` here.
But that means all usage in tests needs to specify this value nearly always. Hence my doubts about making this utility public.

api/src/main/java/org/ehcache/expiry/Expirations.java
henri-tremblay: The builder should be in the `builder` package or the `builder()` method as well? If the later, it will be strange for the users since other builder methods are already in `Expirations`

api/src/main/java/org/ehcache/expiry/Expirations.java
henri-tremblay: Sorry. Was probably badly indented and got magically fixed

api/src/main/java/org/ehcache/expiry/Expirations.java
henri-tremblay: What does `null` mean? For me it is quite frequent to want something on one access type and no expiration on everything else. Thus the infinite.

henri-tremblay:
The default has been changed to `null` for access and update. All tests are still working

henri-tremblay:
test this please

api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -57,11 +57,24 @@
     return new TimeToIdleExpiry(timeToIdle);
   }
 
+  /**
+   * Builder allowing to specify TTL for creation, access and update. Unspecified values will be set to {@code Duration.INFINITE}
```
ljacomet: Javadoc update proposal:
> Fluent API for creating an Expiry instance where you can specify constant values for creation, access and update time.
>
> Unspecified values will be set to {@code Duration.INFINITE} for create and {@code null} for access and update, matching the {@link #noExpiration() no expiration} expiry.

PR #1908, opened 2017-03-09
Track messages properly when passive is restarted

AbfrmBlr:


clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -80,6 +85,17 @@ public boolean isEmpty() {
...
+
+    // This happens when a passive is started after Active has moved on and
+    // passive starts to see msgIDs starting from a number > 0.
+    if (highestContiguousMsgId == -1L && nonContiguousMsgIds.size() > 100) {
+      Long min = Collections.min(nonContiguousMsgIds);
+      LOGGER.info("Setting highestContiguousMsgId to {} from -1", min);
+      highestContiguousMsgId = min;
+    }
+
+    nonContiguousMsgIds.removeIf(x -> x <= highestContiguousMsgId);
```
albinsuresh: Why is this line outside the iif block? Shouldn't it be inside it following the highestContiguousMsgId is reset?

AbfrmBlr: yes it should be

AbfrmBlr: No its needed, since the point when `highestContiguousMsgId ` is first set, nonContiguousMsgIds might not have seen those msgIDs less than highestContiguousMsgId. Thus its needs to be cleaned up.

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -80,6 +85,17 @@ public boolean isEmpty() {
    * Remove the contiguous seen msgIds from the nonContiguousMsgIds and update highestContiguousMsgId
    */
   private void reconcile() {
+
+    // This happens when a passive is started after Active has moved on and
+    // passive starts to see msgIDs starting from a number > 0.
+    if (highestContiguousMsgId == -1L && nonContiguousMsgIds.size() > 100) {
```
ljacomet: I do not like that magic number of `100`.

Also when looking more deeply at the class, we should replace the `Set` by a `ConcurrentSkipListSet` which would allow us to simply iterate over it instead of having the `contains` and separated `remove` calls.

ljacomet: I believe the way to get rid of that magic number is to pass the information during passive sync. From my reading of the code, the active is doing tracking as well so it knows what's the highest contiguous value per client UUID.

chrisdennis: Can't we add a field to the end of the message?  3.3.0 passives would not read it and use their existing logic, and would just skip over it... 3.3.1 passives would look for it but default to the 'magic' number if they don't see it.  So at least the problem would be fixed for fully 3.3.1 systems.

AbfrmBlr: @ljacomet nope active don't track message ids because its not required. Only Passives track it, thus there nothing to sync from Active to Passive. 
Also 100 is not a magic number, I have run tests to come to that, also we can discuss that I can explain you

PR #1921, opened 2017-03-16
[master] Verify that all caches have different aliases (close #1919)

### LINKED ISSUE #1919: ehcache.xml with multiple caches with the same alias should fail
Loading this file works perfectly. The last cache is the one used it seems.

It should be forbidden.

```xml
<config
    xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
    xmlns='http://www.ehcache.org/v3'
    xsi:schemaLocation="http://www.ehcache.org/v3 http://www.ehcache.org/schema/ehcache-core.xsd">

  <cache alias="foo">
    <heap>100</heap>
  </cache>

  <cache alias="foo">
    <heap>100</heap>
  </cache>

</config>
```

henri-tremblay:


xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -76,6 +76,7 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IllegalFormatException;
```
ljacomet: Unused import?

henri-tremblay: Yep :-) That's because I can't use the automatic import organizer :-p

ljacomet:
Are all 3 commits needed? Wouldn't this be cleaner in a single commit?

henri-tremblay:
One is a different topic (missing fail). It's a bug in the tests I've noticed. The two others could have been but I've decided to keep them separated. One is the actual fix. One is about the wrong configurations we had in our samples.

PR #1922, opened 2017-03-16
Closes #1920 Compact and replace long chains on mutative operations

### LINKED ISSUE #1920: Compress long chains on cache mutative operations
Cache mutative operations like put, putIfAbsent must try to compress and replace the chain on server if the incoming chain is too long. Remove should always try to replace the incoming chain with an empty one so that server side resource consumption decreases and not increases as it is right now after performing removes.

albinsuresh:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
```diff
@@ -73,7 +79,11 @@ public Chain getCompactedChain() {
 
     @Override
     public boolean isCompacted() {
-      return compacted;
+      return compactedSize > 0;
+    }
+
+    public int getCompactedSize() {
```
ljacomet: this getter reads like it is the compacted chain size, but the use shows it is something different.
What about `getCompactionCount` or something else?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/operations/ChainResolverExpiryTest.java
```diff
@@ -81,6 +81,7 @@ public void testGetExpiryForAccessIsIgnored() {
     verify(expiry, times(1)).getExpiryForUpdate(anyLong(), any(ValueSupplier.class), anyString());
 
     assertThat(resolvedChain.isCompacted(), is(true));
+    assertThat(resolvedChain.getCompactedSize(), is(2));
```
ljacomet: I recommend against duplicating that assertion in all the places. Let's have one set of tests that verifies it behaves as expected and assume it is correct for others.

ljacomet:
Also this should be a daggy fix against 3.2 as well.

albinsuresh:
Daggy fixed as suggested.

PR #1926, opened 2017-03-17
Closes #1925 Fix message tracker thread safety issue

### LINKED ISSUE #1925: Message tracker thread safety issue in seen
In MessageTracker.java, the following check is not thread safe as it is done outside the lock and is not atomic (as there are two checks)

boolean seen = (msgId <= highestContiguousMsgId) || nonContiguousMsgIds.contains(msgId);

Reconcile() method may adjust the same in the following order (under a reconcile lock)
       nonContiguousMsgIds.remove(msgId);
        highestContiguousMsgId = msgId;

The above will cause a thread safety issue where isDuplicate() checks returns false negatives.

rkavanap:
Closes a thread safety issue in the message tracker. Very minor window with very minor possibility of happening, since active does not do message tracking.

Still fixing this for correctness reasons.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
@@ -19,7 +19,6 @@ package org.ehcache.clustered.server.state;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
 import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -66,7 +65,7 @@ public class MessageTracker {
    * @return true if the given msgId is already tracked otherwise false.
    */
   public boolean seen(long msgId) {
-    boolean seen = (msgId <= highestContiguousMsgId) || nonContiguousMsgIds.contains(msgId);
+    boolean seen = nonContiguousMsgIds.contains(msgId) || msgId <= highestContiguousMsgId;
     tryReconcile();
     return seen;
   }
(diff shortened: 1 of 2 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -66,7 +65,7 @@ public void track(long msgId) {
    * @return true if the given msgId is already tracked otherwise false.
    */
   public boolean seen(long msgId) {
-    boolean seen = (msgId <= highestContiguousMsgId) || nonContiguousMsgIds.contains(msgId);
+    boolean seen = nonContiguousMsgIds.contains(msgId) || msgId <= highestContiguousMsgId;
```
ljacomet: This change makes it very inefficient. Since we can't really have a `track` and `seen` for the same `msgId` concurrently, maybe we should have a fast path then the full path, like:
```java
boolean seen = true;
if (msgId > highestContiguousMsgId) {
  seen = nonContiguousMsgIds.contains(msgId) || msgId <= highestContiguousMsgId;
}
```

rkavanap: Unfortunately, 99.99% of the cases msgId will be greater than highestContigousMsgId.. So that guard is really not helping.  as track() is never called on the active... the hope is contains() is almost always on an empty set except at the beginning of a failover..

Also rather than fixing this at this level, I think we should have the isDuplicate() check itself optimized such that on an active (before failover), map lookups are removed.. Today, it will try a map lookup on a concurrent hash map as follows:

   if (clientUUIDMessageTrackerMap.get(clientId) == null) {
      return false;
    }

After the first failover from a passive to active, the map lookup will find an entry.. and in those cases isDuplicate() check comes to this part of the code, but with msgId almost always > than highestContiguousMsgId. But after a while (post failover) nonContiguousMsgId should always be an emptySet.

PR #1927, opened 2017-03-17
Tiering documentation (close #1905)

### LINKED ISSUE #1905: Improve tiering / sizing documentation
While we added content around tiering through #1548, we are still missing some relevant information around the descriptions of the different config options and their meaning.
Also we should see about leveraging the content that remains applicable from the [Ehcache 2.10.x documentation](http://www.ehcache.org/generated/2.10.3/html/ehc-all/#page/Ehcache_Documentation_Set%2Fto-size_sizing_storage_tiers.html%23).

henri-tremblay:


docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -231,3 +169,23 @@ A number of advanced topics that were discussed here have been moved out, see be
...
+
+=== Off-heap
+
+See <<tiering.adoc#off-heap,the tiering documentation>> for more information on this feature.
+
+=== Disk persistence
+
+See <<tiering.adoc#disk-persistence,the tiering documentation>> for more information on this feature.
+
+=== Three tiers
```
ljacomet: I do not feel that one is needed, I would instead but the section id somewhere inside the storage tiers section. A good place would be the `include` of the three tiers example.

impl/src/test/java/org/ehcache/docs/Tiering.java
```diff
@@ -35,6 +51,18 @@
   public TemporaryFolder folder = new TemporaryFolder();
 
   @Test
+  public void tierSizing() {
+    // tag::heap[]
+    ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES); // <1>
+    ResourcePoolsBuilder.newResourcePoolsBuilder().heap(10); // <2>
```
ljacomet: No, that heap method is static and not an instance method on the builder. So let's indicate that. Otherwise its usage in a chain will effectively silently create a new builder. It should only ever be used as the first point and was actually a mistake to add it.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -38,27 +38,118 @@ The following possibilities are valid configurations:
...
+----
+include::{sourcedir33}/impl/src/test/java/org/ehcache/docs/Tiering.java[tag=byteSizedTieredCache]
+----
+
+<1> This will limit the amount of memory used by the heap tier for storing key-value pairs.
+There is a cost associated to sizing objects.
+<2> The settings are only used by the heap tier. So off-heap won't use it at all.
+<3> The sizing can also be further restrained by 2 additional configuration settings:
+The first one specifies the maximum number of objects to traverse while walking the object graph, the second defines the maximum size of a single object.
+If the sizing goes above any of these two limits, the entry won't be store in cache.
```
ljacomet: > stored in the cache

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -38,27 +38,118 @@ The following possibilities are valid configurations:
...
+<3> Defines a resource pool for the disk that will be used by the cache.
+The third parameter is a boolean value which is used to set whether the disk pool is persistent.
+When set to true, the pool is persistent. When the 2 parameters version `disk(long, MemoryUnit)` is used, the pool is not persistent.
+
+The example above allocates a very small amount of disk storage. You will normally use a much bigger storage.
+
+Persistence means the cache will survive a JVM restart. Everything that was in the cache will still be there after restarting
+the JVM and creating a `CacheManager` disk persistence at the same location.
+
+NOTE: A disk tier can't be shared between JVMs. A persistence directory is dedicated to one JVM at the time.
```
ljacomet: The persistence directory cannot even be shared between multiple cache managers

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -38,27 +38,118 @@ The following possibilities are valid configurations:
...
+* You are interested in persistence
+
+NOTE: Ehcache 3 only offers persistence in the case of clean shutdowns (`close()` was called). If the JVM crashes there
+is no data integrity guarantee. At restart, Ehcache will detect that the `CacheManager` wasn't cleanly closed and will wipe the disk
+storage before using it.
+
+=== Clustered
+
+A clustered tier means the client is connecting to a remote Terracotta server where the cached data is put. It is also
+as way to have a shared cache between JVMs. Having so much new possibilities has let it have <<clustered-cache.adoc#,its own section>>
```
ljacomet: > Having so much new possibilities warrants it having

integration-test/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Samples to get started with Ehcache 3. Java 8 syntax
+ */
+@SuppressWarnings("unused")
+public class GettingStarted {
+
+  @Test
+  public void cachemanagerExample() {
+    // tag::java8Example[]
+    try(CacheManager cacheManager = newCacheManagerBuilder() // <1>
```
henri-tremblay: Huh? That's because you have in mind that we never open and close like that. But in fact we do once in a while and don't want to mess with the `UserManagedCached` and its missing services. But I agree that the example isn't showing much so I can remove it.

docs/src/docs/asciidoc/user/caching-concepts.adoc
```diff
@@ -39,22 +39,23 @@ and technical decision based upon the requirements and assumptions of your appli
...
 The caching tier can itself be composed of more than one storage area.
 The _hottest_ data is kept in the caching tier, which is typically less abundant but faster than the authority tier.
 All the data is kept in the authority tier, which is slower but more abundant.
 
 Data stores supported by Ehcache include:
 
 * On-Heap Store - Utilizes Java's on-heap RAM memory to store cache entries. This tier utilizes the same heap memory as
-your Java application, all of which must be scanned by the JVM's garbage collector.  The more heap space your JVM
-utilizes, the more your application's performance will be impacted by garbage collection pauses.  This store is
+your Java application, all of which must be scanned by the JVM garbage collector.  The more heap space your JVM
```
henri-tremblay: It was like this before. It don't want the line to look more modified than needed and be hard to compare. I've only removed the "'s"

docs/src/docs/asciidoc/user/caching-concepts.adoc
```diff
@@ -39,22 +39,23 @@ and technical decision based upon the requirements and assumptions of your appli
...
+utilizes, the more your application performance will be impacted by garbage collection pauses.  This store is
 extremely fast, but is typically your most limited storage resource.
 * Off-Heap Store - Limited in size only by available RAM.
 Not subject to Java garbage collection (GC).
-Is quite fast, yet slower than the On-Heap Store because data must be moved to and from the JVM's heap as it is stored and re-accessed.
+Is quite fast, yet slower than the On-Heap Store because data must be moved to and from the JVM heap as it is stored and re-accessed.
 * Disk Store - Utilizes a disk (file system) to store cache entries.
-This type of storage resource is typically very abundant but much slower than the RAM-based stores.
+This type of storage resource is typically very abundant but much slower than the RAM-based stores. As for all application using disk
+storage, it is recommended to use a fast (e.g. SSD) and dedicated disk to optimize the throughput.
```
henri-tremblay: Ok

integration-test/src/test/java/org/ehcache/docs/GettingStarted.java
henri-tremblay: One thing that is missing: I want to show that a `CacheManager` should be closed. And that `build()` can init it. So I think I will change the example to showcase that instead of Java 7.

docs/src/docs/asciidoc/user/getting-started.adoc
henri-tremblay: Why not. Ok

impl/src/test/java/org/ehcache/docs/Tiering.java
henri-tremblay: I don't get it. The 3 lines of code are of course not valid. They are there to show the 3 ways to create a heap tier. I guess that wasn't clear in the documentation. Let me try something

docs/src/docs/asciidoc/user/tiering.adoc
henri-tremblay: Done

docs/src/docs/asciidoc/user/tiering.adoc
henri-tremblay: done

docs/src/docs/asciidoc/user/tiering.adoc
henri-tremblay: done

henri-tremblay:
The changes were done. I've changed the part with the 3 heap initialization possibilities. Tell me if the doc is clearer.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -88,31 +179,98 @@ With the above into account, the following possibilities are valid configuration
...
-If you want the disk tier to be persistent, it needs to be configured as such, see examples below.
-Note that the disk tier persistence is not crash proof.
-It requires a clean shutdown of the cache manager through the invocation of `CacheManager.close()`.
+Tiers are configured using resource pools. Most of the time using a `ResourcePoolsBuilder`. Let's revisit an example from
+the <<getting-started.adoc#three-tiers, Getting Started>>.
 
 [source,java,indent=0]
 ----
-include::{sourcedir33}/impl/src/test/java/org/ehcache/docs/Tiering.java[tag=diskPersistent]
+include::{sourcedir33}/impl/src/test/java/org/ehcache/docs/GettingStarted.java[tag=threeTiersCacheManager]
```
ljacomet: Since this fragment has callouts, it looks weird in the rendered doc. Why not duplicate it once in code in a version with less or no callouts?
After all, the reason we keep these in code is to make sure they get hit by a refactoring. But I do not mind at all have a duplicate code sections between two docs source files.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -88,31 +179,98 @@ With the above into account, the following possibilities are valid configuration
...
 * heap + clustered
 
-== Persistent tiers
+Here is an example using heap, offheap and clustered.
 
-The _disk_ and _clustered_ tiers introduce the concept of persistence.
-It means that the data they contain can survive a restart of the Java application hosting the cache instance.
+[source,java,indent=0]
+----
+include::{sourcedir33}/clustered/client/src/test/java/org/ehcache/clustered/client/docs/Tiering.java[tag=threeTiersCacheManager]
```
ljacomet: See the other comment on an include with callouts that is not followed by their description.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -231,3 +172,19 @@ A number of advanced topics that were discussed here have been moved out, see be
...
+
+See <<tiering.adoc#update-resourcepools,the tiering documentation>> for more information on this feature.
+
+=== Off-heap
+
+See <<tiering.adoc#off-heap,the tiering documentation>> for more information on this feature.
+
+=== Disk persistence
+
+See <<tiering.adoc#disk-persistence,the tiering documentation>> for more information on this feature.
```
ljacomet: Wrong anchor, I believe you want `#disk` here.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -231,3 +172,19 @@ A number of advanced topics that were discussed here have been moved out, see be
...
 
 See <<usermanaged.adoc#,the user managed cache documentation>> for more information on this feature.
+
+=== Byte-sized heap
+
+See <<tiering.adoc#byte-sized-heap,the tiering documentation>> for more information on this feature.
+
+=== Update ResourcePools
+
+See <<tiering.adoc#update-resourcepools,the tiering documentation>> for more information on this feature.
```
ljacomet: Should this and the two following ones be changed to:
> See the relevant section in the tiering documentation for more information on this feature.

with _relevant section_ being the link.

henri-tremblay: Ok

docs/src/docs/asciidoc/user/getting-started.adoc
henri-tremblay: Indeed

docs/src/docs/asciidoc/user/tiering.adoc
henri-tremblay: Hickes... You are right

PR #1931, opened 2017-03-27
Add missing serializer. Use different names for caches (close #1930)

### LINKED ISSUE #1930: 3.2 branch build is broken with test failures
org.ehcache.docs.GettingStarted.xmlConfigSample test is failing because of multiple caches with the same name "foo" in getting-started.xml

henri-tremblay:
Required because #1919 is now checking that.

### DIFF
--- a/xml/src/test/resources/configs/docs/getting-started.xml
+++ b/xml/src/test/resources/configs/docs/getting-started.xml
@@ -22,6 +22,7 @@
   <!-- tag::gettingStarted[] -->
   <cache alias="foo"> <!--1-->
     <key-type>java.lang.String</key-type> <!--2-->
+    <value-type>java.lang.String</value-type> <!--2-->
     <resources>
       <heap unit="entries">2000</heap> <!--3-->
       <offheap unit="MB">500</offheap> <!--4-->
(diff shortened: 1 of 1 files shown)


xml/src/test/resources/configs/docs/getting-started.xml
```diff
@@ -22,6 +22,7 @@
   <!-- tag::gettingStarted[] -->
   <cache alias="foo"> <!--1-->
     <key-type>java.lang.String</key-type> <!--2-->
+    <value-type>java.lang.String</value-type> <!--2-->
```
albinsuresh: Same callout <2> for both lines. Not sure if the added line is required. But I see that it was added to master some time back. So I guess we need to keep it for consistency.

xml/src/test/resources/configs/docs/getting-started.xml
```diff
@@ -42,7 +43,7 @@
   <!-- end::gettingStarted[] -->
 
   <!-- tag::expiry[] -->
-  <cache alias="foo">
```
albinsuresh: The following 2 changes are not consistent with what was done for master. These 2 caches were removed on master.

xml/src/test/resources/configs/docs/getting-started.xml
henri-tremblay: It is. Since it's a offheap store, you need to specify the type. The callout is on purpose because it's the same comment for both

xml/src/test/resources/configs/docs/getting-started.xml
henri-tremblay: Yes. Because they were moved somewhere else to match the documentation. I didn't want to do that here.

PR #1934, opened 2017-03-28
[master] Closes #1920 Compact and replace long chains on mutative operations

### LINKED ISSUE #1920: Compress long chains on cache mutative operations

albinsuresh:


albinsuresh:
Rebased

PR #1935, opened 2017-03-28
Checkstyle fix

albinsuresh:


integration-test/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -32,8 +31,9 @@
   @Test
   public void cachemanagerExample() {
     // tag::java7Example[]
-    try(CacheManager cacheManager = newCacheManagerBuilder() // <1>
-      .withCache("preConfigured", newCacheConfigurationBuilder(Long.class, String.class, heap(10))) // <2>
+    try(CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder() // <1>
```
henri-tremblay: That's not right. The goal was to show that we can use static imports. So remove it doesn't work

henri-tremblay:
Now fine.

integration-test/src/test/java/org/ehcache/docs/GettingStarted.java
```diff
@@ -19,9 +19,11 @@
 import org.ehcache.CacheManager;
 import org.junit.Test;
 
+// CSOFF: AvoidStaticImport
```
ljacomet: Given how this works, I'd rather have the sample in its own class file named to indicate this is a static import example.
Otherwise we risk other samples to start using static imports silently.

PR #1938, opened 2017-03-28
[master] Fix #1767 Null check in BaseCacheConfiguration

### LINKED ISSUE #1767: No null-check for keyType and valueType in CacheConfigurationBuilder constructor
Because there is no null check for Class<K> keyType, Class<V> valueType in the constructor of  CacheConfigurationBuilder, it is possible to have a CacheConfigurationBuilder with null as key and value types.
This resulted in the following exception when trying to create a Cache with null types, which didn't clearly indicate what the problem was.

```
java.lang.AssertionError: java.lang.IllegalStateException: Cache 'cache_0' creation in EhcacheManager failed.
Caused by: java.lang.IllegalStateException: Cache 'cache_0' creation in EhcacheManager failed.
	at org.ehcache.core.EhcacheManager.createCache(EhcacheManager.java:288)
	at org.ehcache.core.EhcacheManager.createCache(EhcacheManager.java:238)
	at com.terracotta.qa.framework.CacheConfigurationsProvider.apply(CacheConfigurationsProvider.java:77)
	at com.terracotta.qa.framework.BaseTest.singleThreadedTest(BaseTest.java:112)
	... 34 more
Caused by: java.lang.NullPointerException
	at java.lang.Class.isAssignableFrom(Native Method)
	at org.ehcache.impl.internal.spi.serialization.DefaultSerializationProvider.getSerializerClassFor(DefaultSerializationProvider.java:131)
	at org.ehcache.impl.internal.spi.serialization.DefaultSerializationProvider.createSerializer(DefaultSerializationProvider.java:99)
	at org.ehcache.impl.internal.spi.serialization.DefaultSerializationProvider.createKeySerializer(DefaultSerializationProvider.java:79)
	at org.ehcache.core.EhcacheManager.getStore(EhcacheManager.java:448)
	at org.ehcache.core.EhcacheManager.createNewEhcache(EhcacheManager.java:311)
	at org.ehcache.core.EhcacheManager.createCache(EhcacheManager.java:260)
	... 37 more
```

ljacomet:
Key type, value type and resource pools cannot be null.

### DIFF
--- a/core/src/test/java/org/ehcache/core/config/BaseCacheConfigurationTest.java
+++ b/core/src/test/java/org/ehcache/core/config/BaseCacheConfigurationTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.core.config;
+
+import org.ehcache.config.ResourcePools;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.mockito.Mockito.mock;
+
+/**
+ * BaseCacheConfigurationTest
+ */
+public class BaseCacheConfigurationTest {
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  @Test
+  public void testThrowsWithNullKeyType() {
+    expectedException.expect(NullPointerException.class);
+    expectedException.expectMessage("keyType");
+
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/ehcache/core/config/BaseCacheConfigurationTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Collections;
+
+import static org.ehcache.config.ResourceType.Core.HEAP;
+
+/**
+ * BaseCacheConfigurationTest
+ */
+public class BaseCacheConfigurationTest {
+
+  @Test(expected = NullPointerException.class)
```
henri-tremblay: I really don't like the JUnit expected feature. Because you never know if the NPE is because of you code or because of some other issue. I would prefer an `ExpectedException` rule with an `expectMessage` to make sure it is what we want.

core/src/test/java/org/ehcache/core/config/BaseCacheConfigurationTest.java
```diff
@@ -0,0 +1,57 @@
...
+      null, null, getValidResourcePools());
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testThrowsWithNullResourcePools() {
+    new BaseCacheConfiguration<Long, String>(Long.class, String.class, null,
+      null, null, null);
+  }
+
+  private ResourcePoolsImpl getValidResourcePools() {
```
henri-tremblay: Returning a ResourcePools would be enough. Also, you don't really need a fully initialized pool. `return mock(ResourcePools.class)` would work perfectly I think.

ljacomet:
Adressed review remarks

PR #1941, opened 2017-03-29
[3.4] Allow to specify the number of segments created by a disk store (close #1932)

### LINKED ISSUE #1932: ehcache crashes whole JVM if a diskstore tier is utilized on more than 4000 caches
Issue details reported on ehcache-users forum: https://groups.google.com/forum/#!searchin/ehcache-users/ehcache$20crashes$20whole$20JVM%7Csort:relevance/ehcache-users/nPGHVqL_0kI/sDuFgHa8BQAJ

In short. Disk storages are separated in 16 segments. We would like to have it configurable.

henri-tremblay:
Do no merge now. I will add a unit test and some documentation first.

However, to prevent loosing my time, I would like a global validation on the implementation. An attribute was added to `OffHeapDiskStoreConfiguration`. The other thing is that the XSD is modified to accept this new attribute. So we might have to wait for a 3.4 release.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/config/store/disk/OffHeapDiskStoreConfiguration.java
+++ b/impl/src/main/java/org/ehcache/impl/config/store/disk/OffHeapDiskStoreConfiguration.java
@@ -24,11 +24,26 @@ import org.ehcache.spi.service.ServiceConfiguration;
  */
 public class OffHeapDiskStoreConfiguration implements ServiceConfiguration<OffHeapDiskStore.Provider> {
 
+  public static final int DEFAULT_WRITER_CONCURRENCY = 1;
+  public static final int DEFAULT_DISK_SEGMENTS = 16;
+
   private final String threadPoolAlias;
   private final int writerConcurrency;
+  private final int diskSegments;
 
   /**
-   * Creates a new configuration instance using the provided parameters.
+   * Creates a new configuration instance using the provided {@code diskSegments}. Other attributes are set to their default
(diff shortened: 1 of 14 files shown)


chrisdennis:
This looks correct based on my rapidly bit-rotting knowledge.

impl/src/main/java/org/ehcache/impl/config/store/disk/OffHeapDiskStoreConfiguration.java
```diff
@@ -61,6 +80,15 @@ public int getWriterConcurrency() {
   }
 
   /**
+   * Returns the number of disk segments created
+   *
+   * @return the number of segments
+   */
+  public int getDiskConcurrency() {
```
ljacomet: I would call this `getDiskSegments` or `getDiskSegmenting` since that's the name used in javadoc and it would prevent confusion with the `writerConcurrency`.

henri-tremblay: Yes. I was using diskSegments at first. Then thought it was too low level and used the parameter name to `EhcachePersistentConcurrentOffHeapClockCache` which is concurrency. But wasn't quite happy either.

So I can go back to getDiskSegments.

I will add the tests and the documentation. Then we can wait for 3.4.

ljacomet:
Once rebased, this can go into master

henri-tremblay:
Rebased

PR #1944, opened 2017-03-30
[master] Issue #1723 OSGi cleanups

ljacomet:
Handled a number of issues from #1723 to no longer have `ehcache-clustered` require `internal` packages from `ehcache`.

This is however not complete.


henri-tremblay:
Why is there a "Merge branch 'master' into issue-1723-master" commit?

henri-tremblay:
A general question would be: Why do we need a `ConcurrentWeakIdentityHashMap`? I don't feel like the weak part is that needed (but I'm probably missing something) and the identity part can be done by not having any `equals` or hashcode` on stores.

ljacomet:
@henri-tremblay This is daggy fixed against 3.2 as well and did require some conflict solving around imports and one moved line of code.

As for the `ConcurrentWeakIdentityHashMap` we can discuss but it has its use.

henri-tremblay:
Ok. I missed the fact that it was a daggy fix. Good for me

PR #1952, opened 2017-03-31
Moving the highestContiguousMsgId after passive sync is completed.

ravichaturvedi:


clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -79,17 +83,26 @@ public boolean isEmpty() {
...
-    if (highestContiguousMsgId == -1L && nonContiguousMsgIds.size() > 100) {
-      Long min = nonContiguousMsgIds.first();
+    if (highestContiguousMsgId == -1L && nonContiguousMsgIds.size() > 100 && isSyncCompleted) {
+      Long min = nonContiguousMsgIds.last();
       LOGGER.info("Setting highestContiguousMsgId to {} from -1", min);
       highestContiguousMsgId = min;
+      nonContiguousMsgIds.clear();
     }
 
     for (long msgId : nonContiguousMsgIds) {
```
rkavanap: I would suggest putting a warning message here in case the set size grows beyond 200 and never comes down.

clustered/server/src/main/java/org/ehcache/clustered/server/state/MessageTracker.java
```diff
@@ -79,17 +84,32 @@ public boolean isEmpty() {
...
    * Remove the contiguous seen msgIds from the nonContiguousMsgIds and update highestContiguousMsgId
    */
   private void reconcile() {
 
     // This happens when a passive is started after Active has moved on and
     // passive starts to see msgIDs starting from a number > 0.
-    if (highestContiguousMsgId == -1L && nonContiguousMsgIds.size() > 100) {
-      Long min = nonContiguousMsgIds.first();
+    // Once the sync is completed, fast forward highestContiguousMsgId.
+    if (highestContiguousMsgId == -1L && isSyncCompleted) {
```
ravichaturvedi: Removed the number 100 check, as it is no longer required. Since after sync completion all future msgIds should arrive at passive.

PR #1953, opened 2017-04-03
Compat with async management message receiving + Upgrade to 5.3.0-pre1

mathieucarbou:
Depends on https://github.com/Terracotta-OSS/terracotta-platform/pull/300

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
@@ -115,12 +115,12 @@ public class ClusterTierManagerClientEntityFactory {
     boolean finished = false;
 
     try {
-      EntityRef<InternalClusterTierManagerClientEntity, ClusterTierManagerConfiguration> ref = getEntityRef(identifier);
+      EntityRef<InternalClusterTierManagerClientEntity, ClusterTierManagerConfiguration, Void> ref = getEntityRef(identifier);
       try {
         while (true) {
           ref.create(new ClusterTierManagerConfiguration(identifier, config));
           try {
-            InternalClusterTierManagerClientEntity entity = ref.fetchEntity();
+            InternalClusterTierManagerClientEntity entity = ref.fetchEntity(null);
             try {
               entity.setTimeouts(entityTimeouts);
               finished = true;
@@ -177,7 +177,7 @@ public class ClusterTierManagerClientEntityFactory {
 
     InternalClusterTierManagerClientEntity entity;
     try {
-      entity = getEntityRef(identifier).fetchEntity();
+      entity = getEntityRef(identifier).fetchEntity(null);
     } catch (EntityVersionMismatchException e) {
       LOGGER.error("Unable to retrieve cluster tier manager for id {}", identifier, e);
       silentlyUnlock(fetchHold, identifier);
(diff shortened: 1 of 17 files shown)


mathieucarbou:
@ChrisGreenaway @anthonydahanne @ljacomet : I've released tc-platform `5.3.0-pre1`.
I've done in this PR the changes necessary for the management changes.

@ChrisGreenaway : can I let you add the necessary changes for the api change for Ehcache ? My branch is opened.

Thanks!

clustered/integration-test/build.gradle
```diff
@@ -24,6 +24,12 @@ configurations {
 }
 
 dependencies {
+  // not required by gradle but required by the IDE because 'dist' do not has any transitive dependencies
+  testCompile project(':clustered:client')
+  testCompile project(':clustered:common')
+  testCompile project(':impl')
+  testCompile project(':xml')
+
```
chrisdennis: Can we do these changes in a separate PR?

mathieucarbou: yes

anthonydahanne:
test this please
latest platform jars are now published; but I expect it will fail because of the fetchEntity change of API

clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
```diff
@@ -157,23 +157,19 @@ public static void beforeClass() throws Exception {
...
-    assertThat(counts.get("SERVER_ENTITY_UNFETCHED"), hasSize(3));
-
-    assertThat(readMessages(), hasSize(0));
+    waitForAllNotifications(
+      "CLIENT_CONNECTED",
+      "CLIENT_REGISTRY_AVAILABLE",
+      "CLIENT_TAGS_UPDATED",
+      "EHCACHE_CLIENT_VALIDATED",
+      "EHCACHE_RESOURCE_POOLS_CONFIGURED",
+      "EHCACHE_SERVER_STORE_CREATED", "EHCACHE_SERVER_STORE_CREATED", "EHCACHE_SERVER_STORE_CREATED",
```
chrisdennis: Would be real nice if this could be done with `nCopies(...)`.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
```diff
@@ -115,12 +115,12 @@ public void create(final String identifier, final ServerSideConfiguration config
     boolean finished = false;
 
     try {
-      EntityRef<InternalClusterTierManagerClientEntity, ClusterTierManagerConfiguration> ref = getEntityRef(identifier);
+      EntityRef<InternalClusterTierManagerClientEntity, ClusterTierManagerConfiguration, Object> ref = getEntityRef(identifier);
```
chrisdennis: If we're not using this type it should probably be `Void`.

anthonydahanne: ah, it could be, true

clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
mathieucarbou: I will even do it without nCopies if we could do our ehcache tests in groovy.

PR #1956, opened 2017-04-10
[master] Add byte sizing maximum defaults

henri-tremblay:


docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -82,7 +82,8 @@ include::{sourcedir33}/impl/src/test/java/org/ehcache/docs/Tiering.java[tag=byte
 There is a cost associated to sizing objects.
 <2> The settings are only used by the heap tier. So off-heap won't use it at all.
 <3> The sizing can also be further restrained by 2 additional configuration settings:
-The first one specifies the maximum number of objects to traverse while walking the object graph, the second defines the maximum size of a single object.
+The first one specifies the maximum number of objects to traverse while walking the object graph (default: 1000), the second
+defines the maximum size of a single object (default: unlimited).
```
ljacomet: Is it not `Long.MAX_VALUE`?

ljacomet: Let's be truthful saying what the limit is and adding that in practice it means unlimited.

henri-tremblay: done

ljacomet: I do not see that change.

henri-tremblay: Refresh. I do see it.

Issue #1960, opened 2017-04-14
Errors in documentation


ljacomet:
* http://www.ehcache.org/documentation/3.3/getting-started.html#creating-a-cache-manager-with-clustering-support - callouts do not match code lines.
* http://www.ehcache.org/documentation/3.3/clustered-cache.html#clustered-unspecified-inheritance - review the text and clarify the `clustered()` feature
* http://www.ehcache.org/documentation/3.3/thread-pools.html#let-s-start-with-a-bit-of-theory - review the section and the last sentence in particular
* http://www.ehcache.org/documentation/3.3/serializers-copiers.html#how-is-a-serializer-configured - by _class name_ should be by _class_ 
* http://www.ehcache.org/documentation/3.3/serializers-copiers.html#lifecycle-instances-vs-class-names - confusing sentence "It will need to dispose of any resource the serializer might hold upon and/or persisting and reloading the serializer's state."
* http://www.ehcache.org/documentation/3.3/serializers-copiers.html#writing-your-own-serializer - section before code extract has some duplication, check
* http://www.ehcache.org/documentation/3.3/eviction-advisor.html - review text in before last note, unclear
* http://www.ehcache.org/documentation/3.3/cache-event-listeners.html - Improve table headers
* http://www.ehcache.org/documentation/3.3/cache-event-listeners.html#event-processing-queues - missing callout text
* http://www.ehcache.org/documentation/3.3/tiering.html#disk - callouts out of order
* http://www.ehcache.org/documentation/3.3/tiering.html#clustered - review sentence
* http://www.ehcache.org/documentation/3.3/tiering.html#destroy-persistent-tiers - some weird sentences
* http://www.ehcache.org/documentation/3.3/xa.html#xa-cache-with-three-tiers-and-persistence - in sample, use `persistence` method

PR #1965, opened 2017-04-22
Adjust callouts to match explanations #1960

jdsandifer:
Moved callout 4 to location of previous callout 5 for clusteredCacheManagerExample() to match explanation in GettingStarted doc.

### DIFF
--- a/docs/src/docs/asciidoc/user/getting-started.adoc
+++ b/docs/src/docs/asciidoc/user/getting-started.adoc
@@ -119,7 +119,8 @@ include::{sourcedir33}/clustered/client/src/test/java/org/ehcache/clustered/clie
     The sample URI provided in the example points to the clustered storage with clustered storage identifier *my-application* on the Terracotta server (assuming the server is running on localhost and port *9510*); the query-param `auto-create`
     creates the clustered storage in the server if it doesn't already exist.
 <3> Returns a fully initialized cache manager that can be used to create clustered caches.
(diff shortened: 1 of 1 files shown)


terracotta-org:

jdsandifer:
Thanks, Henri. I added the explanation for #4 and re-numbered the last one (now #5) in the doc.

To revert the code change I made, I just added another commit that fixed it (I'm just working straight from GitHub.com right now and don't know how to revert to an earlier commit). Will those commits go away after the merge or is there a way for me to remove the first two commits so they don't pollute the commit history?

henri-tremblay:
In github I don't know. To make it clean, you need to squash them during an interactive rebase. And then push force the change.

More easily, you should have amended your commit (`--amend --no-edit`) and then push force. That would have been easier.

jdsandifer:
Ok, I think I fixed the multiple commits. Does this look the way it should now?

ljacomet:
Hi @jdsandifer,

Yes it looks ok now.

Thank you very much for your interest in contributing to the project! Please read about submitting contributions here: http://www.ehcache.org/community/contribute.html and confirm in a reply comment that you agree to the grant statement under the heading "Agreement and Grant of Copyright"

jdsandifer:
I agree to the grant statement.

ljacomet:
Commit included in #2035 and #2036

PR #1970, opened 2017-04-25
[master] Fix #1963 Support multiple servers in cluster URI

### LINKED ISSUE #1963: Too strict URI regex for cluster connections
The regexp doing XSD validation for cluster URI is too strict. In a HA setup, the following is a valid Terracotta connection URL:
`terracotta://server1:port,server2:port/myCacheManager`

However this is marked as invalid by the current XSD.

ljacomet:


clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -77,7 +77,7 @@
 
   <xs:simpleType name="connectionUrl">
     <xs:restriction base="xs:anyURI">
-      <xs:pattern value="\w+://([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(/[^\?#]*)?(\?[^#]*)?(#.*)?"/>
+      <xs:pattern value="\w+://([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?(,([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?)*(/[^\?#]*)?(\?[^#]*)?(#.*)?"/>
```
chrisdennis: I think this is right but it might not be the simplest/best way of expressing it.

clustered/client/src/test/java/org/ehcache/clustered/client/HAValidXmlTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.Assert.assertThat;
+
+/**
+ * HAValidXmlTest
+ */
+public class HAValidXmlTest {
+
+  @Test
+  public void testCanLoadXmlConfigValidForHA() {
```
chrisdennis: Might be good to have an negative test too.

ljacomet: You mean a test with an invalid cluster URI? If so, should I include tests with failure in each sections? scheme? servers? path?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
cljohnso: Pattern looks right.  Not sure I agree with @chrisdennis that a simpler expression might be available ... you're kinda stuck since the replicated part -- `(,([^\]\[/?#@]+@)?[^:?#/]+(:[1-9][0-9]{0,4})?)*` -- begins with a comma so the pattern following it needs replication.

clustered/client/src/test/java/org/ehcache/clustered/client/HAValidXmlTest.java
ljacomet: I answered my own question - since any pattern validation is reported through the same exception, a single test is enough and has been added.

PR #1987, opened 2017-05-16
[master] Shutdowns all underlying executors (close #1986)

### LINKED ISSUE #1986: PooledExecutionService does not shut down all the thread pools if manager is closed; blocks application exit
If the PooledExecutionService is used and a thread from one of the pools is used (for example with write back), the application does not close at the end since a threadpool is still running. Please test as application, as junit does not seam to care.

Reason: 
The threadpool in the file [PooledExecutionService.java in line 54](https://github.com/ehcache/ehcache3/blob/master/impl/src/main/java/org/ehcache/impl/internal/executor/PooledExecutionService.java#L54)  is not closed.
If it is closed (by reflection) the application works as intended. If needed I can provide an example with reflection.

Example (taken from the documentation of write back and thread pool):
```
public class WriteBehindTest {


    public static void main(final String[] args) throws URISyntaxException{
        WriteBehindTest test = new WriteBehindTest();
        test.testWriteBehind();
    }

    //@Test // it works from inside the junit test and open threadpools seam to be no problem
    public void testWriteBehind(){


        CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
                .using(PooledExecutionServiceConfigurationBuilder.newPooledExecutionServiceConfigurationBuilder()
                        .defaultPool("dflt", 0, 10)
                        .pool("defaultWriteBehindPool", 1, 3)
                        .build())
                .withDefaultWriteBehindThreadPool("defaultWriteBehindPool") .build(true);

        Cache<Long, String> writeBehindCache = cacheManager.createCache("writeBehindCache",
                CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class, ResourcePoolsBuilder.heap(10))
                .withLoaderWriter(new SampleLoaderWriter<Long, String>(Collections.singletonMap(41L, "zero")))
                .add(WriteBehindConfigurationBuilder
                        .newBatchedWriteBehindConfiguration(1, TimeUnit.SECONDS, 3)
                        .queueSize(3)
                        .concurrencyLevel(1)
                        .enableCoalescing())
                .build());

        assertThat(writeBehindCache.get(41L), is("zero"));
        writeBehindCache.put(42L, "one");
        writeBehindCache.put(43L, "two");
        writeBehindCache.put(42L, "This goes for the record");
        assertThat(writeBehindCache.get(42L), equalTo("This goes for the record"));


        cacheManager.close();

    }

}
```

henri-tremblay:


henri-tremblay:
Daggy fix?

ljacomet:
Yes please, to 3.2, 3.3 and master

ljacomet:
Looks like the build failure is related to the added test.

impl/src/test/java/org/ehcache/impl/internal/executor/PooledExecutionServiceTest.java
```diff
@@ -105,4 +108,23 @@ public void testGetOrderedExecutorSucceedsOnExistingDefaultPool() throws Excepti
...
+      .schedule(new Runnable() {
+        @Override
+        public void run() {
+
+        }}, 10, TimeUnit.HOURS);
+    pooledExecutionService.stop();
+
+    Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
+    for(Thread thread : threadSet) {
+      assertThat(thread.getName()).doesNotContain("pool");
```
chrisdennis: This only matches the string from the underlying ThreadPoolExecutor... we should probably also validate that the "dflt" pool threads have gone away.  (Which would also implying starting them with a task that gets run before we shutdown).

henri-tremblay:
The test should pass now hopefully. And it is now checking all threads. I've given an alias to the scheduled executor and did some refactoring.

impl/src/main/java/org/ehcache/impl/internal/executor/PooledExecutionService.java
```diff
@@ -129,21 +112,27 @@ public void start(ServiceProvider<Service> serviceProvider) {
...
-        if (e.getKey() != null) {
-          destroyPool(e.getKey(), e.getValue());
-        }
+        destroyPool(e.getKey(), e.getValue());
       } finally {
         it.remove();
       }
     }
+    try {
+      scheduledExecutor.awaitTermination(30, SECONDS);
```
chrisdennis: Shouldn't this be in a similar loop to the one used in the `destroyPool` method?

PR #1997, opened 2017-05-31
Including pom.xml and pom.properties in all published jars

akomakom:
Including maven metadata (pom.xml/pom.properties) in ehcache-common.jar only.  (primarily for internal use).

### DIFF
--- a/buildSrc/src/main/groovy/EhDeploy.groovy
+++ b/buildSrc/src/main/groovy/EhDeploy.groovy
@@ -55,6 +55,8 @@ class EhDeploy implements Plugin<Project> {
     }
 
     def artifactFiltering = {
+      pom.scopeMappings.mappings.remove(project.configurations.testCompile)
+      pom.scopeMappings.mappings.remove(project.configurations.testRuntime)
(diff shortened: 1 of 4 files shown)


akomakom:
This is the best way I found, and I made the assumption that this should only be done in one jar.  This could probably be abstracted to be done globally if needed.

henri-tremblay:
I'm curious. You want to use it for what?

GaryWKeim:
So you can crack it open and see useful information about the artifact. I'd also want, somewhere, the location of the repo it was born from. In the olden days it was quite a task to determine from where some artifact came.

akomakom:
It's still quite a task... 

Answer: This change will allow the escrow plugin to identify it as an internal jar and then map it to where it came from.  I can discuss internal details in another channel if you're interested.

henri-tremblay:
Ok. Thanks. I'm looking if we can make something simpler.

akomakom:
Discussed with @ljacomet , I may be moving this to EhDeploy so it applies to all artifacts.

akomakom:
Made a separate plugin, pom.* is now in all jars.

akomakom:
I updated the maven plugin with new mappings so it should be ready once this is merged.

PR #1999, opened 2017-05-31
Close #1996 : use removeCache

### LINKED ISSUE #1996: destroyCache should use the "regular" removeCache method
in [EhcacheManager#683](https://github.com/ehcache/ehcache3/blob/e496850c8bf8191a30d7a783c80944c56d2cdc74/core/src/main/java/org/ehcache/core/EhcacheManager.java#L683) , ```destroyCache``` is calling ```removeAndCloseWithoutNotice``` that unfortunately does not trigger ```cacheRemoved```.

It can be problematic, since the user can destroy a cache programmatically without the monitoring capabilities knowing that a cache was indeed removed

anthonydahanne:


anthonydahanne:
meh, test all threads failed
test this please

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -679,7 +679,7 @@ public void destroyCache(final String alias) throws CachePersistenceException {
     }
 
     try {
-      removeAndCloseWithoutNotice(alias);
+      removeCache(alias);
```
ljacomet: should invoke `removeCache(alias, true);` instead

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -679,7 +679,7 @@ public void destroyCache(final String alias) throws CachePersistenceException {
     }
 
     try {
-      removeAndCloseWithoutNotice(alias);
```
ljacomet: This method is orphaned by removing this call, please delete it.

core/src/main/java/org/ehcache/core/EhcacheManager.java
ljacomet: Not a chance this goes in without tests 😉 
And it needs daggy fix against `release/3.2`, `release/3.3` and `master`

ljacomet:
@henri-tremblay from looking at history and trying to recover from memory, I believe there used to be stricter checks when removing a cache. And looking at the state of the code now, I don't see an issue using the same code and notifying indeed.

core/src/main/java/org/ehcache/core/EhcacheManager.java
anthonydahanne: yup

core/src/main/java/org/ehcache/core/EhcacheManager.java
anthonydahanne: oh, you mean ... we don't already test destroyCache() ? ;-)
I'll have a look. I'll also get a very deep breath for daggy fixing...

henri-tremblay: Yes. But we are not testing that there are no notifications :-) Also, if in maintenance mode from destroyCache, it should stay there all the way through. Not sure this is tested as well.

anthonydahanne:
replaced by #2004 #2005 #2006

PR #2001, opened 2017-06-01
Support for testing (and passing) against Java 9

chrisdennis:


build.gradle
```diff
@@ -82,24 +82,31 @@ if (deployUrl.contains('nexus')) {
 
 // Java 6 build setup
 def java6Error = 'Set the poperty \'java6Home\' in your $HOME/.gradle/gradle.properties pointing to a Java 6 installation'
-assert (JavaVersion.current().isJava8Compatible()) : 'The Ehcache 3 build requires Java 8 to run and a configured Java 6 installation\n' + java6Error
+assert (JavaVersion.current().isJava8()) : 'The Ehcache 3 build requires Java 8 to run and a configured Java 6 installation\n' + java6Error
```
henri-tremblay: I guess java 9 is java 8 compatible? Why do you want java 8 only?

impl/src/test/java/org/ehcache/impl/internal/store/heap/ByteSizedOnHeapStoreByValueSPITest.java
```diff
@@ -37,16 +37,25 @@
...
+import static org.hamcrest.core.StringStartsWith.startsWith;
+import static org.junit.Assume.assumeThat;
 
 public class ByteSizedOnHeapStoreByValueSPITest extends StoreSPITest<String, String> {
 
   private StoreFactory<String, String> storeFactory;
   private static final int MAGIC_NUM = 500;
 
+  @BeforeClass
+  public static void ignoreInJava9() {
```
henri-tremblay: Can we have a comment telling why it's not working?

henri-tremblay: And shouldn't it work with `--permit-illegal-access`?

osgi-test/build.gradle
```diff
@@ -66,6 +66,8 @@ test {
   systemProperty 'ehcache.osgi.jar', project(':dist').jar.archivePath.getPath()
   systemProperty 'ehcache.osgi.jcache.version', parent.jcacheVersion
   systemProperty 'ehcache.osgi.slf4j.version', parent.slf4jVersion
+}.doFirst {
```
henri-tremblay: That is sad. So you know if it's getting fixed? Maybe we should print a warning to tell people building the project that a part is missing

clustered/integration-test/build.gradle
```diff
@@ -70,8 +70,16 @@ compileTestJava {
...
   dependsOn copyServerLibs
-  if (!testJava.javaVersion.isJava8Compatible()) {
+  if (testJava.javaVersion.isJava8Compatible()) {
+    environment 'JAVA_HOME', testJava.javaHome
+  } else {
     executable = Jvm.current().javaExecutable
+    environment 'JAVA_HOME', Jvm.current().javaHome
+  }
+  //TC Server Cannot Parse Java 9 VM Version String
+  if (testJava.javaVersion.isJava9Compatible()) {
```
henri-tremblay: I guess it should be a if isJava9Compatible else if isJava8Compatible else

impl/src/test/java/org/ehcache/impl/internal/store/heap/ByteSizedOnHeapStoreByValueSPITest.java
chrisdennis: You're right - this stuff might work now with the very latest EAs (or with `--permit-illegal-access`).  I'll test to be sure.

osgi-test/build.gradle
chrisdennis: I couldn't get this to work... but I didn't dig too hard.  Nothing is missing here except test coverage though.  I think the best thing to do would be to file an issue to try and fix the test harness (if that is possible).

chrisdennis: Looked at this briefly again, fixing it is a little more complicated.  I currently believe having one version of these tests thats work on both 1.6 and 9 is going to be impossible.  I'll let @ljacomet  decide what he wants to do here.

henri-tremblay: @ljacomet Let's drop 6 then? :-)

chrisdennis: https://issues.apache.org/jira/browse/FELIX-5575

PR #2002, opened 2017-06-01
[master] Check that an alias wasn't already existing when adding it (close #2000)

### LINKED ISSUE #2000: One CacheManager with same name caches in builder
When I create one CacheManager and try to create two caches with the same alias(one clustered and one not clustered) within this CacheManager only 1 cache is actually created!  This doesn't seem like the appropriate behavior in this case.  I create 2 caches but only get 1?  And more importantly I am unaware that I only have 1 cache.  It appears that the builder ends up using the last withCache(...) call and the previous ones are overwritten/ignored.

What is the expected behavior in this situation?  Should this throw an error?

Below is the configuration that exposed this issue:

    String cacheClusteredAlias = "cacheAliasSameName";
    String cacheUnclusteredAlias = "cacheAliasSameName";

    CacheConfiguration<Long, String> cacheConfigClustered = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long, String, newResourcePoolsBuilder()
        .heap(10, MemoryUnit.MB)
        .offheap(50, MemoryUnit.MB)
        .with(ClusteredResourcePoolBuilder.clusteredDedicated("primary-server-resource", 100, MemoryUnit.MB)))
        .build();

    CacheConfiguration<Long, String> cacheConfigUnclustered = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long, String, newResourcePoolsBuilder()
        .heap(10, MemoryUnit.MB)
        .offheap(100, MemoryUnit.MB))
        .build();

    cacheManager1 = CacheManagerBuilder.newCacheManagerBuilder()
        .with(ClusteringServiceConfigurationBuilder.cluster(URI.create(connectionURIStr + "/" + CACHE_MANAGER_NAME))
        .autoCreate()
        .resourcePool("resource-pool-a", 128, MemoryUnit.MB, "secondary-server-resource")
        .resourcePool("resource-pool-b", 128, MemoryUnit.MB))
        .withCache(cacheClusteredAlias, cacheConfigClustered)
        .withCache(cacheUnclusteredAlias, cacheConfigUnclustered)
        .using(new DefaultManagementRegistryConfiguration()
        .setCacheManagerAlias(CACHE_MANAGER_NAME)).build(true);

henri-tremblay:


impl/src/test/java/org/ehcache/config/builders/CacheManagerBuilderTest.java
```diff
@@ -51,16 +54,16 @@ public void testIsExtensible() {
       }
     }).build(true);
 
-    assertThat(cacheManager, nullValue());
-    assertThat(counter.get(), is(1));
+    assertThat(cacheManager).isNull();
```
anthonydahanne: meh, they were fine like this

impl/src/test/java/org/ehcache/config/builders/CacheManagerBuilderTest.java
```diff
@@ -51,16 +54,16 @@ public void testIsExtensible() {
...
+    assertThat(counter.get()).isEqualTo(1);
   }
 
   @Test
   public void testCanOverrideCopierInConfig() {
     @SuppressWarnings("unchecked")
     CacheManagerBuilder<CacheManager> managerBuilder = newCacheManagerBuilder()
         .withCopier(Long.class, (Class) IdentityCopier.class);
-    assertNotNull(managerBuilder.withCopier(Long.class, SerializingCopier.<Long>asCopierClass()));
+    assertThat(managerBuilder.withCopier(Long.class, SerializingCopier.<Long>asCopierClass())).isNotNull();
```
anthonydahanne: not sure that makes it anymore readable

anthonydahanne:
test this please
weird failure that does not seem to be related with that change

impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
```diff
@@ -78,7 +78,9 @@ public Configuration build() {
 
   ConfigurationBuilder addCache(String alias, CacheConfiguration<?, ?> config) {
     Map<String, CacheConfiguration<?, ?>> newCaches = new HashMap<String, CacheConfiguration<?, ?>>(caches);
-    newCaches.put(alias, config);
+    if(newCaches.put(alias, config) != null) {
```
ljacomet: With our immutable builders, we promote sharing them. Should we have a method so you can override a cache with a different configuration?

henri-tremblay: I think the builders reusing can be refactored to prevent adding an existing alias. In the issue will be super rare. So I think adding `withCacheOverride` is overkill

ljacomet: Indeed ... can be added later if requested.

PR #2004, opened 2017-06-01
[master] Close #1996 : use removeCache

### LINKED ISSUE #1996: destroyCache should use the "regular" removeCache method

anthonydahanne:


core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -680,7 +680,7 @@ public void destroyCache(final String alias) throws CachePersistenceException {
     }
 
     try {
-      removeAndCloseWithoutNotice(alias);
```
ljacomet: This method is no longer needed, remove it

core/src/main/java/org/ehcache/core/EhcacheManager.java
```diff
@@ -680,7 +680,7 @@ public void destroyCache(final String alias) throws CachePersistenceException {
...
+      removeCache(alias);
```
ljacomet: You should be using `removeCache(alias, true)`

core/src/main/java/org/ehcache/core/EhcacheManager.java
anthonydahanne: done

ljacomet:
test this please

ljacomet:
Time out in one build - merging anyway

PR #2008, opened 2017-06-05
Spike : move conf before event

anthonydahanne:


anthonydahanne:
here's why (I haven't finished testing it though,I just sent the PR to test if it broke some existing tests) : when you remove a cache from CM, if one of the listeners uses the configuration to trigger a notification, it could send a wrong notification.
Currently, if you remove a cache, your listeners still see the removed cache in the configuration object.
With that change, the listeners will see the updated the configuration (without the cache that was just removed)

henri-tremblay:
The build failed on my flaky test. I will need to revisit it. @anthonydahanne I understand. But I would like a test to showcase it. Also, shouldn't it be done right after the `cacheHolder.retrieve`? And finally, but this is probably a question to @ljacomet. Do we need coherence (so synchronization) between what's in the `caches` and `configuration`?

ljacomet:
This needs to be daggy fixed to 3.2, 3.3 and master

anthonydahanne:
see #2011 #2012 #2013

PR #2024, opened 2017-06-09
Upgrade to new tc-platform version 5.3.0-pre13

mathieucarbou:
Note: NmsAgent is now a permanent entity so it does not need to be created.

### DIFF
--- a/integration-test/build.gradle
+++ b/integration-test/build.gradle
@@ -24,7 +24,7 @@ dependencies {
 }
 
 compileTestJava {
-  options.forkOptions.executable = Jvm.current().javacExecutable
+  options.forkOptions.javaHome = Jvm.current().javaHome
 }
 
 test {
--- a/build.gradle
+++ b/build.gradle
@@ -232,7 +232,7 @@ subprojects {
   tasks.withType(AbstractCompile) {
     options.with {
       fork = true
-      forkOptions.executable = java6.javacExecutable
+      forkOptions.javaHome = java6.javaHome
     }
   }
   tasks.withType(Test) {
--- a/build.gradle
+++ b/build.gradle
@@ -51,7 +51,7 @@ ext {
   ehcache2Version = '2.10.3'
 
   // Clustered
-  terracottaPlatformVersion = '5.3.0-pre12'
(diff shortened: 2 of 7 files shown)


mathieucarbou:
@henri-tremblay @ljacomet : if you can please have a look at the gradle file changes. This is to fix the gradle warning saying that `executable` is deprecated and will be removed in gradle 5.

henri-tremblay:
I will let that to @ljacomet. Because I don't know :-) But can ask around if needed

mathieucarbou:
test this please

mathieucarbou:
build aborted ??? restarting...
test this please

mathieucarbou:
@ljacomet @anthonydahanne : build has been aborted twice by someone or something. So I don't know what to do and I cannot not see anything relevant in the cloudbee logs

ljacomet:
The cloudbees do not reflect the build log, to save space on successful builds. You need to look into the build artifacts and you will find a `gradle.log` that indicates that `org.ehcache.clustered.management.EhcacheManagerToStringTest` ends up never completing and causing the build to be killed after the timeout period has elapsed. Please have a look at that test.

mathieucarbou:
I found which test is failing, i will fix it today. It just takes a lot of time to troubleshoot and fix because of gradle and the fact that it is not easily runnable from within the ide.

integration-test/build.gradle
```diff
@@ -24,7 +24,7 @@ dependencies {
 }
 
 compileTestJava {
-  options.forkOptions.executable = Jvm.current().javacExecutable
+  options.forkOptions.javaHome = Jvm.current().javaHome
```
anthonydahanne: what's the rational for that change ?

chrisdennis:
```
The executable property on ForkOptions has been deprecated and is scheduled to be removed in Gradle 5.0. Please use javaHome instead.
        at build_4hmurz4widgtxhl5sjsel6ff1$_run_closure6$_closure22$_closure30.doCall(/Users/cdennis/src/ehcache3/build.gradle:235)
```

anthonydahanne: okee, thanks

PR #2026, opened 2017-06-09
Fix mockito deprecations (Close #2025)

### LINKED ISSUE #2025: Cleanup Mockito deprecation following version upgrade

henri-tremblay:


107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
```diff
@@ -129,7 +129,7 @@ public void jsr107LoaderGetsRegistered() {
     merger.mergeConfigurations("cache", configuration);
 
     assertThat(factory.called, is(true));
-    verify(cacheLoaderWriterFactory).registerJsr107Loader(eq("cache"), Matchers.<CacheLoaderWriter<Object, Object>>anyObject());
+    verify(cacheLoaderWriterFactory).registerJsr107Loader(eq("cache"), ArgumentMatchers.<CacheLoaderWriter<Object, Object>>any());
```
chrisdennis: Since this is a verification we should probably use `any(CacheLoaderWriter.class)` to ensure the loader writer is not null.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
```diff
@@ -142,7 +142,7 @@ public void jsr107WriterGetsRegistered() {
...
```
chrisdennis: Since this is a verification we should probably use `any(CacheLoaderWriter.class)` to ensure the loader writer is not null.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierActiveEntityTest.java
```diff
@@ -87,9 +87,9 @@
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyString;
-import static org.mockito.Matchers.eq;
+import static org.mockito.ArgumentMatchers.any;
```
chrisdennis: The any usage in `testReplicationMessageAndOriginalServerStoreOpMessageHasSameConcurrency()` should probably be `any(EntityMessage.class)` to ensure we fail if it's null.

impl/src/test/java/org/ehcache/core/events/CacheManagerListenerInteractionsTest.java
```diff
@@ -29,8 +29,8 @@
 import static org.ehcache.config.builders.ResourcePoolsBuilder.newResourcePoolsBuilder;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.eq;
+import static org.mockito.ArgumentMatchers.any;
```
chrisdennis: The two usages of `any` in the class should both `any(Cache.class)` to verify the listener is actually handed a cache.

107/src/test/java/org/ehcache/jsr107/ConfigurationMergerTest.java
henri-tremblay: Right. Since the parameter type is already expecting this type, I will use `isNotNull` which I find more explicit.

Issue #1831, opened 2017-02-06
Ehcache hangs on startup, no error message (Spring webapp on Tomcat) 


darioseidl:
When I started my Webapp in Tomcat, it just hanged at this point:
```
2017-02-06 21:45:26.087 TRACE org.ehcache.core.EhcacheManager                  :     : Initializing
2017-02-06 21:45:26.088 DEBUG org.ehcache.core.internal.service.ServiceLocator :     : Starting 18 Services...
2017-02-06 21:45:26.088 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Starting org.ehcache.impl.internal.spi.event.DefaultCacheEventListenerProvider@3446be60
2017-02-06 21:45:26.088 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Delaying starting org.ehcache.impl.internal.loaderwriter.writebehind.WriteBehindProviderFactory$Provider@e380d82
2017-02-06 21:45:26.088 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Starting org.ehcache.core.DefaultCacheManagerProviderService@5a366562
2017-02-06 21:45:26.088 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Starting org.ehcache.impl.internal.executor.OnDemandExecutionService@3ca06ac4
2017-02-06 21:45:26.088 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Delaying starting org.ehcache.impl.persistence.DefaultDiskResourceService@5404f68
2017-02-06 21:45:26.089 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Delaying starting org.ehcache.impl.internal.store.disk.OffHeapDiskStore$Provider@14438989
2017-02-06 21:45:26.089 TRACE org.ehcache.core.internal.service.ServiceLocator :     : Starting org.ehcache.impl.persistence.DefaultLocalPersistenceService@7fbebc0a
```

I'm using 3.2.0 with a three tier cache (heap, off-heap and disk).

It seems it was because of a `.lock` file in the root dir of my `DefaultPersistenceConfiguration`. After deleting it, the app starts again.

However, it's troubling. I'm not sure if this is to be expected? It was on the first run this machine. An error or info message would be nice.

darioseidl:
Oh, sorry, nevermind. I think I found the cause: there where two instances running (using the same directory). An error message would be nice anyway :grimacing:

chrisdennis:
[...]

PR #2027, opened 2017-06-09
Issue #1831 [master]

chrisdennis:


chrisdennis:
I wanted to make sure everyone was happy with this before I went to the trouble of merging to other branches.

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -89,7 +89,9 @@ private void internalStart() {
         throw new RuntimeException(e);
       }
       try {
-        lock = rw.getChannel().lock();
+        if ((lock = rw.getChannel().tryLock()) == null) {
+          throw new RuntimeException("Couldn't lock rootDir: " + rootDirectory.getAbsolutePath());
```
chrisdennis: Hey @chrisdennis you're double nesting this exception here.  Maybe you should think about doing this differently?

integration-test/src/test/java/org/ehcache/integration/util/JavaExec.java
```diff
@@ -0,0 +1,61 @@
...
+import static java.util.Arrays.asList;
+
+public final class JavaExec {
+
+  private JavaExec() {
+  }
+
+  public static CompletableFuture<Integer> exec(Class klass, String ... args) throws IOException, InterruptedException {
+    String javaHome = System.getProperty("java.home");
+    String javaBin = javaHome + separator + "bin" + separator + "java";
```
chrisdennis: Does this work in Windows?  Maybe we should have some verify it?

impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
chrisdennis: We should probably even be smart about hinting at the difference between inter-process and intra-process locking failures here.

chrisdennis:
Hey @github why can't I review my own PR and block it requesting changes to prevent someone else from merging?

chrisdennis:
@cschanck confirmed that the external process test functions correctly on Windows.

PR #2049, opened 2017-07-12
Fix #2048 Provide default server configuration in kit

### LINKED ISSUE #2048: Adopt new default configuration system from Terracotta server
In an upcoming release, we will be able to provide a default configuration for the server that matches the needs of Ehcache.
See terracotta-oss/terracotta-core#625 for details.

So the ehcache kit will contain a `tc-config.xml` that declares a default off heap resource to be used by clustered caches.

ljacomet:


clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
```diff
@@ -0,0 +1,52 @@
...
+      <!--
+        Note that relative path will be resolved from the location of this configuration file.
+        Also if you end up running more than one server on the same machine for testing purpose, make sure these are different.
+       -->
+      <logs>logs</logs>
+      <!--
+        This port is used by clients to communicate to the server.
+        Its value is actually the default one and could be omitted.
+      -->
+      <tsa-port>9410</tsa-port>
```
chrisdennis: Then why are we not omitting it?

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
```diff
@@ -0,0 +1,52 @@
...
+      <!--
+        This port is used for server to server communication.
+        Its value is actually the default one and is thus omitted.
+      -->
+      <!--<tsa-group-port>9430</tsa-group-port>-->
```
chrisdennis: Why is this one commented but the other one isn't?

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    It also defines a single server, but you can add another one to benefit from high availability.
+  -->
+
+  <plugins>
+    <config>
+      <ohr:offheap-resources>
+        <ohr:resource name="main" unit="MB">512</ohr:resource>
+        <!-- You can define extra offheap resource if you want isolated resource -->
+        <!--<ohr:resource name="other" unit="MB">128</ohr:resource>-->
```
chrisdennis: Do we really think two resources is going to be common enough that we should push the user in the direction?

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
```diff
@@ -0,0 +1,52 @@
...
+      <!--<tsa-group-port>9430</tsa-group-port>-->
+    </server>
+
+    <!--
+      Below a sample server definition that will give HA to the cluster if run on a different host.
+
+      Servers know how to communicate between each others because each version of the config file on each host
+      will list all servers in it.
+    -->
+    <!--<server host="otherhost" name="other-server">-->
```
chrisdennis: Similar to my two offheap resources comment, HA is a serious decision, should we be making it so easy to enable?

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
```diff
@@ -0,0 +1,52 @@
...
+    </config>
+  </plugins>
+
+  <servers>
+    <server host="localhost" name="default-server">
+      <!--
+        Note that relative path will be resolved from the location of this configuration file.
+        Also if you end up running more than one server on the same machine for testing purpose, make sure these are different.
+       -->
+      <logs>logs</logs>
```
chrisdennis: This is horrible, but I don't know what is better.

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
anthonydahanne: there's just 1 defined, the other is commented out

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
anthonydahanne: for documentation purpose ?

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
anthonydahanne: because it's +20 from the other one

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
anthonydahanne: it's not that easy, you still need to replace the hostname and the ports

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
chrisdennis: Right - but by adding a commented resource you are encouraging the user to create a second resource.

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
chrisdennis: What?  Why do we explicitly configure the tsa port to the same value as it's default.  But we let this we be the default.  In other words why do these two sentences conflict:

* Its value is actually the default one and could be omitted.
* Its value is actually the default one and is thus omitted.

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
ljacomet: Right, I will comment it out as well

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
ljacomet: See comment on `tsa-port`

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
ljacomet: Ehcache has indeed client side features to handle that - removed comment and commented XML.

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
ljacomet: Still believe it is a valuable thing to have.

Issue #2053, opened 2017-07-14
Add Cache sizing


Bekreth:
For purposes of debugging and metric collection, it'd be very useful to have a sizing method in the Cache interface.

ljacomet:
Hello,

Thanks for your interest in Ehcache 3.

We request that users start the conversation on the [google group ehcache-users](https://groups.google.com/forum/#!forum/ehcache-users) before opening issues.

***

We currently have no plans on introducing such an API.

First of all, there is not a single cache _size_ to be referred to.
* In heap, cache size could be the number of elements. But is that really interesting when what you are trading is memory for reduced latency? However, it is non trivial to measure real memory size. And while we have the tools to do so, they come at a cost ... even more when the cache grows.
* The moment you move out of heap (off heap, disk or clustered) measuring memory size becomes cheaper but on the opposite maintaining an entry count has its cost.
* Also, do you want to include expired mappings or not?

Forcing all tiers to have either or both of these methods is a serious constraint, not one we are ready to take. For example our storage model for clustered makes it hard to compute both the number of elements the cache holds and the amount of memory really used.

In addition, outside of debugging it has limited use. What matters in a cache are the hit and miss ratio. And while we do not have yet a public statistic API, they will be part of it for sure.

And finally, there is a good reason this is not part of the JCache API itself. And we are aligned with the specification in that context.
Additionally, go back to Ehcache 2.x and check the [`Ehcache.getSize()` javadoc](http://www.ehcache.org/apidocs/2.10.4/net/sf/ehcache/Ehcache.html#getSize()) to understand the kind of issue sizing brings to caching.

I will leave this open for a few days in case you want to comment back but after that will close this and the related pull request #2054.

ljacomet:
As announced, closing

PR #2054, opened 2017-07-14
Add Cache sizing #2053

Bekreth:
This pull request adds the ability to get cache size as per `Add Cache sizing #2053`


terracotta-org:

PR #2064, opened 2017-08-03
Add calculated expiry to ValueHolder (close #2050)

### LINKED ISSUE #2050: Expiry infinite when mixing onHeap and clustered Tiers
Hello,

I configured a cache using onHeap and clustered tiers, with a Custom expiry (I think not relevant) and I saw expiration never occurred in version 3.3.1.
Digging into the code I found the problem and implemented an ugly solution. That's why I don't use pull request.
The problem is that ChainResolver.java checks expiry on line 92, but when returning, the Result doesn't have it.
Later on ClusteredStore.getInternal(K key) it creates a ValueHolder without expiry information, so by default it's -1 -> INFINITE, so element never expires.
There is "hack" for this, calling containsKey expires the element. I think it's beacuse there is a lack here, as containsKey doesn't look first in the onHeap level and it goes directly to the cluster part.

I changed 3 classes to make an unelegant quick fix that worked for me. I share it with you. It basically consist in making ChainResolver return a new Result implementation with expiry, and using it in ClusteredStore. The ugliest part is that I din't want to change Result interface or return type, so I mage this new implementation with expiry. I thought about calling getExpiryForCreation again, but that would break the only-one-call policy on this method.
I attach the 3 source files.

Sorry ni advance if this is not the appropiate way of submitting
[quick-fix.zip](https://github.com/ehcache/ehcache3/files/1145012/quick-fix.zip)

 a bug.
Regards,

Alex

PD: Here I paste part of my cache configuration (it's from a helper to user or not a cluster, suppose there is one):
```
CacheManagerBuilder<? extends CacheManager> cacheManagerBuilder = CacheManagerBuilder.newCacheManagerBuilder();
if (urlCluster != null) {
	cacheManagerBuilder = ((CacheManagerBuilder<CacheManager>) cacheManagerBuilder)
			.with(ClusteringServiceConfigurationBuilder.cluster(URI.create(urlCluster)).autoCreate()
					.resourcePool("resource-pool-" + name, resourcePoolMB, MemoryUnit.MB));
}

//Configuramos la caché
ResourcePoolsBuilder resourcePoolsBuilder = ResourcePoolsBuilder.newResourcePoolsBuilder().heap(maxSize, EntryUnit.ENTRIES);
if (offHeapMB > 0) {
	resourcePoolsBuilder= resourcePoolsBuilder.offheap(offHeapMB, MemoryUnit.MB);
}
if (urlCluster != null) {
	resourcePoolsBuilder= resourcePoolsBuilder.with(ClusteredResourcePoolBuilder.clusteredShared("resource-pool-" + name));
}

cacheManagerBuilder = cacheManagerBuilder.withCache(name,
		CacheConfigurationBuilder.newCacheConfigurationBuilder(claseClave, claseValor, resourcePoolsBuilder)
				// Configuramos la expiración para que dependa de cada elemento
				.withExpiry(new Expiry<K, V>() {
					@Override
					public Duration getExpiryForCreation(K key, V value) {
						return getTimeToLiveDuration(key);
					}

					@Override
					public Duration getExpiryForAccess(K key, ValueSupplier<? extends V> value) {
						return null;
					}

					@Override
					public Duration getExpiryForUpdate(K key, ValueSupplier<? extends V> oldValue, V newValue) {
						return null;
					}
				}));

//Inicializamos el cliente y recuperamos la caché
cacheManager= cacheManagerBuilder.build();
cacheManager.init();
cache = cacheManager.getCache(name, claseClave, claseValor);
```

henri-tremblay:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -182,16 +183,17 @@ public Number call() throws Exception {
...
       getObserver.end(StoreOperationOutcomes.GetOutcome.HIT);
-      return new ClusteredValueHolder<V>(value);
+      return value;
     }
   }
 
-  private V getInternal(K key) throws StoreAccessException, TimeoutException {
-    V value = null;
+  private ValueHolder<V> getInternal(K key) throws StoreAccessException, TimeoutException {
+	ClusteredValueHolder<V> holder = null;
```
ljacomet: Code formatting

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -182,16 +183,17 @@ public Number call() throws Exception {
...
     try {
       Chain chain = storeProxy.get(extractLongKey(key));
       if(!chain.isEmpty()) {
-        ResolvedChain<K, V> resolvedChain = resolver.resolve(chain, key, timeSource.getTimeMillis());
+        long now= timeSource.getTimeMillis();
```
ljacomet: Not used, can't you inline it?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -524,7 +527,7 @@ public void clear() throws StoreAccessException {
     if(mappingFunction instanceof Ehcache.GetAllFunction) {
       Map<K, ValueHolder<V>> map  = new HashMap<K, ValueHolder<V>>();
       for (K key : keys) {
-        V value;
+    	ValueHolder<V> value;
```
ljacomet: Code formatting

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ResolvedChain.java
```diff
@@ -44,12 +44,16 @@
   boolean isCompacted();
 
   /**
-   *
    * @return the number of chain elements that were compacted if there was any compaction
    */
   int getCompactionCount();
 
   /**
+   * @return the unit epoch when the entry should expire
```
ljacomet: That reads really weird.

henri-tremblay: yes... should be unix

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
henri-tremblay: done

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
henri-tremblay: ok

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
henri-tremblay: done

PR #2071, opened 2017-08-08
:bug: Offheap as a middle tier does not seem to be used anymore.

anthonydahanne:
This is a report about some behavior that has changed in ehcache3 very recently.

### Setup : 
* one cache, with 3 tiers : onheap(10 elements max) , offheap (10 MB max), cluster dedicated (50 MB max)
* do 100 puts , each followed by a get on the same key

### What Used to happen : 
* onHeap mapping count : 10, offheap mapping count : 90

### What is happening now : 
* onHeap mapping count : 10, offheap mapping count : **0**

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/Operation.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/Operation.java
@@ -34,10 +34,19 @@ public interface Operation<K, V> {
 
   ByteBuffer encode(Serializer<K> keySerializer, Serializer<V> valueSerializer);
 
+  /**
+   * Time when the operation occurred
+   */
   long timeStamp();
 
+  /**
+   * Does the value installed by this operation have a specific expiry time
+   */
   boolean isExpiryAvailable();
 
+  /**
+   * Time when the operations installed value expires
+   */
   long expirationTime();
 
 }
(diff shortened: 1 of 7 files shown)


anthonydahanne:
I believe the change of behavior was introduced by #2050 (test attached to the PR runs before, stops running after)

chrisdennis:
So this is due to misalignment between the ChainResolver and the ValueHolder as to how "no expiration" is defined.  A hacked up a fix for this locally... but its just one way of resolving the confusion, and it still needs more polish and testing (https://github.com/anthonydahanne/ehcache3/pull/1)

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/operations/Operation.java
```diff
@@ -34,10 +34,19 @@
 
   ByteBuffer encode(Serializer<K> keySerializer, Serializer<V> valueSerializer);
 
+  /**
+   * Time when the operation occured
```
henri-tremblay: occurred

anthonydahanne:
@henri-tremblay @chrisdennis : I've reworded the commits and fixed the typo.
the build passes, let me know if you think it can be merged - or, simply merge it

henri-tremblay:
I've added a unit test and a simpler integration test. I've also revised the code. All fine with me.

anthonydahanne:
meh, you're pretty hardcore in the squashing, I would have kept all commits and added the new tests test on top of it...
also please use "Close #2073" as a prefix to the commit

henri-tremblay:
Ok. Moved the Close at the beginning. One commit was your failing test, one was without a comment. Just saying it was closing the issue. I want fully functional commits. I don't want a commit making the build fail and a commit with code without tests.

anthonydahanne:
that's fine I guess (lost test)
please push your renamed commit and rebase on master, I'll merge

henri-tremblay:
Pushed and rebased

henri-tremblay:
test this please

Issue #2078, opened 2017-08-17
adds missing properties to ehcache configuration


gibsong:
adds missing properties to ehcache configuration.  The configuration was missing the following properties: 
1. server side resource pools
3. defaultResource

gibsong:
PR: https://github.com/ehcache/ehcache3/pull/2079

anthonydahanne:
closed with successful PR

PR #2079, opened 2017-08-17
Issue #2078 adds missing properties to ehcache configuration

gibsong:
In the clustered/common/build.gradle file I added a dependency on core.  However there is probably a better solution to get access to org.ehcache.core.HumanReadable from the clustererd:common module.  Any suggestions?  Perhaps another location for org.ehcache.core.HumanReadable?


clustered/common/build.gradle
```diff
@@ -17,6 +17,7 @@
 apply plugin: EhDeploy
 
 dependencies {
+  compile project(':core')
```
anthonydahanne: I'm wondering why that is necessary...

gibsong: this has been deletetd

Issue #2080, opened 2017-08-18
Fixes findbug in ClusteringServiceConfiguration by using StringBuilder instead of String 


gibsong:

gibsong:
PR: https://github.com/ehcache/ehcache3/pull/2081

anthonydahanne:
closed with successful PR

PR #2081, opened 2017-08-18
Issue #2080 Fixes findbug error in ClusteringServiceConfiguration

gibsong:
Fixes findbug error in ClusteringServiceConfiguration by using StringBuilder instead of String in loop


clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -250,11 +250,10 @@ public String readableString() {
   }
 
   private String readablePoolsString() {
-    String pools = "resourcePools:\n";
+    StringBuffer pools = new StringBuffer("resourcePools:\n");
```
anthonydahanne: please use StringBuilder

PR #2084, opened 2017-08-23
Illegal access is now permitted by default in Java 9

chrisdennis:


build.gradle
```diff
@@ -116,8 +115,8 @@ if (hasProperty('testVM')) {
       } else {
         assert hasProperty('java9Home'): 'Set the poperty \'java9Home\' in your $HOME/.gradle/gradle.properties pointing to a Java 9 installation'
         testJava = Jvm.discovered(new File(java9Home), JavaVersion.VERSION_1_9)
-        break;
       }
+      break;
```
ljacomet: No need for the `;`, be groovy!

chrisdennis: zealot - but done

PR #2099, opened 2017-08-28
Bump to final Terracotta versions

ljacomet:


build.gradle
```diff
@@ -65,13 +65,13 @@ ext {
   ehcache2Version = '2.10.3'
 
   // Clustered
-  terracottaPlatformVersion = '5.3.0-pre35'
+  terracottaPlatformVersion = '5.3.1'
```
chrisdennis: Did we mean for platform and core versions to become misaligned?

ljacomet: We never kept them aligned

PR #2100, opened 2017-08-28
Little typo in "one"

henri-tremblay:


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1295,8 +1295,9 @@ public V apply(final K k) {
           final Map.Entry<? extends K, ? extends V> next = iterator.next();
 
           K computedKey = next.getKey();
-          V computedValue = next.getValue();
           checkKey(computedKey);
+
+          V computedValue = next.getValue();
```
ljacomet: What made you do this change? Careful as your answer may make me say daggy 😉

henri-tremblay: Not a bug. It is just useless to get the value before checking the key.

PR #2102, opened 2017-08-30
Remove client state and move client tracking in management (close #2087)

### LINKED ISSUE #2087: Once the fix to Issue 2041 is merged we can removed the client state map
Once #2041 is fixed the `ClientState` objects and the map they are held in serve no purpose outside of the management code.  This means we can push the map down (or potentially remove it completely) inside the management code.

henri-tremblay:


clustered/server/src/main/java/org/ehcache/clustered/server/management/ClusterTierManagement.java
```diff
@@ -102,37 +106,43 @@ public void init() {
...
   }
 
-
-  public void clientDisconnected(ClientDescriptor clientDescriptor, ClusterTierClientState clientState) {
+  public void clientDisconnected(ClientDescriptor clientDescriptor) {
+    connectedClients.remove(clientDescriptor);
     if (managementRegistry != null) {
       LOGGER.trace("clientDisconnected({})", clientDescriptor);
-      ClusterTierClientStateBinding clientStateBinding = new ClusterTierClientStateBinding(clientDescriptor, clientState);
+      ClusterTierClientStateBinding clientStateBinding = new ClusterTierClientStateBinding(clientDescriptor);
```
mathieucarbou: you can rename the binding classes perhaps ?

henri-tremblay: Yes. I know the name isn't accurate anymore. What do you want? `ClusterTierBinding`?

mathieucarbou: ClientBinding ? DescriptorBinding ?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -653,14 +607,7 @@ public void destroy() {
   }
 
   Set<ClientDescriptor> getConnectedClients() {
-    return unmodifiableSet(connectedClients.keySet());
```
chrisdennis: If we need to know the set of connected clients for non-management reasons then we shouldn't be relying on management to track that set.  I think we're going to need to maintain a set of connected clients here still (unless Voltron can provide a `messageAllClients(...)` API.

henri-tremblay: So you abandon your push up in management idea?

henri-tremblay: In `ClusterTierManagerActiveEntity` it is only used to dump the state. In `ClusterTierActiveEntity` it is used to dump the state.

chrisdennis: We can push the map up... but it looks like we'll need to keep a set in the active entity to handle the invalidation logic.  That set should be temporary since eventually it will get replaced with a specialized statistical set so that we can do smarter/lazier invalidations.

henri-tremblay: Ok. I've moved it back to `ClusterTierActiveEntity`. And removed it from `ClusterTierManagerActiveEntity` since it's not needed. It should be alright now.

PR #2105, opened 2017-09-01
No need to read the message anymore

henri-tremblay:


clustered/server/src/main/java/org/ehcache/clustered/server/ClusterTierManagerActiveEntity.java
```diff
@@ -138,9 +138,7 @@ public EhcacheEntityResponse invokeActive(ActiveInvokeContext invokeContext, Ehc
 
   @Override
   public void handleReconnect(ClientDescriptor clientDescriptor, byte[] extendedReconnectData) {
-    ClusterTierManagerReconnectMessage reconnectMessage = reconnectMessageCodec.decodeReconnectMessage(extendedReconnectData);
```
ljacomet: Given this is no longer needed on the server, we can get rid of the code in org.ehcache.clustered.client.internal.SimpleClusterTierManagerClientEntity#SimpleClusterTierManagerClientEntity that relates to the reconnect data.
However, we can only do that for `3.5`, but could you add a `TODO` so we do not forget?

henri-tremblay: Will do a PR for it with [3.5] in the title

PR #2113, opened 2017-09-06
Return correct result for tracked passive replication messages (close #2112)

### LINKED ISSUE #2112: Dropped value during failover
During failover, the newly elected active server will return `null` for some keys that were set on the passive.

The following message appears in the log:

> org.ehcache.core.spi.store.StoreAccessException: java.lang.RuntimeException: org.ehcache.clustered.client.internal.store.ServerStoreProxyException: Response for getAndAppend operation was invalid : SUCCESS

It doesn't happen all the time and mainly on servers with limited resources.

henri-tremblay:


henri-tremblay:
test this please

PR #2116, opened 2017-09-07
Fix #2114 Resolve task dependency declaration

### LINKED ISSUE #2114: Distribution not correctly updated with :clustered:server binaries
There are two problems actually when I do `gradle :clustered:clustered-dist:distZip`.

First, I get this error:
> Could not get unknown property 'generatePomFileForMavenJavaPublication' for task set of type org.gradle.api.internal.tasks.DefaultTaskContainer

Second (in fact I use a later task (i.e. `:clustered:integration-test:test`) and for some reason I don't get the first error), the `build/distributions/ehcache-clustered-3.4.1-SNAPSHOT-kit.zip` doesn't contain the right `ehcache-server-3.4.1-SNAPSHOT.jar` version. It is an old version (the jar in `clustered/server/build` is correct.

ljacomet:
The `generatePomFileForMavenJavaPublication` task is dynamic and added
late, so anything that wants to depend on it must do so at the last
moment.
Also added a task graph visualisation plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,10 +16,13 @@
 import scripts.*
 import org.gradle.internal.jvm.Jvm
 
-// This adds tasks to auto close or release nexus staging repos
-// see https://github.com/Codearte/gradle-nexus-staging-plugin/
 plugins {
+  // This adds tasks to auto close or release nexus staging repos
+  // see https://github.com/Codearte/gradle-nexus-staging-plugin/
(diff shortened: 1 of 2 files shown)


henri-tremblay:
test this please

henri-tremblay:
The build failed from an issue that is probably linked to message tracker but that is unrelated to the PR. I will look into it. If the other build doesn't pass, someone who has the right can force merge it since it will be helpful to debug other issues.

PR #2117, opened 2017-09-08
[3.5] Remove APPEND since it will probably stay unused forever

henri-tremblay:
In fact, can probably be 3.4 since it removes unused code.


henri-tremblay:
That will mean to have only two `PutOutcome`: `PUT` and `FAILURE`. I am fine with that. However, you will have to discuss with @ljacomet about it.

henri-tremblay:
Ok. So, right now we need PUT, NOOP and FAILURE for 107 statistics. The failing test when not considering NOOP is `CacheMBStatisticsBeanTest.testExpiryOnCreation`.

henri-tremblay:
Might be replaced by #2129

ljacomet:
Closing this as we will go with #2156

PR #2119, opened 2017-09-08
Allow clustered integration tests to run easily in an IDE (close #2118)

### LINKED ISSUE #2118: Can't run Clustered integration tests in IntelliJ
Because of the shading on "dist" and "clustered-dist", some classes are missing from the client classpath.

We should find a way to add them one way or another.

henri-tremblay:


clustered/integration-test/build.gradle
```diff
@@ -37,6 +37,7 @@ dependencies {
   testCompile "com.fasterxml.jackson.core:jackson-databind:2.8.0"
   testRuntime project(':clustered:clustered-dist')
   testRuntime project(':dist')
+  testRuntime "org.terracotta:runnel:$terracottaPlatformVersion"
```
ljacomet: How come this is the only change needed and not around the scope of the `compileOnly` ones?

henri-tremblay: I don't know. I just know that adding only this worked. Unless IntelliJ remembered a bit too much the state with all `compileOnly` removed. Try it for yourself.

ljacomet: Ok did a test, it works if that is declared as `testCompileOnly` which allows it to not leak in the normal gradle build - so please change it.

ljacomet:
When I use IntelliJ and `<ctrl>+<shift>+R` to run a test, the default working directory is the project root and not `<root>/clustered/integration-test` which makes all search of the kit and `gradlew` fail.

So while this is a good idea, it is not fully ready.

henri-tremblay:
Héhé... This is a long standing story with IntelliJ. Everything starts from the root. Which is a problem because Maven (and Gradle from as far as I know) are starting from the module. To solve that, you normally should always set `$MODULE_DIR$` as the default directory for JUnit templates. This aligns them.

I can try to guess where I am but still think this is a sensible default :-)

henri-tremblay:
Ok. I now support both starting directories.

clustered/integration-test/build.gradle
henri-tremblay: Now in `testCompileOnly`

henri-tremblay:
test this please

PR #2121, opened 2017-09-08
Return the real response from StateRepositoryMessages (close #2120)

### LINKED ISSUE #2120: ClassCastException on PUT_IF_ABSENT reponse after failover
PUT_IF_ABSENT should return a `MapValue` object in response. However, the message tracker currently records a `Success` which causes a `ClassCastException` on the client.

Running `BasicClusteredCacheOpsReplicationMultiThreadedTest` and looking at the output if a good way to reproduce it.

henri-tremblay:
The PR should be 3.4 compatible.

Two things I noticed that I would like advice on:
* Comment "Reaching here means a lifecycle or sync operation failed" is talking about lifecycle. But no lifecycle message is handled. Only sync messages
* PassiveReplication messages are logging errors and sync messages are throwing an exception. It feels odd. Is that normal?

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/replication/BasicClusteredCacheOpsReplicationMultiThreadedTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/replication/BasicClusteredCacheOpsReplicationMultiThreadedTest.java
@@ -36,27 +36,32 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameter;
 import org.junit.runners.Parameterized.Parameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terracotta.testing.rules.Cluster;
 
 import java.io.File;
 import java.io.Serializable;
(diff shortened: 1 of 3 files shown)


ljacomet:
@henri-tremblay the comment is most likely a left over when doing the single entity -> multi entity split a while ago.

With regards to exceptions, I believe the logic is that killing a passive while it is syncing makes sense since letting him move forward means having an invalid content. While a transient failure during message passing is ok as we expect a later mutation to still install the proper chain. Remember that we have single clustered tier entity -> segmented off heap maps -> multiple user keys can end up in the same key.

ljacomet:
test this please

henri-tremblay:
Ok. The comment about the lifecycle is now removed

ljacomet:
Again that duplicate test failure

henri-tremblay:
timeout increased to see how it goes

henri-tremblay:
Can `BasicClusteredCacheOpsReplicationMultiThreadedTest.testClear` be longer than 3 minutes? I see nothing in the log telling that something went wrong.

henri-tremblay:
Seeing the build history, it smells like something is wrong.

Issue #2122, opened 2017-09-11
Update revised message tracker from terracotta platform


ljacomet:
In terracotta-oss/terracotta-platform#389 support for segmenting the message tracker synchronisation has been added.

Integration in Ehcache needs to adapt to this while preserving compatibility with a `3.4.0` server.

PR #2124, opened 2017-09-12
Revert "Issue #2122 Backward compatibility with 3.4.0"

ljacomet:
This reverts commit 89995d8ae27681fd021214869b23d110d7066702.
On the 3.5 line, the backward compatible handling is no longer needed.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheMessageTrackerMessage.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheMessageTrackerMessage.java
@@ -31,8 +31,6 @@ import static java.util.stream.Collectors.toMap;
  */
 public class EhcacheMessageTrackerMessage extends EhcacheSyncMessage {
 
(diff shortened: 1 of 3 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheMessageTrackerMessage.java
```diff
@@ -31,8 +31,6 @@
  */
 public class EhcacheMessageTrackerMessage extends EhcacheSyncMessage {
 
-  public static final int UNKNOWN_SEGMENT = -1;
-
   private final int segmentId;
```
albinsuresh: Even this `segmentId` can be removed from the message and be computed on the passive from the concurrency key in `InvokeContext`.

ljacomet: Let's do that in a different PR since this is simply a revert of a given commit if that's ok with you.

PR #2125, opened 2017-09-12
Remove client and transaction id from messages since they are now in the InvokeContext

henri-tremblay:
Note: Many message have now no payload. Just a type.


ljacomet:
test this please

PR #2127, opened 2017-09-13
Add system for client - server backward compatibility testing

ljacomet:
The idea behind this PR is to have an easy way of partially testing backwards compatibility.

Unless others vote for it, I do not plan to add this task to PR builds, as it would add a few minutes for each. I would rather have a weekly build that runs it.

The main downside is that it requires a bit of maintenance. When switching to a new minor version, the added versions need to be disabled but I could not find a way to have the build pass silently then. It always fails at version resolution.
And of course, during the release of the new minor `3.X.0`, versions have to be set to the reference versions of that release.

Interested in comments.


ljacomet:
Needs more work, will be picked up later

PR #2129, opened 2017-09-19
Migrate to append for puts

chrisdennis:
Got bored last night... hacked this up.


ljacomet:
Looks good to me too.
@anthonydahanne @mathieucarbou what's the consequence of removing that stat outcome from a monitoring perspective?

ljacomet:
Had a chat with @chrisdennis and this is not yet fully ready.
* Invokes can be reworked to really ignore the response for `APPEND`
* With the change in `put` the chain on the server can really grow. We may need a server initiated mechanism to ask a client to resolve and compact said chains.

PR #2108, opened 2017-09-01
Cleaning up unused MNM events

mathieucarbou:
@henri-tremblay @anthonydahanne @ljacomet : this is all the stuff that I think we do not need anymore since we can just stick with the raw voltron events to do the same thing.

__IMPORTANT__: We need to check in other projects first before merging that! I've just created this PR to show the work that has to be done

PR #2130, opened 2017-09-19
Cleaning up unused MNM events

mathieucarbou:
This is like #2108 but daggy-fix based on @henri-tremblay 's commit (https://github.com/ehcache/ehcache3/pull/2102/commits) and merged onto master.


mathieucarbou:
replaced by #2132

PR #2142, opened 2017-10-03
Upgrade Mockito and AssertJ

henri-tremblay:


ljacomet:
Damn looks like we have another unstable test ...

test this please

PR #2149, opened 2017-10-05
Upgrade to new tc-platform & tc-core version

mathieucarbou:


mathieucarbou:
@anthonydahanne @ljacomet : I am not sure... I need to do a daggy fix right ?

PR #2156, opened 2017-10-06
Switch to append for Cache.put(...)

chrisdennis:
Two remaining issues with this.  One that I need to look in to, and one that should lead to a Voltron issue.

1. I'm not convinced that this will work against an existing 3.4 client because if the chain managed to get too long (due to hash collision) then the server can send a message to a client that will not be understood (a resolve request).  I need to dig in on what that will do to a client when it cannot be unmarshalled.
2. Currently the put append message is not just waiting for received since it seems you cannot timeout a wait for receive message... instead it is having to wait for complete, since then the `Future` returns without blocking and we can timeout on the `Future.get()`.

Please review with these issues in mind - i.e. likely don't merge.


ljacomet:
For point 2 the API is not available since we upgraded to the latest versions. See `org.terracotta.entity.InvocationBuilder#invokeWithTimeout`

ljacomet:
Closing, work is now in #2165 instead

PR #2157, opened 2017-10-09
Fix #2068 Add support for java.time.Duration

### LINKED ISSUE #2068: Support java.time.Duration
Ehcache has its own [`org.ehcache.expiry.Duration`](https://github.com/ehcache/ehcache3/blob/master/api/src/main/java/org/ehcache/expiry/Duration.java)

It would be nice to support [`java.time.Duration`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html). This could be supported in addition to `org.ehcache.expiry.Duration` for some period of time. Likely it would make sense to deprecate `org.ehcache.expiry.Duration` as there's not a strong reason to have a second non-standard implementation

ljacomet:
* Add static method for converting a java.time.Duration into an Ehcache
Duration
* Add methods on the Expirations utility taking java.time.Duration
parameters

### DIFF
--- a/api/src/main/java/org/ehcache/expiry/Duration.java
+++ b/api/src/main/java/org/ehcache/expiry/Duration.java
@@ -48,6 +48,40 @@ public final class Duration {
     return new Duration(length, timeUnit);
   }
 
+  public static Duration of(java.time.Duration duration) {
+    if (duration == null) {
+      return null;
+    }
+    if (duration.isNegative()) {
+      throw new IllegalArgumentException("Ehcache duration cannot be negative and so does not accept negative java.time.Duration: " + duration);
+    }
+    if (duration.isZero()) {
+      return Duration.ZERO;
+    } else {
+      long nanos = duration.getNano();
+      if (nanos == 0) {
+        return of(duration.getSeconds(), TimeUnit.SECONDS);
+      }
+      long seconds = duration.getSeconds();
+      long secondsInNanos = TimeUnit.SECONDS.toNanos(seconds);
+      if (secondsInNanos != Long.MAX_VALUE && Long.MAX_VALUE - secondsInNanos > nanos) {
+        return of(duration.toNanos(), TimeUnit.NANOSECONDS);
+      } else {
+        long secondsInMicros = TimeUnit.SECONDS.toMicros(seconds);
+        if (secondsInMicros != Long.MAX_VALUE && Long.MAX_VALUE - secondsInMicros > nanos / 1_000) {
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/ehcache/expiry/Duration.java
```diff
@@ -48,6 +48,40 @@ public static Duration of(long length, TimeUnit timeUnit) {
...
+      throw new IllegalArgumentException("Ehcache duration cannot be negative and so does not accept negative java.time.Duration: " + duration);
+    }
+    if (duration.isZero()) {
+      return Duration.ZERO;
+    } else {
+      long nanos = duration.getNano();
+      if (nanos == 0) {
+        return of(duration.getSeconds(), TimeUnit.SECONDS);
+      }
+      long seconds = duration.getSeconds();
```
henri-tremblay: I think we should consider that `Duration.of(Long.MAX_VALUE, ChronoUnit.NANOS)` is the infinite and give a constant to represent it.

api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -45,6 +45,19 @@
   }
 
   /**
+   * Get a time-to-live (TTL) {@link Expiry} instance for the given {@link java.time.Duration}.
+   *
+   * @param timeToLive the TTL duration
+   * @return a TTL expiry
+   */
+  public static Expiry<Object, Object> timeToLiveExpiration(java.time.Duration timeToLive) {
+    if (timeToLive == null) {
```
henri-tremblay: You can use `Objects.requireNonNull` here

api/src/main/java/org/ehcache/expiry/Expirations.java
```diff
@@ -45,6 +45,19 @@
   }
 
   /**
+   * Get a time-to-live (TTL) {@link Expiry} instance for the given {@link java.time.Duration}.
```
henri-tremblay: I think we should deprecate the old methods use our duration object. And in fact, would prefer that we convert to `java.time.Duration` instead of the other was around.

benmccann: +1

api/src/main/java/org/ehcache/expiry/Duration.java
benmccann: Yes, the `java.time.Duration` javadocs suggests essentially that in the saying that `Duration.of(Long.MAX_VALUE, ChronoUnit.NANOS)` is greater than the current estimated age of the universe.

ljacomet:
This will not get merged, going with the approach in #2166

PR #2166, opened 2017-10-12
Issue #2068 Ehcache expiry based on java.time.Duration

ljacomet:
This deprecates the existing expiry related APIs and replaces them with
APIs based on java.time.Duration

For compatibility reasons, the builders still accept the old API and so
does the XML.

This is an alternate proposal to #2157 - Expecting comments on which is the right way


107/src/main/java/org/ehcache/jsr107/ExpiryPolicyToEhcacheExpiry.java
```diff
@@ -55,28 +53,22 @@
...
         return null;
       }
-      if (duration.isEternal()) {
-        return org.ehcache.expiry.Duration.INFINITE;
-      }
-      return new org.ehcache.expiry.Duration(duration.getDurationAmount(), duration.getTimeUnit());
+      return convertDuration(duration);
     } catch (Throwable t) {
-      return org.ehcache.expiry.Duration.ZERO;
+      return java.time.Duration.ZERO;
```
henri-tremblay: Not related to your changes but isn't super weird to swallow the exception whatever it is?

api/src/main/java/org/ehcache/expiry/ExpiryPolicies.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+
+      final BaseExpiryPolicy that = (BaseExpiryPolicy) o;
+
+      if (access != null ? !access.equals(that.access) : that.access != null) return false;
```
henri-tremblay: Objects.equals

api/src/main/java/org/ehcache/expiry/ExpiryPolicies.java
```diff
@@ -0,0 +1,229 @@
...
+      if (create != null ? !create.equals(that.create) : that.create != null) return false;
+      if (update != null ? !update.equals(that.update) : that.update != null) return false;
+
+      return true;
+    }
+
+    @Override
+    public int hashCode() {
+      int result = create != null ? create.hashCode() : 0;
```
henri-tremblay: Objects.hash

107/src/main/java/org/ehcache/jsr107/ExpiryPolicyToEhcacheExpiry.java
ljacomet: It is part of the contract, this behaviour is described in the `ExpiryPolicy` javadoc

ljacomet:
@henri-tremblay That's a good point! Less "code" in the API module, I like it. but pushing it out to builders which are in `impl` will be annoying for `core` usage. Let me think about it a bit

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -0,0 +1,209 @@
...
+    return new TimeToLiveExpiryPolicy(timeToLive);
+  }
+
+  /**
+   * Get a time-to-idle (TTI) {@link ExpiryPolicy} instance for the given {@link Duration}.
+   *
+   * @param timeToIdle the TTI duration
+   * @return a TTI expiry
+   */
+  public static ExpiryPolicy<Object, Object> timeToIdleExpiration(Duration timeToIdle) {
```
henri-tremblay: Shouldn't these return `ExpiryPolicy<K, V>` (inferred by the context)

api/src/main/java/org/ehcache/expiry/ExpiryPolicy.java
```diff
@@ -0,0 +1,115 @@
...
+ * @param <K> the key type for the cache
+ * @param <V> the value type for the cache
+ *
+ */
+public interface ExpiryPolicy<K, V> {
+
+  /**
+   * A {@link Duration duration} that represents an infinite time.
+   */
+  Duration INFINITE = Duration.ofSeconds(Long.MAX_VALUE);
```
henri-tremblay: To be coherent with cluster timeouts (and be `duration.toNanos()` safe), this should be `Duration.ofNanos(Long.MAX_VALUE)`

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
ljacomet: We always consume an `ExpiryPolicy<? super K, ? super V>` so I do not thing the type inference here would buy us much. And since the values are not used, it is correct to say it is an `ExpiryPolicy<Object, Object>`

ljacomet:
@henri-tremblay Regarding the `ExpiryPolicyBuilder` my last update gets rid of the generic type altogether by having it extend `ExpiryPolicy<Object, Object>`

api/src/main/java/org/ehcache/expiry/ExpiryPolicy.java
ljacomet: I understand the request ... not sure I like going that way though. I would feel better keeping at least `Duration.ofMillis(Long.MAX_VALUE)`.
@chrisdennis wdyt?

henri-tremblay: I much prefer Duration.ofNanos to never crash on infinite. As mentioned, we might refactor that differently later. But right now, it's the safer side. You could probably even do what I did for timeouts and move expiration bigger than that to this value.

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
henri-tremblay: Fine with me.

api/src/main/java/org/ehcache/expiry/ExpiryPolicy.java
ljacomet: So changing that is more work than expected ... as I hardcoded `INFINITE` to be `Duration.getSeconds() == Long.MAX_VALUE` in a number of places as a consequence of `or.ehcache.expiry.Duration.isInfinite()` going away. Fixing that mistake so that we have only a limited number of places where we "decide" what `INFINITE`.

PR #2172, opened 2017-10-25
Migrate to Java 9 CHM

henri-tremblay:
Now using Java 9 CHM migrated back to Java 8. We need to benchmark to see what we should use.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/concurrent/ConcurrentHashMap.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/concurrent/ConcurrentHashMap.java
@@ -22,10 +22,12 @@
 
 package org.ehcache.impl.internal.concurrent;
 
+import sun.misc.Unsafe;
+
+import org.ehcache.config.EvictionAdvisor;
+
 import java.io.ObjectStreamField;
 import java.io.Serializable;
-import static java.lang.Integer.rotateLeft;
-
 import java.lang.reflect.ParameterizedType;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/ehcache/impl/internal/concurrent/ConcurrentHashMap.java
```diff
@@ -1,31 +1,45 @@
 /*
- * Copyright Terracotta, Inc.
+ * ORACLE PROPRIETARY/CONFIDENTIAL. Use is subject to license terms.
```
chrisdennis: It would be much safer to pick this up from the 166 repo rather than the OpenJDK one, then the only header will be:
```
/*
 * Written by Doug Lea with assistance from members of JCP JSR-166
 * Expert Group and released to the public domain, as explained at
 * http://creativecommons.org/publicdomain/zero/1.0/
 */
```
Our copyright statement is also still accurate since there are TC changes in the file

henri-tremblay: They are probably the same apart from the header. Do you know how to know which jsr166 version went into the latest Java 9? Current HEAD is complicated to integrate.

henri-tremblay: All good now. I've picked HEAD. Was fine after all.

PR #2173, opened 2017-11-01
Design document on resilience

henri-tremblay:
This is the design on resilience. It contains the current implementation and some thoughts left for later.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -22,7 +22,7 @@ In the second launch the live reload.
 
   ./gradlew -t :docs:livereload
 
-The local port will be printed on the console. The usual URL is link:http:localhost:35729/user[http:localhost:35729/user].
+The local port will be printed on the console. The usual URL is link:http://localhost:35729[http://localhost:35729].
 
 == Add a section
 
--- a/docs/src/docs/asciidoc/developer/design.resilience.adoc
+++ b/docs/src/docs/asciidoc/developer/design.resilience.adoc
@@ -87,7 +87,7 @@ Three different types of timeouts have been defined.
 * Write: Operations modifying data on the server
 * Connection: Establishing the connection with the server
 
-By default, all are set to 5 seconds.
+By default, all operation timeouts are set to 5 seconds. The default connection timeout is 150 seconds.
 
 Note that there is no lifecyle timeout. It doesn't feel useful. For a cache creation, not being able to create the cache
 means you can't do anything anyway. For destruction, well... we rarely destroy. The leasing will take care of the timeout.
--- a/docs/src/docs/asciidoc/developer/design.resilience.adoc
+++ b/docs/src/docs/asciidoc/developer/design.resilience.adoc
@@ -1,0 +1,182 @@
+= Resilience
+
+:toc:
+
+== Configuration
+
+This is a summary of all the possible configurations. It's useful to you if you already have read the rest of the document
+and want to see the configuration summarized in one place. If you haven't read the rest of the document yet, you won't
+understand a thing.
+
+Read timeout::
+  How long to wait on a read to the cluster. Applies to `get`, `getAll` and `containsKey`
+Write timeout::
+  How long to wait on a write operation on the cluster. Applies to everything else
+Connection timeout::
+  How long to wait when first establishing the connection to the server
+Lease time::
+  Duration of a lease between a client and a server as given by the server
+Resilience strategy::
+  Interface implementation telling how to answer when a given cache method fails due to a failing tier
+Loader-writer resilience strategy::
+  A resilience strategy that is aware of the loader-writer and so can take advantage of it in case of failure
+Ledger queue length::
+  How many mutative elements we keep on the client when waiting for the server to come back
+Caching tier behavior in case of failure::
+  How should the caching tier behave when the authoritative tier is gone (clear, server when possible, process updates locally)
+Caching tier behavior in case of lease expiration and reconnect::
+  How should the caching tier behave when the lease expires or when reconnecting to the authoritative tier (clear, reconcile, keep)
+
+== What can go wrong
+
+* Server fails to get an answer erratically because of network or something
+* Server fails to send messages to the client erratically
+* Server goes down
+* Server failover
+* Loader-writer backend fails
+
+== Strong vs Eventual
+
+Strong means: If the client asks for a value that was updated by some other server, it will always get the
+latest value.
+
+In particular, it means that as soon as an update from the server might be missed, we need to clear
+the caching tier. It also means that if we fail to set a value to the authoritative tier, we can't use
+the cache anymore. On this key at least.
+
+Eventual means the correct value will eventually be there. So if we lose contact with the server, it is possible to keep
+using a possibly stale cache entry. This means that a client with eventual consistency shouldn't need to clear its cache
+at all. However, it means two things:
+
+* We need to be able to update the caching tier in that case
+* We need to get updates from the server when the situation resumes
+
+=== Being too eventual
+
+We can argue that if a client stays disconnected for too long, it means the caching tier content is now too obsolete. There
+are two ways to handle that:
+
+* Expiration should take care of it. So entries should have an expiration set to their expected time validity.
+* We configure "too long" which will clear the cache
+
+== Clustering
+
+When using a distributed cache, your Ehcache client will connect to a remote Terracotta server (or multiple in case
+of stripping) which act as the authoritative tier.
+
+This server data can be replicated to other servers called mirrors. How your client will handle being cut from
+the cluster is defined in `ClusteringServiceConfiguration`.
+
+=== Lease
+
+A client has a lease with the server. And there's a heartbeat making sure nobody died. So if the client gets no heartbeat
+from the server, it will decide that it is now on its own. When the heartbeat comes back, it will resume operation.
+
+From the other side, it also means that the server won't accept any write operation without getting an acknowledgment from
+the client. So, until the lease expires, the client is guaranteed to stay consistent with the server, even if the server
+is lost.
+
+=== Timeouts
+
+If a client doesn't receive an answer in the configured time, it will timeout. This can happen for different reasons
+including a network cut, a long full GC or a server down.
+
+Three different types of timeouts have been defined.
+
+* Read: Operations reading data from the server
+* Write: Operations modifying data on the server
+* Connection: Establishing the connection with the server
+
+By default, all are set to 5 seconds.
+
+Note that there is no lifecyle timeout. It doesn't feel useful. For a cache creation, not being able to create the cache
+means you can't do anything anyway. For destruction, well... we rarely destroy. The leasing will take care of the timeout.
+
+Internally, while waiting for the timeout, the client might decide to retry, sleep between retries or do whatever it feels
+useful to provide an answer.
+
+=== In case of failure
+
+A store can fail in two ways.
+
+First, it failed by itself. In that case, it will launch a `StoreAccessException` that will be caught by the resilience strategy.
+
+Second, something underneath failed. For instance, a call to a loader-writer failed. In that case, the original exception
+is wrapped in a `StorePassThroughException`. This will let it pass right thought the store and be unwrapped and thrown
+by the cache to the caller.
+
+=== After a failure
+
+We will now consider a server with no mirror.
+
+A failure can just be a hiccup. If that's the case, the call will timeout and resilience strategy will take care of the rest.
+
+If we lost connection, the client won't be able to renew the lease. It will go in resilience mode. It means
+
+* clearing the caching tier
+* answering everything with the resilience strategy
+* try to renew the lease in background
+
+**Unimplemented yet:**
+Instead of clearing the caching tiers, more flexible strategies can be implemented. Here is a beginning of discussion.
+
+It is important to notice that the client won't fallback to the resilience strategy when the caching tier answers.
+It also means that the caching tier might not receive updates from the server and become out of sync.
+
+This is independent of the consistency configured and configurable. You can pick the following strategies:
+
+* Rely on tier forever, so even if the server is officially lost
+* Rely on tier on hiccups. The server will keep the caching tier until it declares the server lost for good
+* Don't rely on tier.
+
+When the lease is renewed, a reconciliation could occur with the server to sync the caching tier.
+
+=== Active and mirror setup
+
+**Note:** Actual behavior to be tested
+
+It works the same as with a single server. Except that during a failover, the client will behaves. like if the underlying
+server is having hiccups. Or is down if the failover takes too long (longer than the lease).
+
+The new active server will notify the client when ready.
+
+=== Reconnection
+
+When a server goes down and back again. On the same URL. The client will silently reconnect to it.
+
+== Loader writer
+
+A failing loader-writer throws exception to caller. A given implementation could implement it's own resilience strategy.
+
+Also, a resilience strategy can use the loader-write to answer. This is what the default Ehcache resilience strategy does
+in presence of a loader-writer.
+
+=== Write behind
+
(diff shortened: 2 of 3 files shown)


docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+We can argue that if a client stays disconnected for too long, it means the caching tier content is now too obsolete. There
+are two ways to handle that
+
+* Expiration should take care of it. So entries should have an expiration set to their expected time validity.
+* We configure "too long" which will clear the cache
+
+== Clustering
+
+When using a distributed cache, your Ehcache client will connect to a remote Terracotta server (or multiple in case
+of stripping) which act as the authoritative tier.
```
chrisdennis: You mean 'striping', 'stripping' is something entirely different.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+When a client timeouts, it might retry right away. That are two parameters for that:
+
+* Number of retries (default is 0)
+* Sleep time between retries (default is 0)
+
+Note that for each retry, the timeout will be enforced. So, far example, 3 retries with a timeout of 5 seconds and a sleep
+of 1 second will mean 20 seconds before returning.
+
+*Note to myself:* Should the retry be on the resilience strategy?
```
chrisdennis: I still don't understand why the user should care about retries?  If the user sets a read timeout of 5 seconds he would expect us to do whatever is prudent to honor the read within that time.  If that means retries then that makes sense, but that would then be an implementation concern.  How would someone decide between:

* 20 second timeout, no retry.
* 5 second timeout, 3 retries.

To me which of those is right is really only known to the implementation, and even then it's only really known when faced with the failure since different failures might be best tackled with different retry approaches.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+If we failed to get an answer, the cache [#Resilience Strategy] will kick in and return whatever it was configured to return.
+
+Resilience strategy is not used for lifecycle events. So in this case you will get a `StoreAccessException`.
+
+=== After a failure
+
+We will now consider a server with no mirror.
+
+The first thing that can happen is a server that has problem answering correctly all the time but that is still there.
+In this case, after 10 timeouts (configurable), the client will consider the server down to prevent waiting on requests.
```
chrisdennis: Isn't this a matter of tuning timeouts correctly, if the cache sometimes answers correctly why would I want to give up on it completely.  If the connection is completely down then I understand not waiting for timeouts and going straight to resilience strategy, but to me making that decision only makes sense when you *know* that the timeout will happen, not when you think there is some significant chance it could happen.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+Meanwhile, it will rely on the resilience strategy to answer.
+
+*Note:* M&M will trigger an alarm if a server is acting erratically.
+
+It is important to notice that the client won't fallback to the resilience strategy when the caching tier answers.
+It also means that the caching tier might not receive updates from the server and become out of sync.
+
+This is independent of the consistency configured and configurable. You can pick the following strategies:
+
+* Rely on tier forever, so even if the server is officially lost
```
chrisdennis: This would just be eventual consistency with no expiration right?  No need for a distinct configuration for it?

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+* Rely on tier on hiccups. The server will keep the caching tier until it declares the server lost for good
```
chrisdennis: This would be strong consistency behavior, the client will dump the cache when it believes it's lost membership in the cluster (so storng puts could be happening without it's consent).

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+* Don't rely on tier.
```
chrisdennis: Don't rely on tier?  Isn't this just strong consistency with a really short fuse on kicking clients?

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+It works the same as with a single server. Except that during a failover, the client will behaves like if the underlying
+server is having hiccups. Or is down if the failover takes too long.
+
+The new active server will notify the client when ready.
+
+=== Reconnection
+
+When a server goes down and back again. On the same URL. The client will silently reconnect to it. The ledger should allow
+to invalidate entries on the caching tier.
```
chrisdennis: Invalidation push should be able to go in both directions right?  I imagine having a resilience strategy that did this would make sense (might even want to make it the default).

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+Cache loader writers have a dedicated resilience strategy. So in case of failure, the resilience strategy will operate and
+the result will be picked up by the Cache resilience strategy.
+
+When using a loader writer, we tend to consider the following:
+
+* We expect the cache never return null. Because the loader will load any missing data. But to enforce that
+in case of server failure, it means the client should wait until the backend is back.
+* We expect that everything cached is saved. So in general an exception should be thrown if we fail to do so.
+
+The default resilience strategy is the following.
```
chrisdennis: I figure we should be able to make the behavior with loader-writers equivalent to the normal behavior of the cache when used in a sensibly coded cache-aside setup.  Is that what you are aiming for here?

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+=== Write behind
+
+The resilience strategy is applied for write behind as well. So make sure it manages errors
+that are asynchronous.
+
+== Interruptions
+
+When waiting on a call to a store, an interruption should allow to get out. It will then probably rely on the resilience
+strategy (I'm not sure about that) or throw an exception right away.
```
chrisdennis: I don't think it can throw... I think a resilience strategy call might make sense here - but I'd be concerned with how the resultant system might behave in various user applications.  Some people are very ill disciplined about interrupting threads.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+== Resilience Strategy
+
+The default resilience strategy (`RobustResilienceStrategy`) will
+
+* Return null on a read
+* Do nothing on a void mutation
```
chrisdennis: Wouldn't it make sense to "remember the key" and attempt to invalidate the cache when comms is restablished?  Also I assume we plan to invalidate our own local tiers?

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+* Throw an exception if the result depends on the result of the mutation
```
chrisdennis: I think this makes sense... do we want to allow for configuring distinct timeouts for the 'void' and 'non-void' mutative methods then?

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+|getAndPut      |Return null, nothing is put
+|getAndRemove   |Return null, nothing is removed
+|getAndReplace  |Return null, nothing is replaced
+|invoke         |Do nothing and return null
+|invokeAll      |Do nothing and return nulls
+|iterator       |Empty iterator
+|loadAll        |Do nothing
+|put            |Do nothing
+|putAll         |Do nothing
+|putIfAbsent    |Do nothing, return null
```
chrisdennis: This seems weird... the null return is at odds with the "do nothing", since a null return implies the method did something.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+* Server goes down
+* Server failover
+* Loader-writer backend fails
+
+== Strong vs Eventual
+
+Strong means: If the client ask for a value that was updated by some other server, it will always get the
+latest value.
+
+In particular, it means that as soon as an update from the server might be missed, we need to clear
```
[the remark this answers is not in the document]
henri-tremblay: But what about other clients?

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: yep. Typo. Will fix

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: Ok. Agreed. We can expose it later anyway if it becomes useful for any reason.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: Yes. If it's erratic, it's better to try. If it's too erratic or down, it's better to try in background and answer locally to prevent the client from waiting and go to degraded mode.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: I thought it would be useful to separate the two. To have a STRONG but in case of failure I can drop a bit of consistency.  In some kind of PACELC spirit.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: Still PACELC. I want to allow to drop STRONG in case of failures.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: Yes. But it could be used for EVENTUAL as well.

docs/src/docs/asciidoc/user/resilience.adoc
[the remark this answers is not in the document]
chrisdennis: I'm just saying this is really about coordinating knowledge of cluster membership such that the client believes it's left the cluster before the server kicks it.  The effect of leaving on the client is that it knows it's local cache is out of sync (so will/can stop using it), and the effect of kicking on the server is to allow blocked puts to continue.  In that sense the important timing here is not how long the client wants to remain a member but rather either:

1. How long the server is prepared to block strong puts.
2. How eventual the server is prepared to allow the cache to be.

docs/src/docs/asciidoc/user/resilience.adoc
chrisdennis: I guess this is an interesting scenario. Are there users who appreciate strong but don't *need* it?  I'd always figured you either needed strong or didn't.  If you don't need it, why wouldn't you drop to something else (currently that would be to eventual I guess).

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: I'm fine with the resilience strategy. For me, if you interrupt, it should interrupt. We can log when it happens to have a quick diagnosis.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: Hum... that is an excellent question... Right now I wasn't but it's true that a readable mutative call could be wanted as more short-fused than a plain put.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: It's the cache not caching. So `putIfAbsent` will consider the cache to be empty but won't cache anything. You can see it like a cache with entries expiring at insertion. Or a cache of size 0 with everything evicted immediately.

docs/src/docs/asciidoc/user/resilience.adoc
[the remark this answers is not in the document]
henri-tremblay: Yes. There is a how strong is strong and how eventual is eventual in term of time concepts here.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: It seems it does exist yes. PNUTS is an example. It stays strong but prefer to still answer is case of failure.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: Yes. The client could also buffer its pending updates. I'm not sure I would put it in the resilience strategy. For now I have envisioned the resilience strategy to be an answering machine to the client in case of failure. To keep it simple. It think the reconciliation process should happen all the time and the algorithm can the depending on STRONG, EVENTUAL and "how eventual".

docs/src/docs/asciidoc/user/resilience.adoc
chrisdennis: Tying a cache topology choice to a cache consistency choice seems strange.  Why can't a cache-through/cache-behind user have the same rules as a cache-aside user?

docs/src/docs/asciidoc/user/resilience.adoc
chrisdennis: I guess that makes sense.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,189 @@
...
+the caching tier. It also means that if we fail to set a value to the authoritative tier, we can't use
+the cache anymore. On this key at least.
+
+Eventual means the correct value will eventually be there. So if we loose contact with the server, it is possible to keep
```
cschanck: lose not loose :-)

henri-tremblay: Done :-) And a bunch of other grammar errors as well

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: What is the difference? Returning null will cause the XA transaction to commit. If the user wants a rollback in case of cache failure, he better use a resilience strategy that matches his needs.

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: They can. Where am I saying they can't? But you are right, the text is highly focused on a single loader writer usage. I need to had some subtlety.

henri-tremblay: Ok. My loader-writer part was actually more complete than what I remember. So I'm not sure what puzzles you. But I think it's my bad writing. There are two resilience strategies:
* CacheResilienceStrategy
* LoaderWriteResilienceStrategy

Here, I am talking about the later.

So, for instance, if a read from the SoR fails, the `LoaderWriteResilienceStrategy` will kick-in and return a result. Which will be received by the `CacheResilienceStrategy` that will return the final result.

My goal is to simplify the resilience strategy by actually having too. And if you are no using a loader-writer. On won't need to know about its resilience strategy. This removes a lot of noise from cache resilience strategy.

chrisdennis:
So what we're looking at here from a configuration point of view is:

Timeouts:
* Read Operation Timeout (default=?)
* Write Operation Timeout (default=?)
* ??? Non-Void Write Operation Timeout (default=?)

Resiliency Strategy:
 * default
 * no-op
 * custom user implementation

What about leasing timeouts... do we derive them from the operation timeouts, or do we need independent controls for them?

PR #2174, opened 2017-11-01
Add write and connection timeouts to the API

henri-tremblay:
Some thoughts:
- Should we add convenience methods taking the builder in parameter instead of `Timeouts`?
- We are now using ChronoUnit and TimeUnit in our API. I think some will disappear because of #2166
- I've tried to keep deprecated methods to keep compatibility. Do we need to in this case?
- I'm not a big fan of mutative as a naming. We use write timeout in general. It's not as accurate but more classic and prettier. WDYT?


clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -54,31 +55,38 @@
...
+  /**
+   * Creates a {@code ClusteringServiceConfiguration} from the properties provided.
+   *
+   * @param clusterUri the non-{@code null} URI identifying the cluster server
+   * @param operationTimeouts the {@link Timeouts} specifying the time limit for clustered cache operations
    *
    * @throws NullPointerException if {@code clusterUri} is {@code null}
    * @throws IllegalArgumentException if {@code clusterUri} is not URI valid for cluster operations
    */
-  public ClusteringServiceConfiguration(final URI clusterUri, final TimeoutDuration readOperationTimeout) {
```
ljacomet: This is API breakage. We should keep these versions around and add new ones. There is not even a naming conflict so that should be easy.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -219,13 +220,25 @@ public ServerSideConfiguration getServerConfiguration() {
...
+
   /**
    * The timeout for cache read operations.
    *
-   * @return the cache read operation timeout; may be {@code null} indicating the default timeout is used
+   * @return the cache read operation timeout
+   *
+   * @deprecated Use {@link #getOperationTimeouts()}
    */
-  public TimeoutDuration getReadOperationTimeout() {
```
ljacomet: Again API breakage that we could easily prevent.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -90,18 +98,18 @@ public ServerSideConfigurationBuilder expecting() {
    *
    * @throws NullPointerException if {@code unit} is {@code null}
    * @throws IllegalArgumentException if {@code amount} is negative
+   *
+   * @deprecated
    */
+  @Deprecated
   public ClusteringServiceConfigurationBuilder readOperationTimeout(long duration, TimeUnit unit) {
-    return new ClusteringServiceConfigurationBuilder(this, TimeoutDuration.of(duration, unit));
+    Duration readTimeout = Duration.of(duration, toChronoUnit(unit));
```
ljacomet: This can now throw.
And again about API stability, if we decide to drop the methods on the configuration class, this one should go as well. Note that I am _not_ advocating that solution, rather making sure we take a consistent attitude.

clustered/client/src/main/java/org/ehcache/clustered/client/config/Timeouts.java
```diff
@@ -0,0 +1,150 @@
...
+        ", mutativeOperationTimeout=" + mutativeOperationTimeout +
+        ", lifecycleOperationTimeout=" + lifecycleOperationTimeout +
+        '}';
+  }
+
+  /**
+   * Constructs instances of {@link Timeouts}.  When obtained from
+   * {@link Timeouts#builder()}, the default values are pre-set.
+   */
+  public static final class Builder {
```
ljacomet: This could implement  `org.ehcache.config.Builder<Timeouts>`

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -76,7 +68,23 @@ public ServerSideConfigurationBuilder autoCreate() {
...
+   * Mutative operations which time out won't do anything.
+   * Lifecycle operations which time out will fail with exception
+   *
+   * @param timeouts the amount of time permitted for all operations
+   *
+   * @return a clustering service configuration builder
+   *
+   * @throws NullPointerException if {@code timeouts} is {@code null}
+   */
+  public ClusteringServiceConfigurationBuilder operationTimeouts(Timeouts timeouts) {
```
ljacomet: We should provide an overloaded method that takes a `Builder<Timeouts>` and just builds it.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
henri-tremblay: Yes, we can. The thought (I think with @chrisdennis ) was that it's probably unused right now so keeping compatibility is not important. But if you think it is, we can.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
henri-tremblay: Same answer :-)

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: Ok. That was one of my questions in the description of the PR. Let's do it.

clustered/client/src/main/java/org/ehcache/clustered/client/config/Timeouts.java
henri-tremblay: ok

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: Yes. I'm ok with coherence of course. We just need to decide

henri-tremblay:
Fix for #2196

henri-tremblay:
Build will fail because of Terracotta-OSS/terracotta-core#800

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -76,7 +68,39 @@ public ServerSideConfigurationBuilder autoCreate() {
...
+   * Mutative operations which time out won't do anything.
+   * Lifecycle operations which time out will fail with exception
+   *
+   * @param timeoutsBuilder the builder for amount of time permitted for all operations
+   *
+   * @return a clustering service configuration builder
+   *
+   * @throws NullPointerException if {@code timeouts} is {@code null}
+   */
+  public ClusteringServiceConfigurationBuilder timeouts(Timeouts.Builder timeoutsBuilder) {
```
chrisdennis: The parameter here should really be `Builder<? extends Timeouts>`.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
chrisdennis: If this doesn't go we should also think about defining what mixing calls to this and the new method does.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
```diff
@@ -114,14 +115,13 @@ public void create(final String identifier, final ServerSideConfiguration config
...
-      EntityRef<InternalClusterTierManagerClientEntity, ClusterTierManagerConfiguration, Void> ref = getEntityRef(identifier);
+      EntityRef<ClusterTierManagerClientEntity, ClusterTierManagerConfiguration, Void> ref = getEntityRef(identifier);
       try {
         while (true) {
           ref.create(new ClusterTierManagerConfiguration(identifier, config));
           try {
-            InternalClusterTierManagerClientEntity entity = ref.fetchEntity(null);
+            ClusterTierManagerClientEntity entity = ref.fetchEntity(null);
             try {
-              entity.setTimeouts(entityTimeouts);
```
chrisdennis: Without the two calls, I'm wondering if some of the "finished" ugliness can be cleaned up?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
```diff
@@ -35,73 +32,56 @@
...
-    invokeInternal(timeouts.getLifecycleOperationTimeout(), messageFactory.validateStoreManager(config), false);
+    invokeInternal(messageFactory.validateStoreManager(config), false);
   }
 
   @Override
   public Set<String> prepareForDestroy() {
     try {
-      PrepareForDestroy response = (PrepareForDestroy) invokeInternal(timeouts.getLifecycleOperationTimeout(), messageFactory
-        .prepareForDestroy(), true);
+      PrepareForDestroy response = (PrepareForDestroy) invokeInternal(messageFactory.prepareForDestroy(), true);
```
chrisdennis: This shouldn't timeout now right? so the TimeoutException shouldn't need to be caught here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
```diff
@@ -35,73 +32,56 @@
...
-
   @Override
   public void close() {
     endpoint.close();
   }
 
   @Override
   public void validate(ServerSideConfiguration config) throws ClusterException, TimeoutException {
-    invokeInternal(timeouts.getLifecycleOperationTimeout(), messageFactory.validateStoreManager(config), false);
+    invokeInternal(messageFactory.validateStoreManager(config), false);
```
chrisdennis: Is validate a lifecycle thing or a connection thing?  Should this be able to timeout?  If not we should get rid of the TimeoutException throw.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -128,7 +129,19 @@ public URI getNamespace() {
             /*
              * <read-timeout> is an optional element
              */
-            getTimeout = processGetTimeout(fragment, item);
+            getTimeout = processTimeout(fragment, item);
```
chrisdennis: I'm thinking that this has now become complex enough that we should be thinking about a more robust parsing approach?  More JAXB?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -128,7 +129,19 @@ public URI getNamespace() {
...
+
+          } else if ("mutative-timeout".equals(item.getLocalName())) {
+            /*
+             * <mutative-timeout> is an optional element
+             */
+            putTimeout = processTimeout(fragment, item);
+
+          } else if ("lifecycle-timeout".equals(item.getLocalName())) {
```
chrisdennis: connection-timeout?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -128,7 +129,19 @@ public URI getNamespace() {
...
              * <read-timeout> is an optional element
              */
-            getTimeout = processGetTimeout(fragment, item);
+            getTimeout = processTimeout(fragment, item);
+
+          } else if ("mutative-timeout".equals(item.getLocalName())) {
+            /*
+             * <mutative-timeout> is an optional element
+             */
+            putTimeout = processTimeout(fragment, item);
```
chrisdennis: The variable names here should probably be consistent with the other naming (write|mutative).

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -183,12 +200,12 @@ private TimeoutDuration processGetTimeout(Element parentElement, Node timeoutNod
             String.format("Value of XML configuration element <%s> in <%s> exceeds allowed value - %s",
                 timeoutNode.getNodeName(), parentElement.getTagName(), amount));
       }
-      getTimeout = TimeoutDuration.of(amount.longValue(), convertToJavaTimeUnit(timeType.getUnit()));
+      timeout = Duration.of(amount.longValue(), convertToJavaTimeUnit(timeType.getUnit()));
```
chrisdennis: Could just return here.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -57,7 +57,23 @@
...
+      </xs:element>
+      <xs:element name="mutative-timeout" type="eh:time-type" minOccurs="0" maxOccurs="1">
+        <xs:annotation>
+          <xs:documentation>
+            Specifies the amount of time a cache mutative operation will wait for a response from a cluster
+            server before abandoning the cluster operation.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:element>
+      <xs:element name="lifecycle-timeout" type="eh:time-type" minOccurs="0" maxOccurs="1">
```
chrisdennis: This is the connection timeout now right?  Also given it's a connection timeout wouldn't it make sense for this to be more tightly associated with the connection related stuff?

```xml
<tc:connection url="terracotta://example.com:9540,example.com:9640/cachemanager">
  <tc:timeout unit="seconds">30</tc:timeout>
</tc:connection>
```

clustered/client/src/main/java/org/ehcache/clustered/client/config/Timeouts.java
```diff
@@ -0,0 +1,158 @@
...
+ * Describes the timeouts for {@link ClusterTierManagerClientEntity} operations.  Use
+ * {@link #builder()} to construct an instance.
+ */
+public final class Timeouts {
+
+  public static final Duration DEFAULT_OPERATION_TIMEOUT = Duration.ofSeconds(5);
+  public static final Duration INFINITE_TIMEOUT = Duration.ofMillis(Long.MAX_VALUE);
+
+  private final Duration readOperationTimeout;
+  private final Duration mutativeOperationTimeout;
```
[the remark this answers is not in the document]
henri-tremblay: I totally agree. And it is the usual way to say it. So I will modify to `write`

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -54,31 +55,38 @@
...
-   *                            read operations; if {@code null}, the default value is used
+   *                            read operations
+   *
+   * @throws NullPointerException if {@code clusterUri} is {@code null}
+   * @throws IllegalArgumentException if {@code clusterUri} is not URI valid for cluster operations
+   *
+   * @deprecated Use {@link #ClusteringServiceConfiguration(URI, Timeouts)} instead
+   */
+  @Deprecated
+  public ClusteringServiceConfiguration(URI clusterUri, Duration readOperationTimeout) {
```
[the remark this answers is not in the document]
henri-tremblay: Yes. It is weird indeed. I think it happened iteratively... But yes, we should decide.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: Right

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
henri-tremblay: Yes

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -235,7 +248,7 @@ public TimeoutDuration getReadOperationTimeout() {
 
   @SuppressWarnings("unchecked")
   @Override
-  public CacheManagerBuilder<PersistentCacheManager> builder(final CacheManagerBuilder<? extends CacheManager> other) {
+  public CacheManagerBuilder<PersistentCacheManager> builder(CacheManagerBuilder<? extends CacheManager> other) {
```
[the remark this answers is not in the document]
henri-tremblay: Done

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
henri-tremblay: For a manager, we decided that it should wait ever after and that the lease should take care of it. But indeed, it doesn't need the `TimeoutException` anymore

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
henri-tremblay: Yes

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
henri-tremblay: Yes. Done

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
henri-tremblay: It is JAXB underneath

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
henri-tremblay: Yes. Fixed

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
henri-tremblay: Ok

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
henri-tremblay: No. Was renamed connection-timeout and that's it. Need to be discussed?

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
henri-tremblay: Removed as discussed

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: Removed as discussed

henri-tremblay:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/config/Timeouts.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @Override
+  public int hashCode() {
+    int result = readOperationTimeout.hashCode();
+    result = 31 * result + writeOperationTimeout.hashCode();
+    result = 31 * result + connectionTimeout.hashCode();
+    return result;
+  }
+
+  public static Builder builder() {
```
ljacomet: This is not consistent with the way we have builders as top level classes in a different package. I would instead move the builder to `<package>.builders.TimeoutsBuilder` for consistency.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -90,18 +114,19 @@ public ServerSideConfigurationBuilder expecting() {
...
    * @throws NullPointerException if {@code unit} is {@code null}
    * @throws IllegalArgumentException if {@code amount} is negative
+   *
+   * @deprecated Use {@link #timeouts(Timeouts)}. Note that calling this method will override any timeouts previously set
+   * by setting the read operation timeout to the specified value and everything else to its default.
    */
+  @Deprecated
   public ClusteringServiceConfigurationBuilder readOperationTimeout(long duration, TimeUnit unit) {
-    return new ClusteringServiceConfigurationBuilder(this, TimeoutDuration.of(duration, unit));
+    Duration readTimeout = Duration.of(duration, toChronoUnit(unit));
```
ljacomet: This can now throw, something that was not the case before. I believe we should hide that and simply use the largest `java.time.Duration` in case what is passed in is too large, as in the end, it will not make a difference.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
ljacomet: no it is not, we do the parsing by hand for now. We should indeed consider moving to JAXB ... though that can (and should) be done in a separate PR.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -183,12 +202,11 @@ private TimeoutDuration processGetTimeout(Element parentElement, Node timeoutNod
             String.format("Value of XML configuration element <%s> in <%s> exceeds allowed value - %s",
                 timeoutNode.getNodeName(), parentElement.getTagName(), amount));
       }
-      getTimeout = TimeoutDuration.of(amount.longValue(), convertToJavaTimeUnit(timeType.getUnit()));
+      return Duration.of(amount.longValue(), convertToJavaTimeUnit(timeType.getUnit()));
```
ljacomet: This can now throw, same remark as before.

clustered/client/src/main/java/org/ehcache/clustered/client/config/Timeouts.java
henri-tremblay: Like `Expirations` :-) I thought about it and kinda let it slip as an exception. But I can

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: throw an exception you mean? Why do you care? It is deprecated anyway and the `Duration` will throw a nice exception telling what's wrong anyway.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
henri-tremblay: Same comment :-) (ok, this one is not deprecated, but still, they will get a nice error)

henri-tremblay:
@ljacomet For some reason I can't comment on the  JAXB comment. The date-time parsing is made by JAXB from as far as I understood. But yes, I would leave it like this for now.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -131,67 +106,39 @@ public ClusteringServiceConfiguration(final URI clusterUri, final TimeoutDuratio
...
-    validateClusterUri(clusterUri);
-    validateReadOperationTimeout(readOperationTimeout);
-    if (serverConfig == null) {
-      throw new NullPointerException("Server configuration cannot be null");
-    }
-    this.clusterUri = clusterUri;
+  public ClusteringServiceConfiguration(URI clusterUri, Timeouts timeouts, boolean autoCreate, ServerSideConfiguration serverConfig) {
+    this.clusterUri = Objects.requireNonNull(clusterUri, "Cluster URI cannot be null");
     this.autoCreate = autoCreate;
     this.serverConfiguration = serverConfig;
```
chrisdennis: Haven't we ended up dropping the null-checking on serverConfig here?

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -90,18 +114,19 @@ public ServerSideConfigurationBuilder expecting() {
...
+    return timeouts(Timeouts.builder().setReadOperationTimeout(readTimeout).build());
```
chrisdennis: Why can't we add a copying builder and just override the read operation timeout?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
```diff
@@ -110,15 +89,22 @@ private EhcacheEntityResponse invokeInternal(TimeoutDuration timeLimit, EhcacheE
...
+   * interruption flag is then set if needed before returning the get value (or if an exception occurred).
+   *
+   * @param future Future we want to get
+   * @param <T> type of the response
+   * @return the result of the get
+   * @throws EntityException exception that might be thrown by the future in case of error
+   */
+  private static <T extends EntityResponse> T waitFor(InvokeFuture<T> future)
+      throws EntityException {
     boolean interrupted = false;
```
chrisdennis: Not your code but `boolean interrupted = Thread.interrupted()` is a better initialization here.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -171,7 +169,7 @@ public boolean isConnected() {
   @Override
   public void validate(ServerStoreConfiguration clientStoreConfiguration) throws ClusterTierException, TimeoutException {
     try {
-      invokeInternalAndWait(endpoint.beginInvoke(), timeouts.getLifecycleOperationTimeout(), messageFactory.validateServerStore(storeIdentifier , clientStoreConfiguration), false);
+      invokeInternalAndWait(endpoint.beginInvoke(), timeouts.getWriteOperationTimeout(), messageFactory.validateServerStore(storeIdentifier , clientStoreConfiguration), false);
```
chrisdennis: Why is validate a write operation?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
chrisdennis: Fair enough.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: Because it's deprecated. So anybody currently using this will have the same behaviour, and if he wants to add move timeouts, migrating would be a good idea.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
henri-tremblay: Yep. Note that we also have `ExecutorUtil.waitFor` that is also initialized to `false`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
henri-tremblay: Something we decided with @ljacomet. What would you use? Connection?

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
henri-tremblay: Why? It looks like good defensive programming.

clustered/client/src/main/java/org/ehcache/clustered/client/config/Timeouts.java
henri-tremblay: On problem with that is that the constructor can't be private anymore. Before, we were forcing using the builder

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
ljacomet: Since we added back a connection timeout, I would make validate use that one indeed.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
ljacomet: I care because it did not previously, so it is a behaviour change, much more sneaky than an API change. So like I said, I would catch it and use the largest duration we can represent.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
henri-tremblay: Deal. I thought about it but wasn't sure it was making sense. But now I'm convinced :-)

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/TimeoutsBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+  private Duration readOperationTimeout = Timeouts.DEFAULT.getReadOperationTimeout();
+  private Duration writeOperationTimeout = Timeouts.DEFAULT.getWriteOperationTimeout();
+  private Duration connectionTimeout = Timeouts.DEFAULT.getConnectionTimeout();
+
+  /**
+   * Obtain a new get for timeouts
+   *
+   * @return timeout get
+   */
+  public static TimeoutsBuilder get() {
```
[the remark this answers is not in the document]
henri-tremblay: Ok. Make sense

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
henri-tremblay: I found it much sneakier to modify the value set by the user :-(

henri-tremblay: After discussion, I will move anything bigger than `Duration.ofNanos(Long.MAX_VALUE)` to `Duration.ofNanos(Long.MAX_VALUE)` in the `Timeouts` constructor

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
henri-tremblay: Yes. Because null is a valid serverConfig value

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
chrisdennis: I don't like this solution but I think it's the best way forward now.  I'll push for what I'd like to see on the expiry changes PR.

PR #2176, opened 2017-11-08
Using a HashMap to return the invalidated values is a bit heavyweight since we never use the map capabilities

henri-tremblay:
A `List` does the job perfectly and removes all the hashing at insertion, the slow traversal and the superior object allocation


chrisdennis:
This looks sensible... but now that you make me look at it something seems off about this code.  It seems like it's invalidating everything under the bucket, and not just the things that match the hashcode.

PR #2179, opened 2017-11-13
:arrow_up: Bump to latest version

AbfrmBlr:


ljacomet:
Build failed because it timed out - most likely we have a test not working as expected but blocking instead of failing.
@AbfrmBlr you are looking into that, right?

AbfrmBlr:
yes

PR #2189, opened 2017-11-17
Offheap properties should be moved from net.sf.ehcache to org.ehcache

henri-tremblay:


impl/src/main/java/org/ehcache/impl/internal/store/offheap/OffHeapStoreUtils.java
```diff
@@ -46,17 +48,17 @@ public static BufferSource getBufferSource() {
...
-    String globalPropertyKey = "net.sf.ehcache.offheap.config." + property;
+    String globalPropertyKey = PATH_PREFIX + property;
     return MemorySizeParser.parse(System.getProperty(globalPropertyKey, Long.toString(defaultValue)));
   }
 
   public static long getAdvancedLongConfigProperty(String property, long defaultValue) {
-    String globalPropertyKey = "net.sf.ehcache.offheap.config." + property;
-    return Long.parseLong(System.getProperty(globalPropertyKey, Long.toString(defaultValue)));
+    String globalPropertyKey = PATH_PREFIX + property;
+    return Long.getLong(globalPropertyKey, defaultValue);
```
chrisdennis: Was the change from `parseLong` to `getLong` intentional?  It's making the number parsing more tolerant, which isn't bad in principle but it means the long parsing is now more tolerant than the number parsing in the MemorySizeParser, which might be confusing.

henri-tremblay: It was just to have shorter code actually. Shall I go the more tolerant way or the way it was before?

chrisdennis: Not a big deal just stuck out at me.

henri-tremblay: Knowing it's fairly unused, I've put the old code back for coherence

ljacomet:
Good to go once rebased

henri-tremblay:
Rebased

PR #2191, opened 2017-11-27
Issue 2175: No user control over unmarshalling

rahul-mittal:


terracotta-org:

ljacomet:
test this please

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -33,7 +52,8 @@
    * @param valueClass concrete value type
    * @param <K> the key type, must be {@code Serializable}
    * @param <V> the value type, must be {@code Serializable}
+   * @param whiteListedClasses classes whose objects would get deserialized as part of key or value deserialization
    * @return a state holder
    */
-  <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass);
+  <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass, Iterable<Class> whiteListedClasses);
```
chrisdennis: This is a breaking API change, not sure what policy is on this atm.  Would it be better or worse to have the default implementation call the original non-whitelisting method?

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -18,11 +18,30 @@
...
+   *
+   * @param name       the state holder name
+   * @param keyClass   concrete key type
+   * @param valueClass concrete value type
+   * @param <K>        the key type, must be {@code Serializable}
+   * @param <V>        the value type, must be {@code Serializable}
+   * @param whiteListedClasses classes whose objects would get deserialized as part of key or value deserialization
+   * @return a state holder
+   */
+  default <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass, Class ... whiteListedClasses) {
```
chrisdennis: You should be using the wildcard `Class<?>` here instead of the raw type - same goes for every other Class reference added.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ClusteredStateHolder.java
```diff
@@ -38,13 +38,15 @@
   private final Class<K> keyClass;
   private final ValueCodec<K> keyCodec;
   private final ValueCodec<V> valueCodec;
+  private final Iterable<Class> whiteListedClasses;
```
chrisdennis: This variable isn't used.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
```diff
@@ -39,7 +41,7 @@
         || clazz.isPrimitive() || String.class.equals(clazz)) {
```
chrisdennis: Should we care if a primitive wrapper class hasn't been whitelisted?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
```diff
@@ -79,10 +88,14 @@ public T decode(Object input) {
...
       try {
-        try (ObjectInputStream ois = new ObjectInputStream(bais)) {
+        try (ObjectInputStream ois =
+               (whiteListedClasses != null) ? new WhiteListObjectInputStream(bais, whiteListedClasses)
+                                                                  : new ObjectInputStream(bais)) {
           @SuppressWarnings("unchecked")
           T result = (T) ois.readObject();
           return result;
+        } catch (InvalidClassException e) {
+          throw new RuntimeException(e.getMessage());
```
chrisdennis: Why are we obscuring the stacktrace?  Better to just wrap the exception.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
```diff
@@ -79,10 +88,14 @@ public T decode(Object input) {
       ValueWrapper data = (ValueWrapper) input;
       ByteArrayInputStream bais = new ByteArrayInputStream(data.getValue());
       try {
-        try (ObjectInputStream ois = new ObjectInputStream(bais)) {
+        try (ObjectInputStream ois =
+               (whiteListedClasses != null) ? new WhiteListObjectInputStream(bais, whiteListedClasses)
```
chrisdennis: This null special casing is a bit ugly.  Would it be easier to write a "contains-all" set implementation?  You could even just have a Predicate instead of an explicit whitelist.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
```diff
@@ -149,23 +150,27 @@ public ByteBuffer getPayload() {
     };
   }
 
-  public static Object unmarshall(byte[] payload) {
+  public static Object unmarshallWhiteListed(byte[] payload, List<Class> whiteListedClasses) {
```
chrisdennis: No point in renaming these methods since we're not retaining the originals.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
```diff
@@ -149,23 +150,27 @@ public ByteBuffer getPayload() {
...
     ObjectInputStream objectInputStream = null;
     try {
-      objectInputStream = new ObjectInputStream(new ByteArrayInputStream(payload));
+      objectInputStream = new WhiteListObjectInputStream(new ByteArrayInputStream(payload), whiteListedClasses);
       return objectInputStream.readObject();
+    } catch (InvalidClassException e) {
```
chrisdennis: `InvalidClassException` is a subclass of `IOException` - we can probably eliminate the extra catch block.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
```diff
@@ -149,23 +150,27 @@ public ByteBuffer getPayload() {
...
+      throw new RuntimeException(e.getMessage());
     } catch (IOException | ClassNotFoundException ex) {
       throw new IllegalArgumentException(ex);
     } finally {
       closeSilently(objectInputStream);
```
chrisdennis: We should be able to switch to try-with-resources and eliminate the `closeSilently(...)` method.  Can probably also eliminate the duplication between the `byte[]` and `ByteBuffer` variants while we're at it.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/WhiteListObjectInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static java.util.stream.Collectors.toList;
+
+public class WhiteListObjectInputStream extends ObjectInputStream {
+
+  private List<String> whiteListedClassNames;
+
+  public WhiteListObjectInputStream(InputStream in, Iterable<Class> whiteListedClasses) throws IOException {
+    super(in);
+    whiteListedClassNames = StreamSupport.stream(whiteListedClasses.spliterator(), false)
```
chrisdennis: Mapping to names seems a little hazardous.  A class name does not capture it's identity correctly.  It would seem to be safer to load the class (without initializing it) and do the proper check.  That would also allow us to switch to a set contains check which will be quicker.

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
```diff
@@ -96,7 +94,7 @@ private void saveMaps() throws IOException {
   private String createFileName(Map.Entry<String, Tuple> entry) {return HOLDER_FILE_PREFIX + entry.getValue().index + "-" + safeIdentifier(entry.getKey(), false) + HOLDER_FILE_SUFFIX;}
 
   @Override
-  public <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass) {
+  public <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass, Iterable<Class> whiteListedClasses) {
```
chrisdennis: Should this be whitelisting?

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
ljacomet: Because this is API, I would much prefer we deprecate the existing method and add a new one. Also I do not see the benefit of having both the vararg and `Iterable` versions.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
ljacomet: This is more of an optimisation where we do not serialise JDK classes as we can resolve them on the server and their immutability makes them safe.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/WhiteListObjectInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+      .map(Class::getName)
+      .collect(toList());
+  }
+
+  @Override
+  protected Class<?> resolveClass(ObjectStreamClass serialInput) throws IOException, ClassNotFoundException {
+    final String name = serialInput.getName();
+
+    if (!whiteListedClassNames.stream().anyMatch(className -> className.equals(name))) {
```
ljacomet: `noneMatch` instead

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
ljacomet: Yes!

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
rahul-mittal: done

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ClusteredStateHolder.java
rahul-mittal: removed

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
rahul-mittal: retained the original but marked as deprecated, added new one with Collection<Class<?>> as extra parameter with default implementation which ignores the passed in whitelist.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
rahul-mittal: done

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
rahul-mittal: done

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
rahul-mittal: done

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/WhiteListObjectInputStream.java
rahul-mittal: done

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/WhiteListObjectInputStream.java
rahul-mittal: changed to Collection.contains() now

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
rahul-mittal: Are you saying to push this part inside WhitelistObjectInputstream, yes null handling is ugly, did not get the alternative you are proposing here

chrisdennis: Either:
1. Switching to passing a predicate to the WhiteListObjectInputStream.  When whitelisting this would be `whitelist::contains` and when not whitelisting it would be `c -> true`.
2. Writing a "contains-eeverything set implementation, and passing that in when not whitelisting.

Personally I prefer option 1.

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
rahul-mittal: change parameter name to "whitelisting" instead of "whiteListedClasses" is what you are looking at ?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ClusterStateRepository.java
```diff
@@ -40,6 +41,15 @@
 
   @Override
   public <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass) {
-    return new ClusteredStateHolder<>(clusterCacheIdentifier.getId(), composedId + "-" + name, clientEntity, keyClass, valueClass);
+    return new ClusteredStateHolder<>(clusterCacheIdentifier.getId(), composedId + "-" + name, clientEntity, keyClass, valueClass,null);
```
ljacomet: code formatting

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -17,12 +17,30 @@
...
  * A repository allowing to preserve state in the context of a {@link org.ehcache.Cache}.
  */
 public interface StateRepository {
 
+  /**
+   * Gets a named state holder rooted in the current {@code StateRepository}.
+   * <p>
+   * If the state holder existed already, it is returned with its content fully available.
+   *
+   * @deprecated Replaced by {@link #getPersistentStateHolder(String, Class, Class, Collection)} that takes in whiteListedClasses as parameter to be used during deserialization
```
ljacomet: > that takes in a collection of white listed classes authorised for deserialization

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
chrisdennis: No - I mean that there is no enforcement of the whitelist here.

rahul-mittal:
test this please

rahul-mittal:
All review comments addressed.

impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
ljacomet: Had a chat with @rahul-mittal and because this class eagerly loads the saved data, it stays as-is for now.

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
chrisdennis: It seems wrong to have the default implementation not honor the contract properly.  Shouldn't it be the non-whitelisting implementation that has a default implementation that calls the whitelisting one with a "infinite whitelist"?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
chrisdennis: Sure - you'll get a 'supressed' exception instead which is better in all ways.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/WhiteListObjectInputStream.java
chrisdennis: The current version still uses class names to identify classes.  This seems wrong... we should be using Class instances which bind the name, the definition and the defining loader all together.

chrisdennis:
The following still need addressing:

https://github.com/ehcache/ehcache3/pull/2191#discussion_r155271667
https://github.com/ehcache/ehcache3/pull/2191#discussion_r155804091
https://github.com/ehcache/ehcache3/pull/2191#discussion_r155804466
https://github.com/ehcache/ehcache3/pull/2191#discussion_r155805153
https://github.com/ehcache/ehcache3/pull/2191#pullrequestreview-80447991

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
rahul-mittal: Done, taking in Predicate<Class<?>> as input parameter now

api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
rahul-mittal: Added a default implementation for non-whitelisted one now that calls whitelisted one with predicate as c->true and null classLoader.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
rahul-mittal: removed closeSilently()

clustered/common/src/main/java/org/ehcache/clustered/common/internal/store/WhiteListObjectInputStream.java
rahul-mittal: as discussed , taking in Predicate<Class<?>> and classloader as input now

rahul-mittal:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
```diff
@@ -79,11 +91,11 @@ public T decode(Object input) {
       ValueWrapper data = (ValueWrapper) input;
       ByteArrayInputStream bais = new ByteArrayInputStream(data.getValue());
       try {
-        try (ObjectInputStream ois = new ObjectInputStream(bais)) {
```
chrisdennis: Pretty sure the assertion at the end of this method should not be needed, since you shouldn't be ignoring the IOException.  If you mean to assert that the IOException itself is impossible then throw the AssertionError from there wrapping the IOException.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/StateRepositoryOpCodec.java
```diff
@@ -152,8 +155,26 @@ public StateRepositoryOpMessage decode(EhcacheMessageType messageType, ByteBuffe
...
+    java.lang.Integer.class,
+    java.lang.Long.class,
+    java.lang.Float.class,
+    java.lang.Double.class,
+    java.lang.Byte.class,
+    java.lang.Character.class,
+    java.lang.String.class,
+    java.lang.Boolean.class,
+    java.lang.Short.class,
+    java.lang.Void.class,
```
chrisdennis: `Void` isn't `Serializable` why is it here?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/StateRepositoryOpCodec.java
```diff
@@ -152,8 +155,26 @@ public StateRepositoryOpMessage decode(EhcacheMessageType messageType, ByteBuffe
...
 
     ByteBuffer keyBuffer = decoder.byteBuffer(KEY_FIELD);
-    Object key = Util.unmarshall(keyBuffer);
+    Object key = Util.unmarshall(keyBuffer, WHITELIST_PREDICATE);
 
     return new StateRepositoryOpMessage.GetMessage(storeName, mapId, key);
   }
+
+  public static final Predicate<Class<?>> WHITELIST_PREDICATE = Collections.unmodifiableList(Arrays.asList(
+    java.lang.Integer.class,
```
chrisdennis: List contains is O(N)... we should really be using a set contains here.  Also the unmodifiable is unnecessary if you're only retaining the object through the predicate

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ValueCodecFactory.java
rahul-mittal: Removed the uneccessary assertion

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/StateRepositoryOpCodec.java
rahul-mittal: changed to hashset

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/StateRepositoryOpCodec.java
rahul-mittal: removed Void from the list

rahul-mittal:
test this please

PR #2195, opened 2017-12-04
Ignore transaction log files

henri-tremblay:


henri-tremblay:
Ah. Good. It's a better fix actually. Closing this

PR #2206, opened 2017-12-07
Add build scan plugin to gradle

henri-tremblay:


build.gradle
```diff
@@ -25,6 +27,11 @@ plugins {
   id "com.dorongold.task-tree" version "1.3"
 }
 
+buildScan {
+  licenseAgreementUrl = 'https://gradle.com/terms-of-service'
+  licenseAgree = 'yes'
```
chrisdennis: Who is agreeing here?  I imagine it's the person running the gradle build, but can we make that decision for everyone who might build the project.  With that in mind why does this need to be checked in?  Strikes me as something you'd run occasionally locally, but not every time.

henri-tremblay:
It doesn't run everytime. You need to add "--scan" to activate. But it is
now pre-configured

henri-tremblay:
I seem to work perfectly without adding the plugin. So closing this.

PR #2211, opened 2017-12-13
Correctly close cache managers and put temporary files in a temporary folder

henri-tremblay:


ljacomet:
test this please

integration-test/src/test/java/org/ehcache/integration/transactions/xa/XACacheTest.java
```diff
@@ -66,40 +67,61 @@
...
         .setServerId("XACacheTest")
         .setGracefulShutdownInterval(0);
   }
 
   @After
   public void tearDownBtm() throws Exception {
+    if(cacheManager != null) {
+      cacheManager.close();
+    }
+    transactionManager.shutdown();
```
ljacomet: Adding this makes the following block useless? Although it seems the following block is _safer_ in a way.
@lorban Care to comment?

lorban: Right, the `transactionManager.shutdown()` call is redundant with the `if ()` block below.

henri-tremblay: Cool thanks

PR #2212, opened 2017-12-13
Support JCache 1.1.0

ljacomet:
PR to be ready for JCache 1.1 when it gets released as supporting it for Ehcache 3.5 would be ideal.

PR should fail to build until we can reference a released version though.

### DIFF
--- a/107/build.gradle
+++ b/107/build.gradle
@@ -32,8 +32,8 @@ sourceSets {
 dependencies {
   compile project(':impl'), project(':xml')
   provided "javax.cache:cache-api:$parent.jcacheVersion"
-  tckTestRuntime 'javax.cache:cache-tests:1.0.1'
-  tckTestClasses('javax.cache:cache-tests:1.0.1:tests') {
+  tckTestRuntime "javax.cache:cache-tests:$jcacheTckVersion"
+  tckTestClasses("javax.cache:cache-tests:$jcacheTckVersion:tests") {
     transitive = false
   }
 }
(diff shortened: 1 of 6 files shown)


107/build.gradle
```diff
@@ -32,8 +32,8 @@ sourceSets {
 dependencies {
   compile project(':impl'), project(':xml')
   provided "javax.cache:cache-api:$parent.jcacheVersion"
-  tckTestRuntime 'javax.cache:cache-tests:1.0.1'
-  tckTestClasses('javax.cache:cache-tests:1.0.1:tests') {
+  tckTestRuntime "j«avax.cache:cache-tests:$jcacheTckVersion"
```
chrisdennis: rogue character

107/src/test/java/org/ehcache/jsr107/Eh107CacheTypeTest.java
```diff
@@ -79,9 +79,6 @@ public void testRunTimeTypeSafety() throws Exception {
 
     try {
       cacheManager.getCache("cache1");
-      fail("Caches with runtime types should throw illegal argument exception when different types are used in getcache");
```
chrisdennis: This seems really broken... am I missing something here?

ljacomet: See jsr107/jsr107spec#340 for the details behind this change. It effectively covers up a hole in the API.

PR #2217, opened 2017-12-18
Ehcache base class to mutualize basic and loader-writer code

henri-tremblay:
This is the first step to mutualize the code between Ehcache and EhcacheWithLoaderWriter. A lot more can be done but at least we now have a starting point. This mutualization will be highly useful for implementing resilience.

### DIFF
--- a/api/src/main/java/org/ehcache/config/CacheConfiguration.java
+++ b/api/src/main/java/org/ehcache/config/CacheConfiguration.java
@@ -17,7 +17,7 @@
 package org.ehcache.config;
 
 import org.ehcache.Cache;
-import org.ehcache.expiry.Expiry;
+
 import org.ehcache.expiry.ExpiryPolicy;
 import org.ehcache.spi.service.ServiceConfiguration;
 
@@ -92,7 +92,7 @@ public interface CacheConfiguration<K, V> {
    * @deprecated Use {@link #getExpiryPolicy()}
    */
   @Deprecated
-  Expiry<? super K, ? super V> getExpiry();
+  org.ehcache.expiry.Expiry<? super K, ? super V> getExpiry();
 
   /**
    * The {@link ExpiryPolicy} rules for the {@link Cache}.
--- a/core/src/main/java/org/ehcache/core/Ehcache.java
+++ b/core/src/main/java/org/ehcache/core/Ehcache.java
@@ -16,35 +16,20 @@
 
 package org.ehcache.core;
 
-import java.time.Duration;
-import java.util.AbstractMap;
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.EnumMap;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.atomic.LongAdder;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.ehcache.Cache;
-import org.ehcache.Status;
 import org.ehcache.config.CacheConfiguration;
-import org.ehcache.config.CacheRuntimeConfiguration;
 import org.ehcache.core.events.CacheEventDispatcher;
-import org.ehcache.core.internal.resilience.LoggingRobustResilienceStrategy;
-import org.ehcache.core.internal.resilience.RecoveryCache;
-import org.ehcache.core.internal.resilience.ResilienceStrategy;
-import org.ehcache.core.spi.LifeCycled;
 import org.ehcache.core.spi.store.Store;
 import org.ehcache.core.spi.store.Store.PutStatus;
 import org.ehcache.core.spi.store.Store.RemoveStatus;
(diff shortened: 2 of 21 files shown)


impl/src/test/java/org/ehcache/docs/Ehcache3.java
```diff
@@ -41,7 +39,7 @@ public void ehcache3Expiry() throws Exception {
     CacheConfigurationBuilder<Long, String> configuration =
         CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class, ResourcePoolsBuilder
             .heap(100))
-            .withExpiry(new Expiry<Long, String>() {    // <1>
+            .withExpiry(new ExpiryPolicy<Long, String>() {    // <1>
```
ljacomet: Good catch!

core/src/main/java/org/ehcache/core/Ehcache.java
```diff
@@ -21,38 +21,26 @@
 import java.util.ArrayList;
 import java.util.Collection;
```
ljacomet: Imports above this one are unused

core/src/main/java/org/ehcache/core/EhcacheBase.java
```diff
@@ -0,0 +1,504 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Map<BulkOps, LongAdder> getBulkMethodEntries() {
+    return bulkMethodEntries;
+  }
+
+  protected static void checkNonNull(Object thing) {
```
ljacomet: Can't we leverage Java 8 constructs for these?

core/src/main/java/org/ehcache/core/EhcacheWithLoaderWriter.java
```diff
@@ -1019,79 +831,9 @@ public boolean replace(final K key, final V oldValue, final V newValue) throws C
...
-  }
-
-  /**
-   * {@inheritDoc}
-   */
   @Override
-  public Jsr107Cache<K, V> getJsr107Cache() {
-    return jsr107Cache;
+  public Jsr107Cache<K, V> createJsr107Cache() {
+    return new Jsr107CacheImpl();
```
ljacomet: Why this change? There now can be multiple instances of that type for a given Cache. And I am not sure that does not break some (untested) spec wording. So unless you have a strong argument for this, I would tend to leave it as it was - that is a field initialised during construction.

api/src/main/java/org/ehcache/config/CacheConfiguration.java
```diff
@@ -92,7 +92,7 @@
    * @deprecated Use {@link #getExpiryPolicy()}
    */
   @Deprecated
-  Expiry<? super K, ? super V> getExpiry();
+  org.ehcache.expiry.Expiry<? super K, ? super V> getExpiry();
```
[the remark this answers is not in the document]
henri-tremblay: This is a separate commit. It fixes deprecation warnings caused by #2166. Here is need to use the full path because I can't suppress warnings on imports. A stupid Java quirks.

core/src/main/java/org/ehcache/core/Ehcache.java
henri-tremblay: Indeed, merge side effect. Thanks

core/src/main/java/org/ehcache/core/EhcacheBase.java
henri-tremblay: Yes

core/src/main/java/org/ehcache/core/EhcacheRuntimeConfiguration.java
```diff
@@ -100,8 +99,9 @@ public ClassLoader getClassLoader() {
     return this.classLoader;
   }
 
+  @SuppressWarnings("deprecation")
   @Override
-  public Expiry<? super K, ? super V> getExpiry() {
+  public org.ehcache.expiry.Expiry<? super K, ? super V> getExpiry() {
```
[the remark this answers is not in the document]
henri-tremblay: Exactly

core/src/main/java/org/ehcache/core/EhcacheWithLoaderWriter.java
henri-tremblay: I can't. Because I can't this do `super(new Jsr107CacheImpl(this))`. The alternative is to set `this` using a setter. Which is ugly. But no, it doesn't break anything because the JSR wrapper will call `createJsr107Cache` once. It's even better because it is only initialize when JSR107 is used.

api/src/main/java/org/ehcache/config/CacheConfiguration.java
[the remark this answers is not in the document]
ljacomet: But then where is the suppress warnings for that instance?

core/src/main/java/org/ehcache/core/EhcacheWithLoaderWriter.java
ljacomet: Or have a field in subclasses for this ...

henri-tremblay: Yes. That would work too. But I preferred this way. Isn't it more efficient to instantiate it when needed?

api/src/main/java/org/ehcache/config/CacheConfiguration.java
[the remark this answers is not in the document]
henri-tremblay: It seems that since the method itself is deprecated, there's no need to suppress the warning

Issue #2219, opened 2017-12-22
Provide a way to pass connection properties in DefaultClusteringService


saurabhagas:
It would be nice if there could be a way to pass connection level properties in DefaultClusteringService.

PR #2220, opened 2017-12-22
Issue #2219: Provide a way to pass connection properties in DefaultClusteringService

saurabhagas:
Provide a way to pass connection properties in DefaultClusteringService.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
@@ -57,7 +57,6 @@ import org.terracotta.exception.EntityNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


terracotta-org:

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -90,6 +91,7 @@
   DefaultClusteringService(final ClusteringServiceConfiguration configuration) {
     this.configuration = configuration;
     URI ehcacheUri = configuration.getClusterUri();
+    this.properties = configuration.getProperties();
```
henri-tremblay: My problem here is that the timeout is coming from another place (that will override this). I don't like properties coming from two places. I don't have the best answer but I don't think it's a good idea.

saurabhagas:
test this please

AbfrmBlr:
test this please

henri-tremblay:
please rebase

saurabhagas:
test this please

PR #2223, opened 2017-12-27
Replace ValueSupplier by Supplier

henri-tremblay:
Since we just replaced Expiry by ExpiryPolicy, I fell that we can also remove ValueSupplier. It is currently deprecated to keep the Expiry API.


henri-tremblay:
:-) I just read the javadoc and it wasn't that crystal clear but let me cook something into  the `ExpiryPolicy` to make it clearer.

henri-tremblay:
Isn't `Previous values are not accessible directly but are rather available through a value {@code Supplier} to indicate that access can require computation (such as deserialization).` clear enough?

PR #2234, opened 2018-01-04
Made ClusteringServiceConfiguration return a copy of Properties

saurabhagas:


clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -198,7 +198,7 @@ public Timeouts getTimeouts() {
    * @return the connection {@code Properties}
    */
   public Properties getProperties() {
-    return properties;
+    return new Properties(properties);
```
ljacomet: This does not do what you believe it does :wink:

Also, you need to take a copy when assigning it to the field during construction to protect against mutation from the object you received.

saurabhagas: Done

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -144,7 +144,7 @@ public ClusteringServiceConfiguration(URI clusterUri, Timeouts timeouts, boolean
     this.autoCreate = autoCreate;
     this.serverConfiguration = serverConfig;
     this.timeouts = Objects.requireNonNull(timeouts, "Operation timeouts cannot be null");
-    this.properties = Objects.requireNonNull(properties, "Properties cannot be null");
+    this.properties = cloneProperties(Objects.requireNonNull(properties, "Properties cannot be null"));
```
henri-tremblay: You could do `(Properties) Objects.requireNonNull(properties, "Properties cannot be null").clone();` and be done with it.

saurabhagas:
@henri-tremblay, I've done the suggested change.

saurabhagas:
test this please

PR #2235, opened 2018-01-04
Test that StorePassThroughExceptions are handled correctly

henri-tremblay:
It is expected that StorePassThroughExceptions thrown are supposed to have their cause being thrown in the end instead of being wrapped in a StoreAccessException.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
@@ -20,6 +20,7 @@ import org.ehcache.config.ResourcePools;
 import org.ehcache.config.ResourceType;
 import org.ehcache.core.CacheConfigurationChangeListener;
 import org.ehcache.core.collections.ConcurrentWeakIdentityHashMap;
+import org.ehcache.core.exceptions.StorePassThroughException;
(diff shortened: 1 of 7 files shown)


impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
```diff
@@ -84,35 +81,21 @@ public void invalidateAllWithHash(long hash) throws StoreAccessException {
...
-      super(cause);
-    }
-
-    public StoreAccessException getStoreAccessException() {
-      return (StoreAccessException) getCause();
-    }
-
-    @Override
-    public synchronized Throwable fillInStackTrace() {
-      return this;
```
ljacomet: Slightly unrelated, but we should probably do the same trick in `StorePassThroughException`

henri-tremblay: I guess you are right since it disappears out of the store

ljacomet:
Good to go once rebased

henri-tremblay:
Rebased

henri-tremblay:
test this please

henri-tremblay:
The error seems real. I will look into it.

henri-tremblay:
rebased and good to go it seems

PR #2239, opened 2018-01-09
Make sure we do not received unexpected notifications

henri-tremblay:


clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
```diff
@@ -294,5 +298,6 @@ protected static void waitForAllNotifications(String... notificationTypes) throw
     t.interrupt(); // we interrupt the thread that is waiting on the message queue
 
     assertTrue("Still waiting for: " + waitingFor, waitingFor.isEmpty());
+    assertTrue("Unexpected notification: " + missingOnes, missingOnes.isEmpty());
```
ljacomet: Is it really that bad to receive notifications that we do not expect? I am wondering if that does not make the test(s) even more brittle in the face of changes.

henri-tremblay: Yes. We better know exactly what should happen. Because since we are removing notifications without knowing for which entity they are related, the result is pretty random.

I've added it because I was surprised to have to remove an UNFETCHED with removing a FETCHED. It puzzled me.

PR #2241, opened 2018-01-10
Switch plugin for XJC generation

ljacomet:


xml/config/checkstyle-suppressions.xml
```diff
@@ -5,6 +5,6 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-  <suppress files="src[\\/]generated[\\/]java[\\/]" checks=".*"/>
+  <suppress files="build[\\/]xjc[\\/]generated-sources[\\/]" checks=".*"/>
```
henri-tremblay: I was wondering why we were using square braces everywhere. We actually do not need them.

ljacomet: No idea, removing all of them from that file breaks checkstyle, so leaving them 😉

henri-tremblay: Weird, worked for me. Anyway. If it works...

chrisdennis: It's a regular expression thats why Louis has no idea what it's doing.  It's matching one of "/" or "\", for windows or *nix.

henri-tremblay: Oh! Of course... Too much Java, I thought the backslash was escaped and then escaping the slash...

henri-tremblay:
build might pass when rebased

PR #2244, opened 2018-01-11
Change default timeout

AbfrmBlr:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -173,8 +173,9 @@ private void createEntityFactory() {
   private void initClusterConnection() {
     try {
       properties.put(ConnectionPropertyNames.CONNECTION_NAME, CONNECTION_PREFIX + entityIdentifier);
-      properties.put(ConnectionPropertyNames.CONNECTION_TIMEOUT, Long.toString(timeouts.getConnectionTimeout().toMillis()));
+      properties.put(ConnectionPropertyNames.CONNECTION_TIMEOUT, Long.toString(Timeouts.INFINITE_TIMEOUT.toMillis()));
```
henri-tremblay: We should still use the connection timeout no? I missed something?

AbfrmBlr: No you are right, thats waht happens when you just cherry pick

AbfrmBlr:
Changes done

PR #2249, opened 2018-01-17
Fixes #2213 : Extract OversizeMappingException logic in to a common method

### LINKED ISSUE #2213: OffHeapServerStore has mistakes in it's OversizeMappingException handling
There are a handful of errors in the various mutative methods of `OffHeapServerStore` where the `OversizeMappingException` exception handling the implements cross segment eviction is broken.  We should look in to getting this fixed up at a minimium.  It may even be worthwhile extracting the logic in to a single common method to prevent these kinds of mistakes.

chrisdennis:


clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -226,6 +144,37 @@ private void writeUnlockAll() {
...
 
+  private void handleOversizeMappingException(long key, OversizeMappingException e, LongConsumer operation) {
+    handleOversizeMappingException(key, e, k -> {
+      operation.accept(k);
+      return null;
+    });
+  }
+
+  private <R> R handleOversizeMappingException(long key, OversizeMappingException e, LongFunction<R> operation) {
+    if (handleOversizeMappingException(key)) {
```
henri-tremblay: I'm all in for this refactoring. However, I don't get what it's doing. So maybe commenting on the implementation will be useful. We try to evict until it fits if I got it right. But why are we still throwing an exception after looping to evict? And not trying to reapply the operation. Is this part still broken but at least in a single place?

chrisdennis: I've added some clarifying javadoc.  The current code is I believe correct. One of the removed duplicates had a bug.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -148,57 +108,15 @@ public void replaceAtHead(long key, Chain expect, Chain update) {
     try {
       segmentFor(key).replaceAtHead(key, expect, update);
```
chrisdennis: Bug here... there should be a `return` after this line.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -88,29 +90,7 @@ public void append(long key, ByteBuffer payLoad) {
     try {
       segmentFor(key).append(key, payLoad);
```
chrisdennis: Bug here... there should be a return after this line.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -148,57 +108,15 @@ public void replaceAtHead(long key, Chain expect, Chain update) {
...
     }
   }
 
   public void put(long key, Chain chain) {
     try {
       segmentFor(key).put(key, chain);
     } catch (OversizeMappingException e) {
-      if (handleOversizeMappingException(key)) {
-        try {
-          segmentFor(key).put(key, chain);
```
chrisdennis: Bug here!  There should be a return after this line.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -226,10 +144,51 @@ private void writeUnlockAll() {
...
+    try {
+      OversizeMappingException e;
+      do {
+        try {
+          return operation.apply(key);
+        } catch (OversizeMappingException ex) {
+          e = ex;
+        }
+      } while (tryShrinkOthers(key));
+      throw e;
```
henri-tremblay: I still don't get with we are throwing `e`. In the end, if the `operation.apply` worked, aren't we happy? Also, if no amount of shrinking can safe us, aren't we in an infinite loop? Finally, `tryShrinkOthers` try to shrink every other segments. If it fails the first time, isn't it meaning we are doomed already? And so throw an `OversizeMappingException`? No need to have a while loop. Some line comments can possible help me understand.

chrisdennis: If the `operation.apply(key)` worked then the method will have returned.  We'll only get to `throw e` if we failed to shrink other segments. We've shrunk as much as possible and we still can't complete the operation.  Make sense?

henri-tremblay: Yes... I think I missed the return.

Issue #2197, opened 2017-12-05
Add lease mechanism to Ehcache


henri-tremblay:
The concept of a lease is to let a client or server know that the other party is gone if the lease can be renewed. An Ehcache client realizing that will go in resilience mode.

- Have a setup telling how long the lease should last
- When the lease can't be renewed, clear the caching tier, stop the clustered tier from calling the TC server (by throwing an exception each time to go to the resilience strategy? Slow but should work as a first step)
- Rely only on the resilience strategy
- Note that failing calls to the server won't trigger this mode. Only the lease will. They are considered hiccups and will be answered by the resilience strategy

PR #2250, opened 2018-01-18
Adding lease mechanism #2197

AbfrmBlr:


clustered/client/src/test/java/org/ehcache/clustered/client/internal/PassthroughLeasedConnectionService.java
```diff
@@ -0,0 +1,64 @@
...
+import org.terracotta.exception.EntityNotProvidedException;
+import org.terracotta.lease.connection.LeasedConnection;
+import org.terracotta.lease.connection.LeasedConnectionService;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.Properties;
+
+@OverrideService("org.terracotta.lease.connection.LeasedConnectionServiceImpl")
+public class PassthroughLeasedConnectionService implements LeasedConnectionService {
```
ljacomet: Why is this here and not in `terracotta-platform` or `passthrough`?

Note: I was very lazy and did not go back into the code before asking

AbfrmBlr: You are right this should go in passthrough

AbfrmBlr: Raised https://github.com/Terracotta-OSS/tc-passthrough-testing/pull/160

AbfrmBlr:
In api, user has to use Timeouts.connectiontimeout, that will be used for lease time out

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
```diff
@@ -0,0 +1,132 @@
...
+      UnitTestConnectionService.remove(CLUSTER_URI);
+    } catch (IllegalStateException e) {
+      assertThat(e.getMessage(), is("Connection already closed"));
+    }
+  }
+
+  @Test
+  public void testConnectionClosed() throws Exception {
+
+    final CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder =
```
henri-tremblay: Please remove these useless and noisy `final`

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
```diff
@@ -0,0 +1,132 @@
...
+import static org.junit.Assert.assertThat;
+
+@RunWith(Parameterized.class)
+public class ConnectionClosedTest {
+
+  private static final URI CLUSTER_URI = URI.create("terracotta://connection.com:9540/timeout");
+
+  @Parameters
+  public static ResourcePoolsBuilder[] data() {
+    return new ResourcePoolsBuilder[]{ResourcePoolsBuilder.newResourcePoolsBuilder()
```
henri-tremblay: why parameterized since you have only one parameter?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
```diff
@@ -0,0 +1,132 @@
...
+  public void removePassthroughServer() throws Exception {
+    try {
+      UnitTestConnectionService.remove(CLUSTER_URI);
+    } catch (IllegalStateException e) {
+      assertThat(e.getMessage(), is("Connection already closed"));
+    }
+  }
+
+  @Test
+  public void testConnectionClosed() throws Exception {
```
henri-tremblay: Can you explain in the method name the purpose of this test?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
```diff
@@ -0,0 +1,132 @@
+/*
```
henri-tremblay: This is not testing the lease right? Just that the connection still works. Can you test lease expiration and renewal?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
AbfrmBlr: No it has parameters, two types of resource pools one with local tier, one without it

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
AbfrmBlr: Lease expiry and renewal are already tested the lease code.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
henri-tremblay: Just got it.... Can you please indent like
```
  public static ResourcePoolsBuilder[] data() {
    return new ResourcePoolsBuilder[]{
        .with(clusteredDedicated("primary-server-resource", 2, MemoryUnit.MB)),
        .heap(10, EntryUnit.ENTRIES)
        .with(clusteredDedicated("primary-server-resource", 2, MemoryUnit.MB))
    };
  }
```

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
henri-tremblay: And you are super sure we don't need integration tests at the ehcache level?

AbfrmBlr: In integration test, as of now I don't see any way to control lease expiry/ renewal stuff. Since lease mechanism is as good as a normal connection timeout at the entity layer, since lease is hidden right?

PR #2251, opened 2018-01-18
Resilience strategy moving to the API (Fix #2198)

### LINKED ISSUE #2198: Separate the resilience strategy in two and push it to the API
Concerns the actual `ResilienceStrategy` class. It should be move to the API and separated in two:
- Cache resilience strategy, which contains the same methods as the Cache interface (with an exception added to the parameters)
- LoaderWriter resilience strategy, which contains the same methods as the CacheLoaderWriter interface (with an exception added to the parameters)

henri-tremblay:
It's not finished yet, I need to convert every methods. Changes in the architecture are
* The resilience strategy is moved to the API. Please check the packaging is fine
* The resilience strategy is simplified to have one method per cache method
* In case of a loader-writer, a lot of code was moved to a strategy dedicated to EhcacheWithLoaderWriter
* The `cacheLoadingObserver` was removed. Implementations of `CacheLoaderWriter` can monitor themselves

Things that will not be addressed by this PR are:
* The resilience strategy dedicated to the `CacheLoaderWriter`
* Make the resilience strategy configurable easily


core/src/main/java/org/ehcache/core/resilience/RobustLoaderWriterResilienceStrategy.java
```diff
@@ -64,12 +66,11 @@ public boolean containsKeyFailure(K key, StoreAccessException e) {
...
   public void putFailure(K key, V value, StoreAccessException e) {
     cleanup(key, e);
-  }
-
-  @Override
-  public void putFailure(K key, V value, StoreAccessException e, CacheWritingException f) {
-    cleanup(key, e);
-    throw f;
+    try {
+      loaderWriter.write(key, value);
```
chrisdennis: I'm thinking this should write then cleanup.  With this ordering we could cleanup, have another thread populate through the loader and then write.  That leaves the cache with the wrong contents, right?

henri-tremblay: Yes

core/src/main/java/org/ehcache/core/exceptions/StorePassThroughException.java
```diff
@@ -66,7 +66,7 @@ public synchronized Throwable fillInStackTrace() {
    * @return StoreAccessException to be thrown
    * @throws RuntimeException if {@code re} is a {@code StorePassThroughException} containing a {@code RuntimeException}
    */
-  public static StoreAccessException handleRuntimeException(RuntimeException re) {
+  public static StoreAccessException handleRuntimeException(Exception re) {
```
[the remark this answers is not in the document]
henri-tremblay: Yep

core/src/main/java/org/ehcache/core/EhcacheBase.java
```diff
@@ -263,6 +264,40 @@ public void clear() {
...
+      } else if (inCache == null) {
+        putIfAbsentObserver.end(PutIfAbsentOutcome.HIT);
+        return null;
+      } else {
+        putIfAbsentObserver.end(PutIfAbsentOutcome.HIT);
+        return inCache.get();
+      }
+    } catch (StoreAccessException e) {
+      try {
+        return resilienceStrategy.putIfAbsentFailure(key, value, e); // FIXME: We can't know if it's absent or not
```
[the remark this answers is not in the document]
henri-tremblay: This one is actually obsolete.  There are no "knownPresent" anymore

core/src/main/java/org/ehcache/core/EhcacheBase.java
```diff
@@ -263,6 +264,40 @@ public void clear() {
...
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public V putIfAbsent(final K key, final V value) {
+    putIfAbsentObserver.begin();
+    statusTransitioner.checkAvailable();
+    checkNonNull(key, value);
+
+    AtomicBoolean put = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
henri-tremblay: boolean[] is enough indeed

core/src/main/java/org/ehcache/core/EhcacheBase.java
```diff
@@ -221,7 +221,31 @@ public void remove(K key) throws CacheWritingException {
...
+    try {
+      removed = doRemoveInternal(key);
+      if (removed) {
+        removeObserver.end(RemoveOutcome.SUCCESS);
+      } else {
+        removeObserver.end(RemoveOutcome.NOOP);
+      }
+    } catch (StoreAccessException e) {
+      try {
+        resilienceStrategy.removeFailure(key, e);
```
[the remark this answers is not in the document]
henri-tremblay: That is a good question....

core/src/main/java/org/ehcache/core/EhcacheBase.java
```diff
@@ -118,53 +119,59 @@
...
+        } else {
+          getObserver.end(GetOutcome.HIT);
+          return valueHolder.get();
+        }
+      } catch (StoreAccessException e) {
+        V value = resilienceStrategy.getFailure(key, e);
         getObserver.end(GetOutcome.FAILURE);
+        return value;
       }
+    } catch(Exception e) {
```
chrisdennis: This should be `} catch (Throwable) {` right?

core/src/main/java/org/ehcache/core/internal/util/CollectionUtil.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Used to create a new collection with the correct size. Given an iterable, will try to see it the iterable actually
+   * have a size and will return it. If the iterable has no known size, we return the best bet.
+   *
+   * @param iterable the iterable we will try to find the size of
+   * @param bestBet our best bet for the size if the iterable is not sizeable
+   * @return the size of the iterable if found or null
+   */
+  public static int findBestCollectionSize(Iterable<?> iterable, int bestBet) {
```
chrisdennis: I'm wondering if the calling code might read nicer if we used:
```java
public static OptionalInt sizeOf(Iterable<?> iterable) {
  return iterable instanceof Collection ? OptionalInt.of(((Collection<?>) iterable).size()) : OptionalInt.empty();
}
```

core/src/main/java/org/ehcache/core/internal/util/CollectionUtil.java
```diff
@@ -50,4 +52,78 @@ public static int findBestCollectionSize(Iterable<?> iterable, int bestBet) {
...
+   * Copy each map entry to a new map but check that each key and value isn't null. Throw
+   * a {@code NullPointerException} if it's the case.
+   *
+   * @param entries entries to copy
+   * @param <K> type of key
+   * @param <V> type of value
+   * @return cloned map
+   * @throws NullPointerException if a key or value is null
+   */
+  public static <K, V> Map<K, V> copyMapButFailOnNull(Map<? extends K, ? extends V> entries) {
```
chrisdennis:
```java
public static <K, V> Map<K, V> nullSafeCopy(Map<? extends K, ? extends V> entries) {
  return entries.entrySet().stream().collect(toMap(
    e -> requireNonNull(e.getKey()),
    e -> requireNonNull(e.getValue())
  ));
}
```

core/src/main/java/org/ehcache/core/internal/util/CollectionUtil.java
```diff
@@ -50,4 +52,78 @@ public static int findBestCollectionSize(Iterable<?> iterable, int bestBet) {
...
+  /**
+   * Create a map with one key and a corresponding value.
+   *
+   * @param key0 the key
+   * @param value0 the value
+   * @param <K> key type
+   * @param <V> value type
+   * @return map with one entry
+   */
+  public static <K, V> Map<K, V> map(K key0, V value0) {
```
chrisdennis: Why not just inline this?

chrisdennis: On second thoughts... looks like these method are only used in unit tests for this class?  Are they needed?

core/src/main/java/org/ehcache/core/EhcacheWithLoaderWriter.java
```diff
@@ -405,17 +403,11 @@ protected void doRemoveAll(final Set<? extends K> keys) throws BulkCacheWritingE
...
   @Override
-  public boolean remove(final K key, final V value) throws CacheWritingException {
-    conditionalRemoveObserver.begin();
-    statusTransitioner.checkAvailable();
-    checkNonNull(key, value);
-    final AtomicBoolean hit = new AtomicBoolean();
-    final AtomicBoolean removed = new AtomicBoolean();
-    final BiFunction<K, V, V> remappingFunction = memoize((k, inCache) -> {
+  protected Store.RemoveStatus doRemove(K key, V value) throws StoreAccessException {
+    AtomicBoolean hit = new AtomicBoolean();
```
chrisdennis: More `boolean[1]` possibilities (I've probably missed some others too).

core/src/main/java/org/ehcache/core/exceptions/ExceptionFactory.java
```diff
@@ -47,4 +47,30 @@ public static CacheWritingException newCacheWritingException(Exception e) {
...
   }
+
+  /**
+   * Creates a new {@code CacheWritingException} with the provided exception as cause and a suppressed one.
+   *
+   * @param e the cause
+   * @param suppressed the suppressed exception to add to the new exception
+   * @return a cache writing exception
+   */
+  public static CacheWritingException newCacheWritingException(Exception e, Exception suppressed) {
```
chrisdennis: Why do we need this factory?  Wouldn't we better off just adding a fluent `supressing` method to `CacheWritingException` and having: `throw new CacheWritingException(e1).supressing(e);`

henri-tremblay: The factory was there already. I almost deleted it and then needed the suppression. Isn't it weird to add the fluent method on these exceptions? It feels like something that should be on some `BaseEhcacheException` or something.

core/src/main/java/org/ehcache/core/EhcacheWithLoaderWriter.java
henri-tremblay: I did it for all `EhcacheLoaderWriter`. There are some opportunities in `Ehcache` as well. Two questions: Are we sure it's better? And are we sure the functions can never execute in parallel? (we could imagine a store doing a getAll in parallel)

core/src/main/java/org/ehcache/core/internal/util/CollectionUtil.java
henri-tremblay: I use them in tests indeed. The singleton one is there for completude. I didn't one to add a new CollectionUtil for test only. And they are in Java 9 JDK. I don't think it's harmful to keep them there.

core/src/main/java/org/ehcache/core/internal/util/CollectionUtil.java
henri-tremblay: I don't think it's null safe. I think it doesn't like null. I've avoided the stream on purpose, I don't want the related object instantiation. Not using `requireNonNull` was also on purpose.

core/src/main/java/org/ehcache/core/internal/util/CollectionUtil.java
henri-tremblay: Possibly. But again, if I can avoid creating an optional, I've happy to

core/src/main/java/org/ehcache/core/EhcacheBase.java
henri-tremblay: I always feel not at ease when catching throwable... But since we really want to finish the thing, I guess you are right.

core/src/main/java/org/ehcache/core/EhcacheWithLoaderWriter.java
```diff
@@ -118,40 +119,19 @@ public EhcacheWithLoaderWriter(CacheConfiguration<K, V> configuration, Store<K,
...
-   */
   @Override
-  public void put(final K key, final V value) throws CacheWritingException {
-    putObserver.begin();
-    statusTransitioner.checkAvailable();
-    checkNonNull(key, value);
-
-    final BiFunction<K, V, V> remappingFunction = memoize((key1, previousValue) -> {
+  public Store.PutStatus doPut(K key, V value) throws StoreAccessException {
+    BiFunction<K, V, V> remappingFunction = (key1, previousValue) -> {
```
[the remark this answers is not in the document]
henri-tremblay: As discussed, it's fine as is.

henri-tremblay:
@lorban Can you confirm that the `RethrowingStoreAccessException` isn't useful anymore?

chrisdennis:
```
> Task :api:checkstyleMain FAILED
[ant:checkstyle] /Users/cdennis/src/ehcache3/api/src/main/java/org/ehcache/resilience/ResilienceStrategy.java:0: Missing package-info.java file.
```

PR #2259, opened 2018-01-29
Issue #2197 - lease integration test

AbfrmBlr:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -69,18 +70,16 @@
   private final ReconnectMessageCodec reconnectMessageCodec = new ReconnectMessageCodec();
   private final Map<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>> responseListeners =
     new ConcurrentHashMap<>();
+  private final List<DisconnectionListener> disconnectionListeners = new ArrayList<>();
```
henri-tremblay: No concurrency possible on adding listeners? Additions are done from the same thread using the listener? (otherwise, you have visibility issues here)

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
```diff
@@ -120,11 +121,7 @@ public void testCacheOperationThrowsAfterConnectionClosed() throws Exception {
 
     connection.close();
 
-    try {
-      cache.get(1L);
-    } catch (Exception e) {
-      assertThat(e.getCause().getCause(), instanceOf(ConnectionClosedException.class));
-    }
+    assertThat(cache.get(1L), nullValue());
```
henri-tremblay: Why null now? If the connection is closed explicitly by the client, I don't think it should rely on the resilience strategy (if that's the case here)

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    Cache<Long, String> cache = cacheManager.createCache("clustered-cache", config);
+    cache.put(1L, "The one");
+    cache.put(2L, "The two");
+    cache.put(3L, "The three");
+    assertThat(cache.get(1L), equalTo("The one"));
+    assertThat(cache.get(2L), equalTo("The two"));
+    assertThat(cache.get(3L), equalTo("The three"));
+
+    setDelay(6000);
```
henri-tremblay: 6 seconds is too long to wait in a test. Can't be make the lease expiring faster? (or everything to rely on a time source)

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -0,0 +1,199 @@
...
+    Thread.sleep(6000);
+    // We will now have lost the lease
+
+    setDelay(0L);
+
+    assertThat(cache.get(1L), is(nullValue()));
```
henri-tremblay: I was thinking response time here. We should make sure it is fast (not in a performance way, in a "I was not waiting for a timeout" way)

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -0,0 +1,199 @@
...
+    return createURI(proxyPorts);
+  }
+
+  private List<Integer> parsePorts(URI connectionURI) {
+    String uriString = connectionURI.toString();
+    String withoutProtocol = uriString.substring(13);
+    String[] stripes = withoutProtocol.split(STRIPE_SEPARATOR);
+
+    List<Integer> ports = new ArrayList<>(stripes.length);
+    for (String stripe : stripes) {
```
henri-tremblay: Arrays.stream(stripes).map(stripe -> stripe.substring(stripe.indexOf(":") + 1)).mapToInt(Integer::parseInt).collector(Collectors::toList());

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    return ports;
+  }
+
+  private List<Integer> createProxyPorts(int portCount) {
+    PortChooser portChooser = new PortChooser();
+    int firstProxyPort = portChooser.chooseRandomPorts(portCount);
+
+    List<Integer> proxyPorts = new ArrayList<>(portCount);
+    for (int i = 0; i < portCount; i++) {
```
henri-tremblay: IntStream.range(portCount).map(i -> firstProxyPort + i).collector(Collectors::toList())

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -0,0 +1,199 @@
...
+      proxyPorts.add(firstProxyPort + i);
+    }
+
+    return proxyPorts;
+  }
+
+  private URI createURI(List<Integer> proxyPorts) {
+    List<String> stripes = new ArrayList<>();
+
+    for (int proxyPort : proxyPorts) {
```
henri-tremblay: `proxyPorts.stream().map(proxyPort -> "localhost:" + proxyPort).join("terracotta://", STRIPE_SEPARATOR)`  or something similar

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -0,0 +1,199 @@
...
+      String stripe = "localhost:" + proxyPort;
+      stripes.add(stripe);
+    }
+
+    String uri = "terracotta://" + String.join(STRIPE_SEPARATOR, stripes);
+    return URI.create(uri);
+  }
+}
```
henri-tremblay: Add the newline

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
AbfrmBlr: No since these all are one time (at the time of creation only), which is under lock

AbfrmBlr:
Done

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/ConnectionClosedTest.java
henri-tremblay: Was explained. It's the connection that is closed, not the CacheManager. So all good.

PR #2261, opened 2018-01-30
I 2258 : Improve passive sync performance

rkavanap:


rkavanap:
@AbfrmBlr , @chrisdennis.. Can both of you take a look at this PR..? thanks..

rkavanap:
@henri-tremblay Sure...here goes..

I am getting a 3X performance in sync (from 45 minutes to 15 minutes) with these changes with a cache with millions of small entries.

The problem of sequencing building of the sync message (which takes a total of 20 seconds per concurrency) with a send is that the send is a blocking call (and it gets worse as the blocking lasts till the passive processes the message as the same thread send the rcv ack for a given concurrency) and sequencing the building of the passive sync message with the send adds up when the cache has millions of small records.

Ofcourse the bigger help is from the passive sync put (chainmap changes) where the append in a loop (which closes and opens the chain each time) q=was slow,.the change doubles the put performance. This was the biggest bottleneck in passive sync performance.

Btw, This is just a single thread from the pool. But maybe it makes more sense to provide our own threadpool. Suggestions welcome.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -113,6 +116,9 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(ClusterTierActiveEntity.class);
   static final String SYNC_DATA_SIZE_PROP = "ehcache.sync.data.size.threshold";
   private static final long DEFAULT_SYNC_DATA_SIZE_THRESHOLD = 4 * 1024 * 1024;
```
henri-tremblay: Why are these properties  java properties instead of normal Ehcache configuration ones?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -133,6 +139,8 @@
   private volatile List<InvalidationTuple> inflightInvalidations;
   private final Set<ClientDescriptor> connectedClients = ConcurrentHashMap.newKeySet();
   private final int chainCompactionLimit;
+  private volatile Long dataSizeThreshold = null;
```
henri-tremblay: This could be a primitive type initialized right away. `private final long dataSizeThreshold = Long.getLong(SYNC_DATA_SIZE_PROP, DEFAULT_SYNC_DATA_SIZE_THRESHOLD);`

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +594,74 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
-          for (Element element : chain) {
-            size.addAndGet(element.getPayload().remaining());
-          }
-          mappingsToSend.get().put(key, chain);
-          if (size.get() > dataSizeThreshold) {
-            syncChannel.synchronizeToPassive(new EhcacheDataSyncMessage(mappingsToSend.get()));
-            mappingsToSend.set(new HashMap<>());
-            size.set(0);
+      CompletableFuture<Void> asyncGet = CompletableFuture.runAsync(() -> doGetsForSync(segmentId, messageQ));
+      try {
```
henri-tremblay: I will let others comment but this will stuck a thread of the common pool for a bit of time. Also, Ehcache configuration is defining thread-pools. Using them would make sense I think.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +594,74 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
           }
-        });
-      if (!mappingsToSend.get().isEmpty()) {
-        syncChannel.synchronizeToPassive(new EhcacheDataSyncMessage(mappingsToSend.get()));
-        mappingsToSend.set(new HashMap<>());
-        size.set(0);
+        }
+        asyncGet.get();
+      } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException(e);
```
henri-tremblay: Not a fan of the interruption management. I would interrupt the loop instead, and tell the other thread so stop the task ideally.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +594,74 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
-            syncChannel.synchronizeToPassive(new EhcacheDataSyncMessage(mappingsToSend.get()));
-            mappingsToSend.set(new HashMap<>());
-            size.set(0);
+      CompletableFuture<Void> asyncGet = CompletableFuture.runAsync(() -> doGetsForSync(segmentId, messageQ));
+      try {
+        boolean done = false;
+        while (!done) {
+          EhcacheDataSyncMessage msg = messageQ.take();
+          if (msg instanceof LastDataSyncMessage) {
+            done = true;
```
henri-tremblay: No real need for the flag. Just break would be enough

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +594,74 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+  private void doGetsForSync(int segmentId, BlockingQueue<EhcacheDataSyncMessage> messageQ) {
+    if (dataSizeThreshold == null) {
+      // do it only once
+      dataSizeThreshold = Long.getLong(SYNC_DATA_SIZE_PROP, DEFAULT_SYNC_DATA_SIZE_THRESHOLD);
+      dataGetsThreshold = Integer.getInteger(SYNC_DATA_GETS_PROP, DEFAULT_SYNC_DATA_GETS_THRESHOLD);
+    }
+    ServerSideServerStore store = stateService.getStore(storeIdentifier);
+    int numGets = 0;
+    long size = 0;
+    Map<Long, Chain> mappingsToSend = new HashMap<>();
```
henri-tremblay: Should the map be sized to `dataGetsThreshold`?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +594,74 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+          size = 0;
+          numGets = 0;
+        }
+      }
+      if (!mappingsToSend.isEmpty()) {
+        messageQ.put(new EhcacheDataSyncMessage(mappingsToSend));
+      }
+      messageQ.put(new LastDataSyncMessage());
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
```
henri-tremblay: From my point of view, the exception will probably be caught by the common pool and disappear. The interruption should be handled to get out nicely. And throwables should be logged and rethrown I guess.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +594,74 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+    }
+  }
+
+  private static class LastDataSyncMessage extends EhcacheDataSyncMessage {
+    private LastDataSyncMessage() {
```
henri-tremblay: Since it's stateless, we could have a constant for this. No need to instantiate a `LastDataSyncMessage` every time

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: I am guessing we do not want users to play around with these. However keeping it as system properties will give us an opportunity to tune/study the performance. That is my best guess.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: that would cause unnecessary memory usage for very small caches, won't it?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: The current code reads it late. So I wanted to keep it that way (in case we allow system propeties to be set runtime through JMX etc).. I fixed a bug in that area so that I re-read the system property on start of every sync (i.e concurrency key 0).

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: Not really is the number of mappings to send is always `numGets`. Or size is generally reached first?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: I think it would be super duper overkill if that was the original intent.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: Why can't we study performance with a clustered configuration?

rkavanap:
@henri-tremblay , I have updated with review comments,
   For exception and interrupt handling; I have ensured that the sending thread comes out of take and sees the exception of the getting thread. I am not restoring the interrupt status for threads since we are throwing a RuntimeException().. Since this is potentially a very long running operation, it is also not possible to mask the interrupt and forcefully complete the operation before preserving the interrupt status of the thread. Infact the terracotta core itself is internally throwing a runtime exception for syncChannel.synchronizeToPassive is interrupted..So our behaviour is consistent with that.

   For threadpool, I created a single fixed thread pool that is passed to the async method. Later when parallel sync can happen from OSS we can increase the size of that thread pool. As for using threadpool configuration, is any ehcache xml configuration regarding threadpools reaching the server?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: Fair enough then

henri-tremblay:
@rkavanap No. The thread pool isn't available on the server. So I was just stupid.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
chrisdennis: Using interrupts to signal here seems fragile to me too.  I'm thinking we could maybe restructure this asas a SynchronousQueue of functional types.  Then we can push all the sync logic on to one side of the queue and have the consumer just be:

```java
while(synchronousQueue.take().execute());
```

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
chrisdennis: It's not using the common pool though is it?  It's using the `SYNC_GETS_EXECUTOR`.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -112,9 +118,16 @@
...
+  static final String SYNC_DATA_GETS_PROP = "ehcache.sync.data.gets.threshold";
+  private static final int DEFAULT_SYNC_DATA_GETS_THRESHOLD = 16 * 1024;
+
   static final String CHAIN_COMPACTION_THRESHOLD_PROP = "ehcache.server.chain.compaction.threshold";
-  static final int DEFAULT_CHAIN_COMPACTION_THRESHOLD = 8;
+  private static final int DEFAULT_CHAIN_COMPACTION_THRESHOLD = 8;
+
+  private static final LastDataSyncMessage SYNC_MESSAGE_COMPLETION_MARKER = new LastDataSyncMessage();
+  private static final int MAX_SYNC_CONCURRENCY = 1;
+  private static final Executor SYNC_GETS_EXECUTOR = Executors.newFixedThreadPool(MAX_SYNC_CONCURRENCY);
```
chrisdennis: This thread pool is going to keep it's threads alive forever.  That seems wasteful, why can't we have the threads alive only when we're passive syncing?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: Yes. It's not anymore. The code changed since my comment

chrisdennis: My bad, sorry.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: @henri-tremblay Actually you have a point. I did some digging and found out that the keySet size() operation is not expensive at all. So I will upload a new commit soon.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: @chrisdennis , Currently there is no startSync() interface on the active that tells that a sync is beginning. 

I can create the threadpool when segment 0 is syncing, but that would cause issues if there is concurrent syncing, where I will have to make the threadpool creation synchronous. Pls advise on which is better.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: Already done..

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -201,14 +205,7 @@ public int length() {
...
 
-  private static class PrimordialChain implements InternalChain {
-
-    private final ByteBuffer element;
-
-    public PrimordialChain(ByteBuffer element) {
-      this.element = element;
-    }
-
+  private static abstract class Rudimentary implements InternalChain {
```
henri-tremblay: Rudimentary is an adjective. So it is weird for a class name. But I have now clue of the intend to help find a better word. Can you javadoc it to explain what it is?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -468,18 +492,19 @@ private Long createAttachedChain(ByteBuffer keyBuffer, int hash, ByteBuffer elem
...
     }
 
-    try (AttachedInternalChain chain = new AttachedInternalChain(address)) {
-      while (iterator.hasNext()) {
-        if (!chain.append(iterator.next().getPayload())) {
-          chain.free();
-          return null;
+    if (iterator.hasNext()) {
+      try (AttachedInternalChain chain = new AttachedInternalChain(address)) {
+        while (iterator.hasNext()) {
```
henri-tremblay: Can be a `do {} while` since the hasNext is already done

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: I might be naive but why no do this?

```
      ExecutorService executor = Executors.newSingleThreadExecutor();

      try {
        CompletableFuture<Void> asyncGet = CompletableFuture.runAsync(
          () -> doGetsForSync(segmentId, messageQ, Thread.currentThread()), executor);

        while (true) {
          try {
            EhcacheDataSyncMessage msg = messageQ.take();
            if (msg instanceof LastDataSyncMessage) {
              break;
            } else {
              syncChannel.synchronizeToPassive(msg);
            }
            interrupted = true;
            break;
          }
        }

        if (interrupted) {
          // get is finite memory operation..give it time to complete even if async get task is not interrupted
          // for some reason so that we can get the actual cause of exception.
          asyncGet.get(10, TimeUnit.SECONDS);
          // the get threads actual execution status precedes our own interruption status
          throw new InterruptedException();
        } else {
          asyncGet.get();
        }
      } catch (InterruptedException | ExecutionException | TimeoutException e) {
      } finally {
        executor.shutdownNow();
      }
    }
```

rkavanap: @henri-tremblay This will create a thread executor for every concurrency. .. That is 16 times per concurrency

Terracotta core calls this method 16 times during passive sync..Once for every concurrency..Also I am trying to push for the core to do atleast 4 concurrency in parallel for nearly another 3X additional performance. 

Is it ok to create a thread pool executor for each concurrency and shut it down?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
rkavanap: I will change it. I was also not completely happy with that. Will RudimentaryObject sound better?

chrisdennis: It's a chain. Maybe the name should have chain in it somewhere?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
chrisdennis: Just configure the thread pool to have a finite keep alive time.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
rkavanap: I RudimentaryChain and RudimentaryElement extending from Rudimentary 
  Does GenesisChain,  GenesisComplexChain and GenesisSimpleChain sound better?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +603,129 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+      throw new RuntimeException(e);
+    } finally {
+      if (!success) {
+        // ensure waiting peer thread is interrupted, if we run into trouble
+        waitingThread.interrupt();
+      }
+    }
+  }
+
+  private int computeMapCapacity(int remainingSize, int expectedGets) {
```
henri-tremblay: Can you comment on why you did this split or sizing?

rkavanap: @henri-tremblay Not sure I fully understand the question..
    
   Anyway, the logic here is simple: 
               remainingSize is the number of keys left in the chainmap yet to be send
               expectedGets is the max number of items that could be put in a single hashmap
both varies depending on the size of the chains and number of keys. For e.g I may start with an initial size of configured max gets and then soon discover that the size of 2MB is reached for this particular cache before it even reaches 75% of max gets. This will automatically adjust the hashmap capacity for the next iteration. I also re-update the initial size so that for the next concurrency, the initial capacity matches the type of cache  (e.g more data and less keys) we have. 

Ofcourse the logic assumes some sort of symmetry at the concurrency level,, but it is possible that some keys has chain with LARGE data and some keys has small. But on a larger sweep that should even out.. even if it is not, this should perform better as the sizes are reset back to a minimum of 32 and not 16 when a cache is large and when a cache is very small it starts with initial size of 16.

henri-tremblay: Can you write that in a comment in the code? :-)

rkavanap: sure.. I will put a method header for the compute* methods..

rkavanap:
test this please

Looks like a random failure.

rkavanap:
@henri-tremblay , can you merge if found OK? Updated all review comments..

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -201,14 +205,7 @@ public int length() {
...
 
-  private static class PrimordialChain implements InternalChain {
-
-    private final ByteBuffer element;
-
-    public PrimordialChain(ByteBuffer element) {
-      this.element = element;
-    }
-
+  private static abstract class GenesisChain implements InternalChain {
```
henri-tremblay: Still no comment about what it is

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -228,6 +225,34 @@ public boolean replace(Chain expected, Chain replacement) {
     public void close() {
       //no-op
     }
+
+    protected abstract Iterator<Element> iterator();
+  }
+
+  private static class GenesisLink extends GenesisChain {
```
henri-tremblay: And the link as well

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +603,168 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+      }
+      if (!mappingsToSend.isEmpty()) {
+        putMessage(messageQ, syncChannel, mappingsToSend);
+      }
+      // put the last message indicator into the queue
+      putMessage(messageQ, null, null);
+    } catch (Throwable e) {
+      // ensure waiting peer thread is interrupted, if we run into trouble
+      waitingThread.interrupt();
+      // rethro
```
henri-tremblay: w

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierActiveEntityTest.java
```diff
@@ -877,6 +877,7 @@ public void testSyncToPassiveBatchedByDefault() throws Exception {
 
   @Test
   public void testDataSyncToPassiveCustomBatchSize() throws Exception {
+    System.setProperty(ClusterTierActiveEntity.SYNC_DATA_SIZE_PROP, "512");
```
henri-tremblay: You need to unset this in a `finally` otherwise it will stick for ever after

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,40 +603,168 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+      } catch (InterruptedException | ExecutionException | TimeoutException e) {
+        throw new RuntimeException(e);
       }
     }
     sendMessageTrackerReplication(syncChannel, concurrencyKey - 1);
 
     LOGGER.info("Sync complete for concurrency key {}.", concurrencyKey);
   }
 
+  private void doGetsForSync(int segmentId, BlockingQueue<DataSyncMessageHandler> messageQ,
```
henri-tremblay: Can you please add a unit test for the interruption and exception mechanism?

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierActiveEntityTest.java
rkavanap: Good find. This was always like that. I just moved this statement a few lines above.

But it is a must fix.. I will do it,

rkavanap: @henri-tremblay Will fix this..

Actually the clear property was already there at the end of the test..So it is actually not all that bad.  It was my bad that I did not move the try { when I moved the set property up, But behaviour wise was everything looks ok and the other tests were running without this property 

Anyway, the key point is tests was running properly all this time as the clearProperty was there at the end.

rkavanap:
@henri-tremblay, Can you please take a final look and merge?

henri-tremblay:
@rkavanap I've approved. @chrisdennis needs to approve as well to merge.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,38 +603,166 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+      BlockingQueue<DataSyncMessageHandler> messageQ = new SynchronousQueue<>();
       int segmentId = concurrencyKey - DEFAULT_KEY - 1;
-      Long dataSizeThreshold = Long.getLong(SYNC_DATA_SIZE_PROP, DEFAULT_SYNC_DATA_SIZE_THRESHOLD);
-      AtomicLong size = new AtomicLong(0);
+      Thread thisThread = Thread.currentThread();
+      CompletableFuture<Void> asyncGet = CompletableFuture.runAsync(
+        () -> doGetsForSync(segmentId, messageQ, syncChannel, thisThread), SYNC_GETS_EXECUTOR);
+      try {
+        try {
+          while (messageQ.take().execute()) ;
```
chrisdennis: I was hoping that moving to this model would allow you to eliminate most of the interrupt logic. If the async get thread hits an issue it can just push a task to the queue that throws the same exception, thereby causing the termination of the writer thread.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -586,38 +603,166 @@ public void synchronizeKeyToPassive(PassiveSynchronizationChannel<EhcacheEntityM
...
+    } else {
+      return dataMapInitialCapacity;
+    }
+  }
+
+  /**
+   * Executes message sending asynchronously to preparation of message.
+   */
+  @FunctionalInterface
+  private interface DataSyncMessageHandler {
```
chrisdennis: Can't we just use a Callable<Boolean>?

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: @chrisdennis , yes I did get your intention. But since it is a synchronousQueue (where the put succeeds only if other side is waiting on a take - the other side could be in the synchronizeToPassive), it becomes hard to do that. Also to add to the complexity, I have to catch the interrupt exception again inside throwable. 

May have to use some other queue other than a synchronous queue to cleanly do that, which brings other complexities.

Thought it was simpler to still use synchronous queue (as it is the simplest way to do a double buffering mechanism). The error handling works as the test I added suggests. 

Besides any error here is highly unlikely and results in a RuntimeException which results in a server shutdown.

Again suggestions welcome, as I might have missed something.

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
rkavanap: Yes..Will do that.

chrisdennis: Up to you... this is a minor nit for me.

PR #2267, opened 2018-02-01
Do not log too much when on the resilience strategy

henri-tremblay:


core/src/main/java/org/ehcache/core/internal/util/Pacer.java
```diff
@@ -0,0 +1,62 @@
...
+package org.ehcache.core.internal.util;
+
+import org.ehcache.core.spi.time.TimeSource;
+
+/**
+ * Class used to pace a call to prevent it from happening too frequently.
+ */
+public class Pacer {
+
+  private volatile long nextLogTime = 0;
```
chrisdennis: Right now we're going to get a flurry of calls whenever the threshold is reached if we're multithreaded.  Should this be an AtomicLong?

chrisdennis: Also... I think this should be initialized to `TimeSource.getTimeMillis()` otherwise we will be broken with impl's that return negative timestamps.

core/src/main/java/org/ehcache/core/internal/util/Pacer.java
```diff
@@ -0,0 +1,62 @@
...
+  private final TimeSource timeSource;
+  private final long delay;
+
+  /**
+   * Unique constructor
+   *
+   * @param delay delay between each call paced
+   */
+  public Pacer(TimeSource timeSource, long delay) {
```
chrisdennis: Should we be 310-ing it up in here?

core/src/main/java/org/ehcache/core/internal/util/Pacer.java
henri-tremblay: Ok for negative timestamps.

henri-tremblay: I'm not sure the AtomicLong is that necessary since it doesn't matter that much to have many calls when a delay expires but I will pull out a stricter version. Tell me what you think.

core/src/main/java/org/ehcache/core/internal/util/Pacer.java
henri-tremblay: Since it is internal stuff, I don't think it's needed

core/src/main/java/org/ehcache/core/internal/util/Pacer.java
```diff
@@ -0,0 +1,65 @@
...
+    this.timeSource = timeSource;
+    this.delay = delay;
+    this.nextLogTime = new AtomicLong(timeSource.getTimeMillis());
```
chrisdennis: Pedantic me is wondering if this should actually be:
```java
this.nextLogTime = new AtomicLong(timeSource.getTimeMillis());
```
This way even a "static" time source that never changes will log the first call.

PR #2268, opened 2018-02-01
Resilience documentation

henri-tremblay:


docs/build.gradle
```diff
@@ -14,17 +14,13 @@
...
-  dependencies {
-    classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.2'
-  }
+plugins {
+  id 'org.asciidoctor.convert' version '1.5.7'
+  id 'org.kordamp.gradle.livereload' version '0.2.1'
 }
 
 apply plugin: 'org.asciidoctor.convert'
+apply plugin: 'org.kordamp.gradle.livereload'
```
chrisdennis: Shouldn't we be using either the plugins dsl or the `apply plugin` stanzas not both?

docs/build.gradle
```diff
@@ -57,3 +53,8 @@ asciidoctor {
     }
   }
 }
+
+liveReload {
+  docRoot asciidoctor.outputDir.canonicalPath
```
chrisdennis: Why `.canonicalPath`?

docs/README.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+== Live reload
+
+To get live reload while editing the documentation, you need two shells.
+
+In the first one, do a continuous buid of the documentation.
+
+  ./gradlew -t :docs:asciidoctor
+
+In the second launch the live reload.
```
chrisdennis: I'm not convinced that the live reload stuff belongs in the buildscript itself.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+Timeouts can be configured using a dedicated builder or in XML.
+
+[source%nowrap,java,indent=0]
+----
+include::{sourcedir34}/clustered/client/src/test/java/org/ehcache/clustered/client/docs/Resilience.java[tag=timeoutsExample]
+----
+
+<1> Start setting timeouts using the build
+<2> Set a read timeout to 10 seconds
```
chrisdennis: 'a read timeout of' or 'the read timeout to'.  Otherwise it sounds like there are multiple read timeouts.

docs/src/docs/asciidoc/user/resilience.adoc
```diff
@@ -0,0 +1,96 @@
...
+A client establishes a lease with the server. A lease is a contract between the client and the server.
+It means the server can't make a change without getting the client acknowledgment.
+
+The client takes care of renewing the lease before it expires.
+The connection is closed by one of the two parties if it fails to do so.
+As soon as this happens, link:tiering{outfilesuffix}[caching tiers] are cleared.
+The resilience strategy will then start answering every call.
+
+By default, a lease last 150 seconds.
+It is decided by the server and can be overridden in the server configuration.
```
chrisdennis: I think this is the first time we've had to do this (which I don't like).  I wonder if referencing the relevant documentation will be possible.

docs/README.adoc
henri-tremblay: Of course it does. It's nicely configured and not doing anything unless the task is launched.

docs/build.gradle
henri-tremblay: I don't know. I followed the official example https://github.com/asciidoctor/asciidoctor-gradle-examples/blob/master/asciidoc-to-html-livereload-example/build.gradle

docs/build.gradle
henri-tremblay: Same answer, it's in the example

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: ok

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: What do you mean? To talk about the server configuration? I don't know

docs/build.gradle
chrisdennis: I'm fairly confident that the official example is wrong... could you test with one or the other and see?

docs/build.gradle
chrisdennis: Fair enough - it looks like they're using strings for file paths rather than file (or path) objects.

docs/src/docs/asciidoc/user/resilience.adoc
chrisdennis: I mean I wonder if it would be possible to give a link to where the server configuration is documented.

docs/build.gradle
henri-tremblay: You are right. It seems that `plugins {}` will declare an apply the plugin at the same time. See https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block. Not complicated at all... (sight)

docs/src/docs/asciidoc/user/resilience.adoc
henri-tremblay: It would be better. But where is it? If it's nowhere, I think that's the best I can do right now. Or you meant here? http://www.ehcache.org/documentation/3.4/clustered-cache.html

chrisdennis: I'm more thinking long term this is something we should fix... don;t think there is anything we can do now to make this better.

Issue #2203, opened 2017-12-05
Implement reconciliation client-server after failure


henri-tremblay:
When a client manage to reconnect to a server, it should reconcile with it.

It sends a "what did I miss?" message and will receive a list of entries to invalidate on the caching tier.

Pending updates on the client are sent to the server and applied if not obsolete and will be processed if not obsolete.

AbfrmBlr:
When clients get a disconnect event, we already invalidate everything on the caching tier and the connection is closed, which means all the pending updates would be lost.
Infact in some cases, the requests waiting for responses might end up having a successful change in state on the server, but client never managed to see the response. 
In those cases, client will get a StoreAccessException, but the operation might have been successful.
Connecting back actually means a fresh client server for the server. where client has not state and is starting from scratch(as good as that.)

henri-tremblay:
I don't think this should be closed. The idea was really to performa reconciliation. That it can't be done right now is another problem.

PR #2278, opened 2018-02-07
Issue #2203 - Reconnection

AbfrmBlr:
PR is still not complete, but good to get comments.
Missing parts:-
- ~~reconnection during cachemanager lifecycle~~
- of course more tests


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectInProgressException.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.client.internal.store;
+
+public class ReconnectInProgressException extends RuntimeException {
+
+  private static final String STRING = "Connection lost to server. Client it trying to reconnect to server";
```
henri-tremblay: No need to have a constant for the String. It will be interned anyway.

AbfrmBlr:
test this please

AbfrmBlr:
Depends on https://github.com/Terracotta-OSS/terracotta-core/pull/846

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -0,0 +1,260 @@
...
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeoutException;
+
+class ConnectionState {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ConnectionState.class);
+
+  private static final String CONNECTION_PREFIX = "Ehcache:";
+
+  private volatile Connection clusterConnection = null;
```
chrisdennis: It seem weird that these things are volatile and the `ConnectionState` reference inside `DefaultClusteringService` was final. I had expected things to be the other way round.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectHandle.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.client.internal.store;
+
+public interface ReconnectHandle {
```
chrisdennis: Should we just use `Runnable` here?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectingServerStoreProxy.java
```diff
@@ -0,0 +1,154 @@
...
+  private final ServerStoreProxy swap;
+  private final ServerStoreProxy delegate;
+  private final AtomicReference<ServerStoreProxy> delegateRef;
+  private final Runnable onReconnect;
+
+  public ReconnectingServerStoreProxy(ServerStoreProxy serverStoreProxy, Runnable onReconnect) {
+    this.delegate = serverStoreProxy;
+    this.delegateRef = new AtomicReference<>(serverStoreProxy);
+    this.onReconnect = onReconnect;
+    this.swap = new SwappableProxy(serverStoreProxy.getCacheId());
```
chrisdennis: Bunch of things seem weird here.

1. Why do we need a create this thing now?  Why not just do it on demand?
2. Why is it called SwappableProxy, isn't it really `ReconnectInProgressProxy`?
3. Lots of code duplication here.

Wondering if we can solve all this with a method something like this:
```java
  private <T> T onStoreProxy(Function<ServerStoreProxy, T> function) {
    ServerStoreProxy storeProxy = proxy();
    try {
      return function.apply(storeProxy);
    } catch (ServerStoreProxyException sspe) {
      if (sspe.getCause() instanceof ConnectionClosedException) {
        if (delegateRef.compareAndSet(storeProxy, new ReconnectInProgress(storeProxy.getCacheId()))) {
          onReconnect.run();
        }
        return onStoreProxy(function);
      } else {
        throw sspe;
      }
    }
  }
```

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -543,7 +547,9 @@ public void setInvalidationValve(InvalidationValve valve) {
 
     private volatile ServiceProvider<Service> serviceProvider;
     private volatile ClusteringService clusteringService;
+    private volatile ExecutionService executionService;
 
+    private final Lock reconnectLock = new ReentrantLock();
```
chrisdennis: Isn't this actually a `connectLock`?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
AbfrmBlr: I know its weird, need to put more though there

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectingServerStoreProxy.java
AbfrmBlr: This is cool

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectHandle.java
AbfrmBlr: Oh it is still lingering around here, I removed it though. :confused:

AbfrmBlr:
Its complete

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
```diff
@@ -40,12 +43,30 @@
...
+      }
+
+      @Override
+      public byte[] createExtendedReconnectData() {
+        return new byte[0];
+      }
+
+      @Override
+      public void didDisconnectUnexpectedly() {
+        LOGGER.info("CacheManager got disconnected from server");
```
chrisdennis: Could we maybe get some identifying information here?  Maybe if we `toString()` the endpoint?  That should know the identity of the entity.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -78,23 +65,20 @@
...
   private final ConcurrentMap<String, ClusteredSpace> knownPersistenceSpaces = new ConcurrentHashMap<>();
-  private final Timeouts timeouts;
-  private final Properties properties;
+  private final ConnectionState connectionState;
 
-  private volatile Connection clusterConnection;
-  private ClusterTierManagerClientEntityFactory entityFactory;
-  private ClusterTierManagerClientEntity entity;
-  private final ConcurrentMap<String, ClusterTierClientEntity> clusterTierEntities = new ConcurrentHashMap<>();
+  private final Set<String> reconnectSet = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
chrisdennis: It feels like the `reconnectSet` is probably a property of a `ConnectionState` - it's the set of caches that the 'previous' connection had retrieved. In that way the new `ConnectionState` would reconnect using the previous state as a seed.  Not something to fix now but something to consider when looking at cleaning this up.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ReconnectingServerStoreProxyTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Rule
+  public ExpectedException exception = ExpectedException.none();
+
+  private final ServerStoreProxyException storeProxyException = new ServerStoreProxyException(new ConnectionClosedException("Connection Closed"));
+
+  private ReconnectingServerStoreProxy serverStoreProxy;
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
```
henri-tremblay: If you want, you can use a MockitoRule and an `@InjectMocks` to do constructor injection. It will allow to remove the entire `setUp()`

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -124,73 +120,30 @@ public void leaseExpiry() throws Exception {
     assertThat(cache.get(2L), equalTo("The two"));
     assertThat(cache.get(3L), equalTo("The three"));
 
-    setDelay(2000);
-    Thread.sleep(2000);
+    setDelay(6000, proxies);
+    Thread.sleep(6000);
```
henri-tremblay: Is there anything you can do to remove/recude this long sleep?

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -124,73 +120,30 @@ public void leaseExpiry() throws Exception {
...
 
-    assertThat(cache.get(1L), is(nullValue()));
-    assertThat(cache.get(2L), is(nullValue()));
-    assertThat(cache.get(3L), is(nullValue()));
+    CompletableFuture<String> future = CompletableFuture.supplyAsync(() -> {
+      while (true) {
+        try {
+          Thread.sleep(200);
+        } catch (InterruptedException e) {
+          //
```
henri-tremblay: You probably shoud rethrow the interruption

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
```diff
@@ -124,73 +120,30 @@ public void leaseExpiry() throws Exception {
...
     // We will now have lost the lease
 
-    setDelay(0L);
+    setDelay(0L, proxies);
 
-    assertThat(cache.get(1L), is(nullValue()));
-    assertThat(cache.get(2L), is(nullValue()));
-    assertThat(cache.get(3L), is(nullValue()));
+    CompletableFuture<String> future = CompletableFuture.supplyAsync(() -> {
+      while (true) {
```
henri-tremblay: In case of error, we need to be able to get out of the loop at some point. Like after 20 seconds or something.

clustered/client/src/main/java/org/ehcache/clustered/client/service/ClusteringService.java
```diff
@@ -59,8 +59,9 @@
    * Releases access to a {@link ServerStoreProxy} and the server-resident {@code ServerStore} it represents.
    *
    * @param serverStoreProxy a {@link ServerStoreProxy} obtained through {@link #getServerStoreProxy}
+   * @param isReconnect whether client is trying to reconnect
    */
-  void releaseServerStoreProxy(ServerStoreProxy serverStoreProxy);
+  void releaseServerStoreProxy(ServerStoreProxy serverStoreProxy, boolean isReconnect);
```
chrisdennis: Not a massive fan of this change... its bleeding a internal detail.  I'll swallow it for the sake of expediency though.

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/BasicCacheReconnectTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    Cache<Long, String> cache = cacheManager.createCache("clustered-cache", config);
+
+    CompletableFuture<Void> future = CompletableFuture.runAsync(() ->
+            ThreadLocalRandom.current()
+            .longs()
+            .forEach(value ->
+                    cache.put(value, Long.toString(value))));
+
+    setDelay(6000, proxies);
```
henri-tremblay: More delay I would like to remove

clustered/integration-test/src/test/java/org/ehcache/clustered/util/TCPProxyUtil.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.net.InetAddress;
+import java.net.URI;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class TCPProxyUtil {
```
henri-tremblay: Can be final and have a private constructor

clustered/integration-test/src/test/java/org/ehcache/clustered/util/TCPProxyUtil.java
```diff
@@ -0,0 +1,89 @@
...
+      InetAddress host = InetAddress.getByName("localhost");
+      TCPProxy proxy = new TCPProxy(proxyPort, host, port, 0L, false, null);
+      proxies.add(proxy);
+      proxy.start();
+    }
+
+    return createURI(proxyPorts);
+  }
+
+  private static List<Integer> parsePorts(URI connectionURI) {
```
henri-tremblay: That kind of parsing isn't done anywhere else?

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/ThrowingResiliencyStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public boolean containsKeyFailure(K key, StoreAccessException e) {
+    throw new RuntimeException(e);
+  }
+
+  @Override
+  public void putFailure(K key, V value, StoreAccessException e) {
+    if (e.getCause() instanceof ReconnectInProgressException) {
```
henri-tremblay: Why not throwing for other things? And something puzzles me. When is the `ReconnectInProgressException` thrown? I would expect it to be transparent to the user

clustered/integration-test/src/test/java/org/ehcache/clustered/util/TCPProxyUtil.java
```diff
@@ -0,0 +1,89 @@
...
+            .mapToInt(Integer::parseInt)
+            .boxed()
+            .collect(Collectors.toList());
+  }
+
+  private static List<Integer> createProxyPorts(int portCount) {
+    PortChooser portChooser = new PortChooser();
+    int firstProxyPort = portChooser.chooseRandomPorts(portCount);
+
+    return IntStream
```
henri-tremblay: An alternative to the ugly boxed and collect is `toArray()` which returns `int[]`

clustered/integration-test/src/test/java/org/ehcache/clustered/util/TCPProxyUtil.java
```diff
@@ -0,0 +1,89 @@
...
+            .map(i -> firstProxyPort + i)
+            .boxed()
+            .collect(Collectors.toList());
+  }
+
+  private static URI createURI(List<Integer> proxyPorts) {
+
+    Optional<String> reduce = proxyPorts.stream()
+            .map(port -> "localhost:" + port)
+            .reduce((x1, x2) -> x1 + STRIPE_SEPARATOR + x2);
```
henri-tremblay: You can do this instead.

```
proxyPorts.stream()
            .map(port -> "localhost:" + port)
            .collect(Collectors.joining(",", "terracotta://", ""));
```

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/ThrowingResiliencyStrategy.java
AbfrmBlr: This is just for this test, not meant for user, The scope is limited for test which is only using put, and I want to only throw Reconnect exception, since some puts might also get timed out during the test run

clustered/integration-test/src/test/java/org/ehcache/clustered/util/TCPProxyUtil.java
AbfrmBlr: I don't think so, plus this refactored out from Lease test

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/BasicCacheReconnectTest.java
AbfrmBlr: Can't do that, since any time less than this does not leave enough time for reconnect to happen in test environment. I think we will have to live with this, These tests are here just for basic integration coverage

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/ThrowingResiliencyStrategy.java
henri-tremblay: I know. But can you comment the code about the intent? The thing that feels weird is that all other methods are throwing the exception whatever it is.

BTW, I did something similar in `DuplicateTest` by using a dynamic proxy.

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/ReconnectingServerStoreProxyTest.java
AbfrmBlr: cool

clustered/integration-test/src/test/java/org/ehcache/clustered/util/TCPProxyUtil.java
AbfrmBlr: cool

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
AbfrmBlr: This is test code, would it really matter

clustered/client/src/main/java/org/ehcache/clustered/client/internal/SimpleClusterTierManagerClientEntity.java
AbfrmBlr: Sadly endpoint does not override toString

clustered/integration-test/src/test/java/org/ehcache/clustered/LeaseTest.java
henri-tremblay: Yes it matters. Because you might get stuck in the loop if the test harness tries to interrupt you

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/BasicCacheReconnectTest.java
henri-tremblay: Fair enough. That's something we might want to address in TCServer

PR #2279, opened 2018-02-07
Upgrade to asm 6, jetty 9 and servlet API 3.1

henri-tremblay:


chrisdennis:
@lorban If I recall correctly you worked on the peeper demo.  Could you have a quick glance at these changes? Thanks.

demos/build.gradle
```diff
@@ -1,26 +1,18 @@
...
-      // It seems jetty has some internal conflict and so those need to be forced
-      force 'org.ow2.asm:asm:5.0.3', 'org.ow2.asm:asm-commons:5.0.3', 'org.glassfish:javax.el:3.0.1-b08'
-    }
-  }
-
   apply plugin: 'war'
   apply plugin: 'org.akhikhl.gretty'
 
   gretty {
-    port = 8080
```
chrisdennis: Is this now the default?  Was it not before?  Whats going on here?

docs/src/docs/asciidoc/user/examples.adoc
```diff
@@ -44,8 +44,7 @@ cd ehcache3/demos/00-NoCache
 ----
 
 This builds the necessary components, starts a http://eclipse.org/jetty/[Jetty] web service,
-and displays the URL of the web server on the console.  The URL will be something like
-+http://localhost:8080/ehcache-demos/00-NoCache/+.
+and displays the URL of the web server on the console. The URL should be +http://localhost:8080/+.
```
chrisdennis: Would it make sense to set the port anyway to ensure these instructions continue to match?

demos/build.gradle
henri-tremblay: It think it was indeed the default before. And it is now. But the port variable has became I don't know what so removing it was my best move.

docs/src/docs/asciidoc/user/examples.adoc
henri-tremblay: If you find the field :-) Otherwise, I'm fine with the default

henri-tremblay: Since I'm a nice guy, I found the new field. So it is now explicit again in the `build.gradle`

PR #2281, opened 2018-02-09
Add index file and align extensions with the user doc

chrisdennis:


docs/src/docs/asciidoc/developer/index.adoc
```diff
@@ -0,0 +1,33 @@
+= Ehcache 3 Developer Documentation
+== Code layout
+
+The code core of Ehcache is split in three core modules:
```
henri-tremblay: core code?

chrisdennis: I just copied this section verbatim from the `Home.asciidoc` of the wiki.  Will fix however.

Issue #2215, opened 2017-12-18
Clustered shared resources are not properly shared.


chrisdennis:
Clustered shared resources are currently only shared on a first come first served basis.  This is likely to cause nasty behaviors when caches are initially loaded synchronously, or lesser problems whenever the cache is loaded in any way other than 'naturally' as a result of it's regular workload missing.

PR #2282, opened 2018-02-12
Issue #2215 Fix shared resource not being shared issue

rahul-mittal:


clustered/client/src/test/java/org/ehcache/clustered/ClusteredSharedResourceTest.java
```diff
@@ -0,0 +1,100 @@
...
+  private static final URI CLUSTER_URI = URI.create("terracotta://example.com:9540/my-application");
+
+  private static PersistentCacheManager cacheManager;
+  private static Cache<Long, String> sharedCache1;
+  private static Cache<String, Long> sharedCache2;
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  @BeforeClass
```
henri-tremblay: No need to use a `@BeforeClass` here. A `@Before` is fine

clustered/client/src/test/java/org/ehcache/clustered/ClusteredSharedResourceTest.java
```diff
@@ -0,0 +1,100 @@
...
+      ResourcePoolsBuilder.newResourcePoolsBuilder()
+        .with(ClusteredResourcePoolBuilder.clusteredShared("resource-pool-a")));
+
+    sharedCache1 = cacheManager.createCache("shared-cache-1", cacheConfigurationBuilder1);
+    sharedCache2 = cacheManager.createCache("shared-cache-2", cacheConfigurationBuilder2);
+  }
+
+  @AfterClass
+  public static void tearDown() {
+    cacheManager.close();
```
henri-tremblay: Please add a `if(cacheManager != null)` to prevent hidden the actual original exception

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -49,12 +49,13 @@
     }
   }
 
-  public OffHeapServerStore(ResourcePageSource source, KeySegmentMapper mapper) {
+  public OffHeapServerStore(ResourcePageSource source, KeySegmentMapper mapper, boolean isShared) {
     this.mapper = mapper;
     segments = new ArrayList<>(mapper.getSegments());
-    long maxSize = getMaxSize(source.getPool().getSize());
+    int minPageSize = KILOBYTES.toBytes(4);
```
henri-tremblay: Why converting a constant?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -49,12 +49,13 @@
...
+    int maxPageSize = isShared ? minPageSize : (int) KILOBYTES.toBytes(getMaxSize(source.getPool().getSize()));
```
henri-tremblay: Like @chrisdennis, I don't know why this would solve the problem. As usual, an explanation around the code (in the code) would be nice. Same thing in the test to explain what it does.

PR #2284, opened 2018-02-12
Test client doesn't freeze when server is down (close #2202)

### LINKED ISSUE #2202: Inflight queue improvement
Currently, as soon as 200 requests are pending for answers from the TC Server (reads or writes), the client is stuck forever.

We need to improve that. Some ideas:
- Make the queue bigger
- Do not leave gets in it
- It should be big enough to never fill up before the lease expiration
- Entries should be removed after a timeout?

henri-tremblay:


clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
```diff
@@ -149,7 +155,11 @@ public static void restoreProperties() {
     if (OLD_PROPERTIES != null) {
       TCProperties tcProperties = TCPropertiesImpl.getProperties();
       for (Map.Entry<String, String> entry : OLD_PROPERTIES.entrySet()) {
-        tcProperties.setProperty(entry.getKey(), entry.getValue());
+        try {
+          tcProperties.setProperty(entry.getKey(), entry.getValue());
+        } catch(NullPointerException e) {
```
chrisdennis: Then we're not setting the result back right?  Would seem this should be a failure?

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
```diff
@@ -585,9 +595,58 @@ Void runTask() {
...
+        Void runTask() {
+          cache.get(1L); // the call that could block
+          return null;
+        }
+      }.run();
+
+      // FIXME: It freezes on the close() called by the try-with-resources if we never restart the server.
+      cluster.getClusterControl().startOneServer();
+    } catch(StateTransitionException e) {
+      // FIXME: swallow for now because we wait for #2203 to cleanly reconnect
```
chrisdennis: Issue #2203 is merged. Does this work now?

clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
```diff
@@ -194,16 +193,28 @@ public static void afterClass() throws Exception {
     if (cacheManager != null && cacheManager.getStatus() == Status.AVAILABLE) {
 
       if (nmsService != null) {
-        Context ehcacheClient = readTopology().getClient(ehcacheClientIdentifier).get().getContext().with("cacheManagerName", "my-super-cache-manager");
-        nmsService.stopStatisticCollector(ehcacheClient).waitForReturn();
+        readTopology().getClient(ehcacheClientIdentifier)
```
chrisdennis: What's this about?

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
```diff
@@ -141,6 +144,9 @@ public static void setProperties() {
     overrideProperty(oldProperties, TCPropertiesConsts.L1_SHUTDOWN_THREADGROUP_GRACETIME, "1000");
     overrideProperty(oldProperties, TCPropertiesConsts.TC_TRANSPORT_HANDSHAKE_TIMEOUT, "1000");
 
+    // Used only by testTerminationFreezesTheClient to be able to fill the inflight queue
+    overrideProperty(oldProperties, TCPropertiesConsts.CLIENT_MAX_PENDING_REQUESTS, "" + CLIENT_MAX_PENDING_REQUESTS);
```
chrisdennis: Pet peave of mine.  I'd prefer `Integer.toString(CLIENT_MAX_PENDING_REQUESTS)`.

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
henri-tremblay: In fact, it will remove the value as I want. And then fail on a NPE caused by a value.trim(). I've raised Terracotta-OSS/terracotta-core#852 about it.

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
henri-tremblay: Yes. I need to try. I've been lazy and wanted to do it in another PR. But I will do it now.

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
henri-tremblay: Ok

clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
henri-tremblay: This was failing because of an earlier failure in the test. So it was hiding the real problem in the build failed we experienced.

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
cschanck: hahaha

Sorry Henri, private joke.

clustered/integration-test/src/test/java/org/ehcache/clustered/TerminatedServerTest.java
henri-tremblay: Actually, some other tests should now pass as well. But the behaviour seems different depending on the lifecycle method (createCache, destroyCache, removeCache). So I'm digging a bit.

chrisdennis: I know @AbfrmBlr still has work to do on the destroy path... so that might explain some of that.  If you want to split out in to a separate PR that is fine.

PR #2285, opened 2018-02-13
Issue #2215 Add note in docs about the limitations of shared pool

rahul-mittal:


docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -143,6 +143,11 @@ In this example, a dedicated pool of 32MB is allocated for clustered-cache from
 <7> Configures another cache (`shared-cache-2`) that shares the resource pool (`resource-pool-a`) with `shared-cache-1`.
 <8> Creates fully initialized cache manager with the clustered caches.
 
+NOTE:  When multiple caches are sharing a shared pool, eviction operation called by a cache evicts elements of that cache
+itself and not just any set of elements in the shared pool. This means that today, when space in a shared pool
+is allocated to a cache, it remains with it for its use only. Improvisation would be made in future to allow eviction of
```
henri-tremblay: Improvisation?!?

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -143,6 +143,11 @@ In this example, a dedicated pool of 32MB is allocated for clustered-cache from
 <7> Configures another cache (`shared-cache-2`) that shares the resource pool (`resource-pool-a`) with `shared-cache-1`.
 <8> Creates fully initialized cache manager with the clustered caches.
 
+NOTE:  When multiple caches are sharing a shared pool, eviction operation called by a cache evicts elements of that cache
```
henri-tremblay: evictions

docs/src/docs/asciidoc/user/clustered-cache.adoc
henri-tremblay: And we should not talk about the future. This is the current documentation

docs/src/docs/asciidoc/user/clustered-cache.adoc
henri-tremblay: Am I right to say that the real thing is that a cache will reserve a block of memory from the shared pool and keep it forever? So eviction or not, this allocated memory will stay with a given cache and that it's. If I am right, that's probably what you should explain.

docs/src/docs/asciidoc/user/clustered-cache.adoc
rahul-mittal: removed

docs/src/docs/asciidoc/user/clustered-cache.adoc
rahul-mittal: nothing on eviction now in the new version

rahul-mittal: revised it, please let me know if it sounds better now

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -143,6 +143,9 @@ In this example, a dedicated pool of 32MB is allocated for clustered-cache from
 <7> Configures another cache (`shared-cache-2`) that shares the resource pool (`resource-pool-a`) with `shared-cache-1`.
 <8> Creates fully initialized cache manager with the clustered caches.
 
+NOTE:  When a cache is allocated a block of memory from a shared pool, it is retained forever and would never get
```
henri-tremblay: I would put NOTE in normal case but in bold **Note:**. And only one space after

rahul-mittal: 'NOTE' here is asciidoc syntax for admonition para, 'Note' will not have the same effect. Removed one extra space though.

henri-tremblay: Cool. Didn't know that.

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -143,6 +143,9 @@ In this example, a dedicated pool of 32MB is allocated for clustered-cache from
 <7> Configures another cache (`shared-cache-2`) that shares the resource pool (`resource-pool-a`) with `shared-cache-1`.
 <8> Creates fully initialized cache manager with the clustered caches.
 
+NOTE: When a cache is allocated a block of memory from a shared pool, it is retained forever and would never get
```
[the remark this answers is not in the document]
henri-tremblay: I'm not fully happy with it either. I'll call it a good start. CAUTION seems a good idea. If you have better explanation of the impacts, please do.

PR #2287, opened 2018-02-13
Put classes using unsafe in a dedicated sourceSet to allow using -Werror on impl

henri-tremblay:
I'm pretty sure 2 things are not working though. 

1. Distribution doesn't contains the unsafe source files
2. Unsafe classes are not in the MANIFEST.MF for OSGI

I'm not exactly sure are to do that cleanly.


chrisdennis:
Replaced by #2618 which is a tweaked and updated version of this fix.  It seems to work now... maybe due to changign the OSGi plugin we use?

PR #2319, opened 2018-03-05
#2318 : create hash map of correct size as hash map size in message

rkavanap:
This should avoid some rehashing for large messages and save some CPU.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheSyncMessageCodec.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheSyncMessageCodec.java
@@ -31,6 +31,7 @@ import org.terracotta.runnel.decoding.StructDecoder;
 import org.terracotta.runnel.encoding.StructEncoder;
 
 import java.nio.ByteBuffer;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -257,11 +258,12 @@ public class EhcacheSyncMessageCodec implements SyncMessageCodec<EhcacheEntityMe
   }
 
   private Map<Long, Chain> decodeChainMapEntries(StructDecoder<Void> decoder) {
-    Map<Long, Chain> chainMap = new HashMap<>();
-
     StructArrayDecoder<? extends StructDecoder<?>> entriesDecoder = decoder.structs(CHAIN_MAP_ENTRIES_SUB_STRUCT);
+
     if (entriesDecoder != null) {
-      for (int i = 0; i < entriesDecoder.length(); i++) {
+      int len = entriesDecoder.length();
+      Map<Long, Chain> chainMap = new HashMap<>((int) ((float) len / 0.75f + 1.0f));
+      for (int i = 0; i < len; i++) {
         StructDecoder<?> entryDecoder = entriesDecoder.next();
         Long key = entryDecoder.int64(KEY_FIELD);
         StructDecoder<?> chainDecoder = entryDecoder.struct(CHAIN_FIELD);
@@ -269,8 +271,9 @@ public class EhcacheSyncMessageCodec implements SyncMessageCodec<EhcacheEntityMe
         chainMap.put(key, chain);
         entryDecoder.end();
       }
+      return chainMap;
+    } else {
+      return Collections.emptyMap();
     }
-    return chainMap;
   }
-
 }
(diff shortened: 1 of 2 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
```diff
@@ -120,7 +120,7 @@
   static final String SYNC_DATA_SIZE_PROP = "ehcache.sync.data.size.threshold";
   private static final long DEFAULT_SYNC_DATA_SIZE_THRESHOLD = 2 * 1024 * 1024;
   static final String SYNC_DATA_GETS_PROP = "ehcache.sync.data.gets.threshold";
-  private static final int DEFAULT_SYNC_DATA_GETS_THRESHOLD = 16 * 1024;
+  private static final int DEFAULT_SYNC_DATA_GETS_THRESHOLD = 8 * 1024;
```
henri-tremblay: Why changing this number?

clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheSyncMessageCodec.java
```diff
@@ -257,20 +258,22 @@ private EhcacheSyncMessage decodeDataSync(ByteBuffer message) {
...
 
   private Map<Long, Chain> decodeChainMapEntries(StructDecoder<Void> decoder) {
-    Map<Long, Chain> chainMap = new HashMap<>();
-
     StructArrayDecoder<? extends StructDecoder<?>> entriesDecoder = decoder.structs(CHAIN_MAP_ENTRIES_SUB_STRUCT);
+
     if (entriesDecoder != null) {
-      for (int i = 0; i < entriesDecoder.length(); i++) {
+      int len = entriesDecoder.length();
+      Map<Long, Chain> chainMap = (len < 16) ? new HashMap<>() : new HashMap<>((int) ((float) len / 0.75f + 1.0f));
```
henri-tremblay: Why default when less than 16? Also, no need to apply a factor to the initial capacity. This is done by the HashMap. Just put the length. We are not expecting new entries to be added don't we?

chrisdennis: Sadly the HashMap pre-size doesn't take the loadfactor in to account:
```java
Map<Integer, Integer> map = new HashMap<>(16);

Field tableField = HashMap.class.getDeclaredField("table");
tableField.setAccessible(true);

for (int i = 0; i < 16; i++) {
  map.put(i, i);
  System.out.println("map.size() = " + map.size() + " : map.table.length = " + ((Object[]) tableField.get(map)).length);
}
```
gives:
```
map.size() = 1 : map.table.length = 16
map.size() = 2 : map.table.length = 16
map.size() = 3 : map.table.length = 16
map.size() = 4 : map.table.length = 16
map.size() = 5 : map.table.length = 16
map.size() = 6 : map.table.length = 16
map.size() = 7 : map.table.length = 16
map.size() = 8 : map.table.length = 16
map.size() = 9 : map.table.length = 16
map.size() = 10 : map.table.length = 16
map.size() = 11 : map.table.length = 16
map.size() = 12 : map.table.length = 16
map.size() = 13 : map.table.length = 32
map.size() = 14 : map.table.length = 32
map.size() = 15 : map.table.length = 32
map.size() = 16 : map.table.length = 32
```

clustered/server/src/main/java/org/ehcache/clustered/server/store/ClusterTierActiveEntity.java
henri-tremblay: Can you comment on that in the code, please.

clustered/server/src/main/java/org/ehcache/clustered/server/internal/messages/EhcacheSyncMessageCodec.java
henri-tremblay: Huh? It kinda does. The size goes to 32 after 12. Which is about 75% of 16. And then, they use a power of 2 for everything.

chrisdennis: Okay, but thats my point.  Ramesh is trying to size the HashMap such that he can fit 'N' things without having to rehash the table at any point.  To achieve that he can't just do `new HashMap<>(N)` since that constructor just sizes the initial table, and the map obeys a default load factor of 0.75.  Instead he needs (at a minimum) to size it at `ceiling(N/0.75)` which he has approximated as `(N/0.75)+1`.

rkavanap:
@henri-tremblay Can this be merged now? Does this look ok?

henri-tremblay:
Can you rebase please.

rkavanap:
test this please

Issue #2316, opened 2018-03-01
Reconnect during destroy of cachemanager


AbfrmBlr:
If destroy of cachemanager fails due to lease expiry, then client should reconnect to complete the destroy since it is a two phase process

AbfrmBlr:
Additionally , reconnect during destroy of cache also needs to be handled

PR #2328, opened 2018-03-12
Handling reconnection during destroy #2316

AbfrmBlr:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectingServerStoreProxy.java
```diff
@@ -39,7 +44,11 @@ public String getCacheId() {
 
   @Override
   public void close() {
-    proxy().close();
+    try {
+      proxy().close();
+    } catch (ConnectionClosedException | ConnectionShutdownException e) {
+      LOGGER.warn("Store was already closed, since connection was closed");
```
chrisdennis: Is this really worthy of a warn?  What action do we expect the user to take to prevent this from being logged?

AbfrmBlr: Hmm. Actually users can't do much here as this will generally happen when n/w is not stable.
But warn, is to let them know that a user action of close was not completed, since connection was already closed due to lease expiry or other n/w problem

chrisdennis: In what way was it not completed?  Is there anything that hasn't been cleaned up?

AbfrmBlr:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -304,6 +304,7 @@ public void releaseServerStoreProxy(ServerStoreProxy storeProxy, boolean isRecon
     if (!isReconnect) {
       storeProxy.close();
     } else {
+      LOGGER.info("Adding to reconnect set");
```
chrisdennis: What would this log mean to a user?

AbfrmBlr: This needs to go, I added for debugging and forgot

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ReconnectingServerStoreProxy.java
AbfrmBlr: "Not Completed" in the sense the user action was not complete, though since connection is closed, all the cleanup as happened at server side, as it is supposed to do.l
To me warn is to just a way to inform though logs that the user initiated "close" action failed, though the server already knows that this client has left so in a way its "complete" as well.

PR #2330, opened 2018-03-13
Fix #2214 : Handle very large puts that may not fit the cache

### LINKED ISSUE #2204: IndexOutOfBoundException during passive replication application
`ChainReplicationMessage.getResult()` behaves badly when the chain it's being asked to replicate is empty.  This can happen when there are multiple clients accessing the system due to cross segment eviction.  There are a number of ways of fixing this, and there is also some overdue yak shaving that can occur in this area.

### LINKED ISSUE #2214: A *very* large put can bring down the active and/or passive server.
If the OffHeapServerStore legitimately throws an OversizeMappingException (because the value presented is too large to fit in the offheap resourcre configured) then the propagating OversizeMappingException will cause the server to terminate.  Although this scenario is highly unlikely in a typical configuration we should prevent propagation of this exception and instead act as if the value was immediately evicted.

rkavanap:
Avoid the effect of clearing the cache if the user tries to put an object that is bigger than the cache.

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/OversizedCacheOpsTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/OversizedCacheOpsTest.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered;
+
+import org.ehcache.Cache;
+import org.ehcache.PersistentCacheManager;
+import org.ehcache.clustered.client.config.builders.ClusteredResourcePoolBuilder;
+import org.ehcache.clustered.client.config.builders.ClusteringServiceConfigurationBuilder;
+import org.ehcache.config.CacheConfiguration;
+import org.ehcache.config.builders.CacheConfigurationBuilder;
+import org.ehcache.config.builders.CacheManagerBuilder;
+import org.ehcache.config.builders.ResourcePoolsBuilder;
+import org.ehcache.config.units.MemoryUnit;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.terracotta.testing.rules.Cluster;
+
+import java.io.File;
+import java.util.Arrays;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThat;
+import static org.terracotta.testing.rules.BasicExternalClusterBuilder.newCluster;
+
+public class OversizedCacheOpsTest extends ClusteredTests {
+
+  private static final String RESOURCE_CONFIG =
+      "<config xmlns:ohr='http://www.terracotta.org/config/offheap-resource'>"
+      + "<ohr:offheap-resources>"
+      + "<ohr:resource name=\"primary-server-resource\" unit=\"MB\">2</ohr:resource>"
+      + "</ohr:offheap-resources>" +
+      "</config>\n";
+
+  @ClassRule
+  public static Cluster CLUSTER =
+      newCluster().in(new File("build/cluster")).withServiceFragment(RESOURCE_CONFIG).build();
+
+  @Test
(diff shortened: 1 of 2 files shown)


clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -175,4 +178,11 @@ public int getReprobeLength() {
 
     throw new UnsupportedOperationException("Not supported yet.");
   }
+
+  private void checkPayLoadSize(ByteBuffer payLoad) {
+    if (payLoad.remaining() > pageSource.getPool().getSize()) {
+      throw new OversizeMappingException("Payload (" + payLoad.remaining() +
```
henri-tremblay: Why is this exception not thrown to the client? Resilience strategy?

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -66,11 +67,13 @@ public Chain get(long key) {
 
   @Override
   public void append(long key, ByteBuffer payLoad) {
+    checkPayLoadSize(payLoad);
```
henri-tremblay: I will let someone else (@chrisdennis) confirm but it feels weird that it's not the server that is responsible to verify this. Also, why can't it happen on `replaceAtHead`, and `put`?

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
rkavanap: yes..default resiliency strategy also evicts in the correct order so did not do an explicit server side eviction, although again I will wait for @chrisdennis to comment.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
rkavanap: @henri-tremblay this is just to prevent a very large put (larger than the cache) act like a cache.clear().. Ofcourse, we may still have a very large put that fits the cache that may evict everything else..but that may be a legitimate thing..

Since we prevent such chains being formed, replaceAtHead (which is not a direct client action) and put (passive only) should not have such large single chain that crosses the pool size on the active and passive.

It is debatable whether to put this check on client side or server side. One advantage of server side is we have got all the overhead in the payload and it is easier to check pool size.

henri-tremblay: The advantage of client-side you mean?

rkavanap: no currently it is server side.. as ServerStoreImpl is on server side..

henri-tremblay: Oh dear... Totally misread. Sorry. All good then.

clustered/integration-test/src/test/java/org/ehcache/clustered/OversizedCacheOpsTest.java
```diff
@@ -0,0 +1,94 @@
...
+      + "<ohr:resource name=\"primary-server-resource\" unit=\"MB\">2</ohr:resource>"
+      + "</ohr:offheap-resources>" +
+      "</config>\n";
+
+  @ClassRule
+  public static Cluster CLUSTER =
+      newCluster().in(new File("build/cluster")).withServiceFragment(RESOURCE_CONFIG).build();
+
+  @BeforeClass
+  public static void waitForActive() throws Exception {
```
chrisdennis: This *shouldn't* be necessary... I'd rather avoid using it when it's not necessary.

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
chrisdennis: I *think* this is okay.  It's probably not how I would have done it, but I think it's valid.

rkavanap:
@henri-tremblay done with all review comments. Pls take a look and approve if found ok..

clustered/integration-test/src/test/java/org/ehcache/clustered/OversizedCacheOpsTest.java
```diff
@@ -0,0 +1,106 @@
...
+    final PersistentCacheManager cacheManager = clusteredCacheManagerBuilder.build(false);
+    cacheManager.init();
+
+    try {
+      CacheConfiguration<Long, String> config = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
+          ResourcePoolsBuilder.newResourcePoolsBuilder()
+              .with(ClusteredResourcePoolBuilder.clusteredDedicated("primary-server-resource", 1, MemoryUnit.MB))).build();
+
+      Cache<Long, String> cache = cacheManager.createCache("clustered-cache", config);
+      cache.put(1L, "The one");
```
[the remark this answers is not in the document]
rkavanap: I have removed the asserts.. But as discussed, I need to put multiple entries to assert that other entries are not evicted when a large entry that does not fit is added.

rkavanap:
test this please

PR #2331, opened 2018-03-14
Fix #2204 : Handle empty chains that may get replicated

### LINKED ISSUE #2204: IndexOutOfBoundException during passive replication application

rkavanap:
Empty chains may get replicated due to possibility of eviction between calls on the active.

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered.replication;
+
+import org.ehcache.Cache;
+import org.ehcache.PersistentCacheManager;
+import org.ehcache.clustered.ClusteredTests;
+import org.ehcache.clustered.client.config.builders.ClusteredResourcePoolBuilder;
+import org.ehcache.clustered.client.config.builders.ClusteringServiceConfigurationBuilder;
+import org.ehcache.config.CacheConfiguration;
+import org.ehcache.config.builders.CacheConfigurationBuilder;
+import org.ehcache.config.builders.CacheManagerBuilder;
+import org.ehcache.config.builders.ResourcePoolsBuilder;
+import org.ehcache.config.units.MemoryUnit;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.terracotta.testing.rules.Cluster;
+
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+
+import static org.terracotta.testing.rules.BasicExternalClusterBuilder.newCluster;
+
+/**
+ * Test the effect of cache eviction during passive sync.
+ */
+public class OversizedCacheOpsPassiveTest extends ClusteredTests {
+  private static final int MAX_PUTS = 3000;
+  private static final int MAX_SWITCH_OVER = 3;
+  private static final int PER_ELEMENT_SIZE = 256 * 1024;
+  private static final int CACHE_SIZE_IN_MB = 2;
+  private static final String LARGE_VALUE = buildLargeString();
+
+  private static final String RESOURCE_CONFIG =
+      "<config xmlns:ohr='http://www.terracotta.org/config/offheap-resource'>"
+      + "<ohr:offheap-resources>"
+      + "<ohr:resource name=\"primary-server-resource\" unit=\"MB\">2</ohr:resource>"
+      + "</ohr:offheap-resources>" +
+      "</config>\n";
+
+  @ClassRule
+  public static Cluster CLUSTER =
+      newCluster(2).in(Paths.get("build", "cluster").toFile())
+        .withSystemProperty("ehcache.sync.data.gets.threshold", "2")
+        .withServiceFragment(RESOURCE_CONFIG)
+        .build();
+
+  @BeforeClass
+  public static void waitForServers() throws Exception {
+    CLUSTER.getClusterControl().waitForActive();
+    CLUSTER.getClusterControl().waitForRunningPassivesInStandby();
+  }
+
+  @Test
+  public void oversizedPuts() throws Exception {
+    CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder
+        = CacheManagerBuilder.newCacheManagerBuilder()
+        .with(ClusteringServiceConfigurationBuilder.cluster(CLUSTER.getConnectionURI().resolve("/crud-cm"))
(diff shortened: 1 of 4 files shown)


clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.terracotta.testing.rules.Cluster;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.concurrent.CompletableFuture;
+
+import static org.terracotta.testing.rules.BasicExternalClusterBuilder.newCluster;
+
+public class OversizedCacheOpsPassiveTest extends ClusteredTests {
+  private static final int MAX_PUTS = 200;
```
rkavanap: To consistently reproduce the problems, MAX_PUTS have to be 10000 and MAX_SWITCH_OVERS have to be > 10, But that will cause delays in test execution. So for now they have been reduced.

rkavanap:
@chrisdennis , I was able to consistently reproduce empty chains getting replicated..But I am unable to get an oversizedmapping exception.. Will keep trying..

However, with a code trick (e.g on an empty chain making storage engine return null), I was able to simulate an oversizedmapping exception on the passive which crashed the passive...Since this was not a valid case that can happen (esp. since put will not pass empty chains down), I am not doing any special handling for oversizedmapping exception on passive yet...We could simply do a catch on oversizedmapping exception and log an error for now in the clustered tier passive entity. wdyt?

clustered/server/src/main/java/org/ehcache/clustered/server/ServerStoreImpl.java
```diff
@@ -80,6 +80,10 @@ public void replaceAtHead(long key, Chain expect, Chain update) {
   }
 
   public void put(long key, Chain chain) {
+    if (chain.isEmpty()) {
+      // holes may exist where empty chains may be replicated..just return in such cases
```
chrisdennis: Should a put of an empty chain map to a remove?  Seems odd to leave existing state in place?

rkavanap: ok.

rkavanap:
test this please

rkavanap:
Already merged through another PR..closing this

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final String RESOURCE_CONFIG =
+      "<config xmlns:ohr='http://www.terracotta.org/config/offheap-resource'>"
+      + "<ohr:offheap-resources>"
+      + "<ohr:resource name=\"primary-server-resource\" unit=\"MB\">2</ohr:resource>"
+      + "</ohr:offheap-resources>" +
+      "</config>\n";
+
+  @ClassRule
+  public static Cluster CLUSTER =
+      newCluster(2).in(new File("build/cluster"))
```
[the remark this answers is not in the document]
rkavanap: you are right..we should ideally use.. Paths.get("build", "cluster")...However, this works on windows..all tests follow this pattern..Did not want to change it just for this..

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -0,0 +1,101 @@
...
+  private void doPuts(CacheManagerBuilder<PersistentCacheManager> clusteredCacheManagerBuilder) {
+    try (PersistentCacheManager cacheManager = clusteredCacheManagerBuilder.build(true)) {
+      CacheConfiguration<Long, String> config = CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
+        ResourcePoolsBuilder.newResourcePoolsBuilder()
+          .with(ClusteredResourcePoolBuilder.clusteredDedicated("primary-server-resource", 2, MemoryUnit.MB)))
+        .build();
+      Cache<Long, String> cache = cacheManager.createCache("clustered-cache", config);
+      for (long i = 0; i < MAX_PUTS; i++) {
+        if (i % 100 == 0) {
+          Thread.yield(); Thread.yield(); Thread.yield();
```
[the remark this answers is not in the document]
rkavanap: just creating a small pause here without sleep...also helps on a uniprocessor machine, other put thread gets a chance..although that is not a strong argument as put call is io intensive.

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    CompletableFuture f1 = CompletableFuture.runAsync(() -> doPuts(clusteredCacheManagerBuilder));
+    CompletableFuture f2 = CompletableFuture.runAsync(() -> doPuts(clusteredCacheManagerBuilder));
+
+    for (int i = 0; i < MAX_SWITCH_OVER; i++) {
+      CLUSTER.getClusterControl().terminateActive();
+      CLUSTER.getClusterControl().waitForActive();
+      CLUSTER.getClusterControl().startOneServer();
+      CLUSTER.getClusterControl().waitForRunningPassivesInStandby();
+      Thread.sleep(100);
```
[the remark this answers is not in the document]
rkavanap: ..sleep gives some small window where some puts happen when no sync is happening..

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final int MAX_SWITCH_OVER = 3;
+
+  private static final String RESOURCE_CONFIG =
+      "<config xmlns:ohr='http://www.terracotta.org/config/offheap-resource'>"
+      + "<ohr:offheap-resources>"
+      + "<ohr:resource name=\"primary-server-resource\" unit=\"MB\">2</ohr:resource>"
+      + "</ohr:offheap-resources>" +
+      "</config>\n";
+
+  @ClassRule
```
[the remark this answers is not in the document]
rkavanap: @henri-tremblay I have incorporated all other comments except this. Since cluster startup is expensive, it is better if someone adds a test in the future we are not accidentally starting the cluster every time just because that person forgot to change this to a class rule later..Also the framework of this test class should allow more tests to be added without having to again refactor stuff.

rkavanap:
test this please

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -0,0 +1,124 @@
...
+  public static Cluster CLUSTER =
+      newCluster(2).in(Paths.get("build", "cluster").toFile())
+        .withSystemProperty("ehcache.sync.data.gets.threshold", "2")
+        .withServiceFragment(RESOURCE_CONFIG)
+        .build();
+
+  @BeforeClass
+  public static void waitForServers() throws Exception {
```
rkavanap: @chrisdennis I had to do a waitForServers for this test case because otherwise the terminateActive that happens immediately in the test is taking platform through different paths which is not the purpose of this test, On slow build machines, this is making the core hit different windows.

rkavanap:
test this please

rkavanap:
@henri-tremblay Can you approve and merge, please?

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
chrisdennis: Okay - makes sense.

PR #2333, opened 2018-03-14
The default class loader should be used to work correctly in a J2EE container

henri-tremblay:


core/src/main/java/org/ehcache/core/internal/util/ClassLoading.java
```diff
@@ -36,7 +36,7 @@ public static ClassLoader getDefaultClassLoader() {
   }
 
   public static <T> ServiceLoader<T> libraryServiceLoaderFor(Class<T> serviceType) {
-    return ServiceLoader.load(serviceType, ClassLoading.class.getClassLoader());
+    return ServiceLoader.load(serviceType, getDefaultClassLoader());
```
chrisdennis: The method is called `libraryServiceLoader(...)`. My understanding is that the intent is to not look for thing outside the scope of the classloader that loaded the Ehcache classes.  Services are a facility for dynamically binding together the various internal modules of Ehcache to form an implementation of the API.  I don't think we want to discovering implementations of services discovered in other a wider classloader context.

henri-tremblay: That kinda defeats the purpose of using the Java service framework.

chrisdennis: No it doesn't. Our use of the Java service framework is intended to be an internal detail, not a supported piece of public API.

chrisdennis:
To me this expands the scope of the service discovery without any real benefit to regular users. I'm also loathe to make much if any effort to make the use of services easier - they're not part of the public API they're internal details of the implementation only.

PR #2338, opened 2018-03-16
Fix #2337 : Ensure we break out of reconnect loop when the entity does not exist

### LINKED ISSUE #2337: Destroy of non-existent clustered cache causes infinite loop

chrisdennis:


clustered/client/src/test/java/org/ehcache/clustered/client/ClusteredCacheDestroyTest.java
```diff
@@ -181,6 +191,14 @@ public void testDestroyCacheWithCacheManagerStopped() throws CachePersistenceExc
     assertThat(persistentCacheManager.getStatus(), is(Status.UNINITIALIZED));
   }
 
+  @Test
+  public void testDestroyNonExistentCacheWIthCacheManagerStopped() throws CachePersistenceException {
```
henri-tremblay: 'I' should be lowercase

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -299,14 +300,16 @@ private void autoCreateEntity() throws ClusterTierManagerValidationException, Il
...
-    } catch (ConnectionClosedException | ConnectionShutdownException e) {
-      LOGGER.info("Disconnected to the server", e);
-      handleConnectionClosedException();
+    while (true) {
+      try {
+        destroyState();
+        reconnect();
+        retrieveEntity();
+        connectionRecoveryListener.run();
+        return;
```
henri-tremblay: could be a `break`. To look like the rest of the code

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -299,14 +300,16 @@ private void autoCreateEntity() throws ClusterTierManagerValidationException, Il
...
+      } catch (ConnectionClosedException | ConnectionShutdownException e) {
+        LOGGER.info("Disconnected to the server", e);
```
henri-tremblay: from the server?

PR #2339, opened 2018-03-16
Give access to the ServiceLocator for power users

henri-tremblay:


anthonydahanne:
@henri-tremblay  can you please add an example of usage of this change ? thanks

henri-tremblay:
My favorite example would be
```
StatisticsService stats = ((InternalCacheManager) cacheManager)
    .getServiceProvider().getService(StatisticsService.class);
```
It is currently impossible to retrieve the `StatisticsService` when using JSR107 to configure the `CacheManager`.

anthonydahanne:
got it, thanks !

chrisdennis:
Isn't the real need here for a public statistics API? Not sure what the value is in having this as a stop-gap? I'd rather just solve the problem properly and leave the service locator as hidden as it currently is.  Doesn't this essentially risk having more users assuming they can rely on the internal details of classes to no change?  Not sure what upside it has as it stands.

henri-tremblay:
Yes, a statistics API is a must. But having useful access to internal features with all the required red flags is a nice way to test and incubate features.

PR #2342, opened 2018-03-16
Fix multiple raw type warnings

henri-tremblay:


clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ExceptionCodec.java
```diff
@@ -91,7 +91,7 @@ public static ClusterException decode(StructDecoder<StructDecoder<Void>> decoder
       element.end();
     }
     arrayDecoder.end();
-    Class clazz = null;
+    Class<?> clazz = null;
```
chrisdennis: This should be `Class<? extends ClusterException>`, right?

Then the whole block would become:

```java
    Class<? extends ClusterException> clazz = null;
    ClusterException exception;
    try {
      clazz = Class.forName(exceptionClassName).asSubclass(ClusterException.class);
    } catch (ClassNotFoundException e) {
      LOGGER.error("Exception type not found", e);
    }
    exception = getClusterException(message, clazz);
    if (exception == null) {
      exception = new UnknownClusterException(message);
    }
    exception.setStackTrace(stackTraceElements);
    return exception;
  }

  private static ClusterException getClusterException(String message, Class<? extends ClusterException> clazz) {
    ClusterException exception = null;
    if (clazz != null) {
      try {
        Constructor<? extends ClusterException> declaredConstructor = clazz.getDeclaredConstructor(String.class);
        exception = declaredConstructor.newInstance(message);
      } catch (NoSuchMethodException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
        LOGGER.error("Failed to instantiate exception object.", e);
      }
    }
    return exception;
  }
```

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -144,7 +144,7 @@ private void writeUnlockAll() {
     }
   }
 
-  private void handleOversizeMappingException(long key, LongConsumer operation) {
+  private void consumeOversizeMappingException(long key, LongConsumer operation) {
```
chrisdennis: Incidental rename?

impl/src/main/java/org/ehcache/impl/config/resilience/DefaultResilienceStrategyProviderConfiguration.java
```diff
@@ -27,8 +27,10 @@
  */
 public class DefaultResilienceStrategyProviderConfiguration extends ClassInstanceProviderConfiguration<String, ResilienceStrategy<?, ?>> implements ServiceCreationConfiguration<ResilienceStrategyProvider> {
 
-  private static final Class<? extends ResilienceStrategy> DEFAULT_RESILIENCE = RobustResilienceStrategy.class;
-  private static final Class<? extends ResilienceStrategy> DEFAULT_LOADER_WRITER_RESILIENCE = RobustLoaderWriterResilienceStrategy.class;
+  @SuppressWarnings("rawtype")
```
chrisdennis: Why is this preferable?

impl/src/main/java/org/ehcache/impl/internal/classes/commonslang/ArrayUtils.java
```diff
@@ -44,7 +44,7 @@
     /**
      * An empty immutable {@code Class} array.
      */
-    public static final Class<?>[] EMPTY_CLASS_ARRAY = new Class[0];
+    public static final Class<?>[] EMPTY_CLASS_ARRAY = new Class<?>[0];
```
chrisdennis: Not fond of modifying these imported classes.

impl/src/test/java/org/ehcache/impl/internal/loaderwriter/writebehind/PooledExecutorWriteBehindTest.java
```diff
@@ -38,7 +40,7 @@
   }
 
   @Override
-  protected CacheManagerBuilder managerBuilder() {
+  protected CacheManagerBuilder<CacheManager> managerBuilder() {
```
chrisdennis: Why not fixing the initial declaration of this method?

impl/src/test/java/org/ehcache/impl/internal/statistics/DefaultTierStatisticsTest.java
```diff
@@ -47,7 +48,7 @@
   public void before() {
     CacheConfiguration<Long, String> cacheConfiguration =
       CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class,
-        newResourcePoolsBuilder().heap(10))
+        newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES))
```
chrisdennis: Why?

integration-test/src/test/java/org/ehcache/integration/OverSizeMappingTest.java
```diff
@@ -123,5 +124,10 @@ public boolean equals(Object obj) {
       }
       return false;
     }
+
+    @Override
+    public int hashCode() {
+      return Arrays.hashCode(arr);
```
chrisdennis: Inconsistent with equals.

integration-test/src/test/java/org/ehcache/integration/SerializersTest.java
```diff
@@ -83,7 +84,7 @@ private void testSerializerWithByRefHeapCache(Serializer<Long> serializer) throw
     CacheManagerBuilder<CacheManager> cmBuilder =
       newCacheManagerBuilder()
         .withCache("heapByRefCache",
-          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10))
+          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES))
```
chrisdennis: Accident?

integration-test/src/test/java/org/ehcache/integration/SerializersTest.java
```diff
@@ -93,7 +94,7 @@ private void testSerializerWithByValueHeapCache(Serializer<Long> serializer) thr
     CacheManagerBuilder<CacheManager> cmBuilder =
       newCacheManagerBuilder()
         .withCache("heapByValueCache",
-          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10))
+          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES))
```
chrisdennis: Accident?

integration-test/src/test/java/org/ehcache/integration/SerializersTest.java
```diff
@@ -114,7 +115,7 @@ private void testSerializerWithHeapOffheapCache(Serializer<Long> serializer) thr
     CacheManagerBuilder<CacheManager> cmBuilder =
       newCacheManagerBuilder()
         .withCache("heapOffheapCache",
-          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10).offheap(2, MemoryUnit.MB))
+          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).offheap(2, MemoryUnit.MB))
```
chrisdennis: Accident?

integration-test/src/test/java/org/ehcache/integration/SerializersTest.java
```diff
@@ -136,7 +137,7 @@ private void testSerializerWithHeapDiskCache(Serializer<Long> serializer) throws
       newCacheManagerBuilder()
         .with(persistence(temporaryFolder.newFolder().getAbsolutePath()))
         .withCache("heapDiskCache",
-          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10).disk(8, MemoryUnit.MB, true))
+          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).disk(8, MemoryUnit.MB, true))
```
chrisdennis: Accident?

integration-test/src/test/java/org/ehcache/integration/SerializersTest.java
```diff
@@ -147,7 +148,7 @@ private void testSerializerWithThreeTierCache(Serializer<Long> serializer) throw
       newCacheManagerBuilder()
         .with(persistence(temporaryFolder.newFolder().getAbsolutePath()))
         .withCache("heapOffheapDiskCache",
-          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10).offheap(2, MemoryUnit.MB).disk(8, MemoryUnit.MB, true))
+          newCacheConfigurationBuilder(Long.class, Person.class, newResourcePoolsBuilder().heap(10, EntryUnit.ENTRIES).offheap(2, MemoryUnit.MB).disk(8, MemoryUnit.MB, true))
```
chrisdennis: Accident?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -408,7 +408,7 @@ public ListenersConfig listenersConfig() {
 
           @Override
           public Iterable<ServiceConfiguration<?>> serviceConfigs() {
-            Map<Class<? extends ServiceConfiguration>, ServiceConfiguration<?>> configsMap =
```
chrisdennis: Why?  Although the code works with the looser type, the tighter type is more correct imho.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ExceptionCodec.java
henri-tremblay: It can. But there's no way to know that `exceptionClassName` is actually of the right type

chrisdennis: Thats what `asSubclass(ClusterException.class)` does (fourth line).

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
henri-tremblay: No. IntelliJ was giving a name clashing warning. So I've preferred renaming.

impl/src/main/java/org/ehcache/impl/internal/classes/commonslang/ArrayUtils.java
henri-tremblay: Yes... But they still have a compilation warning... Not much I can do

impl/src/test/java/org/ehcache/impl/internal/loaderwriter/writebehind/PooledExecutorWriteBehindTest.java
henri-tremblay: Right. Missed it.

impl/src/test/java/org/ehcache/impl/internal/statistics/DefaultTierStatisticsTest.java
henri-tremblay: Another IntelliJ warning about calling a static method from an instance. I agree with it but preferred to use this version of `heap` because when discussing it was agreed that the static version is probably a mistake in the API.

integration-test/src/test/java/org/ehcache/integration/OverSizeMappingTest.java
henri-tremblay: Indeed

integration-test/src/test/java/org/ehcache/integration/SerializersTest.java
henri-tremblay: Nope. Same reason as above

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
henri-tremblay: Since it's a local variable and I can't type `Class<? extends ServiceConfiguration<?>>` since it causes other parts of the code to scream

impl/src/main/java/org/ehcache/impl/config/resilience/DefaultResilienceStrategyProviderConfiguration.java
henri-tremblay: It's a mistake. I should have kept the original typing and add the "rawtypes"

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
chrisdennis: Given that it's getClass() returning the erased type here and thus prevents us from doing the 'right' thing... wouldn't it make more sense to suppress the warning and leave the code as is?

henri-tremblay: Ok

henri-tremblay:
test this please

PR #2346, opened 2018-03-19
Eliminate Eh107InternalCacheManager by resolving services from services

chrisdennis:


henri-tremblay:
I prefer #2339 which eliminates it as well. But without giving some special status to the `StatisticsService`. Because you never know what other services can be useful.

chrisdennis:
There is no special status here though. The Jsr107Service already declares a dependency on the StatisticsService, this just makes sure we retrieve it via the correct channels.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -147,10 +148,10 @@ private Eh107CacheManager createCacheManager(URI uri, Configuration config, Prop
       services.add(new DefaultJsr107SerializationProvider());
     }
 
-    Eh107InternalCacheManager ehcacheManager = new Eh107InternalCacheManager(config, services, !jsr107Service.jsr107CompliantAtomics());
+    org.ehcache.CacheManager ehcacheManager = new EhcacheManager(config, services, !jsr107Service.jsr107CompliantAtomics());
     ehcacheManager.init();
 
-    return new Eh107CacheManager(this, ehcacheManager, properties, config.getClassLoader(), uri,
+    return new Eh107CacheManager(this, ehcacheManager, jsr107Service.getStatistics(), properties, config.getClassLoader(), uri,
```
henri-tremblay: It still feels weird to pass the `StatisticsService` since the `ehcacheManager` is passed and already has it.

henri-tremblay: Why not passing jsr107service and retrieve the statistics inside the constructor? I would prefer that.

chrisdennis:
@henri-tremblay could you re-review this?  It's been tweaked and expanded since you last looked at it.

107/src/main/java/org/ehcache/jsr107/internal/DefaultJsr107Service.java
```diff
@@ -19,23 +19,26 @@
...
 import org.ehcache.spi.service.ServiceProvider;
 import org.ehcache.spi.service.Service;
 
+import static java.util.Objects.requireNonNull;
+
 @ServiceDependencies(StatisticsService.class)
 public class DefaultJsr107Service implements Jsr107Service {
 
   private final Jsr107Configuration configuration;
+  private StatisticsService statisticsService;
```
lorban: Shouldn't that field be volatile?

chrisdennis: Probably not since there the ServiceLocator creates a happens before relationship between the start of a dependency and the start of it's dependents... but it's probably safer to add it anyway.

PR #2347, opened 2018-03-20
Allow to create a CacheManager using an Ehcache builder

henri-tremblay:


107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -156,6 +156,61 @@ private Eh107CacheManager createCacheManager(URI uri, Configuration config, Prop
...
+   * @param uri the URI identifying this cache manager
+   * @param classLoader class loader where the cache manager will be registered
+   * @param builder the builder to create the underlying EhcacheManager
+   *
+   * @return a cache manager
+   */
+  public CacheManager getCacheManager(URI uri, ClassLoader classLoader, CacheManagerBuilder<org.ehcache.CacheManager> builder) {
+    Eh107CacheManager cacheManager;
+
+    synchronized (cacheManagers) {
```
chrisdennis: We can probably extract this common block between here and the existing code in to a combination of a `computeIfAbsent` and a `compute`.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -156,6 +156,61 @@ private Eh107CacheManager createCacheManager(URI uri, Configuration config, Prop
...
+      }
+    }
+
+    return cacheManager;
+  }
+
+  private Eh107CacheManager createCacheManager(URI uri, ClassLoader classLoader, CacheManagerBuilder<org.ehcache.CacheManager> builder) {
+    Eh107CacheLoaderWriterProvider cacheLoaderWriterFactory = new Eh107CacheLoaderWriterProvider();
+
+    Object[] serviceCreationConfigurations = builder.getServices().toArray();
```
chrisdennis: Type safety all messed up here.  The code is indeed looking for a ServiceCreationConfigurations but the builder will return services.  We'll have to fix the "extra services" logic to catch both services instances and creation configs that would result in service instances.  Would be good to have that as common between the two paths as well.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -156,6 +156,61 @@ private Eh107CacheManager createCacheManager(URI uri, Configuration config, Prop
...
+
+    Collection<Service> services = new ArrayList<>(4);
+    services.add(cacheLoaderWriterFactory);
+    services.add(jsr107Service);
+
+    if (ServiceUtils.findSingletonAmongst(DefaultSerializationProviderConfiguration.class, serviceCreationConfigurations) == null) {
+      services.add(new DefaultJsr107SerializationProvider());
+    }
+
+    InternalCacheManager ehcacheManager = (InternalCacheManager) builder.build(true);
```
chrisdennis: Need to actually add the services to the builder.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: Yes. I've been purposefully lazy.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: Which means the `findSingletonAmongst` is strangely permissive and this code doesn't work at all? Or it supports both?

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: Huh?

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
chrisdennis: `findSingletonAmongst` is just conventionally permissive.  This is it's signature:
```java
<T> T findSingletonAmongst(Class<T> clazz, Object ... instances)
```
What we want as an end result is to be certain that the cache manager has a `Jsr107Service` and the `Eh107CacheLoaderWriterProvider`.  We also need to make sure that any existing config is honored, which I think in this context means:
 * use the 107 service creation config if it exists in the builder (that should be a natural thing though)
 * not add a 107 service if there is one already defined
 * not add the special loader writer factory if there is a loader writer already???

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: Oh dear.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: Yep. I agree. Then, for the loader-writer... I'm not sure we actually need a `CacheLoaderWriterProvider`. We seem to just need something to handle `CacheLoader` and `CacheWriter`. The loader-writer is for a cache. So I guess we should ignore it is a `CacheLoaderWriterProvider` is already provided. If we want to keep things simple.

henri-tremblay:
I think all changes were implemented.

107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
```diff
@@ -84,6 +84,10 @@
     this.jsr107Cache = ehCache.createJsr107Cache();
   }
 
+  private StatisticsService getStatisticsService(Eh107CacheManager cacheManager) {
```
chrisdennis: static?

chrisdennis:
Could we separate this from the exposure of the ServiceLocator so that we can consider the two changes independently?

107/src/main/java/org/ehcache/jsr107/Eh107Cache.java
henri-tremblay: Can be indeed

henri-tremblay:
I can. I did it to prevent merging conflicts. Let me see.

henri-tremblay:
It's hard (impossible?) because getting rid of `Eh107InternalCacheManager` makes it possible. Otherwise, I will need a `CacheManagerBuilder.build107CacheManager`

chrisdennis:
Lets see if we can come to agreement on #2346 then, which would allow this to go through.

henri-tremblay:
@chrisdennis Now rebased on top of #2346 . Can you have a look?

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -108,14 +113,53 @@ public CacheManager getCacheManager(URI uri, Configuration config, Properties pr
...
+   * @param classLoader class loader where the cache manager will be registered
+   * @param builder the builder to create the underlying EhcacheManager
+   *
+   * @return a cache manager
+   */
+  public CacheManager getCacheManager(URI uri, ClassLoader classLoader, CacheManagerBuilder<org.ehcache.CacheManager> builder) {
+    return getCacheManager(uri, classLoader, () -> createCacheManager(uri, classLoader, builder));
+  }
+
+  private CacheManager getCacheManager(URI uri, ClassLoader classLoader, Supplier<Eh107CacheManager> cacheManagerCreator) {
```
chrisdennis: It feel like this Supplier should really be a function, possibly taking the `URI` and `ClassLoader` as arguments.

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -128,34 +172,38 @@ Eh107CacheManager getCacheManager(ConfigSupplier configSupplier, Properties prop
...
       }
     }
 
     return cacheManager;
   }
 
-  private Eh107CacheManager createCacheManager(URI uri, Configuration config, Properties properties) {
-    Eh107CacheLoaderWriterProvider cacheLoaderWriterFactory = new Eh107CacheLoaderWriterProvider();
+  private Eh107CacheManager createCacheManager(URI uri, ClassLoader classLoader, CacheManagerBuilder<org.ehcache.CacheManager> builder) {
+    Set<Service> services = builder.getServices();
```
chrisdennis: This method feels like it contains logic that should really be a part of the CacheManagerBuilder.  Fixing this would mean either adding a new method to the builder that only added the service if it didn't already exist, or modifying the contract of the using method to behave more consistently with the using method that take a config object. This would really be moving a failure that will occur anyway when using `using` from being a CM build time to earlier in the builder invocations (arguably a good thing).

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: I tried but it makes things redundant. With lines like `return getCacheManager(uri, classLoader, (u, c) -> createCacheManager(u, c, builder));`. But I see what you mean. It removes the closure for these two params by the caller. But I still have a closure. Still prefer me to change it?

107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
henri-tremblay: What would you do instead? `usingIfNotPresent` or `using(Service service, boolean overrideIfPresent)`? I agree that I don't like the `getServices` either.

chrisdennis: I'm thinking the sensible solution would be to add `usingIfAbsent(...)` methods to the builder. At the same time it might make sense to also align the contracts for the service and service-config related methods since I think they vary a bit at the moment.

chrisdennis:
Given our proximity to release and the associated need for stabilization I'd like to hold off on merging this until after the release goes out. I'd also like to see if we can align more of the using behaviors like previously discussed.

PR #2348, opened 2018-03-20
Upgrade to Gradle 4.6

chrisdennis:


demos/build.gradle
```diff
@@ -1,10 +1,16 @@
-plugins {
-  id 'org.akhikhl.gretty' version '2.0.0' apply false
+buildscript {
```
henri-tremblay: For me, this is supposed to be legacy style. No?

chrisdennis: Yes... but something odd is happening if I try and do this any other way.  Not sure if I'm screwing up, or if something is wrong with the way the forked version of the plugin is being published.

henri-tremblay: When going back to the 2.0.0 version it "works" (jacoco plugin is failing). But I can't start the app with the new version when doing `../../gradlew appStart`

chrisdennis: It's not Jacoco thats broken, it's Jacoco trying to interact with the gretty task. Gretty uses internal gradle APIs that changed in 4.6 and had to updated things.  The broken `appStart` is an oversight, need to make sure the submodules can resolve the runner. Fixing.

henri-tremblay: Seems I tripping on https://github.com/akhikhl/gretty/issues/425

ljacomet: Replacing the whole `buildscript` with 
```
plugins {
  id 'org.gretty' version '2.1.0' apply false
}
```
seems to work fine.

Including launching the demos with `appStart`.

chrisdennis: Right it does today: https://plugins.gradle.org/plugin/org.gretty

> Version 2.1.0 (latest)
> Created 20 March 2018.

I'll update accordingly.

ljacomet: ah did not pay attention to the dates (PR or plugin publication) :-)

henri-tremblay: héhé... yes... because I tried that earlier. It wasn't working because of a missing jetty 9...

demos/build.gradle
```diff
@@ -1,10 +1,10 @@
 plugins {
-  id 'org.akhikhl.gretty' version '2.0.0' apply false
+  id 'org.gretty' version '2.1.0' apply false
 }
 
 subprojects {
   apply plugin: 'war'
-  apply plugin: 'org.akhikhl.gretty'
+  apply plugin: 'org.gretty'
```
henri-tremblay: So far, without adding jcenter, I can't find the gretty jetty9 jar. So it's not working for me.
```
subprojects {
  apply plugin: 'war'
  apply plugin: 'org.gretty'

  repositories {
    jcenter()
  }

  gretty {
    httpPort = 8080
    contextPath = '/'
    servletContainer = 'jetty9'
  }

  dependencies {
    compile 'javax.servlet:javax.servlet-api:3.1.0', project(':impl')
    runtime 'ch.qos.logback:logback-classic:1.2.3', 'com.h2database:h2:1.4.196'
  }
}
```

PR #2350, opened 2018-03-21
Upgrade to ASM 6.1 to work with Java 10

henri-tremblay:


chrisdennis:
My local testing indicated that the Jacoco Upgrade #2349 was the only change needed to get tests passing in Java 10.  What prompted this?

henri-tremblay:
ASM is used during testing. It might not be needed in our case. But to be on the safe side, ASM 6.1 is indeed the version supporting Java 10 officially.

chrisdennis:
Test this please.

PR #2353, opened 2018-03-28
Fix #2352 : support chain map storage engine extensibility

### LINKED ISSUE #2352: Support  offheap chain storage engine extensibility with binary engine support
Chain storage engine extensibility will allow building of other storage engines and plug it to the chain map.

rkavanap:


rkavanap:
test this please

clustered/server/src/test/java/org/ehcache/clustered/server/offheap/ChainMapExtensionTest.java
```diff
@@ -0,0 +1,378 @@
...
+      long val = ((long) hash << 32) | size;
+      writeContext.setLong(HASH_OFFSET, val);
+    }
+
+    private long getLong(int address) {
+      return data.getLong(address);
+    }
+  }
+
+  public static class ExtendedOffHeapChainStorageEngine<K> extends OffHeapChainStorageEngine<K> {
```
henri-tremblay: This class looks a bit too much for a test. Why is it that complicated?

rkavanap: hmm.. this is only way to test the extension capabilities I added..by adding a storage engine that extends the offheap chain storage engine and uses the extended header to keep track of additional stuff.

One thing I could do is to have the extended storage engine to be a separate class in the test package.. So that the test looks smaller and concise. wdyt?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/DefaultOffHeapChainHeaderExtension.java
```diff
@@ -0,0 +1,112 @@
...
+
+      @Override
+      public void flush() {
+        //no-op
+      }
+    };
+  }
+
+  public Long extractXtnAddress(int hash, ByteBuffer keyBuffer) {
+    Long encoding = owner.getEncodingForHashAndBinary(hash, keyBuffer);
```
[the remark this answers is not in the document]
rkavanap: Actually bind is guaranteed to happen before a storage engine is used by the owning map. there is a lot of places where it will fail with null pointer exception if it does not.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
```diff
@@ -43,11 +44,11 @@
   }
 
   private final ReadWriteLockedOffHeapClockCache<K, InternalChain> heads;
-  private final OffHeapChainStorageEngine<K> chainStorage;
+  private final ChainStorageEngine<K> chainStorage;
```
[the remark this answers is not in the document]
rkavanap: hmm.. interesting point,, 

Anyway for everything there is some association to offheap...just wanted to avoid connecting to a concrete class for extensibility and ability to connect different storage engines to the chain map.
OffHeapChainStorageEngine is a concrete class..while ChainStorageEngine is an interface.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -47,35 +48,64 @@
...
   private final OffHeapStorageArea storage;
   private final Portability<? super K> keyPortability;
   private final Set<AttachedInternalChain> activeChains = Collections.newSetFromMap(new ConcurrentHashMap<AttachedInternalChain, Boolean>());
+  private final int extendedChainHeaderSize;
+  private final ByteBuffer emptyExtendedChainHeader;
+  private final int totalChainHeaderSize;
+  private final OffHeapChainExtensionHeader extendedHeader;
 
   private StorageEngine.Owner owner;
   private long nextSequenceNumber = 0;
```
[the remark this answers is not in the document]
rkavanap: Actually everything in the storage engine assumes appropriate locks are held on the map and since writeMapping etc.. are all under write lock we should be fine here. besides this is not a code associated with the current change.

clustered/server/src/main/java/org/ehcache/clustered/server/state/EhcacheStateContext.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.ehcache.clustered.server.state;
+
+import com.tc.classloader.CommonComponent;
+
+/**
+ * Marker interface to pass context between begin and end message processing.
+ */
+@CommonComponent
+public interface EhcacheStateContext extends AutoCloseable {
```
chrisdennis: This seems awfully heavyweight just to avoid catching an Exception in the outer log... can't we just use AutoCloseable here and amend the exception handling logic in the call sites?

clustered/server/src/main/java/org/ehcache/clustered/server/state/EhcacheStateService.java
```diff
@@ -67,4 +68,5 @@
 
   void loadExisting(ServerSideConfiguration configuration);
 
+  EhcacheStateContext beginProcessing(EhcacheMessageType messageType, String name);
```
chrisdennis: If we're beginning processing... why don't we pass the whole message?  That would read much more naturally, right?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -29,14 +29,15 @@
...
 import org.terracotta.offheapstore.storage.PointerSize;
 import org.terracotta.offheapstore.storage.StorageEngine;
 import org.terracotta.offheapstore.storage.portability.Portability;
+import org.terracotta.offheapstore.util.Factory;
 
 import static java.util.Collections.unmodifiableList;
 
-class OffHeapChainStorageEngine<K> implements StorageEngine<K, InternalChain> {
-
+public class OffHeapChainStorageEngine<K> implements ChainStorageEngine<K>, BinaryStorageEngine, StorageEngineExtensions {
```
chrisdennis: Something feels off about this class... I can't quite put my finger on it though.  There's something going on with the way we're mixing inlined composition, OO composition and then later OO inheritance that makes me feel uncomfortable.  I shall have a further think and see if I can explain this better.

clustered/server/src/test/java/org/ehcache/clustered/server/offheap/ChainMapExtensionTest.java
henri-tremblay: No. I won't change anything. It is the complexity of `ExtendedOffHeapChainStorageEngine` that I don't like.

clustered/server/src/main/java/org/ehcache/clustered/server/state/EhcacheStateContext.java
rkavanap: @chrisdennis , this simplifies exception handling in ClusterTierPassiveEntity nicely as otherwise I need to differentiate between cluster exception and all other exceptions. Thought it was more readable/extensible as well with future possibilities to extend the interface to add additional behaviour to the context (e.g notifying stages of processing, if any).

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
rkavanap: done changes as discussed

rkavanap:
test this please

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainExtensionHeader.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.terracotta.offheapstore.storage.portability.WriteContext;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Support extending header and allowing to read the extended area
+ */
+public interface OffHeapChainExtensionHeader {
+  ByteBuffer getExtensionHeader(long xtnAddress);
```
chrisdennis: Can't we just move these methods in to the ChainStorageEngine interface.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainExtensionHeader.java
```diff
@@ -0,0 +1,30 @@
...
+  WriteContext getExtensionWriteContext(long xtnAddress);
+  long chainAddressToExtensionAddress(long chainAddress);
```
chrisdennis: This (and the reverse) shouldn't need to be exposed outside of the OffHeapChainStorageEngine right?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainExtensionHeader.java
```diff
@@ -0,0 +1,30 @@
...
+package org.ehcache.clustered.server.offheap;
+
+import org.terracotta.offheapstore.storage.portability.WriteContext;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Support extending header and allowing to read the extended area
+ */
+public interface OffHeapChainExtensionHeader {
```
chrisdennis: Acting on both of these comments would eliminate this interface, and it's implementations.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/StorageEngineExtensions.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package org.ehcache.clustered.server.offheap;
+
+public interface StorageEngineExtensions {
+  default void chainFreed(long chainAddress) {}
+  default void chainModified(long chainAddress) {}
+  default void chainMoved(long fromChainAddress, long toChainAddress) {}
+
+  OffHeapChainExtensionHeader getExtendedHeader();
```
chrisdennis: If we eliminate this interface (as described in other comments) then this method goes, and therefore this interface can go too.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/StorageEngineExtensions.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.server.offheap;
+
+public interface StorageEngineExtensions {
+  default void chainFreed(long chainAddress) {}
```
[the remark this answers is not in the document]
rkavanap: ok..will change it.. I was also caught between putting dummy protected methods and exposing it publicly. I guess putting dummy protected methods is better.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainExtensionHeader.java
rkavanap: @chrisdennis should these be in the ChainStorageEngine interface as these are only for other storage engines that may extend this storage engine for reuse.? The methods in ChainStorageEngine is used by other clients such as OffheapChainMap.

I have made them all protected inside OffheapChainStorageEngine and removed the interface classes.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -175,6 +339,56 @@ public boolean shrink() {
...
+      }
+
+      @Override
+      public void flush() {
+        //no-op
+      }
+    };
+  }
+
+  protected long chainAddressToExtensionAddress(long chainAddress) {
```
chrisdennis: Why does this method and it's counterpart need to be exposed?  Shouldn't where the extension header gets stored be a concern only for *this* class and not for the subclass that just needs to write and read the header, not care about where it gets put.

clustered/server/src/main/java/org/ehcache/clustered/server/state/EhcacheStateContext.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package org.ehcache.clustered.server.state;
+
+import com.tc.classloader.CommonComponent;
+
+/**
+ * Marker interface to pass context between begin and end message processing.
+ */
+@CommonComponent
```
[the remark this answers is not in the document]
chrisdennis: Neither AutoCloseable nor Closeable are annotated as such, why is this one special?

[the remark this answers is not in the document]
henri-tremblay: Because `EhcacheStateContext` is implemented by a lambda. It implies that we expect it for be used as a functional interface (the annotation will enforce that). `Closeable` is not a functional interface. It is added to tell an object is closeable. But we expect the object to do other things.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainStorageEngine.java
```diff
@@ -175,6 +339,53 @@ public boolean shrink() {
     return storage.shrink();
   }
 
+  protected ByteBuffer getExtensionHeader(long chainAddress) {
```
rkavanap: @chrisdennis done.. I was able to remove both chainToExtensionAddress and extensionToChainAddress.. Please take a look.

rkavanap:
test this please

PR #2356, opened 2018-04-13
Adapt to new ConnectionFactory and ConnectionService APIs

saurabhagas:


chrisdennis:
Test this please since the platform pre-release is now available.

saurabhagas:
@chrisdennis, this PR depends on https://github.com/Terracotta-OSS/terracotta-platform/pull/461.

chrisdennis:
Right - I falsely assumed that since the first pre-release of 5.5.0 was out that that PR would be in it.

saurabhagas:
You're welcome to review the platform PR :-)

saurabhagas:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -142,6 +148,30 @@ public ClusteringServiceConfiguration(URI clusterUri, Timeouts timeouts, boolean
...
+   * @param timeouts the {@link Timeouts} specifying the time limit for clustered cache operations
+   * @param autoCreate {@code true} if server components should be auto created
+   * @param serverConfig  the server side entity configuration required
+   *
+   * @throws NullPointerException if {@code servers} is {@code null}
+   */
+  public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String cacheManagerName, Timeouts timeouts,
+                                        boolean autoCreate, ServerSideConfiguration serverConfig, Properties properties) {
+    this.clusterUri = null;
+    this.servers = StreamSupport.stream(Objects.requireNonNull(servers, "Servers cannot be null").spliterator(), false).collect(toList());
```
chrisdennis: This seems a bit profligate... and the stream construction around a spliterator seems a bit unusual.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -144,7 +161,11 @@ private void reconnect() {
   }
 
   private void connect() throws ConnectionException {
-    clusterConnection = LeasedConnectionFactory.connect(clusterUri, connectionProperties);
+    if( clusterUri != null){
```
chrisdennis: It seems like all this "If the old state is here do this, otherwise do that" is hiding the fact that there's a type hierarchy that should exist here.

chrisdennis:
Isn't the IPv6 list that this API was introduced to support applicable for OSS active/passive cases too?  In which case shouldn't we be exposing new API in the builders?  If thats the case then should we be deprecating the old APIs?

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
saurabhagas: Changed

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
saurabhagas: Created a `ConnectionSource` object which contains this information.

saurabhagas:
test this please

saurabhagas:
@chrisdennis, done. I haven't deprecated the old APIs however.

saurabhagas:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -41,7 +44,8 @@
     HumanReadable {
 
   public static final boolean DEFAULT_AUTOCREATE = false;
-  private final URI clusterUri;
+  private final ConnectionSource connectionSource;
+  private final String cacheManagerName;
```
mathieucarbou: I think this should be renamed to `entityName` or `clusterTierManagerEntity` or something like that.
Because this is confusing since there is no cache manager name, and in management, we have a cache manager alias attribute on the Management configuration ;-)

PR #2358, opened 2018-04-19
XML translation of core service and service provider configurations

albinsuresh:
XML translation of core `ServiceCreationConfiguration`s and `ServiceConfiguration`s.


xml/src/main/java/org/ehcache/xml/CacheManagerServiceConfigurationParser.java
```diff
@@ -34,4 +34,8 @@
   URI getNamespace();
 
   ServiceCreationConfiguration<T> parseServiceCreationConfiguration(Element fragment);
+
+  Class<T> getServiceCreationConfigurationType();
```
chrisdennis: This is type to return the service type and not the configuration type... which did you want?

xml/src/main/java/org/ehcache/xml/CacheManagerServiceConfigurationParser.java
```diff
@@ -34,4 +34,8 @@
...
+
+  Element translateServiceConfiguration(ServiceCreationConfiguration<T> serviceCreationConfiguration);
```
chrisdennis: Document might be easier than element here, and is arguably more correct.  We know that the configs this thing is generating must honor an xsd so must be able to be validating document in their own right, not just fragments.

chrisdennis: You might want to confer with @chaktatha since I recommended switching to unparse in his PR.

albinsuresh:
test this please

PR #2359, opened 2018-04-20
Handling OSME in passive: remove the key and catch.

AlbertGevorgyan:
Currently can only reproduce OSME in passive while putting, so the key is removed in a nested catch clause in OffHeapServerStore#put, and the exception is rethrown to be caught and logged in ClusterTierPassiveEntity#invokePassiveInternal. Not sure whether the same should be done in other methods potentially throwing OSME, e.g. in OffHeapServerStore#replaceAtHead - can't reproduce it there since the failure results in a successful eviction.

### DIFF
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
@@ -214,6 +214,10 @@ public class OffHeapChainMap<K> implements MapInternals {
     }
   }
 
+  void remove(K key) {
+    heads.removeNoReturn(key);
+  }
+
   public void clear() {
     heads.writeLock().lock();
     try {
--- a/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
+++ b/clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
@@ -119,9 +119,14 @@ public class OffHeapServerStore implements ServerStore, MapInternals {
 
   public void put(long key, Chain chain) {
     try {
-      segmentFor(key).put(key, chain);
-    } catch (OversizeMappingException e) {
-      handleOversizeMappingException(key, (long k) -> segmentFor(k).put(k, chain));
+      try {
+        segmentFor(key).put(key, chain);
+      } catch (OversizeMappingException e) {
+        handleOversizeMappingException(key, (long k) -> segmentFor(k).put(k, chain));
+      }
+    } catch (Throwable t) {
+      segmentFor(key).remove(key);
+      throw t;
     }
   }
 
--- a/clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
+++ b/clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
@@ -20,6 +20,7 @@ import org.ehcache.clustered.common.Consistency;
 import org.ehcache.clustered.common.PoolAllocation;
 import org.ehcache.clustered.common.ServerSideConfiguration;
 import org.ehcache.clustered.common.internal.ServerStoreConfiguration;
+import org.ehcache.clustered.common.internal.messages.ServerStoreOpMessage;
 import org.ehcache.clustered.common.internal.store.Chain;
 import org.ehcache.clustered.common.internal.store.ClusterTierEntityConfiguration;
 import org.ehcache.clustered.common.internal.store.Util;
@@ -42,8 +43,10 @@ import org.terracotta.monitoring.IMonitoringProducer;
 import org.terracotta.offheapresource.OffHeapResource;
 import org.terracotta.offheapresource.OffHeapResourceIdentifier;
 import org.terracotta.offheapresource.OffHeapResources;
+import org.terracotta.offheapstore.exceptions.OversizeMappingException;
 import org.terracotta.offheapstore.util.MemoryUnit;
 
+import java.nio.ByteBuffer;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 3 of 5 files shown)


clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
```diff
@@ -166,6 +167,24 @@ public void testPassiveTracksMessageDuplication() throws Exception {
...
+    passiveEntity.createNew();
+    TestInvokeContext context = new TestInvokeContext();
+
+    Chain oversizeChain = Util.getChain(true, createPayload(2L, 1024 * 1024));
+    int oversizeKey = 2;
+    PassiveReplicationMessage oversizeMsg = new PassiveReplicationMessage.ChainReplicationMessage(oversizeKey, oversizeChain, 2L, 1L, (long) 3);
+    try {
+      passiveEntity.invokePassive(context, oversizeMsg);
+      fail();
+    } catch (OversizeMappingException e) {
```
rkavanap: Actually the passive entity should hide this exception from platform, otherwise the passive entity will crash in this situation. Although I am not sure whether that is the right thing to do. Maybe since we already have cases where a passive could independently evict, it is OK to hang around without crashing. Needs some discussion though on what is the right behaviour.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
```diff
@@ -106,7 +107,7 @@ public void testCreateSharedServerStore() throws Exception {
 
     assertThat(defaultRegistry.getStoreManagerService().getStores(), containsInAnyOrder(defaultStoreName));
     assertThat(defaultRegistry.getStoreManagerService()
-        .getSharedResourcePoolIds(), containsInAnyOrder(defaultSharedPool));
```
rkavanap: see no difference other than a blank.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -149,42 +133,44 @@ private void writeUnlockAll() {
...
+      return operation.apply(key);
+    } catch (OversizeMappingException e) {
+      return handleOversizeMappingException(key, operation, e);
     }
+  }
+
+  private <R> R handleOversizeMappingException(long key, LongFunction<R> operation, OversizeMappingException exception)
+    throws OversizeMappingException {
 
     writeLockAll();
```
rkavanap: @chrisdennis .. One behaviour difference I see from previous version is that now the writelockall is attempted before a quick and dirty tryShrinkOthers..

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
AlbertGevorgyan: Probably need to discuss this with Chris D - I understood from the conversation with him that this behaviour is valid.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
AlbertGevorgyan: Must be accidental formatting - will revert.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
AlbertGevorgyan: OK now I see the initial intent. The problem I am trying to fix here is that the failing quick and dirty tryShrinkOthers is followed by writeLockAll() and another call to operation.apply(key). This doesn't make sense to me because the shrinking has failed so there is no way the insertion will succeed. Am I wrong?

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
```diff
@@ -214,6 +214,15 @@ public void put(K key, Chain chain) {
     }
   }
 
+  void remove(K key) {
+    heads.writeLock().lock();
+    try {
+      heads.computeIfPresent(key, (k, v) -> null);
+    } finally {
```
AlbertGevorgyan: Calling computeIfPresent since remove is overridden.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
```diff
@@ -121,7 +121,12 @@ public void put(long key, Chain chain) {
     try {
       segmentFor(key).put(key, chain);
     } catch (OversizeMappingException e) {
-      handleOversizeMappingException(key, (long k) -> segmentFor(k).put(k, chain));
+      try {
+        handleOversizeMappingException(key, (long k) -> segmentFor(k).put(k, chain));
+      } catch (OversizeMappingException e1) {
+        segmentFor(key).remove(key);
+        throw e;
```
AlbertGevorgyan: This is ugly but ensures that key is removed only in put. Do we need this catch black around other handleOversizeMappingException too? If yes, the method itself can be refactored.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
```diff
@@ -166,6 +169,39 @@ public void testPassiveTracksMessageDuplication() throws Exception {
...
+
+    int key = 2;
+
+    Chain chain = Util.getChain(true, createPayload(1L));
+    PassiveReplicationMessage message = new PassiveReplicationMessage.ChainReplicationMessage(key, chain, 2L, 1L, 3L);
+    passiveEntity.invokePassive(context, message);
+
+    Chain oversizeChain = Util.getChain(true, createPayload(2L, 1024 * 1024));
+    ServerStoreOpMessage.ReplaceAtHeadMessage oversizeMsg = new ServerStoreOpMessage.ReplaceAtHeadMessage(key, chain, oversizeChain);
+    passiveEntity.invokePassive(context, oversizeMsg);
```
AlbertGevorgyan: This calls replaceAtHead followed by a successful eviction. No OSME is thrown.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
chrisdennis: What do you mean remove is overridden? I was fairly sure the following would work:
```java
public void remove(K key) {
  heads.writeLock().lock();
  try {
    heads.remove(key).close();
  } finally {
    heads.writeLock().unlock();
  }
}
```

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapServerStore.java
chrisdennis: Wouldn't it be easier to do:
```java
try {
  try {
    segmentFor(key).put(key, chain);
  } catch (OversizeMappingException e) {
    handleOversizeMappingException(key, (long k) -> segmentFor(k).put(k, chain));
  }
} catch (Throwable t) {
  segmentFor(key).remove(key);
  throw t;
}
```
Thats the general idea right... the fact that we trap the OSME as harmless further out is an independent thing. The general idea is "something went wrong when putting" not sure what happened, safest thing to do is invalidate.

clustered/server/src/test/java/org/ehcache/clustered/server/store/ClusterTierPassiveEntityTest.java
```diff
@@ -166,6 +169,39 @@ public void testPassiveTracksMessageDuplication() throws Exception {
...
+    // Should be evicted, the value is oversize.
+    assertThat(passiveEntity.getStateService().getStore(passiveEntity.getStoreIdentifier()).get(key).isEmpty(), is(true));
+  }
+
+  @Test
+  public void testOversizeChainReplicationMessage() throws Exception {
+    ClusterTierPassiveEntity passiveEntity = new ClusterTierPassiveEntity(defaultRegistry, defaultConfiguration, DEFAULT_MAPPER);
+    passiveEntity.createNew();
+    TestInvokeContext context = new TestInvokeContext();
+
```
AlbertGevorgyan: Unfortunately can't put a small value first, as it results in a successful eviction afterwards so no OSME is thrown.

clustered/server/src/main/java/org/ehcache/clustered/server/offheap/OffHeapChainMap.java
```diff
@@ -214,6 +214,15 @@ public void put(K key, Chain chain) {
     }
   }
 
+  void remove(K key) {
+    heads.writeLock().lock();
+    try {
+      heads.remove(key);
```
chrisdennis: This should either be `heads.removeNoReturn(key)` or `heads.remove(key).close()`. I'd prefer the former since the later is slightly broken although it shows no symptoms at the moment.

chrisdennis: Probably good to add a unit test to ChainMapTest here at the same time.

PR #2361, opened 2018-04-24
Fix #2360 : Add missing tests

### LINKED ISSUE #2360: oversize mapping exception tests in offheap store test
add missing oversize mapping exception tests in offheap store test

rkavanap:


AbfrmBlr:
There is a compilation failure

PR #2363, opened 2018-04-25
Issue#2291: Adapt to the new ConnectionFactory.connect API

saurabhagas:
New APIs and xsd changes to enable us to specify multiple IPv6 server addresses.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
@@ -18,12 +18,14 @@ package org.ehcache.clustered.client.config;
 
 import org.ehcache.CacheManager;
 import org.ehcache.PersistentCacheManager;
+import org.ehcache.clustered.client.internal.ConnectionSource;
 import org.ehcache.clustered.client.service.ClusteringService;
 import org.ehcache.config.builders.CacheManagerBuilder;
 import org.ehcache.config.builders.CacheManagerConfiguration;
 import org.ehcache.core.HumanReadable;
 import org.ehcache.spi.service.ServiceCreationConfiguration;
 
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.time.Duration;
 import java.util.Objects;
@@ -34,14 +36,13 @@ import org.ehcache.clustered.common.ServerSideConfiguration;
 /**
  * Specifies the configuration for a {@link ClusteringService}.
  */
-// TODO: Should this accept/hold a *list* of URIs?
 public class ClusteringServiceConfiguration
     implements ServiceCreationConfiguration<ClusteringService>,
     CacheManagerConfiguration<PersistentCacheManager>,
     HumanReadable {
 
   public static final boolean DEFAULT_AUTOCREATE = false;
-  private final URI clusterUri;
+  private final ConnectionSource connectionSource;
   private final boolean autoCreate;
   private final ServerSideConfiguration serverConfiguration;
   private final Timeouts timeouts;
@@ -59,6 +60,10 @@ public class ClusteringServiceConfiguration
     this(clusterUri, Timeouts.DEFAULT);
   }
 
+  public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String clusterTierManager) {
+    this(servers, clusterTierManager, Timeouts.DEFAULT);
+  }
+
   /**
    * Creates a {@code ClusteringServiceConfiguration} from the properties provided.
    *
@@ -70,6 +75,10 @@ public class ClusteringServiceConfiguration
    */
   public ClusteringServiceConfiguration(URI clusterUri, Timeouts timeouts) {
     this(clusterUri, timeouts, null);
+  }
+
+  public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String clusterTierManager, Timeouts timeouts) {
+    this(servers, clusterTierManager, timeouts, null);
   }
 
   /**
@@ -99,6 +108,11 @@ public class ClusteringServiceConfiguration
     this(clusterUri, timeouts, DEFAULT_AUTOCREATE, serverConfig);
   }
 
+  public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String clusterTierManager, Timeouts timeouts,
+                                        ServerSideConfiguration serverConfig) {
+    this(servers, clusterTierManager, timeouts, DEFAULT_AUTOCREATE, serverConfig);
+  }
+
   /**
    * Creates a {@code ClusteringServiceConfiguration} from the properties provided.
    *
@@ -111,6 +125,11 @@ public class ClusteringServiceConfiguration
    */
   public ClusteringServiceConfiguration(URI clusterUri, boolean autoCreate, ServerSideConfiguration serverConfig) {
     this(clusterUri, Timeouts.DEFAULT, autoCreate, serverConfig);
+  }
+
+  public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String clusterTierManager, boolean autoCreate,
+                                        ServerSideConfiguration serverConfig) {
+    this(servers, clusterTierManager, Timeouts.DEFAULT, autoCreate, serverConfig);
   }
 
   /**
@@ -128,6 +147,11 @@ public class ClusteringServiceConfiguration
     this(clusterUri, timeouts, autoCreate, serverConfig, new Properties());
   }
 
+  public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String clusterTierManager, Timeouts timeouts,
+                                        boolean autoCreate, ServerSideConfiguration serverConfig) {
+    this(servers, clusterTierManager, timeouts, autoCreate, serverConfig, new Properties());
+  }
+
   /**
    * Creates a {@code ClusteringServiceConfiguration} from the properties provided.
    *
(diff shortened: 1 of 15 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -0,0 +1,132 @@
...
+  public abstract String toString();
+
+  public static class ClusterUri extends ConnectionSource {
+
+    private final URI clusterUri;
+    private final String cacheManagerName;
+
+    public ClusterUri(URI clusterUri) {
+      this.clusterUri = Objects.requireNonNull(clusterUri, "Cluster URI cannot be null");
+      this.cacheManagerName = extractCacheManager(clusterUri);
```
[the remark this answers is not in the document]
mathieucarbou: I think it is wrong to use `cacheManagerName` because a cache manager does not have a name. Except for management, where we have an alias that enables us to identify the client part but this is a specific use case. The cache manager is only linked through the URI to an `entityName`. And the entity name could be something completely different than the cache manager alias set in management.

When connecting to a cluster, the cluster clientId used is composed of the entityName:

`connectionProperties.put(ConnectionPropertyNames.CONNECTION_NAME, CONNECTION_PREFIX + entityIdentifier);`

So I think yes it should be renamed everywhere because I think using a variable name `cacheManagerName` is wrong semantically speaking. 

But I'll let @chrisdennis and @cljohnso  and @cschanck decide ;-)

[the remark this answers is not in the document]
chrisdennis: @mathieucarbou I concur.  The two terms we tend to use internally (or we should use) are "clusterTierManager" when describing the thing that you are addressing here, and "entityIdentifier" the referring to the Voltron specific concept. I would lean towards referring to the `clusterTierManager` in public methods, and the `entityIdentifier` when moving closer to the implementation and the actual Voltron code.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -163,7 +207,26 @@ protected ClusteringServiceConfiguration(ClusteringServiceConfiguration baseConf
...
+  public ConnectionSource getConnectionSource() {
+    return connectionSource;
+  }
+
+  /**
+   * Cache manager name for the cache.
+   *
+   * @return the cache manager name
+   */
+  public String getCacheManagerName() {
```
chrisdennis: This shouldn't need to be exposed here since you already expose `ConnectionSource`

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -128,7 +135,7 @@ public ClusteringServiceConfigurationBuilder readOperationTimeout(long duration,
 
   @Override
   public ClusteringServiceConfiguration build() {
-    return new ClusteringServiceConfiguration(clusterUri, timeouts, autoCreate, null);
```
chrisdennis: If you want to extract out the commonality here I would just have this method call `build(null)`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -0,0 +1,132 @@
...
+
+public abstract class ConnectionSource {
+
+  public abstract String getCacheManagerName();
+
+  public abstract LeasedConnection connect(Properties connectionProperties) throws ConnectionException;
+
+  public abstract URI getClusterUri();
+
+  public abstract Iterable<InetSocketAddress> getServers();
```
chrisdennis: I don't think you need to expose this (and certainly not on both implementations).

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public abstract String toString();
```
chrisdennis: No need to define this, I understand it can be useful to force an override when anticipating lots of subclasses in the future, but we don't anticipate that.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -118,13 +123,51 @@ public URI getNamespace() {
...
-              final String urlValue = urlAttribute.getValue();
-              try {
-                connectionUri = new URI(urlValue);
-              } catch (URISyntaxException e) {
-                throw new XmlConfigurationException(
-                  String.format("Value of %s attribute on XML configuration element <%s> in <%s> is not a valid URI - '%s'",
-                    urlAttribute.getName(), item.getNodeName(), fragment.getTagName(), connectionUri), e);
+              final Attr serversAttribute = ((Element) item).getAttributeNode("servers");
+              final Attr cacheManagerAttribute = ((Element) item).getAttributeNode("cacheManager");
+              if (urlAttribute != null && serversAttribute != null) {
```
chrisdennis: These conditions should be able to be enforced by the XSD.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -82,13 +82,22 @@
...
   <xs:complexType name="connection-spec">
-    <xs:attribute name="url" type="tc:connectionUrl" use="required">
+    <xs:attribute name="url" type="tc:connectionUrl">
       <xs:annotation>
         <xs:documentation xml:lang="en">
           Identifies a single cluster member by URL.
         </xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="servers" type="tc:servers">
```
chrisdennis: Why not make this syntax mirror the syntax we use in tc-config (at least structurally).

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -82,13 +82,22 @@
   </xs:complexType>
 
   <xs:complexType name="connection-spec">
```
chrisdennis: We should really be enforcing one or the other of these in the XSD itself.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -82,13 +82,22 @@
...
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="servers" type="tc:servers">
+      <xs:annotation>
+        <xs:documentation xml:lang="en">
+          Identifies servers in the cluster.
+        </xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
+    <xs:attribute name="cacheManager" type="xs:string">
```
chrisdennis: If we have different element types for the two configuration types then we can make this attribute required on the server list form.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
saurabhagas: `ClusteringServiceConfiguration::readableString` depends on the `toString` implementation of `ConnectionSource`. I did this to force subclasses to implement `toString`. I'll remote it.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
saurabhagas: `ConnectionSource` represents the source of the connection, which could either be a URI, or an iterable of servers. I don't understand why having `getServers` in `ConnectionSource` is bad, or unneeded..

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
saurabhagas: Indeed. Will remove.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
chrisdennis: Because only one of your subclasses can implement it.  If you want you could make it a method on the subclass. It's opposite only exists because it predates the addition.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
chrisdennis: Right - it's like scaffolding around a building... useful when building the building... can be removed after that.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
saurabhagas: Will check.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
saurabhagas: Will do.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
saurabhagas: Done

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
saurabhagas: Done

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
saurabhagas: Done. (Ditto for other xsd-related comments as well).

Now the servers can be specified as:
```
<servers-connection>
  <servers>
    <server host="some.host" port="1234">
    <server host="other.host" port="1234">
    ...
  </servers>
  <cluster-tier-manager name="cM"/>
</servers-connection>
```

Note that the above is mutually exclusive (within the schema) with `<connection url="terracotta://some.host/cM" />`.

I had to create the new tag `<servers-connection>` because I couldn't find a way to specify mutually exclusive (at the XSD level) attributes and sub elements. I was look for
`<connection url="terracotta://some.host/cM" />` , OR
`<connection>
  <servers>
    <server host="some.host" port="1234">
    <server host="other.host" port="1234">
    ...
  </servers>
  <cluster-tier-manager name="cM"/>
</connection>`

If there is indeed a way to do the above in XSD, please let me know. If not, the current approach should be fine. I'm open to naming suggestions to replace `<servers-connection>`.

chrisdennis: I think the mutually exclusive attribute and element are indeed impossible. If I where you I'd take out a nesting level on the server list though leaving something like:

```xml
<connection url="terracotta://some.host/cM" />
```
or
```xml
<cluster cluster-tier-manager="cM">
  <server host="some.host" port="1234">
  <server host="other.host" port="1234">
</cluster>
```

(Naming obviously flexible)

saurabhagas: I made it to:
```
<cluster-connection cluster-tier-manager="cM">
  <server host="some.host" port="1234">
  <server host="other.host" port="1234">
</cluster-connection>
```
because this element is a sub-element of `<cluster>`.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -97,6 +95,51 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:complexType name="cluster-connection-spec">
+    <xs:sequence maxOccurs="1" minOccurs="1">
```
chrisdennis: These are the default min and max values no need to specify them.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -97,6 +95,51 @@
...
+      <xs:element name="server" type="tc:server" maxOccurs="unbounded" minOccurs="1"/>
```
chrisdennis: minOccurs=1 is the default so it's not needed.

saurabhagas: Indeed. Done.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
saurabhagas: Indeed. Done.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -127,6 +132,24 @@ public URI getNamespace() {
...
+            case "cluster-connection":
+              clusterTierManager = ((Element) item).getAttribute("cluster-tier-manager");
+              final NodeList serverNodes = item.getChildNodes();
+              for (int j = 0; j < serverNodes.getLength(); j++) {
+                final Node serverNode = serverNodes.item(j);
+                final String host = ((Element) serverNode).getAttributeNode("host").getValue();
+                final Attr port = ((Element) serverNode).getAttributeNode("port");
+                InetSocketAddress address;
+                if (port == null) {
+                  address = InetSocketAddress.createUnresolved(host, 0);
```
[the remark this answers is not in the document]
saurabhagas: `0`, yes. `AbstractConnectionService::connect` does the same.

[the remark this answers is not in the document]
saurabhagas: The default is set in `L2Data` class :)

PR #2369, opened 2018-04-26
Issue#2291: Added javadoc to the newly added public APIs

saurabhagas:
Added javadoc.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
@@ -60,6 +60,14 @@ public class ClusteringServiceConfiguration
     this(clusterUri, Timeouts.DEFAULT);
   }
 
+  /**
+   * Creates a {@code ClusteringServiceConfiguration} from the properties provided.
+   *
+   * @param servers the non-{@code null} iterable of servers in the cluster
+   * @param clusterTierManager the non-{@code null} cluster tier manager identifier
+   *
+   * @throws NullPointerException if {@code servers} is {@code null}
+   */
   public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, String clusterTierManager) {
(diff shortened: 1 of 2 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -60,6 +60,14 @@ public ClusteringServiceConfiguration(URI clusterUri) {
     this(clusterUri, Timeouts.DEFAULT);
   }
 
+  /**
+   * Creates a {@code ClusteringServiceConfiguration} from the properties provided.
+   *
+   * @param servers the non-{@code null} Iterable<InetSocketAddress> identifying the servers in the cluster
```
chrisdennis: I think stating the type like that is a little verbose maybe make it a little more prose like: `an iterable of the servers in the cluster`?

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -60,6 +60,14 @@ public ClusteringServiceConfiguration(URI clusterUri) {
...
+   * @param clusterTierManager the non-{@code null} clusterTierManager identifying the cache manager
```
chrisdennis: This (and the other instances) is a little too close to referring to the cache manager name.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
saurabhagas: Done

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
saurabhagas: Changed.

PR #2373, opened 2018-05-04
Xml refactoring

albinsuresh:
Refactoring the XML parsing logic from the mammoth classes `XmlConfiguration` and `ConfigurationParser` to smaller individual classes. Individual parsers were added for Service provider configurations, service configurations and resource configurations.

### DIFF
--- a/xml/src/main/java/org/ehcache/xml/model/CacheSpec.java
+++ b/xml/src/main/java/org/ehcache/xml/model/CacheSpec.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.xml.model;
+
+import org.ehcache.config.ResourcePool;
+import org.ehcache.config.ResourceUnit;
+import org.ehcache.config.units.EntryUnit;
+import org.ehcache.core.config.SizedResourcePoolImpl;
+import org.ehcache.spi.service.ServiceConfiguration;
+import org.ehcache.xml.CacheResourceConfigurationParser;
+import org.ehcache.xml.CacheServiceConfigurationParser;
+import org.ehcache.xml.CoreResourceConfigurationParser;
+import org.ehcache.xml.JaxbHelper;
+import org.ehcache.xml.resource.DefaultResourceConfigurationParser;
+import org.w3c.dom.Element;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import javax.xml.bind.Unmarshaller;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptySet;
+import static java.util.Collections.singleton;
+import static java.util.Optional.ofNullable;
+import static java.util.function.Function.identity;
+
+public class CacheSpec implements CacheTemplate {
+
+  protected final List<BaseCacheType> sources;
+  private final Map<URI, CacheServiceConfigurationParser<?>> serviceConfigParsers;
+
+  CoreResourceConfigurationParser coreResourceConfigurationParser;
+
+  public CacheSpec(Map<URI, CacheServiceConfigurationParser<?>> serviceConfigParsers,
+                   Map<URI, CacheResourceConfigurationParser> resourceConfigParsers, Unmarshaller unmarshaller, BaseCacheType... sources) {
+    this.serviceConfigParsers = serviceConfigParsers;
+    this.coreResourceConfigurationParser = new DefaultResourceConfigurationParser(resourceConfigParsers, unmarshaller);
+    this.sources = asList(sources);
+  }
+
+  @Override
+  public String keyType() {
+    return key().map(CacheEntryType::getValue).orElseGet(() -> extract(source -> JaxbHelper.findDefaultValue(source, "keyType")).orElse(null));
+  }
+
+  @Override
+  public String keySerializer() {
+    return key().map(CacheEntryType::getSerializer).orElse(null);
+  }
+
+  @Override
+  public String keyCopier() {
+    return key().map(CacheEntryType::getCopier).orElse(null);
+  }
+
+  private Optional<CacheEntryType> key() {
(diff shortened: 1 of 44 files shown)


xml/src/main/java/org/ehcache/xml/service/DefaultCacheEventListenerConfigurationParser.java
```diff
@@ -0,0 +1,81 @@
...
+    ListenersConfig listenersConfig = cacheDefinition.listenersConfig();
+    if(listenersConfig != null && listenersConfig.listeners() != null) {
+      for (ListenersType.Listener listener : listenersConfig.listeners()) {
+        @SuppressWarnings("unchecked")
+        final Class<CacheEventListener<?, ?>> cacheEventListenerClass =
+          (Class<CacheEventListener<?, ?>>) getClassForName(listener.getClazz(), cacheClassLoader);
+        final List<EventType> eventListToFireOn = listener.getEventsToFireOn();
+        Set<org.ehcache.event.EventType> eventSetToFireOn = new HashSet<>();
+        for (EventType events : eventListToFireOn) {
+          switch (events) {
```
henri-tremblay: Since all enums have the same name, we can replace the switch with `EvenType.valueOf(events.name())`.

xml/src/main/java/org/ehcache/xml/service/DefaultCacheEventListenerConfigurationParser.java
```diff
@@ -0,0 +1,81 @@
...
+                                                                   CacheConfigurationBuilder<K, V> cacheBuilder) throws ClassNotFoundException {
+    ListenersConfig listenersConfig = cacheDefinition.listenersConfig();
+    if(listenersConfig != null && listenersConfig.listeners() != null) {
+      for (ListenersType.Listener listener : listenersConfig.listeners()) {
+        @SuppressWarnings("unchecked")
+        final Class<CacheEventListener<?, ?>> cacheEventListenerClass =
+          (Class<CacheEventListener<?, ?>>) getClassForName(listener.getClazz(), cacheClassLoader);
+        final List<EventType> eventListToFireOn = listener.getEventsToFireOn();
+        Set<org.ehcache.event.EventType> eventSetToFireOn = new HashSet<>();
+        for (EventType events : eventListToFireOn) {
```
henri-tremblay: Remove the s of `events`

xml/src/test/java/org/ehcache/xml/provider/DefaultSerializationProviderConfigurationParserTest.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.xml.parsers.ParserConfigurationException;
+
+import static org.ehcache.config.builders.ConfigurationBuilder.newConfigurationBuilder;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+import static org.junit.Assert.*;
+
+public class DefaultSerializationProviderConfigurationParserTest {
+
+  ClassLoader classLoader = this.getClass().getClassLoader();
```
henri-tremblay: Can't we have a base test class to remove all the redundancy?

xml/src/test/java/org/ehcache/xml/provider/DefaultSerializationProviderConfigurationParserTest.java
```diff
@@ -0,0 +1,73 @@
...
+    Configuration xmlConfig = managerBuilder.build();
+
+    assertThat(xmlConfig.getServiceCreationConfigurations().size(), is(1));
+
+    ServiceCreationConfiguration configuration = xmlConfig.getServiceCreationConfigurations().iterator().next();
+
+    assertThat(configuration, instanceOf(DefaultSerializationProviderConfiguration.class));
+
+    DefaultSerializationProviderConfiguration factoryConfiguration = (DefaultSerializationProviderConfiguration) configuration;
+    assertThat(factoryConfiguration.getDefaultSerializers().size(), is(4));
```
henri-tremblay: assertj would be so much nicer here ;-) `assertThat(factoryConfiguration.getDefaultSerializers()).hasSize(4);`

xml/src/test/resources/configs/disk-persistent-cache.xml
```diff
@@ -0,0 +1,34 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<ehcache:config
+  xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
+  xmlns:ehcache='http://www.ehcache.org/v3'
+  xsi:schemaLocation="http://www.ehcache.org/v3 ../../../main/resources/ehcache-core.xsd">
+
+  <ehcache:persistence directory="some/dir"/>
```
henri-tremblay: I never understood why we use the `ehcache` namespace everywhere instead of setting it as a default namespace

xml/src/main/java/org/ehcache/xml/model/CacheTemplate.java
```diff
@@ -0,0 +1,55 @@
...
+
+  Collection<ResourcePool> resourcePools();
+
+  CacheLoaderWriterType.WriteBehind writeBehind();
+
+  DiskStoreSettingsType diskStoreSettings();
+
+  SizeOfEngineLimits heapStoreSettings();
+
+  class Impl extends BaseCacheSpec {
```
henri-tremblay: Why do we need `Impl` classes everywhere? Why not just having an implementation instead of an interface?

xml/src/main/java/org/ehcache/xml/model/CacheSpec.java
```diff
@@ -46,15 +45,15 @@
 import static java.util.Optional.ofNullable;
 import static java.util.function.Function.identity;
 
-public class BaseCacheSpec implements CacheTemplate {
+public class CacheSpec implements CacheTemplate {
```
henri-tremblay: Should it be abstract? Will it be instantiated directly?

xml/src/test/java/org/ehcache/xml/provider/CacheEventDispatcherFactoryConfigurationParserTest.java
```diff
@@ -17,47 +17,35 @@
...
-import java.net.URL;
 
 import javax.xml.bind.JAXBException;
 import javax.xml.parsers.ParserConfigurationException;
 
-import static org.ehcache.config.builders.ConfigurationBuilder.newConfigurationBuilder;
-import static org.hamcrest.core.Is.is;
-import static org.hamcrest.core.IsInstanceOf.instanceOf;
-import static org.junit.Assert.*;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
```
henri-tremblay: Nice! the import should actually be `org.assertj.core.api.Assertions` which will alias at the right place :-)

xml/src/main/java/org/ehcache/xml/model/CacheSpec.java
albinsuresh: It is used directly now after replacing all those interface usages with concrete classes following ur previous comment.

henri-tremblay: Good. Perfect

albinsuresh:
test this please

PR #2374, opened 2018-05-07
Always check key and value at the beginning of a Store method to be coherent.

henri-tremblay:


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1465,8 +1465,7 @@ protected long byteSized() {
   }
 
   @FindbugsSuppressWarnings("QF_QUESTIONABLE_FOR_LOOP")
-  protected void enforceCapacity() {
-    StoreEventSink<K, V> eventSink = storeEventDispatcher.eventSink();
```
chrisdennis: This intentionally creates a new event sink.  I can understand how a bunch of code around this area could be reworked such that this wasn't necessary, but right now it is.

henri-tremblay: Yes. It shouldn't be in this PR. It is a commit mistake. Will remove.

impl/src/main/java/org/ehcache/impl/internal/util/CheckerUtil.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @author Henri Tremblay
+ */
+public final class CheckerUtil {
+
+  public static void checkKey(Class<?> keyType, Object keyObject) {
+    Objects.requireNonNull(keyObject);
+
+    if (!keyType.isAssignableFrom(keyObject.getClass())) {
```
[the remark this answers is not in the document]
henri-tremblay: Funny. This is the original code. Let's fix it

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1523,21 +1527,11 @@ boolean evict(StoreEventSink<K, V> eventSink) {
...
   }
 
   private void checkValue(V valueObject) {
-    if (valueObject == null) {
-      throw new NullPointerException();
-    }
-    if (!valueType.isAssignableFrom(valueObject.getClass())) {
-      throw new ClassCastException("Invalid value type, expected : " + valueType.getName() + " but was : " + valueObject.getClass().getName());
-    }
+    CheckerUtil.checkKey(valueType, valueObject);
```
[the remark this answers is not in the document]
henri-tremblay: oups.

impl/src/test/java/org/ehcache/impl/internal/events/InvocationScopedEventSinkTest.java
```diff
@@ -73,4 +80,46 @@ public void testReset() {
...
   }
 
+  /**
+   * Make sure an interrupted sink sets the interrupted flag and keep both event queues in the state
+   * as of before the event that was interrupted.
+   *
+   * @throws InterruptedException
+   */
+  @Test
+  public void testInterruption() throws InterruptedException {
```
chrisdennis: We claimed this was broken right?  We figured the `removeLast()` call was wrong in a multithreaded environment?

henri-tremblay: I thought there should be not `removeLast()`. But in fact, it's ok. Since the event is already in `events` on interruption, it needs to be removed. I realized that while coding the test.

Then, is the class supposed to be thread-safe? Or thread local? I thought it was thread local.

chrisdennis:
Test this please.

PR #2375, opened 2018-05-08
Implement Java to XML translation of extension config elements

chaktatha:
The idea is to generate an XML config file from a ServiceCreationConfiguration and ServiceConfiguration.

This PR does the following:

1) SPIs and dummy implementation for the SPIs.
2) Basic orchestrator for the translation flow. This is the file ConfigurationTranslator.
3) Implementation in which xml documents will be generated from a given ServiceCreationConfiguration and from a given ServiceConfiguration.

### DIFF
--- a/107/src/main/java/org/ehcache/jsr107/internal/Jsr107CacheConfigurationParser.java
+++ b/107/src/main/java/org/ehcache/jsr107/internal/Jsr107CacheConfigurationParser.java
@@ -69,4 +69,14 @@ public class Jsr107CacheConfigurationParser implements CacheServiceConfiguration
           fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
     }
   }
+
+  @Override
+  public Class<Jsr107Service> getServiceType() {
+    return null;
+  }
+
+  @Override
+  public Element unparseServiceConfiguration(ServiceConfiguration<Jsr107Service> serviceConfiguration) {
+    throw new UnsupportedOperationException("Not yet implemented");
+  }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
@@ -13,23 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.ehcache.clustered.client.internal.config.xml;
 
-import org.ehcache.clustered.client.config.ClusteredStoreConfiguration;
 import org.ehcache.clustered.client.config.ClusteringServiceConfiguration;
 import org.ehcache.clustered.client.config.Timeouts;
 import org.ehcache.clustered.client.config.builders.TimeoutsBuilder;
-import org.ehcache.clustered.client.internal.store.ClusteredStore;
 import org.ehcache.clustered.client.service.ClusteringService;
-import org.ehcache.clustered.common.Consistency;
 import org.ehcache.clustered.common.ServerSideConfiguration;
-import org.ehcache.clustered.common.ServerSideConfiguration.Pool;
 import org.ehcache.config.units.MemoryUnit;
-import org.ehcache.spi.service.ServiceConfiguration;
 import org.ehcache.spi.service.ServiceCreationConfiguration;
 import org.ehcache.xml.CacheManagerServiceConfigurationParser;
-import org.ehcache.xml.CacheServiceConfigurationParser;
 import org.ehcache.xml.exceptions.XmlConfigurationException;
 import org.ehcache.xml.model.TimeType;
 import org.w3c.dom.Attr;
@@ -56,7 +49,8 @@ import javax.xml.bind.Unmarshaller;
 import javax.xml.transform.Source;
 import javax.xml.transform.stream.StreamSource;
 
-import static org.ehcache.clustered.client.internal.config.xml.ClusteredCacheConstants.*;
+import static org.ehcache.clustered.client.internal.config.xml.ClusteredCacheConstants.NAMESPACE;
+import static org.ehcache.clustered.client.internal.config.xml.ClusteredCacheConstants.XML_SCHEMA;
 import static org.ehcache.xml.XmlModel.convertToJavaTimeUnit;
 
 /**
@@ -64,11 +58,7 @@ import static org.ehcache.xml.XmlModel.convertToJavaTimeUnit;
  *
  * @see ClusteredCacheConstants#XSD
  */
-public class ClusteringServiceConfigurationParser implements CacheManagerServiceConfigurationParser<ClusteringService>,
-                                                              CacheServiceConfigurationParser<ClusteredStore.Provider> {
-
-  public static final String CLUSTERED_STORE_ELEMENT_NAME = "clustered-store";
-  public static final String CONSISTENCY_ATTRIBUTE_NAME = "consistency";
+public class ClusteringCacheManagerServiceConfigurationParser implements CacheManagerServiceConfigurationParser<ClusteringService> {
 
   @Override
   public Source getXmlSchema() throws IOException {
@@ -80,19 +70,6 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
     return NAMESPACE;
   }
 
-  @Override
-  public ServiceConfiguration<ClusteredStore.Provider> parseServiceConfiguration(Element fragment) {
-    if (CLUSTERED_STORE_ELEMENT_NAME.equals(fragment.getLocalName())) {
-      if (fragment.hasAttribute(CONSISTENCY_ATTRIBUTE_NAME)) {
-        return new ClusteredStoreConfiguration(Consistency.valueOf(fragment.getAttribute("consistency").toUpperCase()));
-      } else {
-        return new ClusteredStoreConfiguration();
-      }
-    }
-    throw new XmlConfigurationException(String.format("XML configuration element <%s> in <%s> is not supported",
-        fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
-  }
-
   /**
    * Complete interpretation of the top-level elements defined in <code>{@value ClusteredCacheConstants#XSD}</code>.
    * This method is called only for those elements from the namespace set by {@link ClusteredCacheConstants#NAMESPACE}.
@@ -100,7 +77,6 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
    * This method presumes the element presented is valid according to the XSD.
    *
    * @param fragment the XML fragment to process
-   *
    * @return a {@link org.ehcache.clustered.client.config.ClusteringServiceConfiguration ClusteringServiceConfiguration}
    */
   @Override
@@ -108,7 +84,7 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
 
     if ("cluster".equals(fragment.getLocalName())) {
 
-      ServerSideConfig serverConfig = null;
+      ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig serverConfig = null;
       URI connectionUri = null;
       List<InetSocketAddress> serverAddresses = new ArrayList<>();
       String clusterTierManager = null;
@@ -122,7 +98,7 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
               /*
                * <connection> is a required element in the XSD
                */
-              final Attr urlAttribute = ((Element) item).getAttributeNode("url");
+              final Attr urlAttribute = ((Element)item).getAttributeNode("url");
               final String urlValue = urlAttribute.getValue();
               try {
                 connectionUri = new URI(urlValue);
@@ -134,12 +110,12 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
 
               break;
             case "cluster-connection":
-              clusterTierManager = ((Element) item).getAttribute("cluster-tier-manager");
+              clusterTierManager = ((Element)item).getAttribute("cluster-tier-manager");
               final NodeList serverNodes = item.getChildNodes();
               for (int j = 0; j < serverNodes.getLength(); j++) {
                 final Node serverNode = serverNodes.item(j);
-                final String host = ((Element) serverNode).getAttributeNode("host").getValue();
-                final Attr port = ((Element) serverNode).getAttributeNode("port");
+                final String host = ((Element)serverNode).getAttributeNode("host").getValue();
+                final Attr port = ((Element)serverNode).getAttributeNode("port");
                 InetSocketAddress address;
                 if (port == null) {
                   address = InetSocketAddress.createUnresolved(host, 0);
@@ -181,7 +157,7 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
             default:
               throw new XmlConfigurationException(
                 String.format("Unknown XML configuration element <%s> in <%s>",
-                              item.getNodeName(), fragment.getTagName()));
+                  item.getNodeName(), fragment.getTagName()));
           }
         }
       }
@@ -213,47 +189,27 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
       }
     }
     throw new XmlConfigurationException(String.format("XML configuration element <%s> in <%s> is not supported",
-        fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
+      fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
   }
 
-  private Timeouts getTimeouts(Duration getTimeout, Duration putTimeout, Duration connectionTimeout) {
-    TimeoutsBuilder builder = TimeoutsBuilder.timeouts();
-    if (getTimeout != null) {
-      builder.read(getTimeout);
-    }
-    if(putTimeout != null) {
-      builder.write(putTimeout);
-    }
-    if(connectionTimeout != null) {
-      builder.connection(connectionTimeout);
-    }
-    return builder.build();
+  @Override
+  public Class<ClusteringService> getServiceType() {
+    return ClusteringService.class;
   }
 
-  private Duration processTimeout(Element parentElement, Node timeoutNode) {
-    try {
-      // <xxx-timeout> are direct subtype of ehcache:time-type; use JAXB to interpret it
-      JAXBContext context = JAXBContext.newInstance(TimeType.class.getPackage().getName());
-      Unmarshaller unmarshaller = context.createUnmarshaller();
-      JAXBElement<TimeType> jaxbElement = unmarshaller.unmarshal(timeoutNode, TimeType.class);
-
-      TimeType timeType = jaxbElement.getValue();
-      BigInteger amount = timeType.getValue();
-      if (amount.compareTo(BigInteger.valueOf(Long.MAX_VALUE)) > 0) {
-        throw new XmlConfigurationException(
-            String.format("Value of XML configuration element <%s> in <%s> exceeds allowed value - %s",
-                timeoutNode.getNodeName(), parentElement.getTagName(), amount));
-      }
-      return Duration.of(amount.longValue(), convertToJavaTimeUnit(timeType.getUnit()));
-
-    } catch (JAXBException e) {
-      throw new XmlConfigurationException(e);
-    }
+  /**
+   * Translates a {@link ServiceCreationConfiguration} to an xml element
+   *
+   * @param serviceCreationConfiguration
+   */
+  @Override
+  public Element unparseServiceCreationConfiguration(final ServiceCreationConfiguration<ClusteringService> serviceCreationConfiguration) {
+    throw new UnsupportedOperationException("Not yet implemented");
   }
 
-  private ServerSideConfig processServerSideConfig(Node serverSideConfigElement) {
-    ServerSideConfig serverSideConfig = new ServerSideConfig();
-    serverSideConfig.autoCreate = Boolean.parseBoolean(((Element) serverSideConfigElement).getAttribute("auto-create"));
+  private ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig processServerSideConfig(Node serverSideConfigElement) {
+    ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig serverSideConfig = new ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig();
+    serverSideConfig.autoCreate = Boolean.parseBoolean(((Element)serverSideConfigElement).getAttribute("auto-create"));
     final NodeList serverSideNodes = serverSideConfigElement.getChildNodes();
     for (int i = 0; i < serverSideNodes.getLength(); i++) {
       final Node item = serverSideNodes.item(i);
@@ -277,14 +233,14 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
             quantity = Long.parseLong(quantityValue);
           } catch (NumberFormatException e) {
             throw new XmlConfigurationException("Magnitude of value specified for <shared-pool name=\""
-                + poolName + "\"> is too large");
+                                                + poolName + "\"> is too large");
           }
 
-          Pool poolDefinition;
+          ServerSideConfiguration.Pool poolDefinition;
           if (fromResource == null) {
-            poolDefinition = new Pool(memoryUnit.toBytes(quantity));
+            poolDefinition = new ServerSideConfiguration.Pool(memoryUnit.toBytes(quantity));
           } else {
-            poolDefinition = new Pool(memoryUnit.toBytes(quantity), fromResource);
+            poolDefinition = new ServerSideConfiguration.Pool(memoryUnit.toBytes(quantity), fromResource);
           }
 
           if (serverSideConfig.pools.put(poolName, poolDefinition) != null) {
@@ -296,9 +252,44 @@ public class ClusteringServiceConfigurationParser implements CacheManagerService
     return serverSideConfig;
   }
 
+  private Duration processTimeout(Element parentElement, Node timeoutNode) {
+    try {
+      // <xxx-timeout> are direct subtype of ehcache:time-type; use JAXB to interpret it
+      JAXBContext context = JAXBContext.newInstance(TimeType.class.getPackage().getName());
+      Unmarshaller unmarshaller = context.createUnmarshaller();
+      JAXBElement<TimeType> jaxbElement = unmarshaller.unmarshal(timeoutNode, TimeType.class);
+
+      TimeType timeType = jaxbElement.getValue();
+      BigInteger amount = timeType.getValue();
+      if (amount.compareTo(BigInteger.valueOf(Long.MAX_VALUE)) > 0) {
+        throw new XmlConfigurationException(
+          String.format("Value of XML configuration element <%s> in <%s> exceeds allowed value - %s",
+            timeoutNode.getNodeName(), parentElement.getTagName(), amount));
+      }
+      return Duration.of(amount.longValue(), convertToJavaTimeUnit(timeType.getUnit()));
+
+    } catch (JAXBException e) {
+      throw new XmlConfigurationException(e);
+    }
+  }
+
+  private Timeouts getTimeouts(Duration getTimeout, Duration putTimeout, Duration connectionTimeout) {
+    TimeoutsBuilder builder = TimeoutsBuilder.timeouts();
+    if (getTimeout != null) {
+      builder.read(getTimeout);
+    }
+    if (putTimeout != null) {
+      builder.write(putTimeout);
+    }
+    if (connectionTimeout != null) {
+      builder.connection(connectionTimeout);
+    }
+    return builder.build();
+  }
+
   private static final class ServerSideConfig {
     private boolean autoCreate = false;
     private String defaultServerResource = null;
-    private final Map<String, Pool> pools = new HashMap<>();
+    private final Map<String, ServerSideConfiguration.Pool> pools = new HashMap<>();
   }
 }
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
@@ -18,17 +18,21 @@ package org.ehcache.clustered.client.internal.config.xml;
 import org.ehcache.clustered.client.config.ClusteringServiceConfiguration;
 import org.ehcache.clustered.client.config.Timeouts;
 import org.ehcache.clustered.client.config.builders.TimeoutsBuilder;
+import org.ehcache.clustered.client.internal.ConnectionSource;
 import org.ehcache.clustered.client.service.ClusteringService;
 import org.ehcache.clustered.common.ServerSideConfiguration;
 import org.ehcache.config.units.MemoryUnit;
 import org.ehcache.spi.service.ServiceCreationConfiguration;
 import org.ehcache.xml.CacheManagerServiceConfigurationParser;
+import org.ehcache.xml.DomUtil;
 import org.ehcache.xml.exceptions.XmlConfigurationException;
 import org.ehcache.xml.model.TimeType;
 import org.w3c.dom.Attr;
+import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
 
 import java.io.IOException;
 import java.math.BigInteger;
@@ -41,11 +45,15 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
+import javax.xml.XMLConstants;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBElement;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Unmarshaller;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.Source;
 import javax.xml.transform.stream.StreamSource;
 
@@ -59,6 +67,28 @@ import static org.ehcache.xml.XmlModel.convertToJavaTimeUnit;
  * @see ClusteredCacheConstants#XSD
  */
 public class ClusteringCacheManagerServiceConfigurationParser implements CacheManagerServiceConfigurationParser<ClusteringService> {
+
+  public static final String TC_CLUSTERED_NAMESPACE_PREFIX = "tc";
+  public static final String CLUSTER_ELEMENT_NAME = "cluster";
+  public static final String CONNECTION_ELEMENT_NAME = "connection";
+  public static final String CLUSTER_CONNECTION_ELEMENT_NAME = "cluster-connection";
+  public static final String CLUSTER_TIER_MANAGER_ATTRIBUTE_NAME = "cluster-tier-manager";
+  public static final String SERVER_ELEMENT_NAME = "server";
+  public static final String HOST_ATTRIBUTE_NAME = "host";
+  public static final String PORT_ATTRIBUTE_NAME = "port";
+  public static final String READ_TIMEOUT_ELEMENT_NAME = "read-timeout";
+  public static final String WRITE_TIMEOUT_ELEMENT_NAME = "write-timeout";
+  public static final String CONNECTION_TIMEOUT_ELEMENT_NAME = "connection-timeout";
+  public static final String URL_ATTRIBUTE_NAME = "url";
+  public static final String DEFAULT_RESOURCE_ELEMENT_NAME = "default-resource";
+  public static final String SHARED_POOL_ELEMENT_NAME = "shared-pool";
+  public static final String SERVER_SIDE_CONFIG = "server-side-config";
+  public static final String AUTO_CREATE_ATTRIBUTE_NAME = "auto-create";
+  public static final String UNIT_ATTRIBUTE_NAME = "unit";
+  public static final String NAME_ATTRIBUTE_NAME = "name";
+  public static final String FROM_ATTRIBUTE_NAME = "from";
+  public static final String DEFAULT_UNIT_ATTRIBUTE_VALUE = "seconds";
+  public static final String COLON = ":";
 
   @Override
   public Source getXmlSchema() throws IOException {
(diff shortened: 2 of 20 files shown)


henri-tremblay:
Can you please improve the title and add a description to help us understand what this is about? Thanks.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+    doc.appendChild(rootElement);
+    return rootElement;
+  }
+
+  private Element createServerElement(Document doc, ClusteringServiceConfiguration clusteringServiceConfiguration) {
+    if (!ConnectionSource.ServerList.class.isInstance(clusteringServiceConfiguration.getConnectionSource())) {
+      throw new IllegalArgumentException("When connection URL is null, source of connection MUST be of type ConnectionSource.ServerList.class");
+    }
+    ConnectionSource.ServerList servers = (ConnectionSource.ServerList)clusteringServiceConfiguration.getConnectionSource();
+    if (servers.getServers() == null || !servers.getServers().iterator().hasNext()) {
```
henri-tremblay: I don't think there is a good reason for ServerList.getServers to return an `Iterable`. So can switch it to Collection or Set

henri-tremblay: The list of servers can't be null

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+    }
+    Element rootElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + CLUSTER_ELEMENT_NAME);
+    rootElement.setAttributeNS(XMLConstants.XMLNS_ATTRIBUTE_NS_URI, "xmlns:" + TC_CLUSTERED_NAMESPACE_PREFIX, NAMESPACE.toString());
+    Element connElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + CLUSTER_CONNECTION_ELEMENT_NAME);
+    connElement.setAttribute(CLUSTER_TIER_MANAGER_ATTRIBUTE_NAME, clusteringServiceConfiguration.getConnectionSource()
+      .getClusterTierManager());
+    servers.getServers().forEach(server -> {
+      Element serverElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + SERVER_ELEMENT_NAME);
+      serverElement.setAttribute(HOST_ATTRIBUTE_NAME, server.getHostName());
+      if ((server.getPort() < 0)) {
```
henri-tremblay: too many parenthesis

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+    if (clusteringServiceConfiguration.getServerConfiguration() != null) {
+      ServerSideConfiguration serverSideConfiguration = clusteringServiceConfiguration.getServerConfiguration();
+      String defaultServerResource = serverSideConfiguration.getDefaultServerResource();
+      if (!(defaultServerResource == null || defaultServerResource.trim().length() == 0)) {
+        Element defaultResourceElement = createDefaultServerResourceElement(doc, defaultServerResource);
+        serverSideConfigurationElem.appendChild(defaultResourceElement);
+      }
+      Map<String, Pool> resourcePools = serverSideConfiguration.getResourcePools();
+      if (!(resourcePools == null || resourcePools.isEmpty())) {
+        for (Map.Entry<String, Pool> entry : resourcePools.entrySet()) {
```
henri-tremblay: resourcePools.forEach?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+        }
+      }
+    }
+    return serverSideConfigurationElem;
+  }
 
+  private Element createServerSideConfigurationElement(Document doc, ClusteringServiceConfiguration clusteringServiceConfiguration) {
+    Element serverSideConfigurationElem = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + SERVER_SIDE_CONFIG);
+    serverSideConfigurationElem.setAttribute(AUTO_CREATE_ATTRIBUTE_NAME, clusteringServiceConfiguration.isAutoCreate()
+      ? Boolean.TRUE.toString() : Boolean.FALSE.toString());
```
henri-tremblay: `Boolean.toString(clusteringServiceConfiguration.isAutoCreate())`

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
   }
 
+
+  private Element createSharedPoolElement(Document doc, Map.Entry<String, Pool> entry) {
+    String poolName = entry.getKey();
+    Pool pool = entry.getValue();
+    Element poolElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + SHARED_POOL_ELEMENT_NAME);
+    poolElement.setAttribute(NAME_ATTRIBUTE_NAME, poolName);
+    String from = pool.getServerResource();
+    if (!(from == null || from.trim().length() == 0)) {
```
henri-tremblay: I've seen a lot of trim. But this is Java configuration. Are we sure there can be empty space? Shouldn't then be validated when setting it on the pool in first place?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+      poolElement.setAttribute(FROM_ATTRIBUTE_NAME, from);
+    }
+    long memoryInMB = MemoryUnit.MB.convert(pool.getSize(), MemoryUnit.B);
+    poolElement.setAttribute(UNIT_ATTRIBUTE_NAME, MemoryUnit.MB.toString());
+    poolElement.setTextContent(Long.toString(memoryInMB));
```
henri-tremblay: If it's less than a MB, converting to MB isn't a good idea.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+   *
+   * @param doc
+   *
+   * @param serviceConfiguration
+   */
   @Override
   public void translateServiceConfiguration(Document doc, ServiceConfiguration<ClusteredStore.Provider> serviceConfiguration) {
+    validateParametersForTranslationToServiceConfig(doc, serviceConfiguration);
+    ClusteredStoreConfiguration clusteredStoreConfiguration = (ClusteredStoreConfiguration)serviceConfiguration;
+    Consistency consistency = clusteredStoreConfiguration.getConsistency();
```
henri-tremblay: Same here. Consistency can't be null

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -424,6 +443,227 @@ public void testServersWithClusterTierManagerAndOptionalPorts() throws Exception
...
+      .timeouts(Timeouts.DEFAULT)
+      .autoCreate()
+      .defaultServerResource("main")
+      .resourcePool("primaryresource", 5, MemoryUnit.GB)
+      .resourcePool("seondaryresource", 10, MemoryUnit.GB, "optional")
+      .build();
+    Configuration clusterConfiguration = getClusterConfiguration(serviceConfig);
+    ConfigurationTranslator configurationTranslator = new ConfigurationTranslator(clusterConfiguration);
+    String retValue = configurationTranslator.toXml();
+    System.out.println(retValue);
```
henri-tremblay: Remove

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -424,6 +443,227 @@ public void testServersWithClusterTierManagerAndOptionalPorts() throws Exception
...
+    assertThat(retPoolMap, is(notNullValue()));
+    assertThat(retPoolMap.size(), is(equalTo(2)));
+
+    for (Map.Entry<String, ServerSideConfiguration.Pool> entry : retPoolMap.entrySet()) {
+      String name = entry.getKey();
+      ServerSideConfiguration.Pool value = entry.getValue();
+      assertThat(("primaryresource".equals(name) || "seondaryresource".equals(name)), is(true));
+      if ("primaryresource".equals(name)) {
+        assertThat(5 * 1024 * 1024 * 1024L, is(equalTo(value.getSize())));
+      } else if ("seondaryresource".equals(name)) {
```
henri-tremblay: secondaryresource

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -424,6 +443,227 @@ public void testServersWithClusterTierManagerAndOptionalPorts() throws Exception
...
+
+    assertThat(retPoolMap, is(notNullValue()));
+    assertThat(retPoolMap.size(), is(equalTo(2)));
+
+    for (Map.Entry<String, ServerSideConfiguration.Pool> entry : retPoolMap.entrySet()) {
+      String name = entry.getKey();
+      ServerSideConfiguration.Pool value = entry.getValue();
+      assertThat(("primaryresource".equals(name) || "seondaryresource".equals(name)), is(true));
+      if ("primaryresource".equals(name)) {
+        assertThat(5 * 1024 * 1024 * 1024L, is(equalTo(value.getSize())));
```
henri-tremblay: assertj? :-)

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -424,6 +443,227 @@ public void testServersWithClusterTierManagerAndOptionalPorts() throws Exception
...
+  @Test
+  public void testTranslateServiceCreationConfigurationWithNoServerSideConfig() throws Exception {
+    URI connectionUri = new URI("terracotta://localhost:9510/my-application");
+    ClusteringServiceConfiguration serviceConfig = ClusteringServiceConfigurationBuilder.cluster(connectionUri)
+      .timeouts(Timeouts.DEFAULT)
+      .build();
+    Configuration clusterConfiguration = getClusterConfiguration(serviceConfig);
+    ConfigurationTranslator configurationTranslator = new ConfigurationTranslator(clusterConfiguration);
+    String retValue = configurationTranslator.toXml();
+    System.out.println(retValue);
```
henri-tremblay: Remove all of them

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -424,6 +443,227 @@ public void testServersWithClusterTierManagerAndOptionalPorts() throws Exception
...
+    String startTag = "<ehcache:config xmlns:ehcache=\"http://www.ehcache.org/v3\"><ehcache:service>";
+    String endTag = "</ehcache:service></ehcache:config>";
```
henri-tremblay: Can you please refactor the start. end tag mechanic into a private method?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
```diff
@@ -455,4 +695,46 @@ private URL makeConfig(final String[] lines) throws IOException {
...
+  private Configuration getClusterConfiguration(final ClusteringServiceConfiguration serviceConfig) {
+    Configuration clusterConfiguration = new Configuration() {
+      @Override
+      public Map<String, CacheConfiguration<?, ?>> getCacheConfigurations() {
+        return null;
+      }
+
+      @Override
+      public Collection<ServiceCreationConfiguration<?>> getServiceCreationConfigurations() {
+        Collection<ServiceCreationConfiguration<?>> configurations = new ArrayList<>();
```
henri-tremblay: return Collections.singleton(serviceConfig)

transactions/src/main/java/org/ehcache/transactions/xa/internal/xml/TxCacheServiceConfigurationParser.java
```diff
@@ -58,4 +59,14 @@ public URI getNamespace() {
...
     }
   }
+
+  @Override
+  public Class<XAStore.Provider> handleServiceConfigurationType() {
+    return null;
+  }
+
+  @Override
+  public void translateServiceConfiguration(Document doc, ServiceConfiguration<XAStore.Provider> serviceConfiguration) {
```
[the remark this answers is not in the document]
chaktatha: It is a dummy implementations will be converted actual implementation very soon. That's why not putting UnsupportedOperationException. It is not exactly unsupported, it is not yet implemented.

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,216 @@
...
+import static org.ehcache.core.internal.util.ClassLoading.libraryServiceLoaderFor;
+
+public class ConfigurationTranslator {
+
+  private static final SchemaFactory XSD_SCHEMA_FACTORY = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+  private static final String INDENT_PROPERTY_KEY = "{http://xml.apache.org/xslt}indent-amount";
+  private static final String INDENT_SPACING = "2";
+  private static final String YES = "yes";
+
+  private static Schema newSchema(Source[] schemas) throws SAXException {
```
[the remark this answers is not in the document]
chaktatha: Sure

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+    for (CacheServiceConfigurationParser<?> parser : libraryServiceLoaderFor(CacheServiceConfigurationParser.class)) {
+      schemaSources.add(parser.getXmlSchema());
+      this.cacheServiceConfigurationParserMap.put(parser.handleServiceConfigurationType(), parser);
+    }
+    // Parsers for /config/cache/resources extensions
+    for (CacheResourceConfigurationParser parser : libraryServiceLoaderFor(CacheResourceConfigurationParser.class)) {
+      schemaSources.add(parser.getXmlSchema());
+      if(parser.handleResourceTypes() != null){
+        parser.handleResourceTypes().stream().forEach(
```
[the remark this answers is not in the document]
chaktatha: Ok, will change

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,216 @@
...
+
+    this.domBuilder = createAndGetDocumentBuilder(schemaSources);
+  }
+
+  public String toXml() {
+    StringBuilder sb = new StringBuilder();
+    if (configuration.getServiceCreationConfigurations() != null) {
+      for (ServiceCreationConfiguration<?> serviceCreationConfiguration : configuration.getServiceCreationConfigurations()) {
+        String xmlPart = getClusterCreationConfigurationXml(serviceCreationConfiguration);
+        if (!(xmlPart == null || xmlPart.trim().length() == 0)) {
```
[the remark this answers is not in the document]
chaktatha: Will change

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
chaktatha: Sure, wll do it

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
chaktatha: Sure

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
chaktatha: Will correct the type-o

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
chaktatha: I did not see any validation other than not null check. Making sure that empty spaces are not accidentally set.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
chaktatha: Thanks :)

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -227,14 +251,197 @@ public URI getNamespace() {
...
+    rootElement.setAttributeNS(XMLConstants.XMLNS_ATTRIBUTE_NS_URI, "xmlns:" + TC_CLUSTERED_NAMESPACE_PREFIX, NAMESPACE.toString());
+    Element urlElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + CONNECTION_ELEMENT_NAME);
+    urlElement.setAttribute(URL_ATTRIBUTE_NAME, clusteringServiceConfiguration.getClusterUri().toString());
+    rootElement.appendChild(urlElement);
+    doc.appendChild(rootElement);
+    return rootElement;
+  }
+
+  private Element createServerElement(Document doc, ClusteringServiceConfiguration clusteringServiceConfiguration) {
+    if (!ConnectionSource.ServerList.class.isInstance(clusteringServiceConfiguration.getConnectionSource())) {
```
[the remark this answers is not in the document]
chaktatha: Sure

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,216 @@
...
+      this.cacheServiceConfigurationParserMap.get(serviceConfiguration.getServiceType());
+    Document doc = domBuilder.newDocument();
+    cacheServiceConfigurationParser.translateServiceConfiguration(doc, serviceConfiguration);
+    return getStringFromDocument(doc);
+  }
+
+  protected String getStringFromDocument(Document doc) {
+    try {
+      DOMSource domSource = new DOMSource(doc);
+      StreamResult result = createAndGetStreamResult();
```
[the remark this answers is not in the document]
chaktatha: I will take care of this

xml/src/main/java/org/ehcache/xml/CacheManagerServiceConfigurationParser.java
```diff
@@ -34,4 +34,8 @@
   URI getNamespace();
 
   ServiceCreationConfiguration<T> parseServiceCreationConfiguration(Element fragment);
+
+  Class<T> handleServiceCreationConfigurationType();
```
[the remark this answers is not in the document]
chaktatha: Yes, that was a good option. As I have checked in the dummy implementations and dummy implementations will be converted actual implementation very soon, I am planning not to revert the already changed file. Let me know what you think

chaktatha:
In this project, Jaxb is used to translate core configurations only, not for extended configurations. This class is about extended configurations; thst's why we stuck to the convention and did not use Jaxb

clustered/client/src/test/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParserTest.java
chaktatha: :)
I mistakenly used assertThat from hamcrest, originally the assert used in this file was org.junit.Assert.assertThat. I am bringing back the old assertThat.

With respect to assertj,  if we want to use assertj, then imo, that should be used for entire file (i.e. for the older existing tests as well). Do you want to me to use assertj for the entire file? If yes, then can we take it as a different task as that will be a general refactoring?

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,216 @@
...
+    Collection<Source> schemaSources = new ArrayList<>();
+    schemaSources.add(new StreamSource(CORE_SCHEMA_URL.openStream()));
+
+    initCacheMaps(schemaSources);
+
+    this.domBuilder = createAndGetDocumentBuilder(schemaSources);
+  }
+
+  public String toXml() {
+    StringBuilder sb = new StringBuilder();
```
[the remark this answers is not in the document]
chaktatha: I will add a size

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
chaktatha: ServerList.gerServers was introduced as part of IPv6 support. Platform api takes an Iterable of InetSocketAddress as an input for creating connection and that's why an Iterable is kept inside ServerList.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
```diff
@@ -93,6 +119,16 @@ public URI getNamespace() {
         fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
   }
 
+  @Override
+  public Class<ClusteredStore.Provider> handleServiceConfigurationType() {
```
albinsuresh: This single class is so huge since it's still handling the parsing and translation of both `ClusteredStoreConfiguration` and `ClusteringServiceConfiguration`. Splitting this class to handle both separately would get us some smaller simpler classes. I know that it wasn't your decision to make it implement both `CacheManagerServiceConfigurationParser` and `CacheServiceConfigurationParser`. But looking at the current size it feels like it's time to split.

chaktatha:
If you are happy with all the changes. Mark it approved but please do not merge. I will squash the commits and convert the changes to sequence of meaningful (i.e. each commit will have related changes) commits. Please merge after that.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
chaktatha: I can make the changes. Please let me know if you want the refactoring to be done as part of same PR

albinsuresh: It can be a follow-up PR if that's too much work. But I feel that it isn't.

transactions/src/main/java/org/ehcache/transactions/xa/internal/xml/TxCacheServiceConfigurationParser.java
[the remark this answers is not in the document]
henri-tremblay: Still, right now it isn't. You can put a message in the exception "Not implemented yet"

xml/src/main/java/org/ehcache/xml/CacheManagerServiceConfigurationParser.java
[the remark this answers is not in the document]
henri-tremblay: If all the implemented will be, that's fine. And you probably don't need a default method then.

xml/src/main/java/org/ehcache/xml/CacheManagerServiceConfigurationParser.java
```diff
@@ -34,4 +34,8 @@
...
+
+  void translateServiceConfiguration(Document doc, ServiceCreationConfiguration<T> serviceCreationConfiguration);
```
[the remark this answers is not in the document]
henri-tremblay: unmarshal?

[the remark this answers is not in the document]
chrisdennis: Huh? In this context would it be:
`marshal == unparse`
`parse == unmarshal`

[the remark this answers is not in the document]
henri-tremblay: I'm slightly lost but
* From xml to bean = unmarshal
* From bean to xml = marshal

Are we in sync?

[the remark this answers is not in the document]
chrisdennis: Yes I think we are... but just to confirm if you wanted to switch to the marshal terminology:
```java
Document marshalServiceCreationConfiguration(ServiceCreationConfiguration<T> config);
ServiceCreationConfiguration<T> unmarshalServiceCreationConfiguration(Document config);
```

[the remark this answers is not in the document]
chaktatha: So I will alter the api and new api will look like:

void marshalServiceConfiguration(Document doc, ServiceCreationConfiguration<T> serviceCreationConfiguration); 

Let me know if it is otherwise

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringServiceConfigurationParser.java
chaktatha: Ok I will split this into two classes. one implementing CacheManagerServiceConfigurationParser and other CacheServiceConfigurationParser.

107/src/main/java/org/ehcache/jsr107/internal/Jsr107CacheConfigurationParser.java
```diff
@@ -52,7 +54,7 @@ public URI getNamespace() {
   }
 
   @Override
-  public ServiceConfiguration<Jsr107Service> parseServiceConfiguration(Element fragment) {
+  public ServiceConfiguration<Jsr107Service> unmarshalServiceConfiguration(Element fragment) {
```
chrisdennis: You can't change the name of an SPI method without good reason they're considered semi-stable. Not as stable as public API, but we don't break things without good reason.

chrisdennis: This is why I recommended parse and unparse - it fits with the current naming scheme.

107/src/main/java/org/ehcache/jsr107/internal/Jsr107CacheConfigurationParser.java
```diff
@@ -69,4 +71,14 @@ public URI getNamespace() {
...
     }
   }
+
+  @Override
+  public Class<Jsr107Service> getServiceType() {
+    return null;
+  }
+
+  @Override
+  public Node marshalServiceConfiguration(Document doc, ServiceConfiguration<Jsr107Service> serviceConfiguration) {
```
chrisdennis: Why are you passing the whole document to the method?  It doesn't need (and indeed shouldn't have) access to anything else about the configuration.

chrisdennis: Also, isn't it required that this return an element, a Node isn't precise enough.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
```diff
@@ -0,0 +1,466 @@
...
+      } catch (IllegalArgumentException e) {
+        throw new XmlConfigurationException(e);
+      }
+    }
+    throw new XmlConfigurationException(String.format("XML configuration element <%s> in <%s> is not supported",
+      fragment.getTagName(), (fragment.getParentNode() == null ? "null" : fragment.getParentNode().getLocalName())));
+  }
+
+  @Override
+  public Class<ClusteringService> getServiceCreationType() {
```
chrisdennis: It's not a service creation type... it's a service type.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
```diff
@@ -0,0 +1,466 @@
...
+    Element rootElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + CLUSTER_ELEMENT_NAME);
+    rootElement.setAttributeNS(XMLConstants.XMLNS_ATTRIBUTE_NS_URI, "xmlns:" + TC_CLUSTERED_NAMESPACE_PREFIX, NAMESPACE.toString());
+    Element connElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + CLUSTER_CONNECTION_ELEMENT_NAME);
+    connElement.setAttribute(CLUSTER_TIER_MANAGER_ATTRIBUTE_NAME, clusteringServiceConfiguration.getConnectionSource()
+      .getClusterTierManager());
+    servers.getServers().forEach(server -> {
+      Element serverElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + SERVER_ELEMENT_NAME);
+      serverElement.setAttribute(HOST_ATTRIBUTE_NAME, server.getHostName());
+      if (server.getPort() < 0) {
+        throw new IllegalArgumentException("Port for server:" + server.getHostName() + " must not be negative");
```
chrisdennis: The schema will validate this for us won't it?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
```diff
@@ -0,0 +1,466 @@
...
+    rootElement.appendChild(urlElement);
+    return rootElement;
+  }
+
+  private Element createServerElement(Document doc, ClusteringServiceConfiguration clusteringServiceConfiguration) {
+    if (!(clusteringServiceConfiguration.getConnectionSource() instanceof ConnectionSource.ServerList)) {
+      throw new IllegalArgumentException("When connection URL is null, source of connection MUST be of type ConnectionSource.ServerList.class");
+    }
+    ConnectionSource.ServerList servers = (ConnectionSource.ServerList)clusteringServiceConfiguration.getConnectionSource();
+    if (!servers.getServers().iterator().hasNext()) {
```
chrisdennis: Can't we let the xml schema validate this for us?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
```diff
@@ -0,0 +1,466 @@
...
+
+    if (clusteringServiceConfiguration.getServerConfiguration() != null) {
+      ServerSideConfiguration serverSideConfiguration = clusteringServiceConfiguration.getServerConfiguration();
+      String defaultServerResource = serverSideConfiguration.getDefaultServerResource();
+      if (!(defaultServerResource == null || defaultServerResource.trim().length() == 0)) {
+        Element defaultResourceElement = createDefaultServerResourceElement(doc, defaultServerResource);
+        serverSideConfigurationElem.appendChild(defaultResourceElement);
+      }
+      Map<String, ServerSideConfiguration.Pool> resourcePools = serverSideConfiguration.getResourcePools();
+      if (!(resourcePools == null || resourcePools.isEmpty())) {
```
chrisdennis: No need to call `isEmpty()` the forEach works just as well on an empty collection. Also, is it legal for getResourcePools() to return null?  If not why allow it - just let this code blow up with an NPE.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
```diff
@@ -0,0 +1,466 @@
...
+      .isAutoCreate()));
+    return serverSideConfigurationElem;
+  }
+
+
+  private Element createSharedPoolElement(Document doc, String poolName, ServerSideConfiguration.Pool pool) {
+    Element poolElement = doc.createElement(TC_CLUSTERED_NAMESPACE_PREFIX + COLON + SHARED_POOL_ELEMENT_NAME);
+    poolElement.setAttribute(NAME_ATTRIBUTE_NAME, poolName);
+    String from = pool.getServerResource();
+    if (!(from == null || from.trim().length() == 0)) {
```
chrisdennis: Why are we declaring strings of white space characters to be the same as requesting the default pool?

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,190 @@
...
+  private static final URL CORE_SCHEMA_URL = XmlConfiguration.class.getResource("/ehcache-core.xsd");
+
+  private final DocumentBuilder domBuilder;
+  private final Configuration configuration;
+
+  private final Map<Class<? extends Service>, CacheManagerServiceConfigurationParser<?>> cacheManagerServiceConfigurationParserMap;
+  private final Map<Class<? extends Service>, CacheServiceConfigurationParser<?>> cacheServiceConfigurationParserMap;
+  private final Map<ResourceType<?>, CacheResourceConfigurationParser> cacheResourceConfigurationParserMap;
+
+  public ConfigurationTranslator(Configuration configuration) throws IOException, SAXException, ParserConfigurationException {
```
chrisdennis: I''m confused about this class... this PR add xml generation for a subset of the full xml configuration prior to the calling code being ready.  In that sense this class is going to be wholly replaced when Albin's code is merged?  In an ideal world those two PRs would have arrived in the opposite order to how we're getting them.  In this reality wouldn't it make more sense to either:

1. Have this PR be raised against Albin's branch.
2. Leave this class out of this PR so that it's 'correct' but detached, rather than using this class to 'attach' it, but in an invalid way.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chaktatha: Can a pool-resource be such that it is defined only using white spaces? If not, then pointing it to default one might be an option. Another way can deal with it is: If null then go for default; otherwise if the from property is a space only string, throw exception. What do you suggest?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chaktatha: I wanted to make explicit check so that I can throw an exception with custom message, instead a generic parser validation error. Let me know if I should depend only on schema validator.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chaktatha: Same as above. 
I wanted to make explicit check so that I can throw an exception with custom message, instead a generic parser validation error. Let me know if I should depend only on schema validator.

107/src/main/java/org/ehcache/jsr107/internal/Jsr107CacheConfigurationParser.java
chaktatha: I was not sure whether to go for marshal/unmarshal or parse/unparse. Now it is clear. I will make the change. Methods names will start with parse/unparse

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chaktatha: I will make the change

107/src/main/java/org/ehcache/jsr107/internal/Jsr107CacheConfigurationParser.java
chaktatha: We were getting problem if the Document object was not passed from the caller. If a document object was not passed and it was being created inside the method (to create an element from the document), then it was getting error; reference to elements in ehcache-core.xsd was not getting visible. We found that while writing the unit tests. 

Node is more general. That's why I put it in the return signature. Can change it to Element

chrisdennis: The method you are looking for is: Document.importNode(...)

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
chaktatha: This class was required to do some flow testing. Myself and Albin were working simultaneously. As I already wrote a big part of it, we decided to push it in my PR. Albin could add functionalities on top of it. Anyways, as you said, I will ask Albin to put this change in his PR

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chrisdennis: I would always lean toward having these kinds of rules in as few places as possible, then lets see what the errors look like... if they are to impenetrable we can make a decision then.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chrisdennis: For now I would let the existing codification of these rules enforce them rather than duplicating that information.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
chrisdennis: Via xml maybe not, but programaticaly it probably works... regardless if you see it the only correct behaviors are "do the right thing" or "fail", "do something entirely different" is not a valid option.

PR #2383, opened 2018-05-18
Xml translation API and implementation

albinsuresh:


impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -193,13 +193,13 @@ public String toString() {
     }
   }
 
-  private static final class TimeToLiveExpiryPolicy extends BaseExpiryPolicy {
+  public static final class TimeToLiveExpiryPolicy extends BaseExpiryPolicy {
```
chrisdennis: Can we do this without exposing these classes?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -72,14 +73,16 @@ private static Schema newSchema(Source[] schemas) throws SAXException {
     }
   }
 
-  private static final URL CORE_SCHEMA_URL = XmlConfiguration.class.getResource("/ehcache-core.xsd");
+  public static final URL CORE_SCHEMA_URL = XmlConfiguration.class.getResource("/ehcache-core.xsd");
```
chrisdennis: I suspect there is some super-type we can pull out of this and the ConfigurationTranslator. It also seems weird that the translator has an unmarshaller, and the parser has a marshaller.

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
```diff
@@ -0,0 +1,154 @@
...
+  private final Unmarshaller unmarshaller;
+
+  private final Map<Class<?>, CacheManagerServiceConfigurationParser> xmlParsers = new HashMap<>();
+  private final Map<Class<?>, CacheServiceConfigurationParser> cacheXmlParsers = new HashMap<>();
+  private final ResourceConfigurationParser resourceConfigurationParser;
+
+  public ConfigurationTranslator(Configuration configuration) throws JAXBException, IOException {
+    this.configuration = configuration;
+    JAXBContext jaxbContext = JAXBContext.newInstance(CORE_SCHEMA_JAXB_MODEL_PACKAGE);
+    marshaller = jaxbContext.createMarshaller();
```
chrisdennis: We should probably be validating the configuration we generate against the schema when marshalling.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -0,0 +1,100 @@
...
+      if (expiryPolicy == ExpiryPolicy.NO_EXPIRY) {
+        expiryType.withNone(new ExpiryType.None());
+      } else if (expiryPolicy instanceof ExpiryPolicyBuilder.TimeToLiveExpiryPolicy) {
+        Duration ttl = expiryPolicy.getExpiryForCreation(null, null);
+        expiryType.withTtl(new TimeType().withValue(BigInteger.valueOf(ttl.getSeconds())).withUnit(TimeUnit.SECONDS));
+      } else if (expiryPolicy instanceof ExpiryPolicyBuilder.TimeToIdleExpiryPolicy) {
+        Duration tti = expiryPolicy.getExpiryForCreation(null, null);
+        expiryType.withTti(new TimeType().withValue(BigInteger.valueOf(tti.getSeconds())).withUnit(TimeUnit.SECONDS));
+      } else {
+        expiryType.setClazz(expiryPolicy.getClass().getName());
```
chrisdennis: This isn't generally true right?  We don't know how this class works and therefore don't know if it's safe to instantiate in this way?

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public CacheType unparseConfiguration(CacheConfiguration<?, ?> cacheConfiguration, CacheType cacheType) {
+    ExpiryPolicy<?, ?> expiryPolicy = cacheConfiguration.getExpiryPolicy();
+    if (expiryPolicy != null) {
+      ExpiryType expiryType = new ExpiryType();
+      if (expiryPolicy == ExpiryPolicy.NO_EXPIRY) {
+        expiryType.withNone(new ExpiryType.None());
+      } else if (expiryPolicy instanceof ExpiryPolicyBuilder.TimeToLiveExpiryPolicy) {
+        Duration ttl = expiryPolicy.getExpiryForCreation(null, null);
```
chrisdennis: I think you can pull this out and then use it to generate a test instance through the builder and check for equality of the implementation class.  In essence we don't care *what* the impl class is, just that it matches this one.

xml/src/main/java/org/ehcache/xml/ResourceConfigurationParser.java
```diff
@@ -0,0 +1,176 @@
...
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Marshaller;
+import javax.xml.bind.Unmarshaller;
+import javax.xml.parsers.ParserConfigurationException;
+
+import static org.ehcache.xml.ConfigurationParser.CORE_SCHEMA_NAMESPACE;
+
+public class ResourceConfigurationParser {
+
+  private final Map<org.ehcache.config.ResourceType<?>, CacheResourceConfigurationParser> extensionResourceParsers = new HashMap<>();
```
chrisdennis: Why is this code pulled out here?  Whats so special about it?

xml/src/test/resources/configs/ehcache-complete.xml
```diff
@@ -66,13 +67,15 @@
...
       <ehcache:disk unit="GB" persistent="true">100</ehcache:disk>
+      <baz:baz xmlns:baz="http://www.example.com/baz"/>
     </ehcache:resources>
     <ehcache:heap-store-settings>
       <ehcache:max-object-graph-size>10</ehcache:max-object-graph-size>
       <ehcache:max-object-size unit="MB">100</ehcache:max-object-size>
     </ehcache:heap-store-settings>
     <ehcache:disk-store-settings thread-pool="pool-disk" writer-concurrency="10" disk-segments="2"/>
-    <foo:foo/>
+    <foo:foo xmlns:foo="http://www.example.com/foo"/>
```
chrisdennis: Why did you move the namespace scopings to all be local?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
albinsuresh: The marshaller and unmarshaller are there for the `ResourceConfigurationParser`.  But I guess I can have two different constructors for that class that take the marshaller and unmarshaller separately instead of both.

xml/src/main/java/org/ehcache/xml/ConfigurationTranslator.java
albinsuresh: Will see how I can do that

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
albinsuresh: True. Missed that. But then this will get tricky. Our current API does not support class based instantiation of either `ExpiryPolicy` or the `EvictionAdvisor`. I'm wondering how we can determine if those can be safely instantiated when unparsed back. If not, we might have to say that we don't support translation of custom class based expiry and eviction.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
albinsuresh: Will do

xml/src/main/java/org/ehcache/xml/ResourceConfigurationParser.java
albinsuresh: Those two maps you meant? One map is used by the parsing logic and the other one is for unparse.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
chrisdennis: For now it means that... no way of knowing where the instance came from.  I don't think this will be a huge loss.

xml/src/main/java/org/ehcache/xml/ResourceConfigurationParser.java
chrisdennis: I mean the service and service creation configurations have their maps split between the translator and parser.  For resources you decided to have them both here, and use this class in both translator and parser.  Why the difference?

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
albinsuresh: It can be, as you suggested.

albinsuresh:
test this please

albinsuresh:
test this please

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -194,9 +194,15 @@ public String toString() {
   }
 
   private static final class TimeToLiveExpiryPolicy extends BaseExpiryPolicy {
+
     TimeToLiveExpiryPolicy(Duration ttl) {
       super(ttl, null, ttl);
     }
+
+    @Override
+    public boolean equals(Object o) {
```
chrisdennis: Mistake?

osgi-test/src/test/java/org/ehcache/osgi/SimpleOsgiTest.java
```diff
@@ -23,6 +23,7 @@
 import org.ehcache.impl.copy.ReadWriteCopier;
 import org.ehcache.impl.copy.SerializingCopier;
 import org.ehcache.xml.XmlConfiguration;
+import org.junit.Ignore;
```
chrisdennis: Unused

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -100,49 +119,121 @@ static String replaceProperties(String originalValue, final Properties propertie
...
     for (CacheResourceConfigurationParser parser : ClassLoading.libraryServiceLoaderFor(CacheResourceConfigurationParser.class)) {
       schemaSources.add(parser.getXmlSchema());
-      resourceXmlParsers.put(parser.getNamespace(), parser);
+      resourceXmlParsers.add(parser);
     }
+    this.schema = newSchema(schemaSources.toArray(new Source[schemaSources.size()]));
+    resourceConfigurationParser = new ResourceConfigurationParser(this.schema, resourceXmlParsers);
+  }
+
+  public ServiceCreationConfigurationParser getServiceCreationConfigurationParser() {
```
chrisdennis: I don't think you need some (or all?) of these parsers.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -72,16 +88,19 @@ private static Schema newSchema(Source[] schemas) throws SAXException {
     }
   }
 
-  private static final URL CORE_SCHEMA_URL = XmlConfiguration.class.getResource("/ehcache-core.xsd");
+  public static final URL CORE_SCHEMA_URL = XmlConfiguration.class.getResource("/ehcache-core.xsd");
   public static final String CORE_SCHEMA_NAMESPACE = "http://www.ehcache.org/v3";
   private static final String CORE_SCHEMA_ROOT_ELEMENT = "config";
-  private static final String CORE_SCHEMA_JAXB_MODEL_PACKAGE = ConfigType.class.getPackage().getName();
+  public static final String CORE_SCHEMA_JAXB_MODEL_PACKAGE = ConfigType.class.getPackage().getName();
```
chrisdennis: You can probably reduce visibility on some of these fields.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -0,0 +1,109 @@
...
+    Class<?> klazz = getClassForName(name, classLoader);
+    return klazz.asSubclass(type).newInstance();
+  }
+
+  public CacheType unparseConfiguration(CacheConfiguration<?, ?> cacheConfiguration, CacheType cacheType) {
+    ExpiryPolicy<?, ?> expiryPolicy = cacheConfiguration.getExpiryPolicy();
+    if (expiryPolicy != null) {
+      Duration expiry = expiryPolicy.getExpiryForCreation(null, null);
+      ExpiryType expiryType = new ExpiryType();
+      if (expiryPolicy == ExpiryPolicy.NO_EXPIRY) {
```
chrisdennis: `.equals` is better imho.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -0,0 +1,109 @@
...
+        expiryType.withNone(new ExpiryType.None());
+      } else if (expiryPolicy.equals(ExpiryPolicyBuilder.timeToLiveExpiration(expiry))) {
+        Duration ttl = expiryPolicy.getExpiryForCreation(null, null);
```
chrisdennis: You already have this in `expiry`.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -0,0 +1,109 @@
...
+        expiryType.withTtl(new TimeType().withValue(BigInteger.valueOf(ttl.toNanos())).withUnit(TimeUnit.NANOS));
+      } else if (expiryPolicy.equals(ExpiryPolicyBuilder.timeToIdleExpiration(expiry))) {
+        Duration tti = expiryPolicy.getExpiryForCreation(null, null);
+        expiryType.withTti(new TimeType().withValue(BigInteger.valueOf(tti.toNanos())).withUnit(TimeUnit.NANOS));
```
chrisdennis: This method might be useful here:
```java
static java.util.concurrent.TimeUnit bestUnit(Duration duration) {
  return stream(TimeUnit.values()).map(XmlModel::convertToJUCTimeUnit).sorted(comparing(unit -> unit.convert(duration.toNanos(), NANOSECONDS)))
    .filter(unit -> duration.equals(Duration.of(unit.convert(duration.toNanos(), NANOSECONDS), jucTimeUnitToTemporalUnit(unit)))).findFirst().orElseThrow(AssertionError::new);
}
```

xml/src/main/java/org/ehcache/xml/ResourceConfigurationParser.java
```diff
@@ -0,0 +1,209 @@
...
+      resources.add(element);
+    });
+    return cacheType.withResources(new ResourcesType().withResource(resources));
+  }
+
+  private static org.ehcache.xml.model.ResourceUnit unparseUnit(ResourceUnit resourceUnit) {
+    if (resourceUnit instanceof EntryUnit) {
+      return org.ehcache.xml.model.ResourceUnit.ENTRIES;
+    } else {
+      return org.ehcache.xml.model.ResourceUnit.valueOf(((MemoryUnit) resourceUnit).name());
```
chrisdennis: Going to have issues with `KB` versus `K_B`.

xml/src/main/java/org/ehcache/xml/XmlConfiguration.java
```diff
@@ -373,10 +256,10 @@ public URL getURL() {
   }
 
   @SuppressWarnings("unchecked")
-  private <K, V> CacheConfigurationBuilder<K, V> internalCacheConfigurationBuilderFromTemplate(final String name,
-                                                                                               final Class<K> keyType,
-                                                                                               final Class<V> valueType,
-                                                                                               final ResourcePools resourcePools)
+  private <K, V> CacheConfigurationBuilder<K, V> internalCacheConfigurationBuilderFromTemplate(String name,
```
chrisdennis: Things are and always were a bit busted in here... we should chat about it.

xml/src/main/java/org/ehcache/xml/model/CacheSpec.java
```diff
@@ -133,53 +140,15 @@ public ListenersConfig listenersConfig() {
...
-  ServiceConfiguration<?> parseCacheExtension(final Element element) {
-    URI namespace = URI.create(element.getNamespaceURI());
-    final CacheServiceConfigurationParser<?> xmlConfigurationParser = serviceConfigParsers.get(namespace);
-    if(xmlConfigurationParser == null) {
-      throw new IllegalArgumentException("Can't find parser for namespace: " + namespace);
-    }
-    return xmlConfigurationParser.parseServiceConfiguration(element);
+  public Collection<Element> serviceConfigExtensions() {
+    return sources.stream().flatMap(s -> s.getServiceConfiguration().stream())
+      .collect(Collectors.toMap(Element::getTagName, identity(), (a, b) -> a)).values();
```
chrisdennis: Doesn't the cache definition override the template, and not the other way round?  Shouldn't it therefore be `(a, b) -> b`?

albinsuresh:
test this please

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -382,7 +381,9 @@
                   <xs:attribute name="coalesce" type="xs:boolean" default="false" use="optional"/>
                 </xs:complexType>
               </xs:element>
-              <xs:element name="non-batching"/>
+              <xs:element name="non-batching">
```
chrisdennis: Huh?

albinsuresh: A trick to force JAXB to generate a type for that element to ease my unparsing. And it turned out to be better than the previous definition which allowed any random content and attributes. This is more restrictive.

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -0,0 +1,126 @@
...
+    EvictionAdvisor<?, ?> evictionAdvisor = cacheConfiguration.getEvictionAdvisor();
+    if (evictionAdvisor != null) {
+      throw new XmlConfigurationException("XML translation of eviction advisor is not supported");
+    }
+
+    return cacheType;
+  }
+
+  private static TimeType convertToTimeType(Duration duration) {
+    return Stream.of(DAYS, HOURS, MINUTES, SECONDS, MILLISECONDS, MICROSECONDS, NANOSECONDS)
```
chrisdennis: Why hardcoding the units?  Why not pull the legal unit set from jaxb?

chrisdennis: Hmm... was working for me.  Maybe something import related?

xml/src/main/java/org/ehcache/xml/CoreCacheConfigurationParser.java
```diff
@@ -72,29 +83,47 @@
...
     if (evictionAdvisor != null) {
-      cacheType.withEvictionAdvisor(evictionAdvisor.getClass().getName());
+      throw new XmlConfigurationException("XML translation of eviction advisor is not supported");
     }
 
     return cacheType;
   }
+
+  private static TimeType convertToTimeType(Duration duration) {
+    return Stream.of(java.util.concurrent.TimeUnit.values())
```
chrisdennis: This should really be a stream of the jaxb mapped time units that gets mapped to a stream of `j.u.c.TimeUnit`, right?

albinsuresh: Why? We need the XML time unit only at the end before the final transformation. Until then we're mostly using the `j.u.c.TimeUnit` directly. That's why I started the source from it directly.

PR #2388, opened 2018-05-22
Use a local variable to calculate delta

henri-tremblay:
A put is about 20% faster with this code.


chrisdennis:
I assume the reasoning on the performance improvement is that we're moving to a single update of the size for each operation rather than multiple updates right?

henri-tremblay:
Yes. But actually, even with a byteSized=false, it is faster. So my line of thoughts is that it makes the JIT happier. But I can dig into the assembly to get confirmation.

henri-tremblay:
test this please

henri-tremblay:
Build failed because of TickingTimeSourceTest that is in fact flaky. PR #2432 should fix that.

PR #2389, opened 2018-05-31
Document, cleanup and simplify ServiceUtils

henri-tremblay:


impl/src/main/java/org/ehcache/impl/internal/events/CacheEventDispatcherFactoryImpl.java
```diff
@@ -63,7 +63,7 @@ public void stop() {
   @Override
   public <K, V> CacheEventDispatcher<K, V> createCacheEventDispatcher(Store<K, V> store, ServiceConfiguration<?>... serviceConfigs) {
     String threadPoolAlias = defaultThreadPoolAlias;
-    DefaultCacheEventDispatcherConfiguration config = findSingletonAmongst(DefaultCacheEventDispatcherConfiguration.class, (Object[]) serviceConfigs);
+    DefaultCacheEventDispatcherConfiguration config = findSingletonAmongst(DefaultCacheEventDispatcherConfiguration.class, serviceConfigs);
```
chrisdennis: You sure this (and the other `Object[]` cast removals) doesn't trip a compiler warning around the varargs call?

core/src/main/java/org/ehcache/core/spi/service/ServiceUtils.java
```diff
@@ -17,38 +17,60 @@
...
   public static <T> Collection<T> findAmongst(Class<T> clazz, Collection<?> instances) {
-    return findAmongst(clazz, instances.toArray());
-  }
-
-  public static <T> Collection<T> findAmongst(Class<T> clazz, Object ... instances) {
-    Collection<T> matches = new ArrayList<>();
+    Collection<T> matches = new ArrayList<>(instances.size());
     for (Object instance : instances) {
-      if (instance != null && clazz.isAssignableFrom(instance.getClass())) {
+      if (clazz.isInstance(instance)) {
```
chrisdennis: It's probably cleaner to rewrite this as:
```java
return instances.stream().filter(clazz::isInstance).map(clazz::cast).collect(toList());
```

henri-tremblay: I kept it old school on purpose but if you want cute style I can :-)

impl/src/main/java/org/ehcache/impl/internal/events/CacheEventDispatcherFactoryImpl.java
henri-tremblay: Arg... They were not showing at first and I was surprised. But they are there indeed. Rollbacking.

core/src/main/java/org/ehcache/core/spi/service/ServiceUtils.java
chrisdennis: It so much more compact doing it this way, and this isn't performance sensitive code so I have no issues if it's even a bit slower for these small collections.

henri-tremblay: Yes. I was teasing. I did it the modern way ;-)

henri-tremblay:
test this please

PR #2393, opened 2018-06-06
Changes required by extension services for reverse translation to suc…

mobasherul:
…ceed.

### DIFF
--- a/xml/src/main/java/org/ehcache/xml/ServiceConfigurationParser.java
+++ b/xml/src/main/java/org/ehcache/xml/ServiceConfigurationParser.java
@@ -87,7 +87,14 @@ public class ServiceConfigurationParser {
     }
 
     Map<Class<?>, CacheServiceConfigurationParser<?>> parsers =
-      extensionParsers.stream().collect(toMap(CacheServiceConfigurationParser::getServiceType, identity()));
+      extensionParsers.stream().collect(toMap(CacheServiceConfigurationParser::getServiceType, identity(),
+        (key1, key2) -> {
+          if (key1.getClass().isInstance(key2)) {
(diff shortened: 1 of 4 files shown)


xml/src/main/java/org/ehcache/xml/ServiceConfigurationParser.java
```diff
@@ -87,7 +87,14 @@ CacheType unparseServiceConfiguration(CacheConfiguration<?, ?> cacheConfiguratio
     }
 
     Map<Class<?>, CacheServiceConfigurationParser<?>> parsers =
-      extensionParsers.stream().collect(toMap(CacheServiceConfigurationParser::getServiceType, identity()));
+      extensionParsers.stream().collect(toMap(CacheServiceConfigurationParser::getServiceType, identity(),
+        (key1, key2) -> {
```
henri-tremblay: A method reference here would have been much more readable

xml/src/main/java/org/ehcache/xml/ServiceCreationConfigurationParser.java
```diff
@@ -90,7 +90,14 @@ ConfigType unparseServiceCreationConfiguration(Configuration configuration, Conf
     }
 
     Map<Class<?>, CacheManagerServiceConfigurationParser<?>> parsers =
-      extensionParsers.stream().collect(toMap(CacheManagerServiceConfigurationParser::getServiceType, identity()));
+      extensionParsers.stream().collect(toMap(CacheManagerServiceConfigurationParser::getServiceType, identity(),
+        (key1, key2) -> {
```
henri-tremblay: And would have been reused here

PR #2394, opened 2018-06-06
Gradle Advancements (4.8 & java-library plugin)

chrisdennis:
@ljacomet if you have the time I'd appreciate some review of the changes I made to EhDeploy to both fix it to work with the java-library plugin, and also clean up some of the logic.

@henri-tremblay @anthonydahanne You guys probably want to look at that, along with my attempt at classifying the various dependencies as `api` and `implementation`. Consult this guide if you want more information: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies


clustered/server/build.gradle
```diff
@@ -17,18 +17,18 @@
...
     transitive = false
   }
-  compile group: 'org.terracotta', name: 'offheap-store', version: offheapVersion
-  compile group: 'org.slf4j', name: 'slf4j-api', version: slf4jVersion
-  compile("org.terracotta.management:monitoring-service-api:$terracottaPlatformVersion") {
+  implementation project(':clustered:common')
+  implementation group: 'org.terracotta', name: 'offheap-store', version: offheapVersion
+  implementation group: 'org.slf4j', name: 'slf4j-api', version: slf4jVersion
+  implementation ("org.terracotta.management:monitoring-service-api:$terracottaPlatformVersion") {
     transitive = false
```
henri-tremblay: What is the transitive for?

clustered/server/build.gradle
```diff
@@ -17,18 +17,18 @@
...
   }
-  compile"org.terracotta.management.dist:mnm-common:$terracottaPlatformVersion"
-  compile("org.terracotta:client-message-tracker:$terracottaPlatformVersion")
-  provided "org.terracotta:entity-server-api:$terracottaApisVersion"
-  provided "org.terracotta:standard-cluster-services:$terracottaApisVersion"
-  provided "org.terracotta:runnel:$terracottaPlatformVersion"
+  implementation "org.terracotta.management.dist:mnm-common:$terracottaPlatformVersion"
+  providedImplementation "org.terracotta:entity-server-api:$terracottaApisVersion"
```
henri-tremblay: Totally unrelated but why is there a 's' on 'Api' in 'terracottaApisVersion'?

ljacomet:
I can take a look, but it will be over the weekend. Saw a couple changes that I did not expect, so need to try it out locally.

core/build.gradle
```diff
@@ -17,9 +17,7 @@
 apply plugin: EhDeploy
 
 dependencies {
-  compile project(':api'), "org.slf4j:slf4j-api:$parent.slf4jVersion"
-  compile ("org.terracotta:statistics:$parent.statisticVersion") {
-    exclude group:'org.slf4j', module:'slf4j-api'
-  }
-  testCompile project(':spi-tester')
+  api project(':api')
+  implementation "org.terracotta:statistics:$parent.statisticVersion"
```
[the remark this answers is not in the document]
ljacomet: Because `statistics` was bumped to `slf4j` `1.7.25` so it no longer causes conflict resolution.

clustered/server/build.gradle
chrisdennis: At a guess I'd say its because this should be declared `compileOnly`... but maybe @anthonydahanne can shed some light?

demos/build.gradle
```diff
@@ -17,7 +17,8 @@ subprojects {
   }
 
   dependencies {
-    compile 'javax.servlet:javax.servlet-api:3.1.0', project(':impl')
+    compile project(':impl')
```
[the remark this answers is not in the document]
chrisdennis: This should really not use the `java-library` plugin... because it's not a library.  I guess I could rework the main build.gradle to understand this or I could make this 'work' with the wrong plugin? Opinions?

impl/build.gradle
```diff
@@ -17,12 +17,15 @@
...
-  compile (group: 'org.ehcache', name: 'sizeof', version: parent.sizeofVersion) {
+  api project(':core')
+  implementation group: 'org.terracotta', name: 'offheap-store', version: parent.offheapVersion
+  implementation (group: 'org.ehcache', name: 'sizeof', version: parent.sizeofVersion) {
     exclude group:'org.slf4j', module:'slf4j-api'
   }
-  testCompile project(path: ':core-spi-test'), 'org.ow2.asm:asm:6.1.1', 'org.ow2.asm:asm-commons:6.1.1'
+  implementation ("org.terracotta:statistics:$parent.statisticVersion")
+  testImplementation project(':core-spi-test')
+  testImplementation 'org.ow2.asm:asm:6.1.1'
```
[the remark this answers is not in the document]
chrisdennis: Okay.

clustered/server/build.gradle
chrisdennis: Done.

chrisdennis:
@ljacomet Don't feel obliged, but I'll be mucho grateful if you have the time.

clustered/server/build.gradle
chrisdennis: I guess the person who added it considered the fact it applied to multiple things to warrant a plural.

ljacomet: It was because the GitHub project is named `terracotta-oss/terracotta-apis`

core/build.gradle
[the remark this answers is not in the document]
henri-tremblay: Cool

demos/build.gradle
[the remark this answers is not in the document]
henri-tremblay: I don't mind using `java-library` there. Because having one sub-project not using it will just create a mess in the configuration for no particular gain. However, `compile` is deprecated from what I understood. So those should be `implementation`.

clustered/server/build.gradle
anthonydahanne: I'm also not sure; it reminds me of maven when you wanted only the present module to "see" the dependency; and not its "embedders".
That was added by @mathieucarbou , who, at this time, probably did not know about [compileOnly](https://blog.gradle.org/introducing-compile-only-dependencies)

henri-tremblay: My little finger is telling me we should go for compileOnly.

chrisdennis: Looking at the pom I think what is meant here is that this jar is 'implementation', but that it's transitive dependencies are provided:

```xml
<dependencies>
<!--  all these libs will be in common server classpath  -->
<dependency>
<groupId>org.terracotta.management</groupId>
<artifactId>management-model</artifactId>
<version>${project.version}</version>
</dependency>
<dependency>
<groupId>org.terracotta.management</groupId>
<artifactId>management-registry</artifactId>
<version>${project.version}</version>
</dependency>
<dependency>
<groupId>org.terracotta.management</groupId>
<artifactId>sequence-generator</artifactId>
<version>${project.version}</version>
</dependency>
<dependency>
<groupId>org.terracotta.management</groupId>
<artifactId>cluster-topology</artifactId>
<version>${project.version}</version>
</dependency>
<dependency>
<groupId>org.terracotta</groupId>
<artifactId>statistics</artifactId>
<version>${statistics.version}</version>
</dependency>
<!--  provided by voltron (server)  -->
<dependency>
<groupId>org.terracotta</groupId>
<artifactId>entity-server-api</artifactId>
<scope>provided</scope>
</dependency>
<dependency>
<groupId>org.terracotta</groupId>
<artifactId>standard-cluster-services</artifactId>
<scope>provided</scope>
</dependency>
</dependencies>
```
I think maybe this should be implementation and the upstream pom should be fixed instead? @anthonydahanne @mathieucarbou what say you?

anthonydahanne: > upstream pom should be fixed
how ?

chrisdennis: I'm wondering if all those things should be declared as provided.

chrisdennis:
Test this please

clustered/server/build.gradle
mathieucarbou: @anthonydahanne @chrisdennis @henri-tremblay : I think there is a difference between `compileOnly` and `transitive=false`
- `compileOnly` : is when the module needs a dependency but only for compilation time - we do not need for example to have this dependency at runtime (because it is optional or provided) or we do not want this module to transitively bring it in a client project. Examples: logback, or some annotation jars, `entity-server-api`
- `transitive=false`: this is a normal transitive dependency, so needed at runtime and compile time, but only this dependency will come, not the dependencies of this dependency.

chrisdennis: @mathieucarbou so what bucket does `monitoring-service-api` fall in to?

clustered/client/build.gradle
```diff
@@ -17,24 +17,29 @@
 apply plugin: EhDeploy
 
 dependencies {
-  compileOnly project(':api')
+  compileOnly project(':impl')
```
mathieucarbou: Is it because we know here that we can access everything both in impl and api because the shaded jar at the end will contain both ?

chrisdennis: we get `:api` transitively through `:impl`... it's redundant.

chrisdennis:
@henri-tremblay this is good for merge now I think.

henri-tremblay:
test this please

henri-tremblay:
There is a compilation error: `/org/ehcache/impl/internal/store/heap/OnHeapStore.class: warning: Cannot find annotation method 'value()' in type 'FindbugsSuppressWarnings': class file for org.terracotta.offheapstore.util.FindbugsSuppressWarnings not found`

It's a warning but since we have `-Werror` now...

chrisdennis:
@henri-tremblay fixed

PR #2397, opened 2018-06-12
Fix unchecked warnings in xml code

henri-tremblay:


henri-tremblay:
Replaced by the more complete #2415

PR #2398, opened 2018-06-12
Move to a new xjc plugin

henri-tremblay:
This plugin plus playing with source sets allows to apply `-Werror` only on our code, not on generated code. However, it has the same issue as #2287. The generated code doesn't appear in the source jar and the final classes are not in the binary jar.

### DIFF
--- a/xml/build.gradle
+++ b/xml/build.gradle
@@ -15,7 +15,7 @@
  */
 
 plugins {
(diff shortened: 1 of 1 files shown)


henri-tremblay:
@ljacomet Again, if you have time to look at it, it will be greatly appreciated

henri-tremblay:
Was replaced by #2405

PR #2404, opened 2018-06-19
Switch the time source to a ticking time source

henri-tremblay:


impl/src/main/java/org/ehcache/impl/internal/DefaultTimeSourceService.java
```diff
@@ -33,7 +33,7 @@ public DefaultTimeSourceService(TimeSourceConfiguration config) {
     if (config != null) {
       timeSource = config.getTimeSource();
     } else {
-      timeSource = SystemTimeSource.INSTANCE;
+      timeSource = TickingTimeSource.INSTANCE;
```
chrisdennis: I'm not convinced this should be the default, and I'm not sure how we should be configuring it.

core/src/main/java/org/ehcache/core/spi/time/TickingTimeSource.java
```diff
@@ -0,0 +1,66 @@
...
+  private volatile long lastUpdate;
+
+  private TickingTimeSource() {
+    updateToSystemTime();
+    Timer timer = new Timer("TickingTimeSource-timer", true);
+    timer.scheduleAtFixedRate(new TimerTask() {
+      @Override
+      public void run() {
+        if (currentTime - lastUpdate >= PERIOD_BETWEEN_SYSTEM_SET_IN_MS) {
+          updateToSystemTime();
```
chrisdennis: This could step back in time... can't we fix that?

core/src/main/java/org/ehcache/core/spi/time/TickingTimeSource.java
```diff
@@ -0,0 +1,66 @@
...
+   * Singleton instance of this {@link TimeSource}
+   */
+  public static final TimeSource INSTANCE = new TickingTimeSource();
+
+  private volatile long currentTime;
+  private volatile long lastUpdate;
+
+  private TickingTimeSource() {
+    updateToSystemTime();
+    Timer timer = new Timer("TickingTimeSource-timer", true);
```
chrisdennis: When does this get shutdown?

impl/src/main/java/org/ehcache/impl/internal/DefaultTimeSourceService.java
henri-tremblay: It is configurable. But if it works, I think it should be the default since it's the faster. The only sad thing I see about it is the background thread. Then, I did a hardcoded one. I don't know how slower will be a configurable version. But yes, configuring the `GRANULARITY_IN_MS` and `PERIOD_BETWEEN_SYSTEM_SET_IN_MS` would make sense.

core/src/main/java/org/ehcache/core/spi/time/TickingTimeSource.java
henri-tremblay: Really slightly yes. After a discussion with @lorban, we agreed that we don't care.

core/src/main/java/org/ehcache/core/spi/time/TickingTimeSource.java
henri-tremblay: Right. Let me fix that

impl/src/main/java/org/ehcache/impl/internal/DefaultTimeSourceService.java
henri-tremblay: And having a constant means the thing is spinning and not closeable for everyone

henri-tremblay:
I made a bunch of changes to be able to stop and configure the time source.

impl/src/main/java/org/ehcache/impl/internal/DefaultTimeSourceService.java
```diff
@@ -33,7 +33,7 @@ public DefaultTimeSourceService(TimeSourceConfiguration config) {
     if (config != null) {
       timeSource = config.getTimeSource();
     } else {
-      timeSource = SystemTimeSource.INSTANCE;
+      timeSource = new TickingTimeSource();
```
chrisdennis: I'd much rather this be the other way round, and have the default behavior be the current timesource.

core/src/main/java/org/ehcache/core/spi/time/TickingTimeSource.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Timer;
+import java.util.TimerTask;
+
+/**
+ * The default {@link TimeSource} in Ehcache. it increases the time every millisecond
+ * using a timer. We set it to the system time every second to prevent time drifting.
+ */
+public class TickingTimeSource implements TimeSource, Service {
+
+  private static final long GRANULARITY_IN_MS = 1L;
```
chrisdennis: I'm thinking there must be numerous cases where 100ms would be fine for a granularity... why such a short default?  If we make this require configuration then we could tie it to the other time related configuration parameters of the cache.

impl/src/main/java/org/ehcache/impl/internal/DefaultTimeSourceService.java
henri-tremblay: So be it. I will write a performance documentation page about it.

chrisdennis: I'd talk this over with others (say Chris G & Ludovic) before going to this effort... make sure that we have consensus.

core/src/main/java/org/ehcache/core/spi/time/TickingTimeSource.java
henri-tremblay: You mean no default?

chrisdennis: I mean *if* this becomes a non-default then it might be best to not have a default and force the user to specify. If it becomes the default then thats a different question.

henri-tremblay: Ok. That's what I did

PR #2406, opened 2018-06-20
Fixed testcase failure in Windows environment.

bbyreddy:
Hi,
Please review the changes done to the test, which always fails in Windows. Comments in the code explains the issue and the fix.

Regards,
Bhaskar.

### DIFF
--- a/xml/src/test/resources/configs/ehcache-complete.xml
+++ b/xml/src/test/resources/configs/ehcache-complete.xml
@@ -29,7 +29,7 @@
   <default-copiers>
     <copier type="com.pany.ehcache.copier.Person">com.pany.ehcache.copier.AnotherPersonCopier</copier>
   </default-copiers>
-  <persistence directory="some/dir"/>
+  <persistence directory="some_folder_path"/>
   <thread-pools>
     <thread-pool alias="theone" min-size="1" max-size="4"/>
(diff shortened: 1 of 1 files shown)


xml/src/test/java/org/ehcache/xml/XmlConfigurationTest.java
```diff
@@ -772,7 +772,15 @@ public void testCompleteXmlToString() {
     URL resource = XmlConfigurationTest.class.getResource("/configs/ehcache-complete.xml");
     Configuration config = new XmlConfiguration(resource);
     XmlConfiguration xmlConfig = new XmlConfiguration(config);
-    assertThat(xmlConfig.toString(), isSimilarTo(Input.from(resource)).ignoreComments().ignoreWhitespace().withNodeMatcher(new DefaultNodeMatcher(ElementSelectors.byNameAndText)));
+
```
albinsuresh: This fix is fine. But an even simpler fix would have been to just change the path value in the test xml file to a single directory name without any path separators.

bbyreddy:
Hi Albin,
Changed the config file as discussed. Please review the new changes.

Regards,
Bhaskar.

henri-tremblay:
Can you squash the two comments together?

PR #2407, opened 2018-06-21
Add OWASP dependency checking

chrisdennis:


chrisdennis:
@anthonydahanne consider this is a prototype of how this could be done. We should probably assess the impact on cloudbees before we merge it. In particular I'm wondering if it's going to repull the whole NVD database every time it runs.

clustered/client/build.gradle
```diff
@@ -55,3 +55,7 @@ compileJava {
 compileTestJava {
   options.compilerArgs += ['-Xlint:-serial,-rawtypes']
 }
+
+dependencyCheck {
+  suppressionFile = file("$projectDir/config/owasp-suppressions.xml")
```
anthonydahanne: like @henri-tremblay  suggested, why do we duplicate those lines and exclusion files ?
Also @chrisdennis I wonder if there is a better way to get rid of warnings for entity-api : what about a manifest entry or something to tell the plugin to stop considering entity-api ?

chrisdennis: @anthonydahanne @henri-tremblay : I made separate suppression files for each project because I considered that a suppression is ultimately sensitive to both the dependency and the dependent. For now since the only thing we're suppressing is a falsely flagged artifact we could simplify and use a single suppression file. Thoughts?

@anthonydahanne : I had a quick look through the documentation and I don't see any obvious way of doing what you describe.

anthonydahanne: @chrisdennis yes, I agree with 1 single suppression file for now, since, like you said, it's a false alert.

what's weird with the false detection of entity api, is that no other tools have spotted that yet. Anyway, sure, let's take the exclusion route for now.

anthonydahanne:
I think this PR is a step in the good direction, even if there is some duplicated configuration.
Shall we proceed ?

henri-tremblay:
I would like this to go through. So I approved it. A let you decide if a refactoring is needed.

PR #2419, opened 2018-07-04
Allow to pass functions when creating ExpiryPolicyBuilder

henri-tremblay:


impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -64,6 +65,11 @@
     return new TimeToIdleExpiryPolicy(timeToIdle);
   }
 
+  @FunctionalInterface
+  public interface TriFunction<K, V1, V2, R> {
```
chrisdennis: Could these types be `<T, U, V, R>` to fit with the JDK convention?

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -75,9 +81,9 @@ public static ExpiryPolicyBuilder expiry() {
     return new ExpiryPolicyBuilder();
   }
 
-  private Duration create = ExpiryPolicy.INFINITE;
-  private Duration access = null;
-  private Duration update = null;
+  private BiFunction<Object, Object, Duration> create = (key, value) -> ExpiryPolicy.INFINITE;
```
chrisdennis: This type should really have proper generic typing if we're going to expand in to key/value sensitive expiry policies.

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
henri-tremblay: Yes. I don't know where I took these from. Weird.

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
henri-tremblay: What do you mean?

chrisdennis: See: https://github.com/henri-tremblay/ehcache3/pull/2

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -64,36 +65,67 @@
...
   /**
    * Fluent API for creating an {@link ExpiryPolicy} instance where you can specify constant values for creation, access and update time.
    * Unspecified values will be set to {@link ExpiryPolicy#INFINITE INFINITE} for create and {@code null} for access and update, matching
    * the {@link #noExpiration()}  no expiration} expiry.
    *
    * @return an {@link ExpiryPolicy} builder
    */
-  public static ExpiryPolicyBuilder expiry() {
-    return new ExpiryPolicyBuilder();
+  public static <K, V> ExpiryPolicyBuilder<K, V> expiry() {
```
chrisdennis: I think this is an error in my generics changes... I think this should return `ExpiryPolicyBuilder<Object, Object>`.

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -64,36 +65,67 @@
...
+    this.create = create;
+    this.access = access;
+    this.update = update;
+  }
 
   /**
    * Set TTL since creation
    *
    * @param create TTL since creation
    * @return this builder
```
chrisdennis: These `@return` stanzas are now wrong... the builder methods now return new builder instances (the builder has become immutable).

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -102,12 +134,20 @@ public ExpiryPolicyBuilder create(Duration create) {
...
+    return access((a, b) -> access);
+  }
+
+  /**
+   * Set a function giving the TTL since last access
+   * @param access Function giving the TTL since last access
+   * @return this builder
+   */
+  public <K2 extends K, V2 extends V>ExpiryPolicyBuilder<K2, V2> access(BiFunction<K2, Supplier<? extends V2>, Duration> access) {
+    return new ExpiryPolicyBuilder<>(create, Objects.requireNonNull(access), update);
```
chrisdennis: Should it be legal to make multiple calls to the builder methods?

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
```diff
@@ -116,92 +156,105 @@ public ExpiryPolicyBuilder access(Duration access) {
...
     public int hashCode() {
-      int result = Objects.hashCode(create);
-      result = 31 * result + Objects.hashCode(access);
-      result = 31 * result + Objects.hashCode(update);
-      return result;
+      return ttl.hashCode();
     }
 
     @Override
-    public String toString() {
```
chrisdennis: We've lost this `toString()` implementation... should we add it back (at least for the TTL and TTI subclasses)?

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
henri-tremblay: Why?

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
henri-tremblay: Yes. Why not now that you have added attributes for TTL and TTI.

impl/src/main/java/org/ehcache/config/builders/ExpiryPolicyBuilder.java
chrisdennis: Fair enough... might be good to document that the behavior is to over-write... someone might for example think they can specify multiple "access" rules and have the `earliest` expiration time apply.

henri-tremblay: Documented

chrisdennis:
Needs a clean CI build but is otherwise good to go.

henri-tremblay:
@chrisdennis Ok. I triggered a build. Fingers crossed.

PR #2420, opened 2018-07-05
Throw exception while reverse translation to XML from programmatic co…

mobasherul:
…nfiguration of caches having instances specified for Serializer,Copier,CacheLoaderWriter,CacheEventListener and ResilienceStrategy.

### DIFF
--- a/xml/src/main/java/org/ehcache/xml/service/DefaultCacheEventListenerConfigurationParser.java
+++ b/xml/src/main/java/org/ehcache/xml/service/DefaultCacheEventListenerConfigurationParser.java
@@ -40,6 +40,7 @@ import java.util.stream.Collectors;
 import static java.util.stream.Collectors.toSet;
 import static org.ehcache.core.spi.service.ServiceUtils.findAmongst;
 import static org.ehcache.xml.XmlConfiguration.getClassForName;
+import static org.ehcache.xml.service.SimpleCoreServiceConfigurationParser.checkNoConcreteInstance;
 
 public class DefaultCacheEventListenerConfigurationParser implements CoreServiceConfigurationParser {
 
@@ -78,6 +79,7 @@ public class DefaultCacheEventListenerConfigurationParser implements CoreService
 
       Set<ListenersType.Listener> listeners = serviceConfigs.stream().map(serviceConfig -> {
         ListenersType.Listener listener = new ListenersType.Listener();
+        checkNoConcreteInstance(serviceConfig);
         return listener.withClazz(serviceConfig.getClazz().getName())
           .withEventFiringMode(EventFiringType.fromValue(serviceConfig.firingMode().name()))
           .withEventOrderingMode(EventOrderingType.fromValue(serviceConfig.orderingMode().name()))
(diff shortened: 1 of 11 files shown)


xml/src/main/java/org/ehcache/xml/service/DefaultCacheEventListenerConfigurationParser.java
```diff
@@ -78,6 +79,9 @@ public CacheType unparseServiceConfiguration(CacheConfiguration<?, ?> cacheConfi
 
       Set<ListenersType.Listener> listeners = serviceConfigs.stream().map(serviceConfig -> {
         ListenersType.Listener listener = new ListenersType.Listener();
+        if (serviceConfig.getInstance() != null) {
```
albinsuresh: This logic is repeated in multiple classes that extends `SimpleCoreServiceConfigurationParser`. It can be moved to `SimpleCoreServiceConfigurationParser.unparseServiceConfiguration`.

mobasherul: It's repeated in those classes that don't extend SimpleCoreServiceConfigurationParser and have separate unparsing functions.

albinsuresh: Just the copier and serializer configs don't extend from it. Let the duplication be there. But it can still be avoided from the other two.

xml/src/main/java/org/ehcache/xml/service/DefaultCacheEventListenerConfigurationParser.java
```diff
@@ -78,6 +79,9 @@ public CacheType unparseServiceConfiguration(CacheConfiguration<?, ?> cacheConfi
...
+          throw new XmlConfigurationException("XML translation when instance is provided is not supported");
```
henri-tremblay: I would like the message to tell what instance if missing. "XML translation when a CacheEventListener instance is provided in the configuration is not supported"

xml/src/main/java/org/ehcache/xml/service/DefaultCacheLoaderWriterConfigurationParser.java
```diff
@@ -33,6 +34,11 @@ public DefaultCacheLoaderWriterConfigurationParser() {
       CacheTemplate::loaderWriter,
       (config, loader) -> new DefaultCacheLoaderWriterConfiguration((Class<? extends CacheLoaderWriter<?, ?>>) getClassForName(config, loader)),
       CacheType::getLoaderWriter, CacheType::setLoaderWriter,
-      config -> new CacheLoaderWriterType().withClazz(config.getClazz().getName()));
+      config -> {
+        if (config.getInstance() != null) {
```
henri-tremblay: And I agree with @albinsuresh that you could have a `checkNoConcreteInstance` util method somewhere

xml/src/test/java/org/ehcache/xml/service/DefaultCacheLoaderWriterConfigurationParserTest.java
```diff
@@ -57,4 +58,12 @@ public void unparseServiceConfiguration() {
 
     assertThat(cacheType.getLoaderWriter().getClazz()).isEqualTo(TestCacheLoaderWriter.class.getName());
   }
+
+  @Test(expected = XmlConfigurationException.class)
```
henri-tremblay: I really don't like expected exceptions. Because you never know if the exception was thrown from the right place. And I like to check the message as well. I use assertJ instead: http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion

mobasherul:
@henri-tremblay I have addressed the comments given by you. Please review

PR #2421, opened 2018-07-05
Add performance documentation

henri-tremblay:
It's not complete because other PRs must be merged first. However, you can validate the concept.

### DIFF
--- a/docs/src/docs/asciidoc/user/performance.adoc
+++ b/docs/src/docs/asciidoc/user/performance.adoc
@@ -1,0 +1,138 @@
+---
+---
+= Performance Tuning
+ifndef::sourcedir35[]
+include::common.adoc[]
+endif::sourcedir35[]
+
+ifdef::notBuildingForSite[]
+include::menu.adoc[]
+endif::notBuildingForSite[]
+
+== Introduction
+
+Ehcache is fast. Really fast.
+It was thought and made to be.
+For example, a straight `get(key)` from the cache should be under 500ns.
+That's fast and way sufficient for most purpose.
+
+Keep that in mind. Especially when comparing caching frameworks with a benchmark.
+If something looks 10% faster it actually means 50ns faster.
+Do you care that much about these 50ns to lose your precious time in benchmarking?
+
+That said, the way you configure Ehcache can have an impact on performance.
+This document is a work in progress.
+It will give you the performance impact of classical Ehcache configuration.
+It will also give you some advance tuning possibilities.
+
+In the future, we will add some figures of what "slower" means.
+However, always do your own benchmark.
+
+== Stores
+
+We probably know that the faster store is on-heap.
+Until you overwhelm the garbage collector.
+
+Your next best bet is off-heap.
+
+Try to avoid disk.
+Use a remote drive or even an HDD at your own risk.
+
+We won't talk about clustering here because it's a different realm and its performance are based on many factors.
+
+The next question would be: "Should I use a single tier?" Is using a single-tier off-heap faster than two-tiers?
+The answer depends on what you do with it.
+Having two tiers is a bit slower on writing.
+It is also a bit slower on reading when the data is not found in the caching tier (on-heap).
+However, it will be faster for an entry found is indeed found there.
+
+So again, it depends.
+The more you follow the caching hypothesis that the same data is always reused (and so in the caching
+tier), the more interesting having two tiers will be.
+
+== Byte Sizing
+
+A on-heap tier can be limited to a number of entries or a number of bytes.
+When using bytes, we need to calculated the size of every object added to the cache.
+This is of course much slower than calculating the number of entries.
+
+Size calculation is done using the https://github.com/ehcache/sizeof[SizeOf] library.
+This library uses multiple magic tricks to do so. It selects the fastest one for a given environment.
+Make sure of what is used to confirm you can't use a faster way on your platform.
+
+== Serialization
+
+Off-heap, disk and clustering need to serialize keys and values before storing them.
+By default, the Java serialization is used.
+It is well-known for not being the fastest thing around.
+Ehcache uses it because it is supported out of the box.
+However, you can increase performances by provoding your own serializers.
+
+== Copier
+
+By default, on-heap storage stores the entries by reference.
+You might want to use a copier for whatever reason.
+This can be much slower so watch out.
+
+== Loader-Writer
+
+Loader-writer is interesting for many reasons.
+First, it protects you against the Thundering Herd.
+However, it needs to pass through more complicated code to do so.
+
+We are expecting it to be a tiny slower.
+But nothing noticeable enough to prevent you from using it.
+
+== Expiration
+
+A cache with no expiration will always be faster.
+
+=== Time to Live
+
+If you need to set an expiration time, TTL will be the faster one.
+This is because the expiration time is calculated once when the entry is inserted in the cache.
+But it still requires an expiration check at access time.
+
+So you can expect a 2% drop in performance when using TTL. Not bad.
+
+=== Time to Idle
+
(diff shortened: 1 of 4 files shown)


henri-tremblay:
@lorban @chrisdennis I'm interested in your feedback. In particular, do you think some figures for `slower` and `faster` would be interesting? (which means having a benchmark testing each case)

However,   #2455 and #2449 are required to get the final performance.

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+Off-heap, disk and clustering need to serialize keys and values before storing them.
+By default, the Java serialization is used.
+It is well-known for not being the fastest thing around.
+Ehcache uses it because it is supported out of the box.
+However, you can increase performances by provoding your own serializers.
+
+== Copier
+
+By default, on-heap storage stores the entries by reference.
+You might want to use a copier for whatever reason.
```
albinsuresh:
```suggestion
You might want to use a copier to store the entries by value for whatever reason.
```

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+This can be much slower so watch out.
+
+== Loader-Writer
+
+Loader-writer is interesting for many reasons.
+First, it protects you against the Thundering Herd.
+However, it needs to pass through more complicated code to do so.
+
+We are expecting it to be a tiny slower.
```
albinsuresh:
```suggestion
We are expecting it to be a tiny bit slower.
```

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+But nothing noticeable enough to prevent you from using it.
+
+== Expiration
+
+A cache with no expiration will always be faster.
+
+=== Time to Live
+
+If you need to set an expiration time, TTL will be the faster one.
+This is because the expiration time is calculated once when the entry is inserted in the cache.
```
albinsuresh:
```suggestion
This is because the expiration time of an entry is calculated and updated only when the entry is inserted or updated in the cache.
```

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+But it still requires an expiration check at access time.
+
+So you can expect a 2% drop in performance when using TTL. Not bad.
+
+=== Time to Idle
+
+TTI is slower than TTL.
+We need to recalculate the expiration time each time the entry is accessed.
```
albinsuresh:
```suggestion
We need to recalculate and update the expiration time each time the entry is accessed.
```

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+=== Custom
+
+In general, using a custom `ExpiryPolicy` will be the slowest.
+Ehcache uses intrinsics to handle other cases.
```
albinsuresh: Not sure if **intrinsics** would make much sense to an external reader.
```suggestion
Ehcache has optimised the handling of other cases.
```

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+== Time Source
+
+By default, Ehcache uses a `TimeSource` that will retrieve the system time at every call. It is fast but not super
+duper fast. But it is super duper accurate.
+
+You can trade accuracy for speed by using a `TickingTimeSource`. Please read the javadoc for details but the concept is
+that a timer will increase time instead of always retrieving system time.
+
+Switching to `TickingTimeSource`, even with a granularity of 1ms can improve the performance of a `get` of a high as 30%.
```
albinsuresh: performance of a `get` ** as** ~of a~ high as

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+This can be much slower so watch out.
+
+== Loader-Writer
+
+Loader-writer is interesting for many reasons.
+First, it protects you against the Thundering Herd.
+However, it needs to pass through more complicated code to do so.
+
+We are expecting it to be a tiny slower.
+But nothing noticeable enough to prevent you from using it.
```
mathieucarbou: Perhaps we can add that the impact in perf of using that is much more linked to the underlying SOR because Ehache adds nearly no overhead.

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+You can trade accuracy for speed by using a `TickingTimeSource`. Please read the javadoc for details but the concept is
+that a timer will increase time instead of always retrieving system time.
+
+Switching to `TickingTimeSource`, even with a granularity of 1ms can improve the performance of a `get` of a high as 30%.
+
+The drawback is that a timer will continuously run.
+Also, time might drift from the real time a bit.
+Especially if the granularity or the systemUpdatePeriod is big.
+But in most cases, the drifting doesn't matter much.
```
mathieucarbou: I would like to see little explanation why _doesn't matter much._...
Is it because Ehcache mostly uses timestamps to calculate durations so the "base" time and drifting has little impact ?

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+This document is a work in progress.
+It will give you the performance impact of classical Ehcache configuration.
+It will also give you some advance tuning possibilities.
+
+In the future, we will add some figures of what "slower" means.
+However, always do your own benchmark.
+
+== Stores
+
+We probably know that the faster store is on-heap.
```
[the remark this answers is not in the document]
henri-tremblay: I would say it is detailed lower in the test. When talking about serialization and copier. Isn't it enough?

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+== Byte Sizing
+
+A on-heap tier can be limited to a number of entries or a number of bytes.
+When using bytes, we need to calculated the size of every object added to the cache.
+This is of course much slower than calculating the number of entries.
+
+Size calculation is done using the https://github.com/ehcache/sizeof[SizeOf] library.
+This library uses multiple magic tricks to do so. It selects the fastest one for a given environment.
+Make sure of what is used to confirm you can't use a faster way on your platform.
```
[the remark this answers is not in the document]
henri-tremblay: That's Sizeof documentation. I didn't want to detail it here.

docs/src/docs/asciidoc/user/performance.adoc
henri-tremblay: In fact, no really. Because cache-aside will also call the SOR. So the global time is similar. Ehcache does add a tiny overhead. For example, instead of calling a straight `get` in the `ConcurrentHashMap`, it will do a `compute`. And the compute involves bucket locking. So it's not as multi-threading friendly.

docs/src/docs/asciidoc/user/performance.adoc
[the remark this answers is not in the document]
mathieucarbou: ok, but as a user reading this, how I know that ?

docs/src/docs/asciidoc/user/performance.adoc
mathieucarbou: ok!

albinsuresh:
@henri-tremblay Build failed with a tets compilation failure. Will merge once once u fix it.

PR #2423, opened 2018-07-09
Add productInstanceId to keep track of statistics after a client reconnects

mathieucarbou:
Statistics are tracked at the moment per clientId. Sadly, clientId changes over time for the same cache manager instance in case of a reconnection and we loose the statistic history.

We decided to use a `productInstanceId` that will be set in management layer for each cache manager instance and this ID will be the same even if clientIds change.

We will be able to track statistics based on this `productInstanceId` and no more based on the `clientId` now.


clustered/integration-test/src/test/java/org/ehcache/clustered/management/ManagementClusterConnectionTest.java
```diff
@@ -166,13 +169,29 @@ public void test_reconnection() throws Exception {
...
+        break;
+      } else {
+        Thread.sleep(1_000);
       }
+    }
 
-      Thread.sleep(1_000);
+    if (Thread.currentThread().isInterrupted()) {
+      fail();
+    } else {
```
henri-tremblay: No need for a `else`1 here since it's a fail. And it should be an assert. `assertThat(Thread.currentThread().isInterrupted()).isFalse()`

management/src/main/java/org/ehcache/management/registry/DefaultManagementRegistryConfiguration.java
```diff
@@ -89,30 +102,26 @@ public String getCollectorExecutorAlias() {
...
+      Objects.equals(collectorExecutorAlias, that.collectorExecutorAlias);
   }
 
   @Override
   public int hashCode() {
-    int result = tags.hashCode();
-    result = 31 * result + context.hashCode();
-    result = 31 * result + (collectorExecutorAlias != null ? collectorExecutorAlias.hashCode() : 0);
-    return result;
+    return Objects.hash(tags, productInstanceId, context, collectorExecutorAlias);
```
henri-tremblay: The `productInstanceId` should be enough as an hashcode

mathieucarbou: You know what, to me, having an equal/hashcode in a config class like this is useless... I wanted to keep the original behaviour but if you are fine, I would be pleased to completely remove them!

henri-tremblay: I'm fine to remove them if no one is supposed to put these things in a HashMap. Useless code is useless code.

mathieucarbou: Sorry actually this is used! used to verify that a config is equal to another one - so all attributes need to be there.

management/src/main/java/org/ehcache/management/ManagementRegistryServiceConfiguration.java
```diff
@@ -42,4 +42,8 @@
    */
   Collection<String> getTags();
 
+  /**
+   * @return an identifier used to identify this running instance. It will be the same even if a clustered Ehcache client reconnects (and clientId changes).
+   */
+  String getProductInstanceId();
```
chrisdennis: Should the term here be this 'abstract'? At this point the 'product' is a known thing not an abstract concept.

anthonydahanne: what do you suggest ?

mathieucarbou: getter renamed ;-)

chrisdennis: Well at a minimum would 'EhcacheInstanceId' be better - just substituting in the concrete product name? Or maybe EhcacheClientIdentifier?

mathieucarbou: I have pushed: `getCacheManagerInstanceId()`

mathieucarbou:
Had a meeting with @anthonydahanne and @GaryWKeim : we chose to renamed to `instanceId()` and have the same `instanceId` attribute name across all products.

mathieucarbou:
Also we have decided to not make the instance id configurable: it will be a generated UUID per CM instance.

PR #2427, opened 2018-07-12
Fix #2426 : Abandon leadership if any is acquired before closing conn

### LINKED ISSUE #2426: acquireLeadership to go to maintenance mode does not have a corresponding destroy
When ehcache is put to maintenance mode it does an acquireLeadership..but there is no corresponding call to abandonLeadership. This may cause the VoltronReadWriteLockEntity once the cluster tier manager for a given entity identifier is destroyed.

This may cause problems such as follows;
 
If a client disconnection is slow, another client may not be able to do a maintenance hold as the server may not release the lock until the disconnect event reaches the server

rkavanap:
This PR addresses the issue where a VoltronReadWriteLockEntity for a given cluster tier manager entity may remain in the system even after the cluster tier manager entity has been destroyed.

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
@@ -79,13 +79,19 @@ public class ClusterTierManagerClientEntityFactory {
     }
   }
 
-  public void abandonLeadership(String entityIdentifier) {
+  /**
+   * Proactively abandon any maintenance holds (READ or WRITE) before closing connection.
+   *
+   * @param entityIdentifier the master entity identifier
+   * @return true of abandoned false otherwise
+   */
+  public boolean abandonMaintenanceHolds(String entityIdentifier) {
     Hold hold = maintenanceHolds.remove(entityIdentifier);
-    if (hold == null) {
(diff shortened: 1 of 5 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/ClusterTierManagerClientEntityFactory.java
```diff
@@ -88,6 +88,13 @@ public void abandonLeadership(String entityIdentifier) {
     }
   }
 
+  public void abandonLeadershipIfLeader(String entityIdentifier) {
```
rkavanap: There is a method called abandonLeadership above, but it is not used anywhere except tests..

Also there are methods that does localholds, but local holds are released using try-with-resources... so the problem is only when acquireLeadership is explicitly called.. a corresponding abandon must be issued.

henri-tremblay: Can you please javadoc the usage pattern

AlbertGevorgyan: Should we keep an API method if it is used only in tests?

rkavanap:
test this please

rkavanap:
test this please

PR #2430, opened 2018-07-13
Close #2428: CACHE_MANAGER_CLOSED not received when clustered CM is closed

### LINKED ISSUE #2428: CACHE_MANAGER_CLOSED not received when clustered CM is closed

mathieucarbou:
I did a IT test in this PR to reproduce the problem described into #2428: `CMClosedEventSentTest`.

1. I have fixed the service ordering issue when closing. Instead of closing dependencies first, which leaves some running service not working because their dependencies are closed, I am closing the services that do tot depend on dependencies.

2. I have activated the circular dependency check and re-enabled the test

3. I added support for optional service dependencies: this is how M&M works, it automatically activate hen in the classpath but we still ned to be able to declare a dependency on the clustering service

4. Now the IT test is working and M&M is able to send client-side notifications with data using the clustering service before it closes.


core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -423,6 +425,23 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
...
+          Class<?> dependency = ClassLoading.getDefaultClassLoader().loadClass(className);
+          if (Service.class.isAssignableFrom(dependency)) {
+            @SuppressWarnings("unchecked")
+            Class<? extends Service> serviceDependency = (Class<? extends Service>) dependency;
+            dependencies.add(serviceDependency);
+          } else {
+            throw new IllegalStateException("Service dependency declared by " + className + " is not a Service: " + dependency.getName());
+          }
+        } catch (ClassNotFoundException ignored) {
+          // dependency is optional so we ignore it
```
chrisdennis: This isn't the kind of optional I expected.  This kind of optional is:
> If this service interface is on the classpath then please use it, but if it's not on the classpath then please ignore it.

What I had expected was:
> If this service is in the dependency set for this environment already then please order my start/stop with respect to it. If it's not already in the set then don't pull it in on my account.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -438,11 +457,22 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
   private static Set<Class<? extends Service>> identifyTransitiveDependenciesOf(final Class<?> clazz) {
 
     Set<Class<? extends Service>> dependencies = identifyImmediateDependenciesOf(clazz);
+    for (Class<? extends Service> dep : dependencies) {
```
chrisdennis: We already have circular dependency checking in the startup logic... I understand this fails slightly earlier, but I also think it's going to miss some instances.  The circular dependencies we care about are between implementations... I think this logic will miss things if a service implementation introduces extra links by implementing two service interfaces? Right?

api/src/main/java/org/ehcache/spi/service/OptionalServiceDependencies.java
```diff
@@ -0,0 +1,35 @@
...
+ * Annotation that allows a {@link Service} implementation to declare an optional dependency on other {@code Service}s.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface OptionalServiceDependencies {
+
+  /**
+   * Array of {@link Service} dependency classes
+   */
+  String[] value();
```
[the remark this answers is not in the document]
lorban: That would make the class using this annotation throw `ClassNotFoundException` at runtime which would be counter-productive, wouldn't it?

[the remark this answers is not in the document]
chrisdennis: I'm not sure... I'm *hoping* that this might 'do the right thing' in as similar manner to how if an annotation class is missing at runtime that it just fails to load.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/CMClosedEventSentTest.java
```diff
@@ -0,0 +1,104 @@
...
+      + "<lease:lease-length unit='seconds'>5</lease:lease-length>"
+      + "</lease:connection-leasing>"
+      + "</service>";
+
+  @ClassRule
+  public static Cluster CLUSTER = newCluster().in(new File("build/cluster")).withServiceFragment(RESOURCE_CONFIG).build();
+
+  @Test(timeout = 60_000)
+  public void test_CACHE_MANAGER_CLOSED() throws Exception {
+    AbstractClusteringManagementTest.createNmsService(CLUSTER);
```
henri-tremblay: It's weird to use the NmsService from another class like this without extending it. It feels like there should be an external NmsServiceFixture used by everythign

clustered/integration-test/src/test/java/org/ehcache/clustered/management/CMClosedEventSentTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    CacheManager cacheManager = newCacheManagerBuilder()
```
henri-tremblay: This can be a try-with-resource

clustered/integration-test/src/test/java/org/ehcache/clustered/management/CMClosedEventSentTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    assertThat(cacheManager.getStatus(), equalTo(Status.AVAILABLE));
+    waitFor("CACHE_MANAGER_AVAILABLE");
+
+    cacheManager.close();
+    waitFor("CACHE_MANAGER_CLOSED");
+  }
+
+  private void waitFor(String notifType) throws InterruptedException {
+    while (!Thread.currentThread().isInterrupted()) {
```
henri-tremblay: There should be a waiting limit here. For instance, 30 seconds or 1 minute.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -205,10 +206,11 @@ private boolean hasUnstartedDependencies(Service service, Iterable<Service> unst
     return false;
   }
 
-  private boolean hasRunningDependencies(Service service, Iterable<Service> running) {
-    for (Class<? extends Service> dep : identifyTransitiveDependenciesOf(service.getClass())) {
-      for (Service s : running) {
-        if (dep.isInstance(s)) {
+  private boolean aRunningServiceStillDependsOnMe(Service service, Iterable<Service> running) {
+    for (Service runningService : running) {
+      Set<Class<? extends Service>> dependencies = identifyTransitiveDependenciesOf(runningService.getClass());
```
henri-tremblay: I would say dependencyClasses and dependencyClass to remove the doubt that it is an actual dependency instance

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -423,6 +425,23 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
         }
       }
     }
+    final OptionalServiceDependencies optionaAnnotation = clazz.getAnnotation(OptionalServiceDependencies.class);
```
henri-tremblay: No final needed

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
mathieucarbou: Are you sure ? The circular dependency test was ignored and was failing...
I agree it might miss some things and could call stop() on an implementation id the interface is not found as a dependency. But well... It would requires a bigger code change to do it correctly. This is only support the general use case of 1 service 1 impl at least.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -423,6 +425,23 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
...
+    if (optionaAnnotation != null) {
+      for (String className : optionaAnnotation.value()) {
```
henri-tremblay: Yes. Class<?>[] would be pretty here

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -423,6 +425,23 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
...
+        try {
+          Class<?> dependency = ClassLoading.getDefaultClassLoader().loadClass(className);
```
henri-tremblay: Hum... For me, optional means "I will use the implementation of this service interface if available and otherwise I will do without". But here, we are requesting a concrete implementation. Which is strange.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
henri-tremblay: Again, I don't like the variables naming. Those are classes, not dependencies

core/src/test/java/org/ehcache/core/internal/service/ServiceLocatorTest.java
```diff
@@ -239,8 +239,7 @@ public void testRedefineDefaultServiceWhileDependingOnIt() throws Exception {
     serviceLocator.startAllServices();
   }
 
-  @Test
-  @Ignore
+  @Test(expected = IllegalStateException.class)
```
henri-tremblay: As you know, I'm not a bit fan of expected exceptions. I prefer a good old try, fail or assertj

clustered/integration-test/src/test/java/org/ehcache/clustered/management/CMClosedEventSentTest.java
mathieucarbou: yes agree. A previous committer has already widened these methods and used like this in other test. I have reused.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/CMClosedEventSentTest.java
mathieucarbou: no need: `nmsService.waitForMessage()` is blocking

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
mathieucarbou: was there in existing code above.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
mathieucarbou: This is the trick - not possible... Because otherwise they would need to be in the classpath. Dependency is optional

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
mathieucarbou: It is loading the interface defined as an optional dependency: not the impl:

```
@ServiceDependencies({CacheManagerProviderService.class, StatisticsService.class, TimeSourceService.class, ExecutionService.class})
@OptionalServiceDependencies({
  "org.ehcache.clustered.client.service.EntityService",
  "org.ehcache.clustered.client.service.ClusteringService"})
public class DefaultManagementRegistryService extends DefaultManagementRegistry implements ManagementRegistryService, CacheManagerListener {
```

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
mathieucarbou: I have kept teh existing naming that what was already there elsewhere in the same class.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/CMClosedEventSentTest.java
henri-tremblay: Ok

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
henri-tremblay: You are not a good boy scout -:-)

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
henri-tremblay: Ok. So the interface service is optional then. It's a bit non-conventional but I see why you need it. So ok.

PR #2436, opened 2018-07-16
Cast errors correctly to errors (close #2434)

### LINKED ISSUE #2434: Trying to cast a StoreCastException to RuntimeException, swallowing underlying error
Hi everyone,

When using a get operation, I get this exception:

```
java.lang.ClassCastException: java.lang.NoClassDefFoundError incompatible with java.lang.RuntimeException
	at org.ehcache.impl.internal.store.tiering.TieredStore.get(TieredStore.java:98)
	at org.ehcache.core.Ehcache.doGet(Ehcache.java:82)
	at org.ehcache.core.EhcacheBase.get(EhcacheBase.java:127)
```

This is the problematic line:

https://github.com/ehcache/ehcache3/blob/master/impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java#L98

The method catches a StoreAccessException. If it is not a StorePassThroughException, it will be casted to RuntimeException and thrown.

Since StoreAccessException is not a RuntimeException, we see the error above. And this is why we cannot see why the StoreAccessException is being thrown in the first place.

Proposed solutions (implement one of them):
* Make StoreAccessException a RuntimeException, if possible (and feasible).
* Wrap into another exception
* Change the signature of the TieredStore.get() method to allow throwing StoreAccessException.

Thanks!

henri-tremblay:


henri-tremblay:
test this please

impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
```diff
@@ -299,6 +293,17 @@ private void swapBackCachingTiers() {
...
+    Throwable cause = ce.getCause();
+    if (cause instanceof StorePassThroughException) {
+      throw (StoreAccessException) cause.getCause();
+    }
+    if (cause instanceof Error) {
+      throw (Error) cause;
+    }
+    throw (RuntimeException) cause;
+  }
+
```
chrisdennis: This still isn't safe... checked versus runtime exceptions are a compile time concept only.  I think you'll need somthing like this no?

```java
Throwable cause = ce.getCause();
if (cause instanceof StorePassThroughException) {
  throw (StoreAccessException) cause.getCause();
}
if (cause instanceof Error) {
  throw (Error) cause;
} else if (cause instanceof RuntimeException) {
  throw (RuntimeException) cause;
} else {
  throw new RuntimeException(cause);
}
```

henri-tremblay: I could. But since only runtimes and errors are supposed to be wrapped in the StoreAccessException, anything else means something weird is happening.

henri-tremblay: Let me do something

henri-tremblay: @chrisdennis Done. Please have a look

henri-tremblay: Oups... no... not done.

henri-tremblay: @chrisdennis All good now

PR #2438, opened 2018-07-18
Fix #2437 : Wait for the client entity to complete init while processing server messages

### LINKED ISSUE #2437: ehcache3 Client side code Ignoring hash invalidation requests when not fully constructed
The following messages are still occassionally seen on ehcache3 clients:
```
[Clients][client0] [WorkerThread(multi_request_ack_stage, 0)] WARN org.ehcache.clustered.client.internal.store.SimpleClusterTierClientEntity - Ignoring the response org.ehcache.clustered.common.internal.messages.EhcacheEntityResponse$ClientInvalidateHash@1a39c335 as no registered response listener could be found.
```
This looks like a race condition situation as after this message typically the client timesout awaiting the hash invalidation message.

rkavanap:
This resolves issue #2437 and #2439

### DIFF
--- a/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
+++ b/clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
@@ -43,13 +43,13 @@ import org.terracotta.entity.MessageCodecException;
 import org.terracotta.exception.EntityException;
 
 import java.time.Duration;
-import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.LongSupplier;
 
@@ -71,6 +71,8 @@ public class SimpleClusterTierClientEntity implements InternalClusterTierClientE
   private final Map<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>> responseListeners =
     new ConcurrentHashMap<>();
   private final List<DisconnectionListener> disconnectionListeners = new CopyOnWriteArrayList<>();
+  private volatile boolean initCompleted = false;
+  private final Object initLock = new Object();
 
   private ReconnectListener reconnectListener = reconnectMessage -> {
     // No op
@@ -118,6 +120,10 @@ public class SimpleClusterTierClientEntity implements InternalClusterTierClientE
   }
 
   private <T extends EhcacheEntityResponse> void fireResponseEvent(T response) {
+    if (!initCompleted) {
+      // wait for init completion before processing any incoming message from server
+      waitForInitCompleted(timeouts.getReadOperationTimeout());
+    }
     @SuppressWarnings("unchecked")
     List<ResponseListener<T>> responseListeners = (List) this.responseListeners.get(response.getClass());
     if (responseListeners == null) {
@@ -170,6 +176,13 @@ public class SimpleClusterTierClientEntity implements InternalClusterTierClientE
   @Override
   public void validate(ServerStoreConfiguration clientStoreConfiguration) throws ClusterTierException, TimeoutException {
     try {
+      // validation is the first step after cluster tier entity creation. Hence assumed to be done only once
+      if (!initCompleted) {
+        synchronized (initLock) {
+          initCompleted = true;
+          initLock.notifyAll();
+        }
+      }
       invokeInternalAndWait(endpoint.beginInvoke(), timeouts.getConnectionTimeout(), messageFactory.validateServerStore(storeIdentifier , clientStoreConfiguration), false);
     } catch (ClusterException e) {
       throw new ClusterTierValidationException("Error validating cluster tier '" + storeIdentifier + "'", e);
@@ -207,6 +220,32 @@ public class SimpleClusterTierClientEntity implements InternalClusterTierClientE
   public EhcacheEntityResponse invokeAndWaitForRetired(EhcacheOperationMessage message, boolean track)
     throws ClusterException, TimeoutException {
     return invokeInternalAndWait(endpoint.beginInvoke().blockGetOnRetire(true), message, track);
+  }
+
+  private void waitForInitCompleted(Duration duration) {
+    boolean interrupted = false;
+    long deadlineTimeout = System.nanoTime() + duration.toNanos();
+    try {
+      while (true) {
+        try {
+          long timeRemainingInMillis = TimeUnit.NANOSECONDS.toMillis(deadlineTimeout - System.nanoTime());
+          if (timeRemainingInMillis <= 0 || initCompleted) {
+            return;
+          }
+          synchronized (initLock) {
+            if (!initCompleted) {
+              initLock.wait(timeRemainingInMillis);
+            }
+          }
+        } catch (InterruptedException e) {
+          interrupted = true;
+        }
+      }
+    } finally {
+      if (interrupted) {
+        Thread.currentThread().interrupt();
+      }
+    }
   }
 
   private EhcacheEntityResponse invokeInternalAndWait(InvocationBuilder<EhcacheEntityMessage, EhcacheEntityResponse> invocationBuilder, EhcacheOperationMessage message, boolean track)
--- a/clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntityTest.java
+++ b/clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntityTest.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.clustered.client.internal.store;
+
+import org.ehcache.clustered.common.internal.ServerStoreConfiguration;
+import org.ehcache.clustered.common.internal.messages.EhcacheEntityMessage;
+import org.ehcache.clustered.common.internal.messages.EhcacheEntityResponse;
+import org.ehcache.clustered.common.internal.messages.EhcacheResponseType;
+import org.junit.Test;
+import org.mockito.stubbing.Answer;
+import org.terracotta.entity.EndpointDelegate;
+import org.terracotta.entity.EntityClientEndpoint;
+import org.terracotta.entity.InvocationBuilder;
+import org.terracotta.entity.InvokeFuture;
+import org.terracotta.entity.MessageCodecException;
+import org.terracotta.exception.EntityException;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests that the cluster tier client entity fires any {@link ClusterTierClientEntity.ResponseListener} only after
+ * the {@link ClusterTierClientEntity#validate(ServerStoreConfiguration)} method is called.
+ */
+public class SimpleClusterTierClientEntityTest {
+  private static final int MAX_TEST_WAIT_TIME_SECONDS = 30;
+
+  @Test
+  public void testFireWithInit() throws Exception {
+    MockEndpointBuilder eb = new MockEndpointBuilder();
+    SimpleClusterTierClientEntity entity = new SimpleClusterTierClientEntity(eb.mockEndpoint);
+    AtomicBoolean responseRcvd = new AtomicBoolean(false);
+    entity.addResponseListener(MockedEntityResponse.class, response -> responseRcvd.set(true));
+    entity.validate(mock(ServerStoreConfiguration.class));
+    eb.mockInvocationBuilder.message(mock(EhcacheEntityMessage.class));
+    assertTrue(responseRcvd.get());
+  }
+
+  @Test
+  public void testFireWithDelayedInit() throws Exception {
+    MockEndpointBuilder eb = new MockEndpointBuilder();
+    SimpleClusterTierClientEntity entity = new SimpleClusterTierClientEntity(eb.mockEndpoint);
+    AtomicBoolean responseRcvd = new AtomicBoolean(false);
+    entity.addResponseListener(MockedEntityResponse.class, response -> responseRcvd.set(true));
+    CountDownLatch beforeLatch = new CountDownLatch(1);
+    CountDownLatch afterLatch = new CountDownLatch(1);
+    Executors.newFixedThreadPool(1).submit(() -> {
+      beforeLatch.countDown();
+      eb.mockInvocationBuilder.message(mock(EhcacheEntityMessage.class));
+      afterLatch.countDown();
+    });
+    assertTrue(beforeLatch.await(MAX_TEST_WAIT_TIME_SECONDS, TimeUnit.SECONDS));
+    // yield the cpu for some time to ensure the message receiver gets into its wait even on hyper slow uni processor machines
+    Thread.sleep(100);
+
+    // verify that response listener must be triggered only after validation is called on the entity
+    assertFalse(responseRcvd.get());
+    entity.validate(mock(ServerStoreConfiguration.class));
+
+    assertTrue(afterLatch.await(MAX_TEST_WAIT_TIME_SECONDS, TimeUnit.SECONDS));
+    assertTrue(responseRcvd.get());
+  }
+
+  @SuppressWarnings("unchecked")
+  private static class MockEndpointBuilder {
+    private final EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> mockEndpoint = mock(EntityClientEndpoint.class);
+    private final InvocationBuilder<EhcacheEntityMessage, EhcacheEntityResponse> mockInvocationBuilder = mock(InvocationBuilder.class);
+
+    public MockEndpointBuilder() {
+      AtomicReference<EndpointDelegate<EhcacheEntityResponse>> delegateRef = new AtomicReference<>();
+
+      when(mockInvocationBuilder.message(any(EhcacheEntityMessage.class))).thenAnswer(
+        (Answer<InvocationBuilder<EhcacheEntityMessage, EhcacheEntityResponse>>) invocationOnMock -> {
+          delegateRef.get().handleMessage(new MockedEntityResponse());
+          return mockInvocationBuilder;
+        });
+
+      try {
+        when(mockInvocationBuilder.invokeWithTimeout(anyLong(), any(TimeUnit.class))).thenAnswer(
+          (Answer<InvokeFuture<EhcacheEntityResponse>>) invocationOnMock -> new MockedInvokeFuture());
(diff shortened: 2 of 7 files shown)


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -288,4 +300,26 @@ private Duration getTimeoutDuration(EhcacheOperationMessage message) {
...
   }
+
+  private static void waitForLatch(long nanos, CountDownLatch latch) {
+    boolean interrupted = false;
+    long deadlineTimeout = System.nanoTime() + nanos;
+    try {
+      while (true) {
+        try {
+          long timeRemaining = deadlineTimeout - System.nanoTime();
+          if (timeRemaining <= 0) {
```
rkavanap: After a timeout, instead of throwing an error, it is better to attempt to look at the response listeners. It is highly unlikely that init will not complete for such a large duration.

rkavanap:
Actually, the ideal fix is to ensure that server does not send invalidation requests to not yet validated clients. But that requires much more deeper discussions and a much bigger change on the server side.

This particular fix is probably the second best approach . Anyway if the validation fails, the assumption is that the client will disconnect without sending any additional message of its own.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -288,4 +300,26 @@ private Duration getTimeoutDuration(EhcacheOperationMessage message) {
       }
     }
   }
+
+  private static void waitForLatch(long nanos, CountDownLatch latch) {
```
henri-tremblay: Can you rename that to waitForInitCompletion or something like that. Why do you need to pass the latch in parameter? Or maybe this method should go in some util class.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -288,4 +300,26 @@ private Duration getTimeoutDuration(EhcacheOperationMessage message) {
...
+    boolean interrupted = false;
+    long deadlineTimeout = System.nanoTime() + nanos;
+    try {
+      while (true) {
+        try {
+          long timeRemaining = deadlineTimeout - System.nanoTime();
+          if (timeRemaining <= 0) {
+            return;
+          }
+          latch.await(timeRemaining, TimeUnit.NANOSECONDS);
```
henri-tremblay: I don't really like the latch hanging around even though it will rarely be useful... Maybe we should just sleep and loop here until `initComplete` or timeout. To remove the latch.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -170,6 +177,11 @@ public boolean isConnected() {
   @Override
   public void validate(ServerStoreConfiguration clientStoreConfiguration) throws ClusterTierException, TimeoutException {
     try {
+      // validation is the first step. assumed to be done only once
```
henri-tremblay: First step of what?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: just followed an existing pattern in this class. See waitFor..Ideally I should just reuse waitFor and pass in a consumer.. will do that..

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: How about wait/notify? A sleep/wakeup loop may not be responsive enough.

Also since this is an cluster tier entity, we should not be expecting more than a few 100s max, right?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: btw, used a wait/notify instead and changed method name to waitForInitCompletion as per your suggestion.

rkavanap:
test this please

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -209,6 +222,32 @@ public EhcacheEntityResponse invokeAndWaitForRetired(EhcacheOperationMessage mes
     return invokeInternalAndWait(endpoint.beginInvoke().blockGetOnRetire(true), message, track);
   }
 
+  private void waitForInitCompleted(long nanos) {
```
henri-tremblay: I would pass the Duration here actually.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -209,6 +222,32 @@ public EhcacheEntityResponse invokeAndWaitForRetired(EhcacheOperationMessage mes
...
+    boolean interrupted = false;
+    long deadlineTimeout = System.nanoTime() + nanos;
```
henri-tremblay: Why playing with nanos and nanoTime if you are then converting to millis. Just use millis directly, The timeout doesn't need to be that precise.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -209,6 +222,32 @@ public EhcacheEntityResponse invokeAndWaitForRetired(EhcacheOperationMessage mes
...
+          if (timeRemaining <= 0 || initCompleted) {
+            return;
+          }
+          synchronized (initLock) {
+            if (!initCompleted) {
+              lock.wait(TimeUnit.NANOSECONDS.toMillis(timeRemaining));
+            }
+          }
+        } catch (InterruptedException e) {
+          interrupted = true;
```
henri-tremblay: Not going out of the loop when interrupted?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -71,6 +71,8 @@
   private final Map<Class<? extends EhcacheEntityResponse>, List<ResponseListener<? extends EhcacheEntityResponse>>> responseListeners =
     new ConcurrentHashMap<>();
   private final List<DisconnectionListener> disconnectionListeners = new CopyOnWriteArrayList<>();
+  private volatile boolean initCompleted = false;
+  private final Object initLock = new Object();
```
henri-tremblay: Can't we use the existing lock or any other private object in this class? And ok, I accept the solution with a notify :-)

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: No. But I am restoring the interrupt state of the thread. Did not want such a crucial wait to be interrupted leaving the overall state in a mess..This way atleast we have a clear exit path on interrupt. Anyway it is not an indefinite wait...

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: For relative time calculations System.nanotime is a bit more precise than System.currentTimeMillis (and in this particular case we have no need to know the exact time).

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: There is a lock object..but its purpose was different..did not want to mix concerns that might cause problems later..

rkavanap:
@henri-tremblay , I have added one more fix #2439 to this PR as a separate commit.

Since platform provides no gurarantees in the order in which entities are released on a disconnection (ehcache does not explictly close entities), a close followed by a destroy may race on the server, where the fetch entity of a cache manager build for maintenance and destroy may overtake the RELEASE_ENTITY of the previous close on a very slow running server. I have logs that prove this theory.

,,And the reason why the above can happen is because the entities themselves are not explictly closed before the connection is dropped (as seen in the following code comment):

``
    /*
     * Entity close() operations must *not* be called; if the server connection is disconnected, the entity
     * close operations will stall attempting to communicate with the server.  (EntityClientEndpointImpl.close()
     * calls a "closeHook" method provided by ClientEntityManagerImpl which ultimately winds up in
     * InFlightMessage.waitForAcks -- a method that can wait forever.)  Theoretically, the connection close will
     * take care of server-side cleanup in the event the server is connected.
     */
``


So we must explicitly release even read locks.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
henri-tremblay: Ok. I agree me way is more hacky

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -209,6 +222,32 @@ public EhcacheEntityResponse invokeAndWaitForRetired(EhcacheOperationMessage mes
...
+          if (timeRemaining <= 0 || initCompleted) {
+            return;
+          }
+          synchronized (initLock) {
+            if (!initCompleted) {
+              initLock.wait(TimeUnit.NANOSECONDS.toMillis(timeRemaining));
+            }
+          }
+        } catch (InterruptedException e) {
+          interrupted = true;
```
henri-tremblay: For me, interrupting here is like having a smaller timeout. So we should go out of the loop directly. Because it means someone has explicitly called interrupt on this thread. If having a too short timeout breaks everything, than we have another problem.

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+  private static final String CACHE_MANAGER_NAME = "/crud-cm";
+  private static final int PRIMARY_SERVER_RESOURCE_SIZE = 4; //MB
+  private static final int NUM_THREADS = 20;
+
+  private AtomicReference<Throwable> exception = new AtomicReference<>();
+
+
+  @Test
+  public void testMulipleClients() throws Throwable {
+    final CountDownLatch latch = new CountDownLatch(NUM_THREADS + 1);
```
henri-tremblay: final flag not needed

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+      t.join();
+    }
+
+    Throwable throwable = exception.get();
+    if (throwable != null) {
+      throw throwable;
+    }
+  }
+
+  private Runnable content(final CountDownLatch latch) {
```
henri-tremblay: No final needed here as well

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+    return () -> {
+      try {
+        PersistentCacheManager cacheManager = createCacheManager(CLUSTER.getConnectionURI());
+
+        latch.countDown();
+        try {
+          latch.await();
```
henri-tremblay: Please set a timeout here. In case of trouble, the test will eventually finish

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+        } catch (InterruptedException e) {
+          // continue
+        }
+
+        cacheManager.init();
+        doSyncAndPut(cacheManager);
+      } catch (Throwable t) {
+        exception.compareAndSet(null, t); // only keep the first exception
```
henri-tremblay: Would it be useful to add the other exceptions as suppressed exceptions?

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+  private void silentPuts(Cache<Long, CustomValue> customValueCache) {
+    customValueCache.put(2L, new CustomValue("value1"));
+    customValueCache.put(2L, new CustomValue("value2"));
+    customValueCache.put(3L, new CustomValue("value3"));
+    customValueCache.put(4L, new CustomValue("value4"));
+    customValueCache.put(2L, new CustomValue("value5"));
+  }
+
+  private static class CustomValue implements Serializable {
```
henri-tremblay: Do you really need the CustomValue? Why not just putting a String? If you do, I'm pretty sure you don't need the equals and hashcode

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+  private static final String PRIMARY_SERVER_RESOURCE_NAME = "primary-server-resource";
+  private static final String CACHE_MANAGER_NAME = "/crud-cm";
+  private static final int PRIMARY_SERVER_RESOURCE_SIZE = 4; //MB
+  private static final int NUM_THREADS = 20;
+
+  private AtomicReference<Throwable> exception = new AtomicReference<>();
+
+
+  @Test
+  public void testMulipleClients() throws Throwable {
```
henri-tremblay: Can you explain what the test is doing in the javadoc? What's expected and why.

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+        cacheManager.init();
+        doSyncAndPut(cacheManager);
+      } catch (Throwable t) {
+        exception.compareAndSet(null, t); // only keep the first exception
+      }
+    };
+  }
+
+  private void doSyncAndPut(PersistentCacheManager cacheManager) throws InterruptedException {
```
henri-tremblay: Complicated question: Could you have a simpler test by adding hooks in the code to let you wait at the right places with a latch to reproduce the bug?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
rkavanap: hmm.. if we get out of the loop faster, won't we get into an exception path or a confusing log before the higher layer gets a chance to handle the interrupt, as the init is not done. Also remember that the init itself is a set of local operations, so it is highly unlikely that it will take more than a few micros, unless the machine itself is overloaded. 

Besides this is the same pattern we follow for all invoke messages (see SimpleClusterTierClientEntity)..we make it an uninterruptible wait but then update the interrupt status of the thread. Not sure whether I should change that pattern here (esp, since this is not even a remote operation we are waiting for).

henri-tremblay: Ok... Let's keep it like this and revisit in a more global way later.

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
rkavanap: I think at some point we should look at tools like byteman to do what you suggested in a clean way..

http://byteman.jboss.org/

henri-tremblay: If you go heavyweight yes. Easier is just to put such a hook in the code.

rkavanap:
test this please

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
rkavanap: Ok. I will try. But I am personally not a big fan of adding test specific code in production code.. If we can mock things and achieve the same it is OK..but in this case, due to the multi-threaded nature of the problem, it is very hard to achieve the same result using mocking. Let me give it a try though.

rkavanap:
test this please

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
rkavanap: @henri-tremblay created a unit test using mocks..Managed to do it without stubs in production code. Pls take a look.

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
rkavanap: added to the test class header

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
henri-tremblay: Can you please comment or fix?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntityTest.java
```diff
@@ -0,0 +1,148 @@
...
+    entity.addResponseListener(MockedEntityResponse.class, response -> responseRcvd.set(true));
+    CountDownLatch beforeLatch = new CountDownLatch(1);
+    CountDownLatch afterLatch = new CountDownLatch(1);
+    Executors.newFixedThreadPool(1).submit(() -> {
+      beforeLatch.countDown();
+      eb.mockInvocationBuilder.message(mock(EhcacheEntityMessage.class));
+      afterLatch.countDown();
+    });
+    assertTrue(beforeLatch.await(MAX_TEST_WAIT_TIME_SECONDS, TimeUnit.SECONDS));
+    Thread.sleep(100);
```
henri-tremblay: I think instead of the sleep, you want 
```
long end = System.currentMillis() + 10_000;
while(responseRcvd.get()) {
  if(System.currentMillis() > end) {
    fail("Response listener was never called");
    break;
  }
}
```

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
henri-tremblay: All good (just a little comment)

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntityTest.java
rkavanap: just to get the other thread some time to enter its wait..so that this thread does not complete its init too fast...

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
rkavanap: Oopss missed that. Btw, when I initially saw this comment,  did not completely understand the value of that in a unit test. Currently the test fails only with one exception. Do you want a log of all other exceptions..? Typically all failures will be logged anyway in the server log.

rkavanap:
test this please

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntityTest.java
henri-tremblay: It is only to force yield?  (then line comment it). I was wondering if there should be a `while(responseRcvd.get())` instead of the assert below.

rkavanap: done

PR #2443, opened 2018-07-25
Generate notification (CACHE_CLEARED) when Cache.clear() is called

jjian77:


terracotta-org:

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -113,6 +121,20 @@ public void cacheRemoved(String alias, Cache<?, ?> cache) {
         EhcacheNotification.CACHE_REMOVED.name()));
   }
 
+  private void cacheCleared(String alias) {
+    collector.onNotification(
```
chrisdennis: Whatever this method does it's synchronous with the clear.  It's not anything expensive is it?

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -113,6 +121,20 @@ public void cacheRemoved(String alias, Cache<?, ?> cache) {
...
+      new ContextualNotification(
+        configuration.getContext().with("cacheName", alias),
+        EhcacheNotification.CACHE_CLEARED.name()));
+  }
+
+  private void registerClearNotification(String alias, Cache<?, ?> cache) {
+    OperationStatistic<ClearOutcome> clear = StatsUtils.findOperationStatisticOnChildren(cache,
+      ClearOutcome.class, "clear");
+    clear.addDerivedStatistic(new OperationResultFilter<>(EnumSet.of(ClearOutcome.SUCCESS),
+      (long time, long latency) -> cacheCleared(alias)));
```
chrisdennis: You shouldn't need to type these parameters.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
```diff
@@ -113,6 +121,20 @@ public void cacheRemoved(String alias, Cache<?, ?> cache) {
...
 
+  private void cacheCleared(String alias) {
+    collector.onNotification(
+      new ContextualNotification(
+        configuration.getContext().with("cacheName", alias),
+        EhcacheNotification.CACHE_CLEARED.name()));
+  }
+
+  private void registerClearNotification(String alias, Cache<?, ?> cache) {
+    OperationStatistic<ClearOutcome> clear = StatsUtils.findOperationStatisticOnChildren(cache,
```
chrisdennis: I assume this is safe from a GC point of view? Anything un-natural with the context stuff makes me jumpy.

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
anthonydahanne: nope; it's pushing an event

management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java
anthonydahanne: ah right
```(long time, long latency) -> cacheCleared(alias)));```

**Comment by anthonydahanne** [comment]
test this please

File: management/src/main/java/org/ehcache/management/registry/DefaultCollectorService.java [hunk 4]
```diff
@@ -113,6 +121,20 @@ public void cacheRemoved(String alias, Cache<?, ?> cache) {
...
+      ClearOutcome.class, "clear");
+    clear.addDerivedStatistic(new OperationResultFilter<>(EnumSet.of(ClearOutcome.SUCCESS),
```
henri-tremblay: This looks like a hack to me... Using the statistics to get the information about a clear. I don't like it at all. The statistics should not carry business operations. I would prefer the listener to be called by `EhcacheBase.clear()`. But that would mean having it available in `EhcacheBase`. Can't you work with the `CacheEventDispatcher`?

anthonydahanne: hello @henri-tremblay can you please provide guidance on how to this differently ? There is no ```cacheCleared``` in  ```CacheManagerListener```
How could we implement ```cacheCleared``` in ```CacheEventDispatcher``` ?
The "hack" suggested in a 3 lines snippet is totally limited in scope, and actually, pretty readable and understandable, and will provide an important feature for users. If you think there's a "cleaner" way to implement this feature, please share it with us

lorban: This looks like a hack indeed, but a reasonably robust one since it relies on the already existing cache clear statistic. It's also fairly elegant since the hackery is rather self-contained.

Firing this notification from `EhcacheBase` would take a much larger effort which probably isn't worth it since we do not want to expose cache clearing events through the API as the only use-case that makes sense for this notification is M&M.

henri-tremblay: Ok. Going with the family then.

henri-tremblay:
Can you rebase?

PR #2449, opened 2018-08-06
Use an expiration strategy depending on expiration policy

henri-tremblay:
This gives about 5% performance gain when no expiration is set and the SystemTimeSource is used. To get better, some further work is required on the `OnHeapValueHolder`.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
@@ -1378,15 +1378,8 @@ public class OnHeapStore<K, V> extends BaseStore<K, V> implements HigherCachingT
     Objects.requireNonNull(oldValue);
     Objects.requireNonNull(newValue);
 
-    Duration duration = Duration.ZERO;
-    try {
-      duration = expiry.getExpiryForUpdate(key, oldValue, newValue);
-      if (duration != null && duration.isNegative()) {
-        duration = Duration.ZERO;
-      }
-    } catch (RuntimeException re) {
-      LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
-    }
+    Duration duration = strategy.getUpdateDuration(key, oldValue, newValue);
+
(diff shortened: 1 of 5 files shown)


impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStrategy.java
```diff
@@ -0,0 +1,262 @@
...
+          duration = Duration.ZERO;
+        }
+      } catch (RuntimeException re) {
+        LOG.error("Expiry computation caused an exception - Expiry duration will be 0 ", re);
+        duration = Duration.ZERO;
+      }
+      return duration;
+    }
+
+    public OnHeapValueHolder<V> setAccessTimeAndExpiryThenReturnMappingUnderLock(K key, OnHeapValueHolder<V> valueHolder, long now,
```
chrisdennis: This "UnderLock" in this method name is odd in this context. Ut's odd that the name refers to a contract detail related to the calling class.

henri-tremblay:
Both works are not related. However, we can hold of if you think it's better. But why would it be more dangerous now than later? It's more a it works or not.

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -1289,44 +1292,15 @@ public boolean equals(Object obj) {
   }
 
   private void setAccessTimeAndExpiryThenReturnMappingOutsideLock(K key, OnHeapValueHolder<V> valueHolder, long now) {
```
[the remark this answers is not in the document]
henri-tremblay: Yes

impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStrategy.java
henri-tremblay: Ok. I found something

henri-tremblay:
@chrisdennis Can you revisit and tell me if we can merge now?

PR #2452, opened 2018-08-08
Closes #2451 - ignore unresolvable optional dependencies that are present on the classpath

### LINKED ISSUE #2451: Regression between 3.5.2 and master : Failed to find provider with satisfied dependency set for interface
I used to be able to run such code with 3.5.2 : 

```
DefaultManagementRegistryConfiguration managementRegistryConfiguration = new DefaultManagementRegistryConfiguration()
        .setCacheManagerAlias(CM_NAME);
DefaultManagementRegistryService managementRegistry = new DefaultManagementRegistryService(managementRegistryConfiguration);
    managementRegistryConfiguration.addTags("pif", "paf", "pouf");

    CacheManagerBuilder cacheManagerBuilder = CacheManagerBuilder.newCacheManagerBuilder()
        .using(managementRegistry);
    CacheManager cacheManager = cacheManagerBuilder.build();

```

Now, (current master), it will throw when ```cacheManagerBuilder.build()``` is called: 

```
Exception in thread "main" java.lang.IllegalStateException: org.ehcache.core.internal.service.ServiceLocator$DependencyException: Failed to find provider with satisfied dependency set for interface org.ehcache.clustered.client.service.EntityService [candidates []]
	at org.ehcache.core.internal.service.ServiceLocator$DependencySet.build(ServiceLocator.java:321)
	at org.ehcache.core.EhcacheManager.resolveServices(EhcacheManager.java:146)
	at org.ehcache.core.EhcacheManager.<init>(EhcacheManager.java:119)
	at org.ehcache.core.EhcacheManager.<init>(EhcacheManager.java:109)
	at org.ehcache.config.builders.CacheManagerBuilder.newCacheManager(CacheManagerBuilder.java:114)
	at org.ehcache.config.builders.CacheManagerBuilder.build(CacheManagerBuilder.java:70)
	at AnthonyClient.start(AnthonyClient.java:107)
	at AnthonyClient.main(AnthonyClient.java:47)
Caused by: org.ehcache.core.internal.service.ServiceLocator$DependencyException: Failed to find provider with satisfied dependency set for interface org.ehcache.clustered.client.service.EntityService [candidates []]
	at org.ehcache.core.internal.service.ServiceLocator$DependencySet.lookupService(ServiceLocator.java:363)
	at org.ehcache.core.internal.service.ServiceLocator$DependencySet.lookupDependenciesOf(ServiceLocator.java:327)
	at org.ehcache.core.internal.service.ServiceLocator$DependencySet.build(ServiceLocator.java:302)
	... 7 more
```

lorban:
This is the simplest way I've found to ignore optional dependencies.

### DIFF
--- a/core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
+++ b/core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
@@ -30,6 +30,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Deque;
 import java.util.HashMap;
@@ -324,7 +325,16 @@ public final class ServiceLocator implements ServiceProvider<Service> {
 
     ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> requested) throws DependencyException {
       for (Class<? extends Service> dependency : identifyImmediateDependenciesOf(requested)) {
-        resolved = lookupService(resolved, dependency);
+        try {
+          resolved = lookupService(resolved, dependency);
+        } catch (DependencyException de) {
+          OptionalServiceDependencies optionalAnnotation = requested.getAnnotation(OptionalServiceDependencies.class);
+          if (optionalAnnotation != null && Arrays.asList(optionalAnnotation.value()).contains(dependency.getName())) {
+            LOGGER.debug("Skipping optional dependency of {} that cannot be looked up: {}", requested, dependency);
+            continue;
+          }
+          throw de;
+        }
       }
       return resolved;
     }
@@ -425,9 +435,9 @@ public final class ServiceLocator implements ServiceProvider<Service> {
         }
       }
     }
-    OptionalServiceDependencies optionaAnnotation = clazz.getAnnotation(OptionalServiceDependencies.class);
-    if (optionaAnnotation != null) {
-      for (String className : optionaAnnotation.value()) {
+    OptionalServiceDependencies optionalAnnotation = clazz.getAnnotation(OptionalServiceDependencies.class);
+    if (optionalAnnotation != null) {
+      for (String className : optionalAnnotation.value()) {
         try {
           Class<?> dependencyClass = ClassLoading.getDefaultClassLoader().loadClass(className);
           if (Service.class.isAssignableFrom(dependencyClass)) {
--- a/core/src/test/java/org/ehcache/core/internal/service/ServiceLocatorTest.java
+++ b/core/src/test/java/org/ehcache/core/internal/service/ServiceLocatorTest.java
@@ -26,6 +26,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import org.ehcache.core.EhcacheWithLoaderWriter;
 import org.ehcache.core.spi.store.CacheProvider;
+import org.ehcache.spi.service.OptionalServiceDependencies;
 import org.ehcache.spi.service.ServiceProvider;
 import org.ehcache.spi.loaderwriter.CacheLoaderWriterProvider;
 import org.ehcache.spi.service.Service;
@@ -37,7 +38,6 @@ import org.ehcache.core.spi.services.FancyCacheProvider;
 import org.ehcache.core.spi.services.TestProvidedService;
 import org.ehcache.core.spi.services.TestService;
 import org.hamcrest.CoreMatchers;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import static org.ehcache.core.internal.service.ServiceLocator.dependencySet;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -324,7 +325,16 @@ public ServiceLocator build() {
 
     ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> requested) throws DependencyException {
       for (Class<? extends Service> dependency : identifyImmediateDependenciesOf(requested)) {
-        resolved = lookupService(resolved, dependency);
+        try {
```
henri-tremblay: I'm fine with the general concept. The thing I don't get is why blending normal dependencies and optional ones. They could be resolved separately: `identifyImmediateDependenciesOf` and `identifyImmediateOptionalDependenciesOf`. With the exception check only on the later. This is not directly the code of this PR. It is a quirk in the original code. The PR just makes it more obvious.

anthonydahanne:
I'm cool with the changes and the tests too.
Let me run that locally with my sample code

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
lorban: I've refactored the code to introduce `identifyImmediateOptionalDependenciesOf` but I do find the code less readable.

But I'll leave it up to you which form you prefer.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -425,9 +438,28 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
...
+      }
+    }
+
+    dependencies.addAll(identifyImmediateDependenciesOf(clazz.getSuperclass()));
+
+    return dependencies;
+  }
+
+  private static Set<Class<? extends Service>> identifyImmediateOptionalDependenciesOf(final Class<?> clazz) {
+    if (clazz == null) {
```
henri-tremblay: Why would clazz be null?

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -326,6 +327,18 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
       for (Class<? extends Service> dependency : identifyImmediateDependenciesOf(requested)) {
         resolved = lookupService(resolved, dependency);
       }
+      for (Class<? extends Service> dependency : identifyImmediateOptionalDependenciesOf(requested)) {
+        try {
+          resolved = lookupService(resolved, dependency);
+        } catch (DependencyException de) {
```
henri-tremblay: My idea was to move the try/catch inside identifyImmediateOptionalDependenciesOf

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
lorban: Go figure. It's present in `identifyImmediateDependenciesOf` so I kept it in `identifyImmediateOptionalDependenciesOf` too. Better be safe than sorry.

henri-tremblay: Hum... Recursion I guess

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
lorban: That can't be done with the way the code currently is structured as you can only know if an optional dependency can be resolved or not when it is being looked up.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
```diff
@@ -445,11 +477,11 @@ ServiceMap lookupDependenciesOf(ServiceMap resolved, Class<? extends Service> re
 
     for (Class<?> interfaceClazz : clazz.getInterfaces()) {
       if (Service.class.isAssignableFrom(interfaceClazz)) {
-        dependencies.addAll(identifyImmediateDependenciesOf(Service.class.getClass().cast(interfaceClazz)));
+        dependencies.addAll(identifyImmediateOptionalDependenciesOf(Service.class.getClass().cast(interfaceClazz)));
```
henri-tremblay: It feels like we could recurse over `lookupDependenciesOf` to prevent the code duplication between optional and not optional. Or keep the code mechanic and pass a lambda to have an optional/non-optional behavior.

core/src/main/java/org/ehcache/core/internal/service/ServiceLocator.java
henri-tremblay: Ah... Not that useful then

PR #2460, opened 2018-08-20
Do not use derived statistics when not needed

henri-tremblay:


henri-tremblay:
test this please

mathieucarbou:
- The ability to change the histogram window should be kept and moved to the management service config (and perhaps be customizable through xml config like the other management config Params), and the setting should be used inside Standard Ehcache Statistics class.

- the unit tests should be kept and moved to management module.

chrisdennis:
`core/build.gradle` should have an `api` dependency on statistics now since it exposes stats classes through StatisticsService.

core/src/main/java/org/ehcache/core/statistics/CacheStatistics.java
```diff
@@ -41,6 +40,16 @@
...
 
+  /**
+   * Register a derived statistic to one of the existing statistic.
+   *
+   * @param outcomeClass the enum of the possible outcomes
+   * @param statName name of the statistic we are looking for
+   * @param derivedStatistics derived statistic to register
+   * @param <T> type of the outcome
+   */
+  <T extends Enum<T>> void registerDerivedStatistics(Class<T> outcomeClass, String statName, ChainedOperationObserver<? super T> derivedStatistics);
```
chrisdennis: It would make calling code simpler if this was:
```
<T extends Enum<T>, S extends ChainedOperationObserver<? super T>> S registerDerivedStatistic(Class<T> type, String name, S statistic);
```

henri-tremblay: Thanks

henri-tremblay: Hum. It is actually worse... Nothing compiled with this syntax

chrisdennis: How did it not compile?

henri-tremblay: I've changed it to void. Then, I had a compilation issue.  But just realized I was being stupid. You remove the s at the end. So that's the only reason... Doh. Will push the change.

henri-tremblay:
Can't be merged right now. Statistics framework should be updated first

henri-tremblay:
All good now... except for the build

chrisdennis:
What about the core/build.gradle change? I don't see that here.

henri-tremblay:
Which core/build.gradle change?

chrisdennis:
https://github.com/ehcache/ehcache3/pull/2460#issuecomment-417062466

henri-tremblay:
Oh! Sorry. I thought it was already done. Done now.

henri-tremblay:
@mathieucarbou @chrisdennis Could please validate and merge if ok?

mathieucarbou:
@henri-tremblay I will next week.

core/src/main/java/org/ehcache/core/spi/service/StatisticsServiceConfiguration.java
```diff
@@ -21,6 +21,16 @@
...
 
+  /**
+   * When {@code true}, latencies will be calculated for the cache. If {@code null} (the default), latencies will only
+   * be calculated if the management module is used.
+   * <p>
+   * Note: Calculating latencies has a performance impact. Especially in a concurrent environment.
+   *
+   * @return if latencies should be calculated
+   */
+  Boolean getCalculateLatencies();
```
anthonydahanne: hum ... usually boolean begins with "isXXX"

impl/src/main/java/org/ehcache/impl/internal/statistics/DefaultStatisticsService.java
```diff
@@ -28,6 +28,7 @@
 import org.ehcache.core.spi.time.TimeSource;
 import org.ehcache.core.spi.time.TimeSourceService;
 import org.ehcache.core.statistics.CacheStatistics;
+import org.ehcache.spi.service.OptionalServiceDependencies;
```
anthonydahanne: unused import, but fixed in later commit I think

PR #2461, opened 2018-08-22
x.y makes more sense than y.y mathematically

henri-tremblay:


henri-tremblay:
Since it's a doc fix, and I tiny one, I wouldn't do a daggy. Do we deploy old documentation?

chrisdennis:
Rebase and I will merge.

henri-tremblay:
It is rebased no?

PR #2463, opened 2018-08-22
Fix#2437 :- Proper fix for clientinvalidation race

rkavanap:
This is the correct fix for clientinvalidation race..

The previous fix is reverted along with this...


clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/MultiThreadedStrongServerStoreProxyTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+import static org.ehcache.clustered.common.internal.store.Util.createPayload;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+
+public class MultiThreadedStrongServerStoreProxyTest extends AbstractServerStoreProxyTest {
```
rkavanap: Attempts to reproduce the problem on a proper blackbox test was unsuccessful. Hence created a white box test that asserts that invalidate events are fired only after a client is validated.

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -181,11 +181,17 @@ private static PersistentCacheManager createCacheManager(URI clusterURI) {
     return clusteredCacheManagerBuilder.build(false);
   }
 
-  private void silentPuts(Cache<Long, String> customValueCache) {
-    customValueCache.put(2L, "value2");
-    customValueCache.put(2L, "value21");
-    customValueCache.put(3L, "value3");
-    customValueCache.put(4L, "value4");
-    customValueCache.put(2L, "value22");
+  private void parallelPuts(Cache<Long, String> customValueCache) {
```
rkavanap: These changes were put to reproduce locally..

I was able to reproduce the problem only when I put an explicit random sleep inside the connected() method which cannot be raised in a pull request..

But I have manually verified the issue is fixed this time by verifying it with a random sleep inside the connected() method, which was then reverted before raising this pull request.

PR #2464, opened 2018-08-23
Remove hit rate from ValueHolder since it's unused

henri-tremblay:


henri-tremblay:
No. The hitrate was totally unused before.

chrisdennis:
Right - but you've changed the serialized form of this class right?  Doesn't this risk breaking things?

henri-tremblay:
This class isn't serializable. However, `XAValueHolder` is through `SerializedXAValueHolder`. I should at least change the serialVersionUID on it.

`OffHeapValueHolderPortability` also is with our custom serialization.

It means that someone saving the data to disk and migrating to Ehcache 3.6 won't be able to recover. I think it is highly unlikely to happen right now. So I would keep it like this. The alternatives are
1- Keep saving and reading a dummy field to keep compatibility
2- Support the two versions with some versioning. I'm not sure we have the information for that in the serialized value holder though.

chrisdennis:
The `SerializedXAValueHolder` should be okay... as long as we don't change the serialVersionUID the new class will just ignore the extra field in the serialization stream.  For the `OffHeapValueHolderPortability` I think we're safest to just continue to write and read the dummy long... at least until we're moving a major version number... that migration issue would affect all the persistent local caches.

henri-tremblay:
Major would be Ehcache 4.x or 3.7?

chrisdennis:
If we take the official party line I think that would mean 4.x. 3.7 could be a topic for discussion... I'm not sure how our backwards compatibility commitments translate in to the unclustered world.

henri-tremblay:
So, `SerializedXAValueHolder` has the same serialVersionUID. I've put back compatibility to `OffHeapValueHolderPortability`

PR #2465, opened 2018-08-24
Make sure that all projects are using the statistic version from parent

henri-tremblay:
Basically, when the terracotta-statistics version is changed in gradle.properties, the project should compile directly. Right now it fails on tons of dependency conflicts. This PR solves that by ignoring and adding an explicit dependency at the right place.

### DIFF
--- a/clustered/integration-test/build.gradle
+++ b/clustered/integration-test/build.gradle
@@ -29,8 +29,13 @@ dependencies {
   testCompileOnly "org.terracotta:lease-api:$terracottaPlatformVersion"
 
   testImplementation project(':management')
-  testImplementation "org.terracotta.management.dist:mnm-nms:$terracottaPlatformVersion"
-  testImplementation "org.terracotta.management.dist:mnm-nms-agent:$terracottaPlatformVersion"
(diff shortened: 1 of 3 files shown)


chrisdennis:
If we did this it would mask any  divergence between the direct ehcache dependency and the transitive dependency through the management modules. This runs the risk of introducing bugs caused by forcing management to try and operate against the wrong version of statistics. If we're lucky that gets caught by testing before it gets released. If we're unlucky it doesn't. If we want to make the situation easier here the right way to do it is to try and simplify the dependency tree so that statistics is pulled through less paths (if that is indeed possible - or worth the effort).

henri-tremblay:
So you are alright for someone to lose half a day every time he wants to use a different of statistics? Can't we just implement a check to make sure everything is in sync? And yes, in general, a bad version of statistics just crashes the app right away. So it is quite easy to know that something is wrong.

Less path is not the solution. Changing a version in `gradle.properties` should change it for the entire project. We need a way to get there. The entire Spring Boot is working that way and has much much more cross-dependencies than we will ever have.

henri-tremblay:
Ah! That's better. Ok. Let's do this until we figure out how to cleanup the dependencies.

PR #2467, opened 2018-08-24
Remove an impossible possibility during a putIfAbsent

henri-tremblay:


AbfrmBlr:
needs rebase

henri-tremblay:
Rebased

chrisdennis:
Abhilash is doing work that will likely pull the loader-writer logic out in to a store layer.  This is going obviously going to be disruptive of lots of this code. It would seem to make more sense to me to hold off on doing anything here until Abhilash was finished.

henri-tremblay:
No. This is in sync with what @AbfrmBlr is doing. The PR comes from a question he asked.

AbfrmBlr:
Yes, but it would be good if we can hold this for a bit.

core/src/main/java/org/ehcache/core/EhcacheBase.java
```diff
@@ -297,9 +297,6 @@ public V putIfAbsent(final K key, final V value) {
         if (put[0]) {
           putIfAbsentObserver.end(PutIfAbsentOutcome.PUT);
           return null;
-        } else if (inCache == null) {
```
AbfrmBlr: Other than the fact that this part of code was unnecessary, I was more interested in `Consumer<Boolean> put` to go, which is unlikely and we need it. Thus I still need to make `Store` spi change respectively. So lets wait for the Store layer refactoring.

henri-tremblay: Oh! Wait a second. We actually might now that this if was removed. Silly me. Let me look.

henri-tremblay: Done. No consumer anymore

AbfrmBlr: this helps

AbfrmBlr:
This has test failures

chrisdennis:
@AbfrmBlr since this interacts with your stuff I'll let you review/approve/merge this.

henri-tremblay:
to trigger a build

PR #2480, opened 2018-09-11
Ability to autoconfigure Ehcache

mathieucarbou:
We have the requirement to be able to automatically configure a management registry in some case.

So this PR is to add the ability to provide within the classpath a configuration provider that would be discovered and add configs if they are not already there through programmatic or xml config.

This PR is not complete:
- I'll send an email for more details and use case
- If this solution is good, we have to decide where to put the provider interface and depending on this what we document

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/management/ManagementActivator.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/management/ManagementActivator.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2011-2018 Software AG, Darmstadt, Germany and/or Software AG USA Inc., Reston, VA, USA, and/or its subsidiaries and/or its affiliates and/or their licensors.
+ * Use, reproduction, transfer, publication or disclosure is prohibited except as specifically provided for in your License Agreement with Software AG.
+ */
+package org.ehcache.clustered.management;
+
+import org.ehcache.management.ManagementRegistryService;
+import org.ehcache.management.registry.DefaultManagementRegistryConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfigurationProvider;
+
+/**
+ * Activates management automatically if no config has been previously set manually or through XML.
+ *
+ * @author Mathieu Carbou
+ */
+public class ManagementActivator implements ServiceCreationConfigurationProvider<ManagementRegistryService> {
+  @Override
+  public Class<ManagementRegistryService> getServiceType() {
+    return ManagementRegistryService.class;
+  }
+
+  @Override
+  public ServiceCreationConfiguration<ManagementRegistryService> get() {
+    return new DefaultManagementRegistryConfiguration()
+      .addTags("webapp-1", "server-node-1")
+      .setCacheManagerAlias("my-super-cache-manager");
+  }
+
+}
--- a/impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
+++ b/impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
@@ -21,11 +21,18 @@ import org.ehcache.config.CacheConfiguration;
 import org.ehcache.config.Configuration;
 import org.ehcache.core.config.DefaultConfiguration;
 import org.ehcache.spi.service.ServiceCreationConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfigurationProvider;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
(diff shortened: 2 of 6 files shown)


impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
```diff
@@ -122,4 +125,14 @@ public ConfigurationBuilder withClassLoader(ClassLoader classLoader) {
   public boolean containsCache(String alias) {
     return caches.containsKey(alias);
   }
+
+  private void discoverServiceConfiguration() {
+    for (ServiceCreationConfigurationProvider<?> provider : ServiceLoader.load(ServiceCreationConfigurationProvider.class, classLoader)) {
+      // we only add an discovered service
```
henri-tremblay: 'a'. not 'an'

henri-tremblay:
I would love an example indeed. For `ehcache.xml` maybe. For programmatic, I don't see why a `ServiceCreationConfiguration` can't be given directly.

mathieucarbou:
@henri-tremblay : see my email and use case in email ;-) this is for ee

anthonydahanne:
LGTM ; elegant way to let ehcache configure services via META-INF/services ; will be useful for Management and monitoring

henri-tremblay:
My question is: Why do we need a provider? Why not getting a `DefaultManagementRegistryConfiguration` directly in the services? Using a factory method to get your defaults.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ManagementActivator.java
```diff
@@ -0,0 +1,30 @@
...
+import org.ehcache.management.registry.DefaultManagementRegistryConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfiguration;
+import org.ehcache.spi.service.ServiceCreationConfigurationProvider;
+
+/**
+ * Activates management automatically if no config has been previously set manually or through XML.
+ *
+ * @author Mathieu Carbou
+ */
+public class ManagementActivator implements ServiceCreationConfigurationProvider<ManagementRegistryService> {
```
henri-tremblay: So why not a `DefaultManagementRegistryConfiguration.defaultConfiguration()` instead?

impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
```diff
@@ -73,7 +75,8 @@ private ConfigurationBuilder(ConfigurationBuilder builder, ClassLoader classLoad
 
   @Override
   public Configuration build() {
-    return new DefaultConfiguration(caches, classLoader, serviceConfigurations.toArray(new ServiceCreationConfiguration<?>[serviceConfigurations.size()]));
+    List<ServiceCreationConfiguration<?>> serviceCreationConfigurations = addDiscoveredServiceConfiguration();
```
henri-tremblay: Weird to return a list just to go to an array right after

impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
```diff
@@ -122,4 +125,16 @@ public ConfigurationBuilder withClassLoader(ClassLoader classLoader) {
   public boolean containsCache(String alias) {
     return caches.containsKey(alias);
   }
+
+  public List<ServiceCreationConfiguration<?>> addDiscoveredServiceConfiguration() {
+    List<ServiceCreationConfiguration<?>> newServiceConfigurations = new ArrayList<>(serviceConfigurations);
```
henri-tremblay: Why would the discovered ones be overridden by the explicit ones?

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ManagementActivator.java
mathieucarbou: default is just equivalent of `new DefaultManagementRegistryConfiguration()`.

Here, i am just creating a sample use case to show you how it would work ;-) but i need to check the tests because some are failing probably because of the values here.

impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
mathieucarbou: How you would do it more elegantly ? The list not being modifiable, we have to recreate it :-(

impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
mathieucarbou: because we want to have management activated by default in the EE kit but users could still programmatically or through xml change the management config to customize it.

impl/src/main/java/org/ehcache/config/builders/ConfigurationBuilder.java
henri-tremblay: @SuppressWarnings("unchecked")
  public ServiceCreationConfiguration<?>[] addDiscoveredServiceConfiguration() {
    return Stream.concat(serviceConfigurations.stream(), 
      StreamSupport.stream(ServiceLoader.load(ServiceCreationConfigurationProvider.class, classLoader).spliterator(), false)
      .filter(provider -> findServiceByClass(provider.getServiceType()) == null)
      .map((Function<ServiceCreationConfigurationProvider, Object>) Supplier::get))
      .toArray(ServiceCreationConfiguration[]::new);
  }

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ManagementActivator.java
henri-tremblay: I was giving an example for my comment about not needing a provider.

henri-tremblay: Ok. I failed. It is a Java 9 feature. So the provider is necessary.

anthonydahanne:
test this please

mathieucarbou:
test this please

mathieucarbou:
test this please

mathieucarbou:
Note: had 2 times a random failure not related to my changes:

```

```

PR #2493, opened 2018-09-24
Upgrade to Gradle 4.10.1

henri-tremblay:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.8.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.1-bin.zip
```
chrisdennis: We usually use the `all` distribution, maybe we should add the following to the top level build.grade:
```
wrapper {
    distributionType = Wrapper.DistributionType.ALL
}
```

PR #2510, opened 2018-11-05
bump platform versions

myronkscott:


myronkscott:
switching to daggy fix for multiple branches

Issue #2523, opened 2018-11-13
StandardEhcacheStatisticsTest.getCacheGetHitMissLatencies


GaryWKeim:
This tests fails daily on Windows.

Failed : org.ehcache.management.providers.statistics.StandardEhcacheStatisticsTest.getCacheGetHitMissLatencies - Stack Trace
--
Cause : java.lang.AssertionError: Expecting: <149714630L> to be greater than or equal to: <150000000L>

henri-tremblay:
Closed by #2524

PR #2524, opened 2018-11-13
https://github.com/ehcache/ehcache3/issues/2523: use System.nanoTime …

GaryWKeim:
…for better timing.

### DIFF
--- a/management/src/test/java/org/ehcache/management/providers/statistics/StandardEhcacheStatisticsTest.java
+++ b/management/src/test/java/org/ehcache/management/providers/statistics/StandardEhcacheStatisticsTest.java
@@ -224,19 +224,18 @@ public class StandardEhcacheStatisticsTest {
   // In fact, on Windows, it does not sleep for long enough.
   // This method keeps sleeping until the full time has passed.
   private void minimumSleep(long millis) {
-    long start = System.currentTimeMillis();
+    long nanos = TimeUnit.MILLISECONDS.toNanos(millis);
+    long start = System.nanoTime();
 
     while (true) {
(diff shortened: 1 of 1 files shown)


management/src/test/java/org/ehcache/management/providers/statistics/StandardEhcacheStatisticsTest.java
```diff
@@ -224,19 +224,18 @@ public void getCacheGetHitMissLatencies() {
...
+      long nanosLeft = nanos - (System.nanoTime() - start);
 
-      if (millisLeft <= 0) {
+      if (nanosLeft <= 0) {
         break;
       }
 
       try {
-        Thread.sleep(millisLeft);
+        Thread.sleep(TimeUnit.NANOSECONDS.toMillis(nanosLeft));
```
henri-tremblay: You can do `TimeUnit.NANOSECONDS.sleep(nanosLeft)`

management/src/test/java/org/ehcache/management/providers/statistics/StandardEhcacheStatisticsTest.java
```diff
@@ -224,19 +224,18 @@ public void getCacheGetHitMissLatencies() {
   // In fact, on Windows, it does not sleep for long enough.
   // This method keeps sleeping until the full time has passed.
   private void minimumSleep(long millis) {
-    long start = System.currentTimeMillis();
+    long nanos = TimeUnit.MILLISECONDS.toNanos(millis);
```
henri-tremblay: Can you please add a comment telling why nanotime is required here?

PR #2527, opened 2018-11-14
Fixes #2526 : Exclude api module from clustered-dist uberjar

### LINKED ISSUE #2526: Clustered distribution uber-jar packages it's own copies of the API classes.
The clustered distribution uber-jar is shading in independent copies of the API classes.  This is/will cause problems for OSGi users.

chrisdennis:


henri-tremblay:
I will merge (because the build failure is unrelated) but it seems to be based on a strange point in the past. Is it a daggy?

chrisdennis:
Yes. It was broken by one of the commits in the write behind work.

PR #2528, opened 2018-11-15
Timeout the test and perform a thread dump

henri-tremblay:


clustered/integration-test/src/test/java/org/ehcache/clustered/writebehind/BasicClusteredWriteBehindTest.java
```diff
@@ -69,6 +80,14 @@ public static void waitForActive() throws Exception {
     CLUSTER.getClusterControl().waitForActive();
   }
 
+  @After
```
akomakom: Is this "after timeout" or just always after the test?

henri-tremblay: :-) Always. But I will remove the code after we found out what's wrong. So I don't mind doing it all the time. We might also learn so things when the test is just failing.

henri-tremblay: I've added a flag to dump only on lock or failure

PR #2534, opened 2018-11-23
Prepare to update to Gradle 5

henri-tremblay:
This PR makes the build pass under Gradle 5 RC5


henri-tremblay:
Superseded by #2537

PR #2540, opened 2018-12-03
Silently destroy caches in case of reconnect during destroy.

mobasherul:


clustered/integration-test/src/test/java/org/ehcache/clustered/ReconnectDuringDestroyTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.ehcache.clustered.reconnect.BasicCacheReconnectTest.RESOURCE_CONFIG;
+import static org.ehcache.clustered.util.TCPProxyUtil.setDelay;
+import static org.terracotta.testing.rules.BasicExternalClusterBuilder.newCluster;
+
+/**
+ * ReconnectDuringDestroyTest
```
albinsuresh: Document what this test is trying to do as this involves a lot of mimicking of internal flow. It could even be comments within the test code.

clustered/integration-test/src/test/java/org/ehcache/clustered/ReconnectDuringDestroyTest.java
```diff
@@ -0,0 +1,121 @@
...
+    cacheManager.init();
+  }
+
+  @Test
+  public void reconnectDuringDestroyTest() throws Exception {
+    Connection client = CLUSTER.newConnection();
+    VoltronReadWriteLock voltronReadWriteLock = new VoltronReadWriteLock(client, "crud-cm");
+    try (VoltronReadWriteLock.Hold localMaintenance = voltronReadWriteLock.tryWriteLock()) {
+      if (localMaintenance == null) {
+        throw new EntityBusyException("Unable to obtain maintenance lease for ");
```
albinsuresh: The error message is incomplete

clustered/integration-test/src/test/java/org/ehcache/clustered/ReconnectDuringDestroyTest.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+      EntityRef<ClusterTierManagerClientEntity, ClusterTierManagerConfiguration, Void> ref = getEntityRef(client);
+      try {
+        ClusterTierManagerClientEntity entity = ref.fetchEntity(null);
+        entity.prepareForDestroy();
+        entity.close();
+      } catch (EntityNotFoundException e) {
+        Assert.fail();
+      }
+      setDelay(6000, proxies);
```
albinsuresh: what's the relevance of this value 6000?

clustered/integration-test/src/test/java/org/ehcache/clustered/ReconnectDuringDestroyTest.java
```diff
@@ -0,0 +1,121 @@
...
+      Thread.sleep(6000);
+      setDelay(0L, proxies);
+    }
+    voltronReadWriteLock = new VoltronReadWriteLock(client, "crud-cm");
+    try (VoltronReadWriteLock.Hold localMaintenance = voltronReadWriteLock.tryWriteLock()) {
+      if (localMaintenance == null) {
+        throw new EntityBusyException("Unable to obtain maintenance lease for ");
```
albinsuresh: Same as above regarding the message.

clustered/integration-test/src/test/java/org/ehcache/clustered/ReconnectDuringDestroyTest.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+      EntityRef<ClusterTierManagerClientEntity, ClusterTierManagerConfiguration, Void> ref = getEntityRef(client);
+      try {
+        ClusterTierManagerClientEntity entity = ref.fetchEntity(null);
+        entity.prepareForDestroy();
```
albinsuresh: prepare is being done before introducing the delay, right? Why do it here again?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/DestroyInProgressException.java
```diff
@@ -26,4 +26,9 @@
   public DestroyInProgressException(String message) {
     super(message);
   }
+
+  @Override
+  public DestroyInProgressException withClientStackTrace() {
+    return new DestroyInProgressException(this.getMessage());
```
albinsuresh: Why was this needed?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/service/DefaultClusteringServiceTest.java
```diff
@@ -2054,6 +2055,23 @@ public void testReleasePersistenceSpaceIdentifierTwice() throws Exception {
     service.releasePersistenceSpaceIdentifier(cacheIdentifier);
   }
 
+  @Test
+  public void releaseMaintenanceHolds() {
```
albinsuresh: The naming could be better, indicating when the release should have happened

clustered/integration-test/src/test/java/org/ehcache/clustered/ReconnectDuringDestroyTest.java
mobasherul: Because in reconnect path it tries to again repeat the same set of calls for cacheManager.destroy().

clustered/common/src/main/java/org/ehcache/clustered/common/internal/exceptions/DestroyInProgressException.java
mobasherul: This I have done for scenarios in which DestroyInProgressException is caught in the client side and take action accordingly like in retrieveEntity(). But currently this exception can never be caught in client side because the Ehcache codec always throws Lifecycle Exception for that.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -171,6 +171,8 @@ private void silentDestroy() {
     } catch (EntityBusyException e) {
       // Ignore - we have a racy client
       LOGGER.debug("ClusterTierManager {} marked busy when trying to clean it up", entityIdentifier);
+    } catch (ConnectionClosedException | ConnectionShutdownException e) {
+      throw e;
```
albinsuresh: Why catch, if you are just gonna rethow it?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -247,9 +249,13 @@ public void destroy(String name) throws CachePersistenceException {
           throw new CachePersistenceException("Could not connect to the cluster tier manager '" + entityIdentifier
                   + "'; retrieve operation timed out", e);
         } catch (DestroyInProgressException e) {
-          silentDestroy();
-          // Nothing left to do
-          break;
+          try {
```
albinsuresh:
```
          try {
            silentDestroy();
            // Nothing left to do
            break;
          } catch (ConnectionClosedException | ConnectionShutdownException e1) {
            reconnect();
          }
```

This block of code is repeating at multiple places. See if you can extract it out to a method and reuse.

PR #2568, opened 2019-02-07
Issue 2567 : Refresh CompactJavaSerializer state on state holder failures

chrisdennis:


impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -130,57 +133,63 @@ public boolean equals(T object, ByteBuffer binary) throws ClassNotFoundException
...
-        writeLookup.put(new SerializableDataKey(discOsc, true), rep);
-        readLookupLocalCache.put(rep, discOsc);
+      ObjectStreamClass disconnected = disconnect(desc);
+      SerializableDataKey key = new SerializableDataKey(disconnected, true);
+      while (true) {
+        Integer rep = writeLookup.get(probe);
+        if (rep != null) {
+          return rep;
+        }
+        rep = nextStreamIndex++;
```
cljohnso: The single-step probing used in this method for updating the local association cache gives me pause.  

I can envision situations (for example, some other client has added a batch of complex cache objects that are composed of Java objects new to the cache) where many failed puts would be required before the local copy of the class associations is synchronized with the server and `nextStreamIndex` accurately reflects the server state.

impl/src/test/java/org/ehcache/impl/serialization/CompactJavaSerializerTest.java
```diff
@@ -0,0 +1,85 @@
...
+      //expected
+    }
+
+    failing.set(false);
+    ByteBuffer serialized = serializerA.serialize(object);
+    assertThat(serializerA.read(serialized), is(object));
+
+    assertThat(stateHolder.entrySet(), hasSize(1));
+
+    CompactJavaSerializer serializerB = new CompactJavaSerializer(getClass().getClassLoader());
```
cljohnso: Lack of generic types is causing a build failure here.

impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
cljohnso: Discussed ... @chrisdennis is opening  a new issue to address this (needs a new server operation).

Issue #2552, opened 2019-01-03
Cache throws ReconnectInProgressException even after reconnect is complete


AbfrmBlr:
`StrongServerProxy` wraps `ConnectionClosedException` which prevent ReconnectingServerStoreProxy from swapping back to the new Store proxy, thus resulting is cache throwing ReconnectInProgressException even after reconnect is complete

PR #2571, opened 2019-02-11
Don't wrap ConnectionCLosed in RuntimeException #2552

AbfrmBlr:


AbfrmBlr:
This needs to be daggy

Issue #2548, opened 2018-12-17
Race in reconnect after failover & locking in loader-writer cache


AbfrmBlr:
Under high write workload, after failover the client gets stuck probably due to the race in collection of reconnect data & unlock response.

PR #2572, opened 2019-02-11
Clear local lock state #2548

AbfrmBlr:


chrisdennis:
I'll confess I don't fully understand what the bug is/was here... any chance of adding an explanation on the issue? Also do we have a solid test that reproduces the bug... the original issue says the failure was sporadic.

PR #2575, opened 2019-02-13
[DO NOT MERGE] Offheap Iterator Fixes

chrisdennis:
This is the 'proper' fix for the offheap iterator issues around buffer deserialization and detachment.  ~~Right now I've somewhat cleaned the actual 'test' code I used to identify all the unlocked paths. As I see it we have a number of options regarding the test coverage here:~~

~~1. Commit the bug fixes with no test coverage (get rid of the second commit)~~
~~2. Commit the bug fixes with the test code commented out.~~
~~3. Commit the bug fixes and test code as they are in this PR.~~
4. Add additional testing seams to the offheap code to allow injection of a subclass of the `LazyOffheapValueHolder` in enough of the tests to provide coverage.

Edit: So I did the *right* thing and added proper testing coverage.

### DIFF
--- a/impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
+++ b/impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
@@ -69,12 +69,14 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static java.lang.Long.max;
+import static java.util.Arrays.asList;
 import static org.ehcache.core.spi.ServiceLocator.findSingletonAmongst;
 import static org.ehcache.core.spi.ServiceLocator.findSingletonAmongst;
 import static org.terracotta.offheapstore.util.MemoryUnit.BYTES;
@@ -169,11 +171,11 @@ public class OffHeapDiskStore<K, V> extends AbstractOffHeapStore<K, V> implement
       MappedPageSource source = new MappedPageSource(dataFile, false, size);
       try {
         PersistentPortability<K> keyPortability = persistent(new SerializerPortability<K>(keySerializer));
-        PersistentPortability<OffHeapValueHolder<V>> elementPortability = persistent(new OffHeapValueHolderPortability<V>(valueSerializer));
+        PersistentPortability<OffHeapValueHolder<V>> valuePortability = persistent(createValuePortability(valueSerializer));
         DiskWriteThreadPool writeWorkers = new DiskWriteThreadPool(executionService, threadPoolAlias, writerConcurrency);
 
         Factory<FileBackedStorageEngine<K, OffHeapValueHolder<V>>> storageEngineFactory = FileBackedStorageEngine.createFactory(source,
-            max((size / 16) / 10, 1024), BYTES, keyPortability, elementPortability, writeWorkers, false);
+            max((size / 16) / 10, 1024), BYTES, keyPortability, valuePortability, writeWorkers, false);
 
         EhcachePersistentSegmentFactory<K, OffHeapValueHolder<V>> factory = new EhcachePersistentSegmentFactory<K, OffHeapValueHolder<V>>(
             source,
@@ -210,11 +212,11 @@ public class OffHeapDiskStore<K, V> extends AbstractOffHeapStore<K, V> implement
       throw new RuntimeException(e);
     }
     PersistentPortability<K> keyPortability = persistent(new SerializerPortability<K>(keySerializer));
-    PersistentPortability<OffHeapValueHolder<V>> elementPortability = persistent(new OffHeapValueHolderPortability<V>(valueSerializer));
+    PersistentPortability<OffHeapValueHolder<V>> valuePortability = persistent(createValuePortability(valueSerializer));
     DiskWriteThreadPool writeWorkers = new DiskWriteThreadPool(executionService, threadPoolAlias, writerConcurrency);
 
     Factory<FileBackedStorageEngine<K, OffHeapValueHolder<V>>> storageEngineFactory = FileBackedStorageEngine.createFactory(source,
-        max((size / 16) / 10, 1024), BYTES, keyPortability, elementPortability, writeWorkers, true);
+        max((size / 16) / 10, 1024), BYTES, keyPortability, valuePortability, writeWorkers, true);
 
     EhcachePersistentSegmentFactory<K, OffHeapValueHolder<V>> factory = new EhcachePersistentSegmentFactory<K, OffHeapValueHolder<V>>(
         source,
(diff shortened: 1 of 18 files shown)


impl/src/main/java/org/ehcache/impl/internal/store/offheap/LazyOffHeapValueHolder.java
```diff
@@ -91,6 +97,9 @@ void writeBack() {
   void forceDeserialization() {
     if (value == null) {
       try {
+        if (!isBinaryValueAvailable()) {
```
lorban: Ouch, ouch, ouch, my eyes hurt and I cannot unsee that.

impl/src/main/java/org/ehcache/impl/internal/store/offheap/LazyOffHeapValueHolder.java
```diff
@@ -124,4 +134,69 @@ private void writeObject(java.io.ObjectOutputStream out) {
     throw new UnsupportedOperationException("This subclass of AbstractValueHolder is NOT serializable");
   }
 
+  private static final boolean IN_TESTING;
```
lorban: Unless there is absolutely not alternative, I'm very much against testing code in non-testing repos. We should try to move all this out, even if that means modifying the offheap lib to be even more test-friendly.

chrisdennis:
Okay - so I've made the necessary adaptions to test this properly.  I still need to rebase this on the correct commit for a daggy bugfix... but I'll wait until I have review approval before doing that.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -488,12 +488,12 @@ public void initAuthoritativeTier(AuthoritativeTier<?, ?> resource) {
...
-    final Class<?> normalKlazz = normal.getClass();
-    Class<?>[] delegateInterfaces = normalKlazz.getInterfaces();
-    Class<?>[] proxyInterfaces = Arrays.copyOf(delegateInterfaces, delegateInterfaces.length + 1);
-    proxyInterfaces[delegateInterfaces.length] = PersistentPortability.class;
-
-    return (PersistentPortability<T>) Proxy.newProxyInstance(normal.getClass().getClassLoader(), proxyInterfaces, (o, method, os) -> {
+    List<Class<?>> proxyInterfaces = new ArrayList<>();
+    proxyInterfaces.add(PersistentPortability.class);
+    for (Class<?> klazz = normal.getClass(); klazz != null; klazz = klazz.getSuperclass()) {
+      proxyInterfaces.addAll(asList(klazz.getInterfaces()));
```
cljohnso: This may be OK in this circumstance, but there's a restriction on what can be passed to `Proxy.newInstance` -- "No two elements in the interfaces array may refer to identical Class objects".  Doing so will result in an `IllegalArgumentException`.  Is there any chance this restriction will be violated in this code as a surprise?

impl/src/test/java/org/ehcache/impl/internal/store/offheap/AssertingLazyOffHeapValueHolder.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.ByteBuffer;
+import java.util.Collection;
+import java.util.function.Predicate;
+
+import static java.util.Arrays.asList;
+import static java.util.Arrays.stream;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+public class AssertingLazyOffHeapValueHolder<V> extends LazyOffHeapValueHolder<V> {
```
cljohnso: This class could use some comments to future maintainers.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
chrisdennis: Changed to a `Set<Class<?>` should be safe now.

impl/src/test/java/org/ehcache/impl/internal/store/offheap/AssertingLazyOffHeapValueHolder.java
chrisdennis: Added a block comment on the static field explaining what is going on.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
cljohnso: There is an order dependency as well so you should really use a `LinkedHashSet` to maintain inheritance order.

chrisdennis: How does this look?

impl/src/test/java/org/ehcache/impl/internal/store/offheap/AssertingLazyOffHeapValueHolder.java
```diff
@@ -0,0 +1,104 @@
...
+   *    and either force deserialization or detach the value under the pre-existing lock scope.
+   * 3. (Least Likely) You introduced a new method to OffHeapStore that needs locking properly.
+   */
+  private static final Collection<Predicate<StackTraceElement>> LOCK_SCOPES = asList(
+    className("org.terracotta.offheapstore.AbstractLockedOffHeapHashMap").methodName("shrink").lineRange(499, 501),
+    className("org.terracotta.offheapstore.AbstractLockedOffHeapHashMap").methodName("computeWithMetadata").lineRange(519, 521),
+    className("org.terracotta.offheapstore.AbstractLockedOffHeapHashMap").methodName("computeIfPresentWithMetadata").lineRange(541, 543),
+    className("org.ehcache.impl.internal.store.offheap.factories.EhcacheSegmentFactory$EhcacheSegment$EntrySet").methodName("iterator").lineRange(125, 135),
+    className("org.ehcache.impl.internal.store.disk.factories.EhcachePersistentSegmentFactory$EhcachePersistentSegment$EntrySet").methodName("iterator").lineRange(129, 138),
+    className("org.terracotta.offheapstore.AbstractLockedOffHeapHashMap$LockedEntryIterator").methodName("next").lineRange(391, 393)
```
lorban: How about reading the bytecode to discover the line ranges? With Javassist:

```
  static class Range {
    int begin = -1;
    int end = -1;
  }

  private static Range findLockRange(String className, String methodName) throws BadBytecode, NotFoundException {
    ClassPool pool = ClassPool.getDefault();
    CtClass ctClass = pool.get(className);
    CtMethod ctMethod = ctClass.getDeclaredMethod(methodName);
    MethodInfo methodInfo = ctMethod.getMethodInfo2();
    CodeAttribute codeAttribute = methodInfo.getCodeAttribute();
    ConstPool constPool = methodInfo.getConstPool();
    CodeIterator codeIterator = codeAttribute.iterator();

    Range range = new Range();
    while (codeIterator.hasNext()) {
      int pos = codeIterator.next();

      if (range.begin == -1) {
        range.begin = lineNumberOfInvokeInterface(codeIterator, pos, constPool, methodInfo,"java.util.concurrent.locks.Lock","lock");
        if (range.begin != -1) {
          range.begin++;
        }
      } else if (range.end == -1) {
        range.end = lineNumberOfInvokeInterface(codeIterator, pos, constPool, methodInfo,"java.util.concurrent.locks.Lock","unlock");
        if (range.end != -1) {
          range.end--;
        }
      }

      if (range.end != -1) {
        return range;
      }
    }
    throw new AssertionError(ctMethod.getName() + " does not contain both calls to Lock.lock and Lock.unlock");
  }

  private static int lineNumberOfInvokeInterface(CodeIterator iter, int pos, ConstPool pool, MethodInfo methodInfo, String intfName, String methodName) {
    int opcode = iter.byteAt(pos);
    if (opcode == INVOKEINTERFACE) {
      int index = iter.u16bitAt(pos + 1);
      String interfaceMethodrefClassName = pool.getInterfaceMethodrefClassName(index);
      if (!interfaceMethodrefClassName.equals(intfName)) {
        return -1;
      }
      String interfaceMethodrefName = pool.getInterfaceMethodrefName(index);
      if (!interfaceMethodrefName.equals(methodName)) {
        return -1;
      }
      return methodInfo.getLineNumber(pos);
    }
    return -1;
  }

```

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
```diff
@@ -488,18 +490,25 @@ public void initAuthoritativeTier(AuthoritativeTier<?, ?> resource) {
...
-
-    return (PersistentPortability<T>) Proxy.newProxyInstance(normal.getClass().getClassLoader(), proxyInterfaces, (o, method, os) -> {
-      if (method.getDeclaringClass().equals(Persistent.class)) {
-        return null;
-      } else {
-        return method.invoke(normal, os);
+    if (normal instanceof PersistentPortability<?>) {
+      return (PersistentPortability<T>) normal;
+    } else {
+      Set<Class<?>> proxyInterfaces = new HashSet<>();
```
cljohnso: I'd still encourage you to use a `LinkedHashSet` here since using a normal `HashSet` is randomizing the interface order.  But I'm not familiar with the objects that would be passed to this method so if you know that classes passed in won't have method collisions, I can go with this.

impl/src/test/java/org/ehcache/impl/internal/store/offheap/AssertingLazyOffHeapValueHolder.java
cljohnso: It would make the test less brittle at the potential cost of an additional test dependency.  Not a bad trade-off.

chrisdennis: Will look in to this.

impl/src/main/java/org/ehcache/impl/internal/store/disk/OffHeapDiskStore.java
lorban: I would go with a `LinkedHashSet` for extra safety; at least bug in there would be consistent across JVM implementations.

impl/src/test/java/org/ehcache/impl/internal/store/offheap/AssertingLazyOffHeapValueHolder.java
cljohnso: The ASM code looks good to me

chrisdennis:
This will need an integration branch... closing.

PR #2580, opened 2019-02-15
Fixes #2345 : Allow Java Language style syntax for arrays, primitives and inner classes

### LINKED ISSUE #2345: Ability to specify arrays and primitive arrays in a non-cryptic way
I think everybody would benefit of having an easier and more intuitive way to specific arrays in key and value types.

Instead of:

```
  <cache alias="cache-1">
    <key-type>java.lang.String</key-type>
    <value-type>[B</value-type>
    [...]
  </cache>
```

I'd like to be able to say:

```
  <cache alias="cache-1">
    <key-type>java.lang.String</key-type>
    <value-type>byte[]</value-type>
   [...]
  </cache>
```

This is not uncommon. Spring also uses that. See:

https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/ClassUtils.java#L225

```
> println ClassUtils.forName('byte[]', Thread.currentThread().getContextClassLoader())
> class [B
```

chrisdennis:
The only wrinkle that I can see with this is that we don't map back to the pretty class names when marshaling a configuration out to XML.

### DIFF
--- a/core/src/test/java/org/ehcache/core/util/ClassLoadingTest.java
+++ b/core/src/test/java/org/ehcache/core/util/ClassLoadingTest.java
@@ -17,6 +17,7 @@
 package org.ehcache.core.util;
 
 import static java.util.Collections.list;
+import static org.ehcache.core.util.ClassLoading.getDefaultClassLoader;
 import static org.hamcrest.collection.IsIterableContainingInOrder.contains;
 import static org.hamcrest.core.Is.is;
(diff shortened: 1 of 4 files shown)


chrisdennis:
Thanks @zemian. Ehcache 3 policy is that the author cannot (merge - one of the reviewers must).  Could you merge for me?  Thanks!

Issue #2170, opened 2017-10-23
Support Path on top of File for disk configuration


henri-tremblay:
Java 7 has the new `Path` class that can be used. We could support it on top of `File`.

chrisdennis:
An initial cut of what this should look like (imho) is in: chrisdennis/ehcache#2170 I think it only makes sense to adopt Path if we're going to use Path internally.  That way we can work against non-default providers properly. The referenced branch does everything for Ehcache 3... but it's incomplete without the equivalent changes in Terracotta-OSS/offheap-store.

PR #2581, opened 2019-02-19
Issue #2170 : Use path for all internal file IO

chrisdennis:


AbfrmBlr:
This changes api, that would mean more process. Let's talk about this

AbfrmBlr:
Do we want to revisit this PR? Has been hanging there for long

chrisdennis:
Right now due to deficiencies in the `java.nio.files` API any migration to the new API would be incomplete and kind of pointless.  Closing this PR, but I'll keep the branch just in case someone wants to revisit this in the future.

PR #2589, opened 2019-02-25
[master] : Don't wrap ServerStoreProxyException in RuntimeException #2552

AbfrmBlr:


chrisdennis:
I guess what you really mean is "Don't wrap ServerStoreProxyException in RuntimeException"?  Could that be expanded to "Don't rewrap thrown RuntimeExceptions"?

PR #2591, opened 2019-02-25
[rel/3.6] : Don't wrap ServerStoreProxyException in RuntimeException #2552

AbfrmBlr:


albinsuresh:
@chrisdennis @akomakom @anthonydahanne Are there any additional restrictions put in place for merges to these release branches? This is a daggy PR and the merge of this same branch to master went just fine. But I'm not able to merge this one. Here it's complaining that this branch is out-of-date, which is fine because it's a daggy commit. That was the same case for master as well but we did not get this warning there. So could you relax the restrictions on merges to these branches?

anthonydahanne:
I've updated the restrictions in release/* branches, to match the ones on master.
should be good now

PR #2616, opened 2019-03-06
Run baseline task on check for 'api' only. Hack up versions to appease OSGi/BND

chrisdennis:


build.gradle
```diff
@@ -240,9 +240,15 @@ subprojects {
         'Bundle-License': 'LICENSE',
         'Bundle-Vendor': 'Terracotta Inc., a wholly-owned subsidiary of Software AG USA, Inc.',
         'Bundle-RequiredExecutionEnvironment': 'JavaSE-1.8',
+        'Bundle-Version': project.version.asType(String).replace("-SNAPSHOT", ".999-SNAPSHOT")
```
lorban: This odd `.999-SNAPSHOT` hack is probably worth a line of comment.

PR #2618, opened 2019-03-07
Put classes using unsafe in a dedicated sourceSet to allow using -Werror on impl

chrisdennis:


henri-tremblay:
Awesome news if this goes through. Build is failing thought. Looks like a flaky test.

PR #2623, opened 2019-03-27
clustered events design asciidoc

lorban:


docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+* replace(3)
+
+The `Store` methods not listed above or below either throw `UnsupportedOperationException`, or delegate to other
+`Store` methods.
+
+== The complicated bits
+
+The following cases are going to be more complicated to implement:
+
+* All clients must be aware of the event firing logic, apply its rules and fire events even when some clients are not
```
chrisdennis: Wouldn't this whole thing be simpler if we layered the event listener feature in the same way that the current clustered caching is layered. That would mean building a "chain event listener" feature that the server implements, and then have the clients build their feature on top.  In this model events are only sourced from the server... there is a slight complication around "expiry" but that complication already exists. The only thing this would cause us to do is to clarify when an expiration event occured, with the constraints here forcing that to be "when a client successfully resolves a value to expired on the server". This may require associating some additional metadata with the resolve operation but I don't think that will prove insurmountable.

docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+registered event listeners.
+* `Store.put()` needs to discriminate between *creation* and *update*. This means `resolve()` must be called, which also
+means that the current `append()` call must be replaced with the slower `getAndAppend()` one. This is going to put more
```
chrisdennis: If we move to a server initiated model then we only need call `append()`.  The initiating client is not special, it's listener will get fired when the server issues notifications that the chain was changed.

docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+=== Recommandations
+
+It must be made clear (documentation?) that the eventing mechanism is going to have a performance impact, sometimes
+a very noticeable one.
+
+Some features are undesirable because they are unlikely to be practical:
+
+* Synchronous events would require waiting for a round-trip to all clients before achieving a cache operation. This
+would pretty much make such cache unusably slow. Such config should throw *UnsupportedOperationException* when attempted.
```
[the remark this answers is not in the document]
lorban: I'm not very worried about the engineering time required to implement that feature; but very much by the time to support users that are unavoidably going to shoot themselves in the foot expecting this to have reasonable performance.

If there's no proven reasonable use-case for this feature, I'd rather explicitly leave it out.

docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+* Most events have to find their source from a client, more specifically from the clustered `Store` implementation:
+*expiry*, *removal*, *update* and *creation* are all client-triggered.
+* One event has to originate from the server: *eviction* is server-triggered.
+
+This means a cluster-wide listener mechanism has to be created with the following features:
+
+* Clients can register and unregister themselves.
+* Events can be fired from any client or any server.
+* Events have to be transported either from a server to all listening clients, or from one client to all clients
+(sometimes including the source client) via a server. AA topologies complicate client-sourced events transport as events
```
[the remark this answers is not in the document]
lorban: Yes, and making sure events don't get duplicated.

docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+An event delivery mechanism must be built to transport the events from the clients and servers to all clients registered
+as listeners. It requires the following:
+
+* API to (un)register a client as a listener
+* API for a client to fire an event
+* API for a server to fire an event
+* Implementation support of multiple servers
+
+A dedicated entity looks like the cleanest way ahead. This entity might need to be general-purpose enough to make
```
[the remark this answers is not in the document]
lorban: That is IMO debatable. I already went ahead the separate entity route and found both advantages and drawbacks in this approach.

docs/src/docs/asciidoc/developer/clustered-events.adoc
lorban: If I get you idea right, the server currently lacks data that would allow the clients to figure out what mapping was modified.

docs/src/docs/asciidoc/developer/clustered-events.adoc
lorban: This could work if the server wasn't a n-keys to operations store. That's going to cause some serious problems too if events are enabled while the clustered cache is being bulk-loaded, although this could be mitigated with extra work.

docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+* Guaranteeing event delivery in all cases would require some form of stable store and a fairly complex and costly
+2-phase logic. This would also have an unsustainable performance impact. Instead, clients should have a way to figure out
+when such hazard happen to compensate for the possible lack of event delivery.
+
+== Technical facts
+
+Because of the current implementation:
+
+* Most events have to find their source from a client, more specifically from the clustered `Store` implementation:
+*expiry*, *removal*, *update* and *creation* are all client-triggered.
```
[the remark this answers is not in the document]
lorban: I'm not sure where you're trying to go here.

docs/src/docs/asciidoc/developer/clustered-events.adoc
```diff
@@ -0,0 +1,90 @@
...
+* Events can be fired from non-listening clients.
+
+== The transport mechanism
+
+An event delivery mechanism must be built to transport the events from the clients and servers to all clients registered
+as listeners. It requires the following:
+
+* API to (un)register a client as a listener
+* API for a client to fire an event
+* API for a server to fire an event
```
anthonydahanne: hum... why does the server need to send events ? I thought only clients would be firing / receving events ?

PR #2628, opened 2019-04-04
Fixes #2627 : Better XML error messages when service implementations are absent

### LINKED ISSUE #2627: XML configuration parser could provide better error messages when XML schemas are missing
When client side extension libraries are missing from the classpath the XML configuration parsing code provides fairly unhelpful error messages:
```
Invalid content was found starting with element 'terracotta:cluster'. One of '{"http://www.ehcache.org/v3":service-creation-configuration}' is expected
```
We should be able to at least hint that when these kind of "abstract typing" errors occur that it might be due to missing libraries.

chrisdennis:


anthonydahanne:
2 failing tests ?

PR #2631, opened 2019-04-08
Upgrade to Asciidoctor Gradle Plugin 2.3.0

chrisdennis:
This is a fairly invasive set of changes to adopt the new 2.x version of the asciidoctor gradle plugin.  This needs to be reviewed for its impact on the website generation before it can be merged as there may need to be compensating changes there. In addition it would be good for a cross-section of developers to confirm their IDE integration is still working okay.


chrisdennis:
@lorban @anthonydahanne I've modified this so that the build will fail if there are any warnings when building the documentation. I'm thinking/hoping that with some work in the website generation we could have the same thing happen there.

PR #2656, opened 2019-06-03
update doc

lorban:
Remove note that states clustered caches don't support event listeners.


jhouserizer:
Looks good, but just wondering, should we leave a note that it doesn't work before version X.Y ?

PR #2660, opened 2019-06-05
Add support for running OSGi tests in Java 11

chrisdennis:


osgi-test/src/test/java/org/ehcache/osgi/OsgiTestUtils.java
```diff
@@ -44,17 +46,37 @@ public static Option baseConfiguration(String ... path) {
...
-        "javax.xml.bind.annotation.adapters;version=2.3.0"
-      ),
       cleanCaches(true),
       workingDirectory(join(File.separator, "build", "osgi-container", join(File.separator, path))),
       junitBundles()
     );
   }
 
+  public static Option jaxbConfiguration() {
+    if (parseInt(getProperty("java.version").split("[^\\d]+")[0]) >= 9) {
```
cljohnso: Any reason you didn't use `\D` instead of `^\d`?

chrisdennis: Because I find the whole concept of `\D` being the inverse of `\d` abhorrent.

PR #2672, opened 2019-07-15
Running tests with java11

mobasherul:


build.gradle
```diff
@@ -88,6 +88,9 @@ if (hasProperty('testVM')) {
 }
 
 subprojects {
+  tasks.withType(Test) {
```
chrisdennis: This block already exists at line 218 in my copy.

clustered/integration-test/build.gradle
```diff
@@ -64,14 +64,9 @@ task copyServerLibs(type: Copy) {
...
   dependsOn copyServerLibs
   environment 'JAVA_HOME', testJava.javaHome
-  if (testJava.javaVersion.isJava9Compatible()) {
-    environment 'JAVA_OPTS', '--add-modules java.xml.bind'
-  }
   //If this directory does not exist, tests will fail with a cryptic assert failure
   systemProperty 'kitInstallationPath', "$unzipKit.destinationDir/${project(':clustered:clustered-dist').archivesBaseName}-$project.version-kit"
   // Uncomment to include client logging in console output
   // testLogging.showStandardStreams = true
-}.doFirst {
```
chrisdennis: I think there should also be changes to the clustered OSGi tests to enable them in Java 11 too.

mobasherul: There is a problem there -
if (testJava.javaVersion.isJava9Compatible()) {
    //Deploying the jaxb-runtime as an OSGi module is currently impossible - don't ask... sigh.
    jvmArgs += ['--add-modules', 'java.xml.bind']
    //Servers need JAXB too.
    jvmArgs += ['-Dtc-server-opts=--add-modules java.xml.bind']
  }
and add modules can't work in java11.

chrisdennis: Have a look at the changes I made to the unclustered OSGi tests in 0c9e97f1df1b9e7265539556fb24094543e690fc.

clustered/osgi-test/src/test/java/org/ehcache/osgi/ClusteredOsgiTest.java
```diff
@@ -81,8 +82,8 @@
       gradleBundle("org.ehcache.modules:api"),
       gradleBundle("org.ehcache.modules:core"),
       gradleBundle("org.ehcache.modules:impl"),
-      gradleBundle("org.ehcache.modules:xml"),
-      gradleBundle("org.ehcache:clustered-dist"),
```
chrisdennis: I don't think `clustered-dist` should need `jaxbConfiguration()` since it doesn't depend on JAXB (plus specifying it twice is redundant).

clustered/osgi-test/src/test/java/org/ehcache/osgi/ClusteredOsgiTest.java
```diff
@@ -98,8 +99,8 @@
   @Configuration
   public Option[] uberJar() {
     return options(
-      gradleBundle("org.ehcache:dist"),
-      gradleBundle("org.ehcache:clustered-dist"),
```
chrisdennis: Again... you shouldn't need the duplicate `jaxbConfiguration()`.

PR #2678, opened 2019-07-22
Dependency check upgrade

chrisdennis:


config/owasp-supressions.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+  <suppress>
+    <notes><![CDATA[file name: Ehcache modules are not Gradle!]]></notes>
+    <packageUrl regex="true">^pkg:maven/org\.ehcache.*@.*$</packageUrl>
+    <cve>CVE-2019-11065</cve>
```
AbfrmBlr: I understand that this loophole existed earlier in gradle , curious if there is any other reason to add this.

build.gradle
```diff
@@ -206,7 +206,8 @@ subprojects {
 
   dependencyCheck {
     failBuildOnCVSS = 0
-    skipConfigurations += ['checkstyle']
+    skipConfigurations += ['checkstyle', 'spotbugs']
```
AbfrmBlr: Sry for limited knowledge but why skip spotbugs

config/owasp-supressions.xml
lorban: I don't get why you added this suppression file at all, it's certainly worth a comment. Also, doesn't it miss a license header?

build.gradle
lorban: I don't mind this change, but it enlarges a bit the scope of this PR.

chrisdennis: Upgrading the plugin somehow triggered checking the `spotbugs` configuration for vulnerabilities when it never has before.  These two configurations are here because we don't care about vulnerabilities in the dependencies in spotbugs/checkstyle or their dependencies since they don't end up in any of our stuff.

config/owasp-supressions.xml
chrisdennis: I'll add a comment to explain the exclusion.

PR #2680, opened 2019-07-23
Fixes #2679 : Delegate the iterator() call to the underlying clustered store

### LINKED ISSUE #2679: Iterator is not wired correctly for clustered caches with loaders/writers
The `DelegatingLoaderWriterStore` incorrectly throws rather than delegating the `iterator()` call.

chrisdennis:


AbfrmBlr:
It's a daggy right?

chrisdennis:
Yes... it's daggy.

PR #2683, opened 2019-07-25
[DO NOT MERGE] proposal: core XML extension to instantiate simple services

lorban:
Currently, the XML config proposes no way to configure custom services without resorting to the whole write-an-extension-xsd-with-parser-provider-factories shebang.

This quick prototype demonstrates that trivial services could be configured with a simple embedded XML config extension parser that uses simple reflection to instantiate the service.

This would allow for instance configuring the documented `TickingTimeSource` (https://www.ehcache.org/documentation/3.8/performance.html#time-source) via XML with a simple block, solving the problem mentioned here https://stackoverflow.com/questions/55866367/how-to-switch-to-tickingtimesource-for-ehcache-3:

```
<service>
  <ss:instance class="org.ehcache.core.spi.time.TickingTimeSource">
    <ss:arg value="10"/>
    <ss:arg value="10000"/>
  </ss:instance>
</service>
```

Opinions & reviews welcome.

### DIFF
--- a/xml/src/main/java/org/ehcache/xml/ss/ReflectionHelper.java
+++ b/xml/src/main/java/org/ehcache/xml/ss/ReflectionHelper.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ehcache.xml.ss;
+
+import org.ehcache.spi.service.Service;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+class ReflectionHelper {
+
+  static Service instantiate(Class<? extends Service> clazz, List<String> unparsedArgs) throws Exception {
+    List<Constructor<?>> potentialCtors = new ArrayList<>();
+
+    for (Constructor<?> constructor : clazz.getConstructors()) {
+      if (constructor.getParameterTypes().length == unparsedArgs.size()) {
+        potentialCtors.add(constructor);
+      }
+    }
+
+    Constructor<?> targetCtor = null;
+    List<Object> targetArgs = new ArrayList<>();
+
+    for (Constructor<?> potentialCtor : potentialCtors) {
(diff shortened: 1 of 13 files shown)


xml/src/test/resources/configs/simple-service.xml
```diff
@@ -0,0 +1,45 @@
...
+<config
+  xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
+  xmlns='http://www.ehcache.org/v3'
+  xmlns:ss="http://www.ehcache.org/v3/ss"
+  xsi:schemaLocation="http://www.ehcache.org/v3 ../../../../../xml/src/main/resources/ehcache-core.xsd
+                        http://www.ehcache.org/v3/ss ../../../../../xml/src/main/resources/ehcache-simple-service-ext.xsd">
+
+  <service>
+    <ss:instance class="org.ehcache.core.spi.time.TickingTimeSource">
+      <ss:arg value="10"/>
```
chrisdennis: This is nice... but it would be better if we could name the parameters in the XML and have them matched against the names of the constructor parameters. Then in cases where the names aren't available via reflection and we have to resort to ordering (and the types are indistinct) we could print a big warning?

xml/src/main/java/org/ehcache/xml/ss/SimpleServiceProvider.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.ehcache.spi.service.Service;
+import org.ehcache.spi.service.ServiceProvider;
+
+import java.util.Collection;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class SimpleServiceProvider implements Service {
+  private final SimpleServiceConfiguration configuration;
+  private volatile Collection<Service> startedServices;
```
chrisdennis: Can the configurations services really change on us?

xml/src/main/java/org/ehcache/xml/ss/ReflectionHelper.java
```diff
@@ -0,0 +1,104 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+class ReflectionHelper {
+
+  static Service instantiate(Class<? extends Service> clazz, List<String> unparsedArgs) throws Exception {
```
chrisdennis: We're already using a repackaged version of `ConstructorUtils` in the class instance provider code for precisely this purpose. You may be able to reuse that logic.

xml/src/test/resources/configs/simple-service.xml
lorban: That sounds like a lot of fuss for little value IMHO.

xml/src/main/java/org/ehcache/xml/ss/SimpleServiceProvider.java
lorban: No, they can't. I should get rid of the `startedServices` variable.

xml/src/main/java/org/ehcache/xml/ss/ReflectionHelper.java
lorban: I would need to modify `ConstructorUtils` to add the parameter type conversion logic. I'll have a look to figure out the best way to merge both these util classes.

PR #2689, opened 2019-08-14
CODE_OF_CONDUCT.md

mathanlalsait:
Added initial code of conduct

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


chrisdennis:
Generally speaking we prefer pull requests to have associated issues which they are addressing before being created. This makes it much easier to discuss the intended changes before time is invested in creating those changes.

In this case there are a number of issues that should be discussed before adopting a formal code of conduct surrounding the codes content and enforcement policies. I'm also almost certainly going to have to engage with Software AG's HR and legal departments since they are the employer for most of the contributors and also own the copyright in all the code. I've taken the liberty of filing an issue for starting this process here: #2690

Issue #2694, opened 2019-08-30
Ehcache reconnect fails (or Ehcache should fail when cluster is gone)


mathieucarbou:
This issue has been reported in Ehcahce forum at https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!category-topic/ehcache-users/1PVADdK84cs

I am able to reproduce his issue in an IT test in the project:

GIST: https://gist.github.com/9c912dcb6101176eaf4076836efa01dc

I have the same error: `Reconnection rejected by L2, no more trying to recon...`

Ehcache never reconnects / recreate the clustered stack.

I would Except Ehcache to:

1) Either fail miserably when all the stripe servers are down after a certain time, and even if servers are back online because it is a new lcuster
2) Either Ehcache should recycle the connection and recreate the whole clustered layer when servers are back online (because it is a new cluster)

Other issue I saw from my IT test:

Ehcache `.close()` is failing with an exception where it should not be:

```
Connection lost to server. Client is trying to reconnect to server
org.ehcache.StateTransitionException: Connection lost to server. Client is trying to reconnect to server
````

Well... As a user i find this exception useless because if the CM is trying to reconnect, then it is NOT connected. So why do I get a failure when doing a proper close of the CM...

So 2 issues here linked to the fact that all the servers of a stripe are gone, and eventually come back again.

Attached the traces and stacks.

- [client.log](https://github.com/ehcache/ehcache3/files/3560788/client.log)
- [jstack.1.log](https://github.com/ehcache/ehcache3/files/3560791/jstack.1.log)
- [jstack.2.log](https://github.com/ehcache/ehcache3/files/3560792/jstack.2.log)

mathieucarbou:
CC @AbfrmBlr @anthonydahanne @myronkscott @cljohnso

PR #2700, opened 2019-09-06
Issue #2694 [master]

chrisdennis:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/PerpetualCachePersistenceException.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Thrown to indicate a perpetual (non-transient) failure in a persistent cache manager.
+ * <p>
+ * Receiving this exception indicates that future interactions with the throwing entity will continue to fail without
+ * corrective action.
+ *
+ * @see CachePersistenceException
+ */
+public class PerpetualCachePersistenceException extends CachePersistenceException {
```
mathieucarbou: I would prefer an exception name that better reflect the fact that the persistent layer cannot be used anymore instead of the fact that this exception will be throwing again and again. Something like: `UnusableCachePersistenceException` or `InvalidCachePersistenceException` or `IllegalStateCachePersistenceException`...

chrisdennis: At the point of throwing this exception that's not really true. The persistence layer could be used again if someone created the relevant entities that are missing. This exception signals to the outer layers of the clustering code that we're hitting a non-transient issue with the persistence code. It chooses to translate that it to a 'graveyard' state where nothing can be made to work again.  I think your alternative exception names might fit better for the exceptions thrown from the `FailedReconnectServerStoreProxy` that gets installed.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/FailedReconnectStoreProxy.java
```diff
@@ -0,0 +1,89 @@
...
+  private final String cacheId;
+
+  public FailedReconnectStoreProxy(String cacheId, Throwable failure) {
+    this.cacheId = cacheId;
+    this.failure = failure;
+  }
+
+  @Override
+  public ChainEntry get(long key) {
+    throw new RuntimeException("Cache " + getCacheId() + " failed reconnecting to cluster", failure);
```
mathieucarbou: shouldn't they be `IllegalState` instead ? Which also would match closer the underlying exception in case we rename it to something else than Perpetual...

chrisdennis: I think you're right that 'IllegalStateException' or maybe something new and Ehcache specific make sense here. I disagree on renaming the perpetual exception though. ;)

clustered/client/src/main/java/org/ehcache/clustered/client/internal/PerpetualCachePersistenceException.java
mathieucarbou: cool!
So that means that we could have a config in Ehcache like "autoRecover(true|false)" that would allow Ehcache to automatically recreate the persistence layer then ?

chrisdennis: That would likely not be a distinct setting but more likely a facet of the behavior of the existing 'autoCreate' option. If the cache isn't auto create already it's either not the right thing  to do to create on reconnect, or we don't even have the necessary information to do the create.

PR #2708, opened 2019-09-20
Issue #2694 Support auto-creation of entities on reconnect

chrisdennis:


clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/AutoCreateOnReconnectTest.java
```diff
@@ -0,0 +1,80 @@
...
+        newCacheConfigurationBuilder(Long.class, String.class, newResourcePoolsBuilder()
+          .with(clusteredDedicated("primary-server-resource", 1, MemoryUnit.MB)))
+          .build());
+
+      cache.put(1L, "one");
+
+      CLUSTER.getClusterControl().terminateAllServers();
+      CLUSTER.getClusterControl().startAllServers();
+
+      long end = System.nanoTime() + TimeUnit.SECONDS.toNanos(20);
```
anthonydahanne: unused variable

chrisdennis: Removed.

clustered/client/src/main/java/org/ehcache/clustered/client/config/builders/ClusteringServiceConfigurationBuilder.java
```diff
@@ -152,7 +153,20 @@ public ServerSideConfigurationBuilder expecting() {
...
-    return new ClusteringServiceConfigurationBuilder(this.connectionSource, this.timeouts, true,
+    return new ClusteringServiceConfigurationBuilder(this.connectionSource, this.timeouts, ClientMode.AUTO_CREATE,
+      serverSideConfig.apply(new ServerSideConfigurationBuilder()), properties);
+  }
+
+  /**
+   * Support connection to an existing entity or create if the entity if absent on initial connection or any subsequent reconnect attempt.
+   * <p>
+   * An empty server-side configuration can be created by performing no operations on the supplied builder:
+   * {@code builder.autoCreate(b -> b)}
```
chrisdennis: Fix to use this method.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/config/xml/ClusteringCacheManagerServiceConfigurationParser.java
```diff
@@ -380,9 +389,21 @@ private Element createDefaultServerResourceElement(Document doc, String defaultS
...
-  private ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig processServerSideConfig(Node serverSideConfigElement) {
+  private ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig processServerSideConfig(Element serverSideConfigElement) {
     ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig serverSideConfig = new ClusteringCacheManagerServiceConfigurationParser.ServerSideConfig();
-    serverSideConfig.autoCreate = Boolean.parseBoolean(((Element)serverSideConfigElement).getAttribute("auto-create"));
+
+    String autoCreateAttr = serverSideConfigElement.getAttribute(AUTO_CREATE_ATTRIBUTE_NAME);
+    String clientModeAttr = serverSideConfigElement.getAttribute(CLIENT_MODE_ATTRIBUTE_NAME);
+    if (clientModeAttr.isEmpty()) {
+      if (!autoCreateAttr.isEmpty()) {
+        serverSideConfig.clientMode = Boolean.parseBoolean(autoCreateAttr) ? ClientMode.AUTO_CREATE : ClientMode.CONNECT;
```
chrisdennis: Should be `EXPECTING` - add a test too.

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -215,7 +215,7 @@ public void unknownClusteredCacheExample()
 
     CacheManagerBuilder<PersistentCacheManager> cacheManagerBuilderAutoCreate = CacheManagerBuilder.newCacheManagerBuilder()
             .with(ClusteringServiceConfigurationBuilder.cluster(URI.create("terracotta://localhost/my-application"))
-              .autoCreate(server -> server  // <1>
```
chrisdennis: Documentation changes likely needed to accompany this.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -44,9 +45,16 @@
     CacheManagerConfiguration<PersistentCacheManager>,
     HumanReadable {
 
-  public static final boolean DEFAULT_AUTOCREATE = false;
+  public enum ClientMode {
+    CONNECT, EXPECTING, AUTO_CREATE, AUTO_CREATE_ON_RECONNECT
```
cljohnso: This appears to be public API -- needs javadoc.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -246,11 +280,20 @@ public ClusteringServiceConfiguration(Iterable<InetSocketAddress> servers, Strin
...
    */
+  @Deprecated
   public ClusteringServiceConfiguration(ConnectionSource connectionSource, Timeouts timeouts, boolean autoCreate,
                                         ServerSideConfiguration serverSideConfiguration, Properties properties) {
+    this(connectionSource, timeouts,
+      autoCreate ? ClientMode.AUTO_CREATE : (serverSideConfiguration == null ? ClientMode.CONNECT : ClientMode.EXPECTING),
+      serverSideConfiguration, properties);
+  }
+
+  public ClusteringServiceConfiguration(ConnectionSource connectionSource, Timeouts timeouts, ClientMode clientMode,
```
cljohnso: Public API -- needs javadoc.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
```diff
@@ -288,9 +331,21 @@ public ConnectionSource getConnectionSource() {
    * Returns {@code true} is server side components should be automatically created.
    *
    * @return {@code true} is auto-create is enabled
+   * @deprecated Deprecated in favor of {@link #getClientMode()}
    */
+  @Deprecated
```
mathieucarbou: why we have to deprecate ? Isn't the method stil doing the exact same thing the javadoc says ?

chrisdennis: It's return type can't discriminate between the full set of behaviors. Anyone using this should probably be migrating to `getClientMode()` since they may care about the distinction between `AUTO_CREATE` and `AUTO_CREATE_ON_RECONNECT`.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -157,13 +158,23 @@
         </xs:annotation>
       </xs:element>
     </xs:sequence>
-    <xs:attribute name="auto-create" type="xs:boolean" use="optional" default="false">
```
mathieucarbou: why don't we keep this to let the user know what's used by default ?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -157,13 +158,23 @@
...
+    <xs:attribute name="auto-create" type="xs:boolean" use="optional">
       <xs:annotation>
-        <xs:documentation> xml:lang="en">
+        <xs:documentation xml:lang="en">
           True if server side components should be automatically created if they are absent.
         </xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="client-mode" type="tc:client-mode-spec" use="optional">
```
mathieucarbou: why don't we set a default to let the user know what's used by default ?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -157,13 +158,23 @@
...
+      <xs:annotation>
+        <xs:documentation> xml:lang="en">
+          The client connection behavior, either:
+           * expecting - expect a matching server configuration (fail otherwise)
+           * auto-create - expect a matching server configuration or create one if none is present
+           * auto-create-on-reconnect - as 'auto-create' but also auto-create on reconnection
+        </xs:documentation>
```
mathieucarbou: Does it means that this won't be possible to have the client auto-create on connect first time, but not auto-create on network failure for example ? I didn't see how we could have the behaviour where we fail if the cluster was gone ?

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
mathieucarbou: but this method meaning is not to get the configuration state but much more to know whether we have to auto-create or not ? It could have been called: `boolean shouldAutoCreate()` right ?

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -19,6 +19,7 @@
            xmlns:eh="http://www.ehcache.org/v3"
            xmlns:tc="http://www.ehcache.org/v3/clustered"
            xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning"
```
cljohnso: This namespace declaration is not used.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
```diff
@@ -157,13 +158,23 @@
         </xs:annotation>
       </xs:element>
     </xs:sequence>
-    <xs:attribute name="auto-create" type="xs:boolean" use="optional" default="false">
+    <xs:attribute name="auto-create" type="xs:boolean" use="optional">
       <xs:annotation>
-        <xs:documentation> xml:lang="en">
+        <xs:documentation xml:lang="en">
           True if server side components should be automatically created if they are absent.
```
cljohnso: It may be worth noting here that using both the `auto-create` and `client-mode` attributes is an error.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
chrisdennis: Sadly the necessary information associated with the DOM that distinguishes between the 'default' value and a value equal to it is elided by the JAXB parser. This means if these attributes have default values in the XSD we can't tell if the user actually defined a value at all.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
chrisdennis: See my other reply.

clustered/client/src/main/resources/ehcache-clustered-ext.xsd
chrisdennis: You can have the behavior you want. That's just 'auto-create' as it stands.

clustered/client/src/main/java/org/ehcache/clustered/client/config/ClusteringServiceConfiguration.java
chrisdennis: This method still isn't good enough in that situation since whether we should 'auto-create' in a given situation is now a function of two things as follows:

mode                                    | connect | reconnect |
---------------------------|---------|------------|
auto-create                          |    true    |      false       |
auto-create-on-reconnect |    true    |       true       |

A single boolean return cannot capture this behavior. As an extension I could clarify the javadoc here to clarify that a true return here means only: "this configuration will auto-create on connect (but might also auto-create at other times).

PR #2725, opened 2019-11-22
Do property substitution using JAXB

chrisdennis:
Lets make a go/no-go decision on this idea.


xml/src/main/resources/ehcache-core.xsd
```diff
@@ -615,9 +615,52 @@
...
+          <xs:whiteSpace value="collapse"></xs:whiteSpace>
+        </xs:restriction>
+      </xs:simpleType>
+    </xs:union>
+  </xs:simpleType>
+
+  <xs:simpleType name="propertyOrPositiveInteger">
+    <xs:annotation>
+      <xs:appinfo>
+        <jaxb:javaType name="java.math.BigInteger" parseMethod="org.ehcache.xml.JaxbParsers.parseSubInteger"/>
```
cljohnso, line 637: Should probably have type-specific validation for `propertyReference` values.

xml/src/main/resources/ehcache-core.xsd
```diff
@@ -615,9 +615,52 @@
...
+      </xs:appinfo>
+    </xs:annotation>
+    <xs:union memberTypes="xs:positiveInteger ehcache:propertyReference"/>
+  </xs:simpleType>
+
+  <xs:simpleType name="propertyOrNonNegativeInteger">
+    <xs:annotation>
+      <xs:appinfo>
+        <jaxb:javaType name="java.math.BigInteger" parseMethod="org.ehcache.xml.JaxbParsers.parseSubInteger"/>
```
cljohnso, line 646: Should probably have type-specific validation for `propertyReference` values.

cljohnso:
We should probably beef up the documentation to add cautions about safely using properties for certain values that should remain stable for the life of a server.

For the numeric types, I think the schema currently protects against composition of values (permitting either a numeric value or a single property) but internal parsing supports composed values.  I don't think its necessary to change the parsing for this (as long as the schema prevents it) but it may be wise to add a test or two to ensure this isn't changed without proper consideration.

cljohnso (changes requested):
The only necessary change is the validation for the specific numeric types when the values are supplied by property substitution.

chrisdennis:
I still need to make documentation changes... but I think the coding/testing changes are complete.

cljohnso (approved):

PR #2727, opened 2019-12-02
bump upstream dependencies

myronkscott:


clustered/server/build.gradle
```diff
@@ -35,4 +35,5 @@ dependencies {
   providedImplementation "org.terracotta:runnel:$terracottaPlatformVersion"
 
   testImplementation project(':clustered:test-utils')
+  testImplementation ("org.terracotta:passthrough-server:$terracottaPassthroughTestingVersion")
```
chrisdennis: Why does this need to be added?

clustered/clustered-dist/build.gradle
```diff
@@ -79,7 +79,7 @@ jar {
     'Bundle-SymbolicName': 'org.ehcache.clustered',
     'Export-Package': '!com.tc.*, !com.terracotta.*, !org.terracotta.*, !org.ehcache.*.internal.*, !sun.misc, ' +
       'org.ehcache.clustered.client.*, org.ehcache.clustered.common.*',
-    'Import-Package': '!sun.misc.*, org.ehcache.xml.*;resolution:=optional, *'
+    'Import-Package': '!sun.misc.*, org.ehcache.xml.*;resolution:=optional, jdk.jfr.*;resolution:=optional, *'
```
chrisdennis: This makes sense. Have you confirmed however that this stuff does work correctly if the bundle is available?

build.gradle
```diff
@@ -215,6 +218,7 @@ subprojects {
     }
   }
 
+
```
chrisdennis: Random whitespace change.

build.gradle
```diff
@@ -149,6 +149,9 @@ subprojects {
     if (parent.isCloudbees) {
       systemProperty 'disable.concurrent.tests', 'true'
     }
+    jacoco {
+        excludes = ['org.terracotta.tripwire.*']
```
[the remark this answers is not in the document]
myronkscott: The JFR custom event classes are already instrumented with some magic from that package.  Jacoco doesn't like that and this error is generated during tests.  Caused by: java.lang.IllegalStateException: Cannot process instrumented class XXX. Please supply original non-instrumented classes

clustered/clustered-dist/build.gradle
myronkscott: I've confirmed everything works if everything is present and a tripwire appender is added to logback.

clustered/server/build.gradle
myronkscott: Offheap is now instrumented with tripwire which is provided by the server.  There are some unit tests that instantiate OffheapProvider without a server so passthrough is added to provide the required libs.  We could just add tripwire but that is one more dependency to manage.

chrisdennis: Okay.

build.gradle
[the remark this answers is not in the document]
chrisdennis: Okay, that makes sense. Could we switch to `excludes += 'org.terracotta.tripwire.*'`?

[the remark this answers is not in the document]
myronkscott: ok

PR #2755, opened 2020-04-02
Adopt files utility and turn on windows clustered testing

chrisdennis:


mathieucarbou:
@chrisdennis @cljohnso : it would seem that you probably have the same issue in Ehcache as we have in tc-platform (https://github.com/Terracotta-OSS/terracotta-platform/pull/645) with the port chooser problem (https://github.com/Terracotta-OSS/galvan/issues/215) ? The Windows build are also reaching the build time limit and get canceled ?

chrisdennis:
Close in favor of #2762

PR #2756, opened 2020-04-07
Ehcache with dynamic config kit

mathieucarbou:


clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
```diff
@@ -66,7 +67,7 @@
 
   @ClassRule
   public static Cluster CLUSTER =
-    newCluster().in(clusterPath()).withServiceFragment(RESOURCE_CONFIG).build();
+    newCluster().in(clusterPath()).withServiceFragment(RESOURCE_CONFIG).startupBuilder(DynamicConfigStartupBuilder::new).build();
```
mathieucarbou, line 70: @saurabhagas : we would need a way to not have to go over all the Galvan tests in the project.
ideas:
- use a META-INF/services ?
- change teh default behavior of galvan ?
- have this class moved somewhere else (in galvan) and used by the rule ?
I will let you see that with @chrisdennis

clustered/integration-test/src/test/java/org/ehcache/testing/DynamicConfigStartupBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.terracotta.testing.config.ConfigConstants.DEFAULT_CLUSTER_NAME;
+
+public class DynamicConfigStartupBuilder extends DefaultStartupCommandBuilder {
```
mathieucarbou, line 33: @saurabhagas :
this class is duplicated from tc-platform and has no link at all with dynamic config code.
Could we move it to galvan ? and remove this from ehcache and tc-platform ?

clustered/osgi-test/src/test/java/org/ehcache/osgi/OsgiTestUtils.java
```diff
@@ -155,7 +114,16 @@ public static Cluster startServer(Path serverDirectory) throws IOException {
       "-Dtc.install-root=" + serverDir,
       "-cp", serverDir.resolve("lib").resolve("tc.jar").toString(),
       "com.tc.server.TCServerMain",
-      "-f", configFile.toString()));
+      "--cluster-name=foo",
```
mathieucarbou, line 117: Note: cluster-name required to pre-activate a node at startup

clustered/server/service/build.gradle
```diff
@@ -22,6 +22,8 @@ plugins {
 dependencies {
   service project(':clustered:server:service-api')
   service "org.terracotta:offheap-resource:$terracottaPlatformVersion"
+  service "org.terracotta:statistics:$statisticVersion"
+  service "org.terracotta.common:common-structures:$terracottaPlatformVersion"
```
mathieucarbou, line 26: @chrisdennis @saurabhagas : structures should come transitively from offheap jar. There is a bug in tc-platform. I will raise a PR and the last line could be removed.

For statistics, this is another matter, it can be left as-is as service right now.

clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
```diff
@@ -64,16 +63,16 @@ public boolean initialize(ServiceProviderConfiguration configuration, PlatformCo
...
     Iterator<OffHeapResources> iterator = extendedConfiguration.iterator();
     if (iterator.hasNext()) {
       offHeapResourcesProvider = iterator.next();
       if (offHeapResourcesProvider.getAllIdentifiers().isEmpty()) {
-        throw new UnsupportedOperationException("There are no offheap-resource defined, this is not supported. There must be at least one!");
+        LOGGER.warn("No offheap-resource defined - this will prevent provider from offering any EhcacheStateService.");
       }
     } else {
-      LOGGER.warn("No offheap-resource defined - this will prevent provider from offering any EhcacheStateService.");
+      throw new UnsupportedOperationException("There are no offheap-resource defined, this is not supported");
```
mathieucarbou, line 75: Same bugfix we did in EE: `OffHeapResources` is ALWAYS there, but with 0..N resources inside now.

clustered/integration-test/src/test/java/org/ehcache/testing/DynamicConfigStartupBuilder.java
myronkscott: platform should have a galvan-support module like tc-core and tc-ee to support its kit.  This class belongs there.

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
myronkscott: platform needs a galvan-support module that supports it's kit.  When this is the direction we go universally, galvan-support should be removed from tc-core.

clustered/integration-test/src/test/java/org/ehcache/testing/DynamicConfigStartupBuilder.java
mathieucarbou: There or in tc-platform in a common module or elsewhere, but the idea is to not have to go over all classes because this is not normal to add a specific behavior whereas dynanmic config becomes the default behavior. Tests in Ehcache should be able to run as-is without any modification.

chrisdennis: That would also be a place to solve the problem with needing to chose this class in every IT.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -263,15 +263,17 @@ public void test_D_server_capabilities_exposed() throws Exception {
     // tms entity
 
     managerCapabilities = readTopology().activeServerEntityStream().filter(serverEntity -> serverEntity.is(tmsServerEntityIdentifier)).findFirst().get().getManagementRegistry().get().getCapabilities().toArray(new Capability[0]);
-    assertThat(managerCapabilities.length).isEqualTo(3);
+    assertThat(managerCapabilities.length).isEqualTo(5);
 
-    assertThat(managerCapabilities[0].getName()).isEqualTo("OffHeapResourceSettings");
-    assertThat(managerCapabilities[1].getName()).isEqualTo("OffHeapResourceStatistics");
-    assertThat(managerCapabilities[2].getName()).isEqualTo("StatisticCollectorCapability");
+    assertThat(managerCapabilities[0].getName()).isEqualTo("DataRootSettings");
```
chrisdennis, line 268: What do the `DataRoot` capabilities do in an OSS system given that they are EE features?

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
chrisdennis: See my comment on the class duplication below.

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
```diff
@@ -1,61 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<tc-config xmlns="http://www.terracotta.org/config"
```
chrisdennis, line 2: The fact that there is no "default" configuration means that OSS users can no longer just run `./start-tc-server.sh` (or it's replacement) and have a functional 'cluster' of one machine. Do we want to consider having a default properties file equivalent?

clustered/clustered-dist/build.gradle
```diff
@@ -93,7 +86,7 @@ jar {
     'Bundle-SymbolicName': 'org.ehcache.clustered',
     'Export-Package': '!com.tc.*, !com.terracotta.*, !org.terracotta.*, !org.ehcache.*.internal.*, !sun.misc, ' +
       'org.ehcache.clustered.client.*, org.ehcache.clustered.common.*',
-    'Import-Package': '!sun.misc.*, org.ehcache.xml.*;resolution:=optional, jdk.jfr.*;resolution:=optional, *'
+    'Import-Package': '!sun.misc.*, org.ehcache.xml.*;resolution:=optional, jdk.jfr.*;resolution:=optional, com.fasterxml.jackson.annotation;resolution:=optional, *'
```
chrisdennis, line 89: Would be good to explore if this can be a straight exclusion rather than an optional import.

clustered/osgi-test/build.gradle
```diff
@@ -91,6 +91,14 @@ task unzipKit(type: Copy) {
 test {
   dependsOn unzipKit
   systemProperty 'kitInstallationPath', "$unzipKit.destinationDir/${project(':clustered:clustered-dist').archivesBaseName}-$project.version-kit"
+
+  // do not cache test execution
+  outputs.cacheIf { false }
+  // adds the ability, like Maven, to run only unit tests but skip IT/system tests
+  if(project.hasProperty('skipITs')) {
```
chrisdennis, line 98: If we're going to do this I'd rather tie the kit tests to a distinct task instance `integrationTest` for example. Then `check.dependsOn integrationTest` and `./gradlew test` only runs unit tests and `integrationTest` or `check` will run the kit tests as well. That said here are the current timings for various executions on my dev machine:
```
./gradlew cleanTest test
...
BUILD SUCCESSFUL in 4m 30s
143 actionable tasks: 36 executed, 107 up-to-date

./gradlew cleanTest test -x :clustered:integration-test:test -x :clustered:osgi-test:test
...
BUILD SUCCESSFUL in 2m 34s
122 actionable tasks: 33 executed, 89 up-to-date
```
Assuming things aren't regressing too badly with this PR then I'm not sure any change here is really needed.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
mathieucarbou: no - we had a meeting with @jhouserizer and you were there and in this meeting we all decided that it was ok to move the data-root resource to oss when moving dynamic config plus all the utility classes (we talked about them in this meeting too).

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
mathieucarbou: This is explained in the presentation, slide 27.
Initially, the goal was to have another script, to highlight this UX change.
Now, keeping the same script, we will have a lot of fuzzy questions like that, part of teh reason I don't like the idea to re-use old stuff.

The answer IMO would be: NO.

Now, the way a server starts, it starts unconfigured. User has to configure it, or start it pre-activated with a cluster name or a config file, and this needs to be documented. **The behavior to start & configure a server has changed.**

We cannot _by default_ have it run `./start-tc-server.s` and make the user get a 1-node activated stripe because he won't be able to use the same script to start an un-configured node and build his cluster (which should be the default, as it was decided 1 year ago with everybody). And it would nott be the same behavior as EE also.

So what we could do though, is provide a default `cluster.properties` in the `conf` fodler that is showing all configurable settings. We could add a default cluster-name there and if the user is pointing to this file att startup, it would pre-activate the node.

But anyway, he will have to know about the `--node-repository-dir` setting, which is not part of the configuration file, and this is one of the most important setting to know now.

clustered/clustered-dist/build.gradle
mathieucarbou: yes, should be solved by https://github.com/Terracotta-OSS/terracotta-platform/pull/647

clustered/osgi-test/build.gradle
mathieucarbou: I will remove from the PR when it will be ready. Don't want to loose time with that.
I just wanted a way to first run the unit tests without the system tests. It helps having such flags in tc-platform and EE because systems tests takes more than 1h in these projects.

clustered/clustered-dist/src/assemble/server/conf/tc-config.xml
mathieucarbou: Added a default config file:

```properties
#Configuration for node 'covid' in stripe ID 1
client-lease-duration=150s
client-reconnect-window=120s
cluster-name=cluster
failover-priority=availability
offheap-resources=primary-server-resource:512MB
stripe.1.node.1.node-bind-address=0.0.0.0
stripe.1.node.1.node-group-bind-address=0.0.0.0
stripe.1.node.1.node-group-port=9430
stripe.1.node.1.node-hostname=localhost
stripe.1.node.1.node-log-dir=%H/terracotta/logs
stripe.1.node.1.node-logger-overrides=
stripe.1.node.1.node-name=covid
stripe.1.node.1.node-port=9410
stripe.1.node.1.node-public-hostname=
stripe.1.node.1.node-public-port=
stripe.1.node.1.tc-properties=
```

Usage:

```
~/.../distributions/ehcache-clustered-3.9-SNAPSHOT-kit ( kit → origin ↑1 ↓1 {8} U:1 ⊘ )
❯  ./server/bin/start-tc-server.sh --config-file server/conf/cluster.properties 
2020-04-07 18:42:34,040 INFO - PID is 25189
2020-04-07 18:42:35,049 INFO - Available Max Runtime Memory: 1822MB
2020-04-07 18:42:35,660 INFO - Terracotta Server instance has started up as ACTIVE node on 0:0:0:0:0:0:0:0:9410 successfully, and is now ready for work.
2020-04-07 18:42:35,660 INFO - Moved to State[ ACTIVE-COORDINATOR ]

```

clustered/integration-test/src/test/java/org/ehcache/clustered/BasicCacheOpsMultiThreadedTest.java
mathieucarbou: @saurabhagas : I have opened TDB-4939. I will let you see with @chrisdennis  / @myronkscott about how to solve that. To me, PR works, is usable, can be merged and should be the priority.

clustered/integration-test/src/test/java/org/ehcache/testing/DynamicConfigStartupBuilder.java
mathieucarbou: @saurabhagas : I have opened TDB-4939. I will let you see with @chrisdennis  / @myronkscott about how to solve that. To me, PR works, is usable, can be merged and should be the priority.

clustered/osgi-test/build.gradle
mathieucarbou: removed

mathieucarbou:
@chrisdennis @saurabhagas : really don't understand the build failures. All works fine locally. Seems to be something with Azure ?

clustered/server/service/build.gradle
mathieucarbou: Updated the PR with release 5.7.4-pre10 to remove the second line

clustered/clustered-dist/build.gradle
mathieucarbou: Put the maven provided scope so that `jackson-annotations` does not come transitively (5.7.4-pre10), but I still need the addition of the line otherwise the test fails.

chrisdennis: Right - but you should be able to use a straight 'exclusion'. In fact to truly assert this isn't needed you should be able to apply a transitive dependency exclusion at the point where you depend on `common-structures` as well.

clustered/clustered-dist/build.gradle
```diff
@@ -119,6 +114,7 @@ distributions {
       }
       into ("server/plugins/lib") {
         from (configurations.serverLibs - configurations.serverApis)
+        duplicatesStrategy DuplicatesStrategy.EXCLUDE
```
chrisdennis, line 117: You'd be better off playing the trick that I use in the enterprise kit and applying this duplicatesStrategy at a higher level, but only for jar files.

clustered/clustered-dist/src/assemble/server/conf/cluster.properties
```diff
@@ -0,0 +1,17 @@
+#Configuration for node 'covid' in stripe ID 1
```
chrisdennis, line 1: Topical, but will likely not age well.

clustered/clustered-dist/src/assemble/server/conf/cluster.properties
```diff
@@ -0,0 +1,17 @@
...
+client-lease-duration=150s
+client-reconnect-window=120s
+cluster-name=cluster
+failover-priority=availability
+offheap-resources=primary-server-resource:512MB
```
chrisdennis, line 6: 'offheap'?

clustered/osgi-test/src/test/java/org/ehcache/osgi/OsgiTestUtils.java
chrisdennis: Could this startup code lean on the new config file for most of its defaults?

clustered/clustered-dist/build.gradle
mathieucarbou: oh! yes true!

clustered/clustered-dist/src/assemble/server/conf/cluster.properties
mathieucarbou: oups! saw the offheap name - forgot to put it back

clustered/clustered-dist/build.gradle
mathieucarbou: used :
```
      filesMatching('**/*.jar') {
        // We can safely exclude JAR duplicates as our dependency strategy is fail on conflict
        duplicatesStrategy DuplicatesStrategy.EXCLUDE
      }
```

clustered/clustered-dist/src/assemble/server/conf/cluster.properties
```diff
@@ -0,0 +1,16 @@
...
+cluster-name=cluster
+failover-priority=availability
+offheap-resources=main:512MB
+stripe.1.node.1.node-bind-address=0.0.0.0
+stripe.1.node.1.node-group-bind-address=0.0.0.0
+stripe.1.node.1.node-group-port=9430
+stripe.1.node.1.node-hostname=localhost
+stripe.1.node.1.node-log-dir=%H/terracotta/logs
+stripe.1.node.1.node-logger-overrides=
+stripe.1.node.1.node-name=covid
```
chrisdennis, line 12: Previous comment applies here too!

clustered/osgi-test/src/test/java/org/ehcache/osgi/OsgiTestUtils.java
mathieucarbou: oh yes we could point to it now!

mathieucarbou: wait! no!
default config file:

```
stripe.1.node.1.node-group-port=9430
stripe.1.node.1.node-port=9410
stripe.1.node.1.node-log-dir=%H/terracotta/logs
```
whereas the CLI here need to point to different generated values.

`--config-file` and CLI parameters are mutually exclusive.

clustered/clustered-dist/src/assemble/server/conf/cluster.properties
mathieucarbou: oh forgot this one!

clustered/osgi-test/src/test/java/org/ehcache/osgi/OsgiTestUtils.java
chrisdennis: Wouldn't it be lovely if the CLI could override settings in the config file? ;)

mathieucarbou: it is not consistent because it means that you could have 2 nodes that are meant to be activated in the same stripe that are activated and started with the same config, but if one of them is overriding the config passed, then they won't have the same topology and it will fail.

When using config a file, it means the user sets a cluster configuration to be used across th node and this config must not be altered. All the nodes must have the same.

When starting with the CLI, you are not configuring a topology, you are just creating a 1-node-cluster based on the CLI parameters.

This is not the same use case.

mathieucarbou:
@chrisdennis : 
- added a commit for the  management test leak
- added a commit to update to platform *-pre11 and core *-pre25. The release of tc-platform -pre11 with core -pre25 is currently in progress

mathieucarbou:
Upgrade to tc-platform *-pre12 with new CLI behaviors (release in progress)

saurabhagas:
I'll do the suggested changes in a follow-up PR, its not a big deal at the moment. @chrisdennis can this PR be merged - I need these changes in this PR so that I can work on the latest ehcache release.

mathieucarbou:
@saurabhagas : merge your PR https://github.com/Terracotta-OSS/terracotta-platform/pull/653 and triggered a new tc-platform release (-pre13)

saurabhagas:
Ignored 2 failing tests on Linux, and recorded them in Issue#2758. Merging this PR, and releasing a new version.

saurabhagas:
> Can we not sort out the `BasicExternalCluster` issues first so this PR doesn't touch every integration test in the repository?

I have higher priority tasks to work on at the moment, for which I need an ehcache release. Once that is done, I'll fix this. There's an issue tracking it already, so that it doesn't slip through.

PR #2759, opened 2020-04-13
terracottaPlatformVersion = 5.7.4-pre14

mathieucarbou:
Release is currently in progress so the build will need to be restarted.
All tests passed locally with the snapshot.


mathieucarbou:
@akomakom : the log collector is failing, but build success otherwise.
@chrisdennis : you could force merge, or @akomakom would have to fix the issue on the build and we wait for that.

PR #2761, opened 2020-04-14
[osgi] avoid javax.xml.bind version range to support Java 8

amarkevich:


chrisdennis:
I'm not very comfortable with the idea of leaving this as a completely open version on the import. I'm also a bit uneasy about expanding the range without adding test coverage to prove it's safe. I'm going to look in to expanding this in to a range that encompasses both 2.2.8 as well our current 2.3.1... probably something like `[2.2,3)` along with testing across both version bounds (at least for the tests in the `:xml` module).

chrisdennis:
Subsumed by #2858

PR #2762, opened 2020-04-18
Fix all the tests... everywhere.

chrisdennis:
All 3 of you are welcome to look at everything but the main things I'm interested in are:
@akomakom please confirm I'm okay to move everything to one job (since everything now runs the same targets)
@mathieucarbou could you look at the MnM test changes and confirm things look reasonable thanks.
@cljohnso if you could look at the utilities lib adoption work and the `TestRetryer` that would be great.


clustered/integration-test/src/test/java/org/ehcache/clustered/management/AbstractClusteringManagementTest.java
```diff
@@ -199,92 +168,46 @@ protected void initCM() throws InterruptedException {
...
-        .map(ServerEntity::getServerEntityIdentifier)
-        .findFirst()
-        .orElse(null);
-      sleep(500);
-    } while (tmsServerEntityIdentifier == null && !Thread.currentThread().isInterrupted());
-
-    do {
-      ehcacheClientIdentifier = readTopology().getClients().values()
+  private static void initIdentifiers() throws Exception {
+    while ((ehcacheClientIdentifier = readTopology().getClients().values()
```
mathieucarbou, line 172: It would be nice to have a version of WaitAssert that is able to return the value from the callable that has completed with the assertion inside, something like:
`ehcacheClientIdentifier = assertThat(() -> { return ... }, is(notNull()))`

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusterWithManagement.java
```diff
@@ -0,0 +1,130 @@
...
+          try {
+            after();
+            throw t;
+          } catch (Exception e) {
+            throw new MultipleFailureException(asList(t, e));
+          }
+        }
+      }
+    }, description);
+  }
```
mathieucarbou, line 103: It would be cleaner to depend on that: https://github.com/Terracotta-OSS/terracotta-utilities/pull/9/files

FYI I have modified the PR to not re-use the JUnit one, that is not able to passe the method description  to after() and before(), and is also not able to throw an exception from after().

So this is a different one now.

`ExtendedTestRule` is currently copied in tc-platform in or `common-test-utilities` module and is used across all our rules.

Also, `ExtendedTestRule` with the description passed to before() is required to correctly replace the superclass of Galvan `BasicExternalCluster` class, which need to get the test description, and to avoid replacing a test failure by a galvan cleanup failure.

Anyway... All of that is cleaned in tc-platform. But it would be nice to have a common behavior.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/DiagnosticTest.java
```diff
@@ -51,9 +51,9 @@ public void test_CACHE_MANAGER_CLOSED() throws Exception {
     int activePort = readTopology().serverStream().filter(Server::isActive).findFirst().get().getBindPort();
 
     Properties properties = new Properties();
-    properties.setProperty(ConnectionPropertyNames.CONNECTION_TIMEOUT, String.valueOf("5000"));
+    properties.setProperty(ConnectionPropertyNames.CONNECTION_TIMEOUT, String.valueOf("10000"));
     properties.setProperty(ConnectionPropertyNames.CONNECTION_NAME, "diagnostic");
-    properties.setProperty(PROP_REQUEST_TIMEOUT, "5000");
+    properties.setProperty(PROP_REQUEST_TIMEOUT, "10000");
```
mathieucarbou, line 56: needed to to the same in tc-platform to make slow azure windows build pass. 10 sec was not enough in the case of tc-platform.

mathieucarbou: I also needed to do that when using management clients directly in case you need: https://github.com/Terracotta-OSS/terracotta-platform/commit/990bd379672d07438f6465a7a95374d395144371#diff-2d8d456b4c6788aff9292dea30c0ff19R93-R99

akomakom:
The last build was affected by [this azure problem](https://developercommunity.visualstudio.com/content/problem/982515/unable-to-access-logs-of-task-after-timeout.html)

azure-pipelines.yml
```diff
@@ -27,10 +27,23 @@ jobs:
 - template: build-templates/gradle-common.yml@templates
   parameters:
     jdkVersion: '1.8'
```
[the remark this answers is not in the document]
chrisdennis: Prefer leaving this since it's explicitly a 'Linux**Java8**' job, and not a 'Linux**Default**' job.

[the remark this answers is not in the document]
akomakom: Sure.  Will probably need to change other things anyway if we move away from 8, plus changing the default won't affect ehcache

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
```diff
@@ -82,11 +84,14 @@
   private volatile boolean connected = true;
   private volatile boolean eventsEnabled;
 
+  private final Executor asyncWorker;
+
   public SimpleClusterTierClientEntity(EntityClientEndpoint<EhcacheEntityMessage, EhcacheEntityResponse> endpoint,
-                                       Timeouts timeouts, String storeIdentifier) {
+                                       Timeouts timeouts, String storeIdentifier, Executor asyncWorker) {
```
cljohnso, line 90: Recommend checking guarding `asyncWorker` from null.

clustered/client/src/test/java/org/ehcache/clustered/client/UnSupportedCombinationsWithClusteredCacheTest.java
```diff
@@ -113,7 +109,7 @@ public void testClusteredCacheWithXA() throws Exception {
                   .withService(new XAStoreConfiguration("xaCache"))
                   .build()
           )
-          .build(true);
+          .build(true).close();
```
cljohnso, line 112: A couple of things here ... is this test missing a `fail`?  Isn't the `StateTransitionException` thrown from `PersistentCacheManager.build` meaning `close` won't be called?  Would this one be better with a try-with-resources?

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
```diff
@@ -229,35 +229,37 @@ public void unknownClusteredCacheExample()
 
     PersistentCacheManager cacheManager1 = cacheManagerBuilderAutoCreate.build(false);
     cacheManager1.init();
+    try {
```
cljohnso, line 232: I know these are samples, but is there any reason you didn't shift them to try-with-resources as you have in the tests?

clustered/integration-test/src/test/java/org/ehcache/clustered/reconnect/BasicCacheReconnectTest.java
```diff
@@ -50,17 +59,6 @@
 import static org.terracotta.testing.rules.BasicExternalClusterBuilder.newCluster;
 
 public class BasicCacheReconnectTest extends ClusteredTests {
-  public static final String RESOURCE_CONFIG =
```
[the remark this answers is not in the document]
chrisdennis: Extracted out the lease and resource xml snippets in to static "builder" methods.

impl/src/main/java/org/ehcache/impl/persistence/FileUtils.java
```diff
@@ -17,33 +17,33 @@
...
 /**
  * A bunch of utility functions, mainly used by {@link DefaultLocalPersistenceService} and
  * {@link FileBasedStateRepository} within this class.
  */
 final class FileUtils {
   private static final Logger LOGGER = LoggerFactory.getLogger(FileUtils.class);
-  private static final Charset UTF8 = forName("UTF8");
   private static final int DEL = 0x7F;
   private static final char ESCAPE = '%';
+  private static final boolean IS_WINDOWS = System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("windows");
```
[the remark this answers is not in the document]
chrisdennis: Removed.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -118,7 +120,8 @@ public void removeClusterTierClientEntity(String cacheId) {
     clusterTierEntities.remove(cacheId);
   }
 
-  public void initClusterConnection() {
+  public void initClusterConnection(Executor asyncWorker) {
+    this.asyncWorker = asyncWorker;
```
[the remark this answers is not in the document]
chrisdennis: Yes. Done.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/SimpleClusterTierClientEntity.java
chrisdennis: Done.

clustered/client/src/test/java/org/ehcache/clustered/client/UnSupportedCombinationsWithClusteredCacheTest.java
chrisdennis: Yes... fail was missing, added. Close should be called if the build erroneously executes cleanly. In the case that it fails (as expected) then it shouldn't since closing an uninitialized cache manager throws an ISE.

clustered/client/src/test/java/org/ehcache/clustered/client/docs/GettingStarted.java
chrisdennis: As discussed given the funky (maybe broken) contract of close I'd rather avoid encouraging the try-with-resources pattern with `.build(false)` cache managers.

test-utilities/src/main/java/org/ehcache/testing/TestRetryer.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+      };
+    } else {
+      isClassRule = true;
+      return new Statement() {
+        @Override
+        public void evaluate() throws Throwable {
+          Iterator<T> iterator = inputs.iterator();
+          while (iterator.hasNext()) {
+            T input = iterator.next();
```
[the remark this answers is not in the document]
chrisdennis: Added some logging here.

PR #2763, opened 2020-04-20
Bumped up tc-platform

saurabhagas:


saurabhagas:
@mathieucarbou `ParallelTestCluster` overrides the `apply` method, which was made `final` in tc-platform.

mathieucarbou:
@chrisdennis : FYI I have updated `ParallelTestCluster` to match the updated Galvan rule that now uses `ExtendedTestRule` as a super class.

mathieucarbou:
> @mathieucarbou `ParallelTestCluster` overrides the `apply` method, which was made `final` in tc-platform.
> 
> Funnily enough, windows builds pass despite test code having compilation issues.

@saurabhagas (and @chrisdennis ) : I have made `ExtendedTestRule` final explicitly because sub classes have better not override this method because this is complex to correctly re-write it with the correct exception handling. If a subclass really needs just to use an apply method, it has better use composition over inheritene and override.

FYI `ExtendedTestRule` now exposes before and after methods **with the Description** as argument, so we have all we need.

mathieucarbou:
@chrisdennis : is it worth moving the `Parallel` to test-utilities ?

mathieucarbou:
All checks passed, but Gihub status was not updated. Merging.

@akomakom : FYI, here is what I saw on tc-platform yesterday. It has happened again gere.

image

BUT:

image

clustered/integration-test/src/test/java/org/ehcache/clustered/util/ParallelTestCluster.java
```diff
@@ -27,7 +28,7 @@
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
-public class ParallelTestCluster extends Cluster {
+public class ParallelTestCluster implements TestRule {
```
chrisdennis, line 31: This has decoupled `ParallelTestCluster` from `Cluster`. This seems like bad OO and is unhelpful from a maintenance POV. Ideally I would like to have seen `Cluster` decoupled from JUnit entirely in the earlier platform PR and then have this class implement both `Cluster` and `TestRule` independently.

mathieucarbou: Yes this is right we should go even further 👍 
This is another change to do then. 
Could you please open an issue in tc-platform describing that and an assign @saurabhagas to it ? 
We can track that in our board in our user story regarding QA/Testing, with low priority.
This is something he could work when I will be out.

PR #2764, opened 2020-04-20
Use a test-specific temp directory as the default used by GlobalFilePortLocker may not writable

saurabhagas:


chrisdennis:
No. You're breaking the whole file based port locker mechanism. This either needs to be fixed in the the port chooser where it needs a fallback position when $TMPDIR is unusable, and maybe it needs to be fix in the jenkins node itself by setting a process specific tmpdir.

PR #2766, opened 2020-04-22
Bumped-up platform version

saurabhagas:


saurabhagas:
Reran builds as the artifacts weren't available in the previous build instance.

saurabhagas:
All checks passed. Status here is stale.

PR #2777, opened 2020-05-11
Upgrade tc-platform to 5.8.0-pre5

mathieucarbou:
This update brings the switch to runnel for the dynamic topology entity.

Note: model is still annotated with jackson annotations, which will create some warnings and fail when compiling with `-Werror`.

There is a WIP PR in tc-platform (https://github.com/Terracotta-OSS/terracotta-platform/pull/716) that is removing all annotations from the model classes and moves them to jackson mixins. This PR will be ready tomorrow I hope but it might be to late for the XPC, that is why we can use the `5.8.0-pre5` release at the moment.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ sizeofVersion = 0.3.0
 jaxbVersion = 2.3.1
 
 # Terracotta clustered
-terracottaPlatformVersion = 5.8.0-pre4
+terracottaPlatformVersion = 5.8.0-pre5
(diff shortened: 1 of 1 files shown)


saurabhagas:
All tests passed - UI isn't refreshed. Merging. Will not release a new version immediately though.

mathieucarbou:
> All tests passed - UI isn't refreshed. A force merge would be required.

@chrisdennis @akomakom : pelase ^^

PR #2785, opened 2020-05-27
Notifying clients about topology changes required for failover.

mobasherul:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -81,22 +90,58 @@ private static URI extractClusterUri(URI uri) {
...
     public String getClusterTierManager() {
       return clusterTierManager;
     }
 
     @Override
     public LeasedConnection connect(Properties connectionProperties) throws ConnectionException {
-      return LeasedConnectionFactory.connect(servers, connectionProperties);
+      LeasedConnection connection = LeasedConnectionFactory.connect(servers, connectionProperties);
+      try {
+        EntityRef<DynamicTopologyEntity, Object, DynamicTopologyEntity.Settings> ref = connection.getEntityRef(DynamicTopologyEntity.class, 1, DynamicTopologyEntityConstants.ENTITY_NAME);
```
chrisdennis, line 117: Where does this version number come from and how do we make sure it stays in sync with the server side?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -81,22 +90,58 @@ private static URI extractClusterUri(URI uri) {
...
+
+          @Override
+          public void onNodeAddition(int stripeId, Node addedNode) {
+            servers.add(addedNode.getNodeAddress());
+          }
+        });
+        return new LeasedConnection() {
+          @Override
+          public void close() throws IOException {
+            dynamicTopologyEntity.close();
```
chrisdennis, line 133: I don't think you can do this... if the connection is down I think this will block the connection from being closed.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
mobasherul: We can introduce some constants for version number as well but currently it doesn't have that.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
mobasherul: Planning to introduce async enity close for handling this.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -81,22 +93,67 @@ private static URI extractClusterUri(URI uri) {
...
+          @Override
+          public void onNodeAddition(int stripeId, Node addedNode) {
+            servers.add(addedNode.getNodeAddress());
+          }
+        });
+        return new LeasedConnection() {
+          @Override
+          public void close() throws IOException {
+            try {
+              final Future<?> close = dynamicTopologyEntity.releaseEntity();
```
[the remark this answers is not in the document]
mobasherul: since entity close is not doing much removing this entiy close while closing conection.

clustered/client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -81,22 +95,66 @@ private static URI extractClusterUri(URI uri) {
...
+          public <T extends Entity, C, U> EntityRef<T, C, U> getEntityRef(Class<T> cls, long version, String name) throws EntityNotProvidedException {
+            return connection.getEntityRef(cls, version, name);
+          }
+
+          @Override
+          public void close() throws IOException {
+            Future<?> close = dynamicTopologyEntity.releaseEntity();
+            try {
+              close.get(10, TimeUnit.SECONDS);
+              connection.close();
```
chrisdennis, line 146: This needs to be in finally block.

PR #2793, opened 2020-06-29
Updating server startup doc

mobasherul:


clustered/client/src/test/resources/configs/docs/cluster.properties
```diff
@@ -0,0 +1,4 @@
+cluster-name=mycluster
```
saurabhagas, line 1: If you do end up retaining this file, this line could be removed.

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -134,7 +134,7 @@ It contains the bare minimum configuration required for the samples in the rest
 
 [source,xml,indent=0]
 ----
-include::{sourcedir38}/clustered/client/src/test/resources/configs/docs/tc-config.xml[]
+include::{sourcedir38}/clustered/client/src/test/resources/configs/docs/cluster.properties[]
 ----
 
 The above configuration defines two named server off-heap resources:
```
saurabhagas, line 140: The callouts at lines 142 and 143 are now dead. You'd need to define callouts in the properties file.

IMO, we don't need a properties file merely to mention `offheap-resources`. I'd include the lone `offheap-resources=primary-server-resource:128MB,secondary-server-resource:96MB` line here, and get rid of the config file. Note that unlike before, having a config file isn't mandatory now - you could specify `offheap-resources` during server startup without needing a config file.

docs/src/docs/asciidoc/user/clustered-cache.adoc
```diff
@@ -151,14 +151,14 @@ On Windows:
 [source,cmd]
 ----
 cd <path/to/terracotta/kit>/server/bin
-start-tc-server.bat -f <path/to/server/config>/tc-config.xml
+start-tc-server.bat --config-file clustered/client/src/test/resources/configs/docs/cluster.properties
```
saurabhagas, line 154: This was just an example of how to use the startup script with a config file. You don't need to specify an actual path here. Saying `start-tc-server.bat -f path/to/config-file` suffices.

I'd collapse the unix example into this one and simply use `start-tc-server.(bat|sh) -f path/to/config-file`.

Also, it would be good idea to mention the usage of the script without the config file.

PR #2804, opened 2020-07-17
Standardize async testing timeouts

chrisdennis:


clustered/integration-test/src/test/java/org/ehcache/clustered/replication/DuplicateTest.java
```diff
@@ -136,8 +136,7 @@ public void duplicateAfterFailoverAreReturningTheCorrectResponse() throws Except
             case "equals":
               return proxy == args[0];
             default:
-              fail("Unexpected method call: " + method.getName());
-              return null;
+              throw new AssertionError("Unexpected method call: " + method.getName());
```
cljohnso, line 139: Any reason you didn't use `org.junit.Assert#fail(java.lang.String)`?

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/DuplicateTest.java
```diff
@@ -148,7 +147,6 @@ private StoreAccessException findStoreAccessException(Object[] objects) {
         return (StoreAccessException) o;
       }
     }
-    fail("There should be an exception somewhere in " + Arrays.toString(objects));
-    return null;
+    throw new AssertionError("There should be an exception somewhere in " + Arrays.toString(objects));
```
cljohnso, line 150: Any reason you didn't use `org.junit.Assert#fail(java.lang.String)`?

clustered/integration-test/src/test/java/org/ehcache/testing/StandardTimeouts.java
```diff
@@ -0,0 +1,27 @@
...
+package org.ehcache.testing;
+
+import org.terracotta.utilities.test.matchers.Eventually;
+
+import java.time.Duration;
+
+public interface StandardTimeouts {
+
+  static Eventually.Timeout eventually() {
+    return Eventually.within(Duration.ofMinutes(1L));
```
cljohnso, line 25: Should the timeout value be alterable by Java system property?

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/DuplicateTest.java
chrisdennis: I think I started with the move to Hamcrest over AssertJ (since the eventually method works with the former). In the process I would have just moved to my preferred pattern. I'm not a fan of unreachable statements.

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/DuplicateTest.java
chrisdennis: Same reason as above.

clustered/integration-test/src/test/java/org/ehcache/testing/StandardTimeouts.java
chrisdennis: I might argue no... this timeout is a trade-off of promptness of "true-positive" versus rate of 'false-positive' - if all tests are passing it should have no effect on test runtimes. For now the unification is to make it easier to bump in the future (if needed). I hope that we can just have a standard value everywhere and never need to tune it for certain environments.

PR #2807, opened 2020-07-22
Upgrade to new tc-platform version

mathieucarbou:
### Requires:

- [x] merge https://github.com/Terracotta-OSS/terracotta-platform/pull/790/
- [x] release tc-platform `5.8.0-pre17`

### DIFF
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
@@ -383,7 +383,7 @@ public class ClusteringManagementServiceTest extends AbstractClusteringManagemen
       .map(ContextualStatistics::getCapability)
       .collect(Collectors.toCollection(TreeSet::new));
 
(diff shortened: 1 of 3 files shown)


clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -383,7 +383,7 @@ public void test_G_stats_collection() throws Exception {
       .map(ContextualStatistics::getCapability)
       .collect(Collectors.toCollection(TreeSet::new));
 
-    assertThat(capabilities).containsOnly("PoolStatistics", "ServerStoreStatistics", "OffHeapResourceStatistics");
+    assertThat(capabilities).contains("PoolStatistics", "ServerStoreStatistics", "OffHeapResourceStatistics");
```
chrisdennis, line 386: Why the switch here?

mathieucarbou: I was testing something related to data dirs and metadata-dir and settings / config, which would end up putting some data dir resources on the server, with their management info. I have relaxed this check because ehcache does not need to enforce that the resources are only these ones. Ehcache only needs to test that these resources are at least there, but there can be other ones too.

PR #2810, opened 2020-07-29
Upgrade to new tc-platform version

mathieucarbou:


mathieucarbou:
Note: release of `5.8.0-pre18` is in progress...

saurabhagas:
> Note: release of `5.8.0-pre18` is in progress...

5.8.0-pre19 in progress now.

clustered/integration-test/src/test/java/org/ehcache/clustered/management/ClusteringManagementServiceTest.java
```diff
@@ -271,17 +271,15 @@ public void test_D_server_capabilities_exposed() throws Exception {
     // tms entity
 
     managerCapabilities = readTopology().activeServerEntityStream().filter(serverEntity -> serverEntity.is(CLUSTER.getTmsServerEntityIdentifier())).findFirst().get().getManagementRegistry().get().getCapabilities().toArray(new Capability[0]);
-    assertThat(managerCapabilities.length).isEqualTo(5);
+    assertThat(managerCapabilities.length).isEqualTo(3);
```
chrisdennis, line 274: Given the sequence of changes here would we be better off just asserting that the array contains the 3 things we care about (being ambivalent about what else it might contain)?

mathieucarbou: Yes, completely agree. but looking at this test to do that cleanly we would need to rework the whole file, add some matchers and methods to correctly find items with a given name in the list. This is not the only place like that.
So for another time... You can open an issue that won't be priority as of now but can be done after the release. 
We need this release asap.

PR #2820, opened 2020-08-14
 (DC config versioning and backward compat)

mathieucarbou:
Includes:
- DC config versioning and backward compat'
- M&M changes from Gary

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ sizeofVersion = 0.3.0
 jaxbVersion = 2.3.1
 
 # Terracotta clustered
-terracottaPlatformVersion = 5.8.0-pre24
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -10,7 +10,7 @@ sizeofVersion = 0.3.0
 jaxbVersion = 2.3.1
 
 # Terracotta clustered
-terracottaPlatformVersion = 5.8.0-pre23
+terracottaPlatformVersion = 5.8.0-pre25
```
mathieucarbou, line 13: @GaryWKeim @chrisdennis : FYI updated to -25 directly (release in progress) to also include Gary's changed for M&M

PR #2833, opened 2020-09-01
Remove client directory exclusion from platform kit

ramsai1729:
client directory exclusion is causing problems with voter script

### DIFF
--- a/clustered/clustered-dist/build.gradle
+++ b/clustered/clustered-dist/build.gradle
@@ -109,8 +109,8 @@ distributions {
           f.path = f.path.replace("platform-kit-$terracottaPlatformVersion/", "")
         }
         exclude { f ->
-          // Exclude tc's client subdir and README.txt - Issue 1273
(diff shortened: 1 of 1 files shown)


ramsai1729:
Adding the kit diff - 

```
~ % diff ~/new.txt ~/old.txt
159d158
< | | |____client-runtime-5.7.0.jar
166,172d164
< | |____logging
< | | |____impl
< | | | |____logback.xml
< | | | |____tc-tripwire-appenders-1.0.0.jar
< | | | |____logback-classic-1.2.3.jar
< | | | |____logback-core-1.2.3.jar
< | | |____slf4j-api-1.7.25.jar
~ %
```

chrisdennis (approved):
This is good. My understanding is that there is work in progress upstream to align on the location of the voter in terracotta-platform, right?

ramsai1729:
> This is good. My understanding is that there is work in progress upstream to align on the location of the voter in terracotta-platform, right?

Yes, we can merge this PR, will release new version after upgrading the core and platform (with voter location changes)

PR #2862, opened 2020-12-08
Send server keys with iterator batches

chrisdennis:


clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
```diff
@@ -69,12 +76,17 @@ public static void encode(StructEncoder<?> encoder, Chain chain) {
     elementsEncoder.end();
   }
 
-  public static Chain decode(byte[] payload) {
+  public static void encodeChainEntry(StructEncoder<?> encoder, Map.Entry<Long, Chain> chain) {
```
cljohnso, line 79: `ChainCodecTest` lacks tests for this method.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/messages/ChainCodec.java
```diff
@@ -110,4 +122,10 @@ public String toString() {
 
     return chainFromList(elements);
   }
+
+  public static Map.Entry<Long, Chain> decodeChainEntry(StructDecoder<?> decoder) {
```
cljohnso, line 126: `ChainCodecTest` lacks tests for this method.

cljohnso (changes requested):
Looks good with the exception of a couple of missing unit test cases.

cljohnso (approved):

PR #2864, opened 2021-01-08
Suppress CVE-2020-13956 for pax-url-aether

chrisdennis:


cljohnso (approved):
If you haven't already, you probably want to add an issue tracking the need to update pax-url-aether when an update address the CVE is available.

PR #2871, opened 2021-01-26
Issue 2866 : Don't pass duplicate schemas to factory

chrisdennis:


xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -314,9 +323,22 @@ public Document configToDocument(Configuration configuration) throws JAXBExcepti
...
+        objectFactory.createServiceConfiguration(null).getName(),
+        objectFactory.createResource(null).getName());
+
+      ABSTRACT_TYPE_FAILURES = asList(
+        //Xerces
+        abstractTypes.stream().map(element -> quote(format("\"%s\":%s", element.getNamespaceURI(), element.getLocalPart())))
+          .collect(collectingAndThen(joining("|", "^\\Qcvc-complex-type.2.4.a\\E.*'\\{.*(?:", ").*\\}'.*$"), Pattern::compile)),
+        //Saxon
+        abstractTypes.stream().map(element -> quote(element.getLocalPart()))
+          .collect(collectingAndThen(joining("|", "^.*\\QThe content model does not allow element\\E.*(?:", ").*"), Pattern::compile)));
```
cljohnso, line 340: The `SAXParseException` message is _stable_ regardless of Locale?

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
```diff
@@ -392,16 +414,26 @@ public static DocumentBuilder documentBuilder(Schema schema) throws ParserConfig
...
+      }
     }
     for (CacheResourceConfigurationParser p : servicesOfType(CacheResourceConfigurationParser.class)) {
-      schemaSources.add(p.getXmlSchema());
+      if (!pluginSchemas.containsKey(p.getNamespace())) {
+        pluginSchemas.put(p.getNamespace(), p.getXmlSchema());
+      }
     }
+
+    List<Source> schemaSources = new ArrayList<>(asList(fixedSources));
```
cljohnso, line 434: Does there need to be a guard against a fixed source being the same as a discovered plugin source?

cljohnso (commented):
Just a few questions ...

clustered/client/src/test/java/org/ehcache/clustered/client/XmlUnknownCacheTest.java
```diff
@@ -53,7 +54,7 @@ public void testGetUnknownCacheInvalidElement() {
       new XmlConfiguration(this.getClass().getResource("/configs/unknown-cluster-cache-invalid-element.xml"));
       fail("Expected XmlConfigurationException");
     } catch(XmlConfigurationException xce) {
-      assertThat(xce.getCause().getMessage(), endsWith("Element 'tc:clustered' must have no character or element information item [children], because the type's content type is empty."));
+      assertThat(xce.getCause().getMessage(), allOf(containsString("haracter"), containsString("clustered"), containsString("empty")));
```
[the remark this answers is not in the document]
chrisdennis: From memory yes. My trial license for Saxon has expired though so I cannot validate (easily).

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
chrisdennis: It almost certainly isn't... but there is no way to handle it. Xerces is nice in that it quotes the relevant section of the spec which is probably stable. What saves us here is that this condition doesn't affect pass/fail behavior it just affects how helpful we can be when we see this failures.

xml/src/main/java/org/ehcache/xml/ConfigurationParser.java
chrisdennis: It would be coding error in Ehcache... and it wouldn't be very easy to enforce since we don't know until they are parsed what the namespaces of the fixed sources are.

cljohnso (approved):

Issue #2876, opened 2021-02-11
With XML-based config the cache cannot be restarted with previous mappings


haqer1:
whereas with Java-API-based config, it can be restarted with previous mappings.

haqer1:
I think it would be a good idea to fix this. To reproduce the issue, feel free to use the parameterized JUnit test from PR #2877.

chrisdennis:
You need to set the disk resource to be persistent: `<disk persistent='true' unit='MB'>1024</disk>`

haqer1:
Good to know. Thanks.

PR #2877, opened 2021-02-11
Parameterized JUnit test for issue #2876 (coded to fail until the issue is fixed: will pass once it is)

haqer1:
If there is a fix to be checked in for the issue, i think this PR should be merged.


haqer1:
To clarify: this test now passes when using Java-API-based-config [1: useXmlBasedConfig = false], and fails when using the same config sourced via XML [0: useXmlBasedConfig = true]. It's coded to fail at present when using XML-based config. Once the issue is fixed (& XML-based config implementation starts to work the same as Java-based one for disk persistence), the test will pass.

P.S. This is at present in the assertion at EhCacheProviderWithXMLOrJavaBasedConfigTest.java:46
(for > testEhCachePutAndGet[0: useXmlBasedConfig = true]). More assertions could be added in the future, but it makes the point i'm trying to make well enough at present.
Ekran Körüntisi - 2021-02-11

chrisdennis:
The `disk` XML element takes a `persistent` attribute that defaults to false. It needs to be set to true for the cache to be persistent.

haqer1:
Makes sense: matches a method param in Java-API-based config. Thanks.

PR #2878, opened 2021-02-11
Parameterized JUnit test for disk persistence with XML-based and Java-API-based configs

haqer1:
It might still make sense to check this in for testing purposes & as an example of configuration(s) for developers.
It's up to you, of course.


chrisdennis:
I have to be honest and say I'm not planning to merge this PR. I don't see this as a valuable enough addition to the test set to justify the extra runtime and maintenance. If you'd like to pivot this in to a set of documentation chnages that would end up on the website then I'd consider it in that form. Making those changes might then lead to modifying existing tests or adding new ones like this to provide code samples/snippets to get embedded in the documentation.

PR #2893, opened 2021-03-24
Fixes #2892 : Capture the ThreadGroup of the starting thread and use it for async workers

### LINKED ISSUE #2892: DefaultClusteringService async executor can create Threads in the TCThreadGroup.
DefaultClusteringService can trigger thread creation inside the TCThreadGroup for the currently active client. This is bad and can lead to issues with client shutdown and during application level reconnect. It should always use the ThreadGroup of the thread that started the service (or some non-destroyed ancestor group).

chrisdennis:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/DefaultClusteringService.java
```diff
@@ -337,8 +337,18 @@ public String toString() {
...
+      ThreadGroup group = initialGroup;
+      while (group != null && group.isDestroyed()) {
+        ThreadGroup parent = group.getParent();
+        if (parent == null) {
+          break;
+        } else {
+          group = parent;
+        }
+      }
+      Thread t = new Thread(group, r, "Async DefaultClusteringService Worker");
```
myronkscott, line 351: Is this racy?  What happens if thread group is destroyed between looping and thread creation?

chrisdennis: Yes, it's racy... but in a worse way than you think. An unstarted thread won't prevent a thread group from being destroyed. So yes I could catch `IllegalThreadStateException` here and continue the loop of trying the parents... but I can't prevent the group from being destroyed between when I create the thread and when the executor starts the actual thread. Since use of thread groups is rare, and when it does happen they are usually (should be?) application context scoped (and therefore lifecycled alongside the cache?) I think we're as safe as we need to be. If you want me to I can narrow the race in the way you describe though.

myronkscott (approved):

chrisdennis:
@myronkscott clean CI finally... merge when you are happy/ready.

PR #2894, opened 2021-03-24
move integrated tests to inline testing

myronkscott:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -149,7 +149,11 @@ private void reconnect() {
   }
 
   private void connect() throws ConnectionException {
-    clusterConnection = connectionSource.connect(connectionProperties);
+    try {
+      clusterConnection = connectionSource.connect(connectionProperties);
+    } catch (ConnectionClosedException closed) {
```
chrisdennis, line 154: I think I'd prefer it if we just caught this in the callers instead. That should only be a small number of places. Should be able to use a multi-catch and treat it the same as the existing `ConnectionException` catches.

clustered/integration-test/build.gradle
```diff
@@ -60,6 +60,8 @@ task copyServerLibs(type: Copy) {
 
 test {
   dependsOn copyServerLibs
+
+  jvmArgs "-Xmx8g", "-XX:+UseG1GC", "-XX:MaxDirectMemorySize=64g"
```
chrisdennis, line 64: For me this passes fine without the tunings. Are we sure we need this?

chrisdennis (changes requested):

clustered/integration-test/build.gradle
myronkscott: yeah.  i get OOME without it on OversizedCacheOpsPassiveTest

chrisdennis: That test has been a thorn in my side for a while. Could you `@Ignore` it, remove the tuning and I'll look at it separately?

myronkscott: ok

myronkscott: Ignored.

myronkscott:
Rebased. Please have a look

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
```diff
@@ -52,10 +52,12 @@
       newCluster(2).in(clusterPath())
         .withSystemProperty("ehcache.sync.data.gets.threshold", "2")
         .withServiceFragment(offheapResource("primary-server-resource", 2))
-        .withSystemProperty("JAVA_OPTS", "-Xms1024m -Xmx8192m")
+        .withServerHeap(2048)
+        .inline(false)
         .build();
 
   @Test
+  @Ignore("OOME on build slaves due to high memory requirements")
```
chrisdennis, line 60: Probably safer to `@Ignore` the class and leave the cluster setup unmodified for now.

clustered/integration-test/build.gradle
```diff
@@ -59,6 +59,7 @@ task copyServerLibs(type: Copy) {
 }
 
 test {
+  jvmArgs '-Xmx512m'
```
chrisdennis, line 62: Could we use `maxHeapSize '512m'` instead? Also could we set `maxParallelForks 8` to cap the total heap usage at 4GB?

chrisdennis: While you're at it could we lower the `maxParallelForks` in the top level `build.gradle` to 16 to give the same cap?

clustered/server/entity/src/test/java/org/ehcache/clustered/server/store/ClusterTierActiveEntityTest.java
```diff
@@ -113,6 +117,9 @@
 public class ClusterTierActiveEntityTest {
 
   private static final KeySegmentMapper DEFAULT_MAPPER = new KeySegmentMapper(16);
+  private static final int MAX_SYNC_CONCURRENCY = 1;
+  private static final ExecutorService SYNC_GETS_EXECUTOR = new ThreadPoolExecutor(0, MAX_SYNC_CONCURRENCY,
```
chrisdennis, line 121: Are entities passive sync-ed in parallel or sequentially? If it's done in parallel we might have introduced a bottleneck. Previously we had one thread per cache... now we have one thread for all the caches. I think given the way the executors are used the equivalent would be `Executors.newCachedThreadPool()`?

chrisdennis (changes requested):

clustered/server/entity/src/test/java/org/ehcache/clustered/server/store/ClusterTierActiveEntityTest.java
myronkscott: Caches are synced sequentially.  A single thread should be sufficient no?

clustered/integration-test/src/test/java/org/ehcache/clustered/replication/OversizedCacheOpsPassiveTest.java
myronkscott: problem is `.withSystemProperty("JAVA_OPTS", "-Xms1024m -Xmx8192m")` does nothing the way galvan sets up the cluster

myronkscott:
@chrisdennis review items addressed I believe.  Please take a second look.

clustered/integration-test/build.gradle
myronkscott: done.

chrisdennis (approved):

PR #2897, opened 2021-04-15
Ensure previous connection is fully closed before starting reconnection

chrisdennis:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
```diff
@@ -139,6 +139,12 @@ public void initClusterConnection(Executor asyncWorker) {
   private void reconnect() {
     while (true) {
       try {
+        try {
+          //Ensure full closure of existing connection
+          clusterConnection.close();
+        } catch (IOException | ConnectionClosedException | IllegalStateException e) {
+          LOGGER.debug("Exception closing previous cluster connection", e);
```
cljohnso, line 146: Would this be better at INFO or WARN since it could have an impact the subsequent `connect` call?  Or are you expecting this to be an irrelevant issue often enough to want to suppress the noise?

cljohnso (commented):
Looks fine but I have an open question ....

clustered/client/src/main/java/org/ehcache/clustered/client/internal/service/ConnectionState.java
cljohnso: The `ConnectionShutdownException` might the the most frequent noise .... you could eat this one and log anything else.

cljohnso (approved):

PR #2905, opened 2021-05-18
Support starting with diagnostic mode and without offheap resources

mathieucarbou:
In relation to: https://github.com/Terracotta-OSS/terracotta-platform/pull/1006

```
> start-tc-server.sh
> start-tc-server.sh -auto-activate
> start-tc-server.sh -config-file ...
> start-tc-server.sh -auto-activate -config-file ...
```

Config file with/without offheap-resources.

### DIFF
--- a/clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
+++ b/clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
@@ -36,6 +36,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
@@ -62,6 +63,8 @@ public class EhcacheStateServiceProvider implements ServiceProvider, Closeable {
   @Override
   public boolean initialize(ServiceProviderConfiguration configuration, PlatformConfiguration platformConfiguration) {
     Collection<OffHeapResources> extendedConfiguration = platformConfiguration.getExtendedConfiguration(OffHeapResources.class);
+    // When a server is activated, there will ALWAYS be one OffHeapResources, that will hold the mapping configured by the user with the "offheap-resources" setting.
+    // In diagnostic mode, no extended configuration is loaded, so there won't be any OffHeapResources. In that case, we ask this service to be discarded (by returning false).
     if (extendedConfiguration.size() > 1) {
       throw new UnsupportedOperationException("There are " + extendedConfiguration.size() + " OffHeapResourcesProvider, this is not supported. " +
         "There must be only one!");
@@ -73,7 +76,7 @@ public class EhcacheStateServiceProvider implements ServiceProvider, Closeable {
         LOGGER.warn("No offheap-resource defined - this will prevent provider from offering any EhcacheStateService.");
       }
     } else {
-      throw new UnsupportedOperationException("There are no offheap-resource defined, this is not supported");
+      return false;
     }
     return true;
   }
--- a/clustered/integration-test/src/test/java/org/ehcache/clustered/NoOffheapTest.java
+++ b/clustered/integration-test/src/test/java/org/ehcache/clustered/NoOffheapTest.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ehcache.clustered;
+
+import org.ehcache.StateTransitionException;
+import org.ehcache.config.units.MemoryUnit;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.terracotta.testing.rules.Cluster;
+
+import static java.util.function.UnaryOperator.identity;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.ehcache.clustered.client.config.builders.ClusteredResourcePoolBuilder.clusteredDedicated;
+import static org.ehcache.clustered.client.config.builders.ClusteringServiceConfigurationBuilder.cluster;
+import static org.ehcache.config.builders.CacheConfigurationBuilder.newCacheConfigurationBuilder;
+import static org.ehcache.config.builders.CacheManagerBuilder.newCacheManagerBuilder;
+import static org.ehcache.config.builders.ResourcePoolsBuilder.newResourcePoolsBuilder;
+import static org.junit.Assert.fail;
+
+
+public class NoOffheapTest extends ClusteredTests {
+
+  @ClassRule
+  public static Cluster CLUSTER = newCluster().in(clusterPath()).build();
+
+  @Test
+  public void testNoOffheap() throws InterruptedException {
+    try {
+      newCacheManagerBuilder().with(cluster(CLUSTER.getConnectionURI().resolve("/no-offheap-cm"))
+        .autoCreate(identity()))
+        .withCache("testNoOffheap", newCacheConfigurationBuilder(Long.class, byte[].class, newResourcePoolsBuilder()
+          .with(clusteredDedicated("primary-server-resource", 1, MemoryUnit.MB))
+        )).build(true).close();
+      fail();
+    } catch (StateTransitionException e) {
+      assertThat(e).hasMessage("Could not create the cluster tier manager 'no-offheap-cm'.");
+    }
+  }
+}
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ sizeofVersion = 0.4.0
 jaxbVersion = [2.2,3)
 
 # Terracotta clustered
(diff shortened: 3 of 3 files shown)


clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
```diff
@@ -61,6 +62,11 @@ public void addStateTo(StateDumpCollector dump) {
 
   @Override
   public boolean initialize(ServiceProviderConfiguration configuration, PlatformConfiguration platformConfiguration) {
+    if (!platformConfiguration.getExtendedConfiguration(TopologyService.class).iterator().next().isActivated()) {
```
chrisdennis, line 65: What are the rules on which services should and shouldn't get activated? If the rules are fixed and could be evaluated by Voltron itself couldn't we just have it not call initialize rather than having providers have to know this contract?

chrisdennis (changes requested):

clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
mathieucarbou: There is no global rule. This is per service and extension. I.e, there are some service that need to be there because they are exposing some mbeans for diagnostic port. Security needs to be there also.

mathieucarbou: FYI, I wanted to have a better / cleaner API (I proposed something with just an annotation added to the service) but it requires a sort of hack by taking control over the `Server` impl with  `Server.getServer()` / `Server.setServer()` to filter the loading of builtin services and configs.

mathieucarbou: 3rd point to consider: the service loading is a bit messy. Right now, core controls the loading of builtin services without giving DC a chance to veto. But DC controls the loading of extended configurations and service provider configurations (those creating service providers at the end). So there is IMO a responsibility issue here and until this gets correctly solved I don't see how we can have a clean solution.

A fix would be to let DC provide through the `StartupConfiguration` the list of builtin services, exactly like DC is doing for the other services right now, and not have core load them.

@myronkscott : FYI discussion above ^^ :-)

chrisdennis: If we're not going to fix this properly by separating core services (like those that support the config system) from the rest of the services, then wouldn't it just be simpler to enumerate the 5 special services somewhere and not force every service implementation to know about the wart?

myronkscott: I’m curious why you care these services are loaded when unconfigured.  It doesn’t have consequences.  We should just fix persistence so that it does the correct action when presented with zero data-dirs.

mathieucarbou: > If we're not going to fix this properly by separating core services (like those that support the config system) from the rest of the services, then wouldn't it just be simpler to enumerate the 5 special services somewhere 

This is exactly similar to what I wanted to do with these enumerations: having a way to define exceptionally some non default rules: i.e. `OnlyWhen({CONFIGURED, ACTIVATED})`. The default being `OnlyWhen(ACTIVATED)`. This change would remove 80% of the changes required. But it also require DC to hook in the `ServerEnv` to control how the built in serviced are loaded.

I don't like enumeration because it is just an ugly workaround that is worse than what I have proposed and this list is also spread amongst EE and OSS... Besides, having such list of built in services to load (or not load) would require a core update if core is still loading itself the builtin services. So ion that case, why no just do things correctly and have core not loaded these services and DC instead and do the annotation filtering.

> and not force every service implementation to know about the wart?

they already know about it ;-) the dependency is there.

chrisdennis: The wart is about the fact that dynamic config shouldn't be a service. It needs to be more tightly coupled (or maybe just differently coupled) to the core code. Services shouldn't need to know about any of that.

chrisdennis: This, and I suspect a huge number of other both discovered and undiscovered warts are all symptomatic of the underlying issue with the way that dynamic configuration integrates with core. I don't see the value in band-aid fixing these symptoms, especially if those band aids are just causing the warts to metastasize like this does. We either need to isolate the warts and not allow them to spread (aka ignore them), or we need to remove them by fixing the root problems.

mathieucarbou: I agree with all that. I would like an approach where we do not touch the services excepts those we specifically want even in diagnostic mode. We can discuss that with @myronkscott later :-) I will ping you guys.

mathieucarbou: > I’m curious why you care these services are loaded when unconfigured. It doesn’t have consequences. We should just fix persistence so that it does the correct action when presented with zero data-dirs.

Base on what we said yesterday, when in diagnostic mode, no extension are loaded. So no data dirs and no offheap. Most of the services right now are relying on plugins to be there (i.e. FRS).

What I would like to avoid also is going to each service and change their code to make them optional if they do not see a plugin because this is wrong: it would imply we can start an activated EE server without any data dirs no persistence, or ehcache without offheap resources. Some are invalid states and some are states that for now we didn't want to have (like no persistence in EE).

That's why I think it would be far easier to let DC control the loading of all the services and have a way to explicitly (annotations ?) mark the ones that should not be filtered out (by default) when running in diagnostic mode.

clustered/clustered-dist/src/assemble/server/conf/cluster.cfg
```diff
@@ -5,14 +5,9 @@ failover-priority=availability
...
-default-node:node-group-port=9430
-default-node:node-hostname=localhost
-default-node:node-log-dir=%H/terracotta/logs
-default-node:node-logger-overrides=
-default-node:node-name=default-server
-default-node:node-port=9410
-default-node:node-public-hostname=
-default-node:node-public-port=
-default-node:tc-properties=
+default-node:bind-address=0.0.0.0
```
chrisdennis, line 8: These changes should really be in a separate commit. Arguably they should be daggy... but the timegap is short so it's probably harmless.

chrisdennis (approved):

clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
```diff
@@ -73,7 +76,8 @@ public boolean initialize(ServiceProviderConfiguration configuration, PlatformCo
         LOGGER.warn("No offheap-resource defined - this will prevent provider from offering any EhcacheStateService.");
       }
     } else {
-      throw new UnsupportedOperationException("There are no offheap-resource defined, this is not supported");
+      offHeapResourcesProvider = new OffHeapResourcesProvider(emptyMap());
+      LOGGER.warn("No offheap-resource defined - this will prevent provider from offering any EhcacheStateService.");
```
mathieucarbou, line 80: @chrisdennis : I think the warnings are misleading... There are 2 choices here:

1) either we log a warning and `return false` to tell core to discard the service
2) or we keep it as-is and fake an empty offheap, the service will stay and offer offheap services with inexisting resources...

What's the right thing for Ehcache ?

mathieucarbou: Forget my comment ;-) since we can add offheap at runtime we cannot discard the service.

clustered/server/service/src/main/java/org/ehcache/clustered/server/state/EhcacheStateServiceProvider.java
```diff
@@ -66,14 +66,14 @@ public boolean initialize(ServiceProviderConfiguration configuration, PlatformCo
...
-      if (offHeapResourcesProvider.getAllIdentifiers().isEmpty()) {
-        LOGGER.warn("No offheap-resource defined - this will prevent provider from offering any EhcacheStateService.");
-      }
-    } else {
-      throw new UnsupportedOperationException("There are no offheap-resource defined, this is not supported");
+    try {
+      // When a server is activated, there will ALWAYS be one OffHeapResources, that will hold the mapping configured by the user with the "offheap-resources" setting.
+      offHeapResourcesProvider = extendedConfiguration.iterator().next();
+    } catch (NoSuchElementException ignored) {
+      // In diagnostic mode, no extended configuration is loaded, so there won't be any OffHeapResources.
```
chrisdennis, line 73: Would we be better off to just keep the existing structure and change the `throw new UnsupportedOperationException(...)` for a logger call - then we also log something different when the "no resources" is due to the "no service".

mathieucarbou: Frankly, what would really be better is to `return false` instead to leverage platform's service loading capabilities and have platform discard the service. In diagnostic mode anyway, nothing is there to call `getService()`. This is what @myronkscott was also suggesting... I would prefer to change to that if possible ?

myronkscott: seems like the correct behavior but does this cascade into failure somewhere else in bootstrapping just in case this is post bootstrap and no off heap is defined?

mathieucarbou: No because nothing is calling `getService()` in diagnostic mode. And when activated, if no offheap, entity creation service is already having `throw new ConfigurationException("Unable to retrieve EhcacheStateService: " + e.getMessage());` in case no service is found.

PR #2915, opened 2021-06-29
Add custom manifest entries in jars

mathieucarbou:
Align manifest entries with https://github.com/Terracotta-OSS/terracotta-platform/pull/1023.

Example of manifest entries generated for clustered server module:

```
Implementation-Title: ehcache-server
Implementation-Revision: 97f13d24b0d600fabef627f344c5d98ef4ed64d4
Implementation-Version: 3.9-SNAPSHOT
Implementation-Vendor-Id: org.ehcache.modules
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ sizeofVersion = 0.4.0
 jaxbVersion = [2.2,3)
 
 # Terracotta clustered
-terracottaPlatformVersion = 5.8.6-pre7
+terracottaPlatformVersion = 5.8.6-pre8
 terracottaApisVersion = 1.8.1
 terracottaCoreVersion = 5.8.2-pre6
 terracottaPassthroughTestingVersion = 1.8.2


buildSrc/src/main/groovy/scripts/Utils.groovy
```diff
@@ -49,6 +49,8 @@ class Utils {
             'provider': 'gradle',
             'Implementation-Title': title,
             'Implementation-Version': "$version $revision",
+            'Bundle-Version': "$version",
```
chrisdennis, line 52: This is an OSGi specific attribute. We shouldn't be using it for our purposes.

buildSrc/src/main/groovy/scripts/Utils.groovy
```diff
@@ -49,6 +49,8 @@ class Utils {
...
+            'BuildInfo-Timestamp': new Date().format("yyyy-MM-dd'T'HH:mm:ssZ"),
```
chrisdennis, line 53: So this is a little awkward from a Gradle point of view... I'd also argue it's not massively helpful from a diagnostic POV. I would think the current `Implementation-Version` attribute format is a more informative thing that the version + timestamp combination.

clustered/common/src/main/java/org/ehcache/clustered/common/internal/util/ExtensionInfoSkeleton.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.jar.Attributes;
+import java.util.jar.JarFile;
+
+/**
+ * @author Mathieu Carbou
+ */
+public abstract class ExtensionInfoSkeleton implements ExtensionInfo {
```
chrisdennis, line 29: Not sure why this class is here and not in platform... it doesn't seem to be Ehcache specific.

chrisdennis (changes requested):

buildSrc/src/main/groovy/scripts/Utils.groovy
mathieucarbou: What's the alternative way to get the version from the manifest without the commit hash ? Should I create a custom entry for that ?

clustered/common/src/main/java/org/ehcache/clustered/common/internal/util/ExtensionInfoSkeleton.java
mathieucarbou: It is... In platform is is implemented another way to retrieve the details from the jar files.
Also, I have no issue in extending the one in platform. But it would introduce a dependency over DC in Ehcache (server-side). Do we want that ?

PR #2933, opened 2021-08-16
Use PortManager for OSGi tests.

chrisdennis:


clustered/osgi-test/src/test/java/org/ehcache/osgi/ClusterSupport.java
```diff
@@ -31,8 +32,10 @@
 
   public static Cluster startServer(Path serverDirectory) throws IOException {
     Path kitLocation = Paths.get(System.getProperty("kitInstallationPath"));
-    int tsaPort = selectAvailableEphemeralPort();
-    int tsaGroupPort = selectAvailableEphemeralPort();
+
+    PortManager portManager = PortManager.getInstance();
+    PortManager.PortRef tsaPort = portManager.reservePort();
+    PortManager.PortRef tsaGroupPort = portManager.reservePort();
```
mathieucarbou, line 38: I don't know how `PortRef` and such are implemented, but the way we use them in Galvan and Angela, we make sure to close these resources at the end to "release" the port so that it could be picked up again. Some port exhaust otherwise could happen since the file is common to several processes (from what I remember from the port manager impl we were using back then).

clustered/osgi-test/src/test/java/org/ehcache/osgi/ClusterSupport.java
```diff
@@ -93,6 +92,8 @@ public void close() {
           serverProcess.waitFor(60, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
           throw new AssertionError(e);
+        } finally {
+          ports.forEach(PortManager.PortRef::close);
```
chrisdennis, line 96: @mathieucarbou `PortRef`s closed here.

mathieucarbou: Sorry! I didn't see the short line ;-)

mathieucarbou (approved):

mathieucarbou:
@chrisdennis : can you rebase it ? I'll merge!

cljohnso (approved):

PR #2971, opened 2021-12-10
Correctly handle the distinction between client and server sourced invalidations

chrisdennis:


clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -803,8 +803,20 @@ public void onAppend(Chain beforeAppend, ByteBuffer appended) {
           }
         }
 
+        public void onAppendInvalidateHash(long hash) {
+          if (clusteredStore.invalidationValve != null) {
+            try {
+              LOGGER.debug("CLIENT: calling invalidation valve for hash {}", hash);
+              clusteredStore.invalidationValve.invalidateAllWithHash(hash);
+            } catch (StoreAccessException sae) {
+              //TODO: what should be done here? delegate to resilience strategy?
```
GaryWKeim, line 812: Is there any other choice? And what would the handler be named? invalidateAllWithHashFailure(long hash, StoreAccessException e)

chrisdennis: Yeah... this issue predates the split. The problem is the failure really needs to be reported back to the original putting thread... but that's on a different client. Depending on the configured consistency it might have left to do other things. I'm not inclined to take fixing this right now.

GaryWKeim (approved):

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/CommonServerStoreProxy.java
```diff
@@ -73,15 +73,14 @@
       long key = response.getKey();
       Chain evictedChain = response.getEvictedChain();
       LOGGER.debug("CLIENT: on cache {}, server requesting hash {} to be invalidated (evicted chain : {})", cacheId, key, evictedChain);
-      invalidation.onInvalidateHash(key, evictedChain);
+      invalidation.onEvictInvalidateHash(key, evictedChain);
```
jhouserizer, line 76: the above log message be reworded a tad to make it more explicit that an eviction happened?  (customer has been asking about enabling logging for those).

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/EventualServerStoreProxyTest.java
```diff
@@ -72,7 +77,12 @@ public void compact(ServerStoreProxy.ChainEntry chain) {
     });
     EventualServerStoreProxy serverStoreProxy2 = new EventualServerStoreProxy("testServerSideEvictionFiresInvalidations", clientEntity2, new ServerCallback() {
       @Override
-      public void onInvalidateHash(long hash, Chain evictedChain) {
+      public void onAppendInvalidateHash(long hash) {
+        //expected
```
cljohnso, line 81: Did you want to check to ensure this actually got called?

clustered/client/src/test/java/org/ehcache/clustered/client/internal/store/StrongServerStoreProxyTest.java
```diff
@@ -80,7 +85,12 @@ public void compact(ServerStoreProxy.ChainEntry chain) {
     });
     StrongServerStoreProxy serverStoreProxy2 = new StrongServerStoreProxy("testServerSideEvictionFiresInvalidations", clientEntity2, new ServerCallback() {
       @Override
-      public void onInvalidateHash(long hash, Chain evictedChain) {
+      public void onAppendInvalidateHash(long hash) {
+        //expected
```
cljohnso, line 89: Ensure called?

clustered/client/src/main/java/org/ehcache/clustered/client/internal/store/CommonServerStoreProxy.java
chrisdennis: I mean I guess we could... I'm more worried about the implication of turning on this logging).

cljohnso (approved):

PR #2975, opened 2021-12-15
OSGi import/export cleanups

chrisdennis:


ehcache-107/src/main/java/org/ehcache/jsr107/Eh107CacheLoaderWriterProvider.java
```diff
@@ -15,21 +15,30 @@
...
+import org.ehcache.spi.loaderwriter.CacheLoaderWriterConfiguration;
+import org.ehcache.spi.loaderwriter.CacheLoaderWriterProvider;
+import org.ehcache.spi.service.Service;
+import org.ehcache.spi.service.ServiceProvider;
 
 /**
  * @author teck
  */
-class Eh107CacheLoaderWriterProvider extends DefaultCacheLoaderWriterProvider {
+class Eh107CacheLoaderWriterProvider implements CacheLoaderWriterProvider {
```
chrisdennis, line 33: We're attempting here to decouple this class from the default cache loader writer provider. This eliminates the need to export an internal package from the implementation OSGi bundle.

ehcache-107/src/main/java/org/ehcache/jsr107/Eh107CacheLoaderWriterProvider.java
```diff
@@ -15,21 +15,30 @@
...
 
   private final ConcurrentMap<String, CacheLoaderWriter<?, ?>> cacheLoaderWriters = new ConcurrentHashMap<>();
+  private final Set<String> cachesWithJsrRegisteredLoaders = new HashSet<>();
+  private final ServiceLocator privateServiceLocator;
+  private CacheLoaderWriterProvider delegate;
 
   public Eh107CacheLoaderWriterProvider() {
-    super(null);
+    this.privateServiceLocator = ServiceLocator.dependencySet().with(CacheLoaderWriterProvider.class).build();;
```
chrisdennis, line 41: This is ugly and it might have some strange side effects if the CacheLoaderWriterProvider has service dependencies (the default one doesn't). Better solutions would include:
1. Making CacheLoaderWriterProvider a plural service so that we can delegate within the same service locator.
2. Allowing this service locator to delegate service lookups to the main loader.

cljohnso (approved):

PR #3009, opened 2022-04-12
default target platform should be 8

mathieucarbou:


mathieucarbou:
will be included into #3011

PR #3020, opened 2022-04-22
issue_2993 : Adding hint about JAXB to the documentation

nnares:


docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,26 @@ ifdef::notBuildingForSite[]
 include::{includedir}/menu.adoc[]
 endif::notBuildingForSite[]
 
+== Getting Ehcache
+
+From Maven
+
+    <dependencies>
```
chrisdennis, line 15: It's best to call out the language explicitly here so that we get syntax highlighting:
```asciidoc
[source,xml]
----
<dependencies>
    <dependency>
      <groupId>org.ehcache</groupId>
      <artifactId>ehcache</artifactId>
      <version>3.10.0</version>
    </dependency>
</dependencies>
----
```

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,26 @@ ifdef::notBuildingForSite[]
...
+        <dependency>
+          <groupId>org.ehcache</groupId>
+          <artifactId>ehcache</artifactId>
+          <version>3.10.0</version>
+        </dependency>
+    </dependencies>
+
+From Gradle
+
+    org.ehcache:ehcache:3.10.0
```
chrisdennis, line 25: Although this is a valid short form syntax for Gradle, it's probably better if we use the more complete:
```asciidoc
[source,groovy]
----
dependencies {
  implementation 'org.ehcache:ehcache:3.10.0'
}
----
```

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,26 @@ ifdef::notBuildingForSite[]
...
+
+ehcache-xml and ehcache modules requires jaxb dependency at runtime, So If you are using java 1.9+ version then you should make jaxb dependency available at runtime.
```
chrisdennis, line 27: I wouldn't bother to reference `ehcache-xml` here since you haven't mentioned it the the snippets above.

I'm not sure the emphasis is correct here either. Since the maven dependency is now `runtime` scoped for most users inclusion of JAXB on the classpath will be automatic. There are however two things that may be worth pointing out:
1. We depend on a version range. This means unless the user picks a version the build will be unstable.
2. If they want to use Jakarta EE artifacts they will need to select the Jakarta variants using the appropriate Maven/Gradle invocation.

chrisdennis (changes requested):
This might be factually correct, but I fear it is a little too terse (even for technical documentation). The documentation still has to read like prose. This is a good example of what we should be aiming for: http://hamcrest.org/JavaHamcrest/distributables

docs/src/docs/asciidoc/user/getting-started.adoc
nnares: Accommodated this review comment and push the changes

docs/src/docs/asciidoc/user/getting-started.adoc
nnares: Accommodated this review comment and push the changes

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,65 @@ ifdef::notBuildingForSite[]
...
+Add the following to the dependencies section of your build.gradle:
+
+[source,groovy]
+----
+dependencies {
+  implementation 'org.ehcache:ehcache:3.10.0'
+}
+----
+
+ehcache-xml and ehcache modules requires jaxb dependency. So If you are using jdk-9+ version then you should make jaxb dependency available at runtime.
```
chrisdennis, line 37: I think from here onwards we could split in to two sections:
- JAXB Dependencies:
- Jakarta Variants

The JAXB dependencies section can talk about the fact that we depend on a JAXB runtime. This is a maven/gradle declared runtime dependency. It will normally appear automatically. The dependency uses a version range though, this means it will float, and can therefore change from build to build. This means if the build must be reproducible then the build system should declare dependency on an explicit version.

The Jakarta variant section should talk about how to depend on the Jakarta variants from both Maven and Gradle, and how that complicates the dependency situation for Maven (the dependencies are declared as optional).

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,65 @@ ifdef::notBuildingForSite[]
...
+
+1. We depend on a version range, Version for jaxb dependency should range between 2.2.0 - 3.0.0
+2. We have upgraded the jaxb api to jakarta version
+
```
jhouserizer, line 41: I see that there were previously some changes made to this section (these few sentences)... and the text is left a bit rough.  Could we please update it to the following?:

```
The modules ehcache-xml and ehcache have a dependency upon JAXB.  If you are using JDK-9 or greater then you need to ensure the JAXB dependency is available at runtime.

Please note:
1. The JAXB dependency declared by Ehcache is with the version range 2.2.0 - 3.0.0.   You will probably want to specify a particular version, in order to get consistent builds of your components.
```

docs/src/docs/asciidoc/user/getting-started.adoc
nnares: Accommodated the review comment.

chrisdennis (changes requested):
See my PR against your branch: https://github.com/nnares/ehcache3/pull/1

linghengqian:
Why does the PR have a bunch of changes unrelated to Hint?

nnares:
this branch got the changes from another local branch, so closing this one and creating a new branch

PR #3031, opened 2022-05-13
issue-2993 : Adding hint about JAXB to the documentation

nnares:


chrisdennis (approved):

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,102 @@ ifdef::notBuildingForSite[]
...
+  <groupId>org.ehcache</groupId>
+  <artifactId>ehcache</artifactId>
+  <version>3.10.0</version>
+  <classifier>jakarta</classifier>
+</dependency>
+----
+
+[WARNING]
+====
+The Maven POM cannot capture the dependencies of the Jakarta variant.
```
cljohnso, line 70: "capture"?  Does that mean "express" or "describe"?  


(While the dependencies of the Jakarta variant of Ehcache can be described using Gradle metadata, there is no corresponding description using a Maven POM.  When using the Jakarta Ehcache variant in a Maven project, you will also need ... )


You might consider just including the following dependencies section along with the Jakarta dependency and with a possible footnote as to why the extras are needed.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -8,6 +8,102 @@ ifdef::notBuildingForSite[]
...
+    <artifactId>jaxb-runtime</artifactId>
+    <version>[3,4)</version>
+  </dependency>
+</dependencies>
+----
+====
+
+==== From Gradle
+
+Add the following to the dependencies section of your `build.gradle`:
```
cljohnso, line 96: Hmmm ... without some discussion of the difference between Gradle and Maven dependency expression here_, you will likely get people unnecessarily adding the above dependencies in a build.gradle...

The mention of differences might flow better if the Gradle version is presented first followed by the Maven version with a sentence as to why the extra stuff is needed.

cljohnso (changes requested):

docs/src/docs/asciidoc/user/getting-started.adoc
chrisdennis: Fixed as discussed.

docs/src/docs/asciidoc/user/getting-started.adoc
chrisdennis: Fixed, and then discussed.

cljohnso (approved):

PR #3041, opened 2022-07-13
issue-2887 : Auto clear invaild cache entry

nnares:


core-spi-test/src/main/java/org/ehcache/internal/store/StoreSPITest.java
```diff
@@ -32,6 +33,7 @@ public void testGetAndCompute() throws Exception {
   }
 
   @Test
+  @Ignore("TieredStoreSPITest#testComputeIfAbsent, TieredStoreWith3TiersSPITest#testComputeIfAbsent" )
   public void testComputeIfAbsent() throws Exception {
```
chrisdennis, line 37: These shouldn't be ignored... they either need to be adjusted (with some fairly strong justification since they enforce the Store contract) or they need to pass unmodified.

ehcache-api/src/main/java/org/ehcache/spi/resilience/StoreAccessRuntimeException.java
```diff
@@ -0,0 +1,76 @@
...
+public class StoreAccessRuntimeException extends RuntimeException {
+
+  private static final long serialVersionUID = 6249505400891654776L;
+
+  /**
+   * Creates a new exception wrapping the {@link Throwable cause} passed in.
+   *
+   * @param cause the cause of this exception
+   */
+  public StoreAccessRuntimeException(Throwable cause) {
```
chrisdennis, line 35: My understand was that this type was only needed to transit a StoreAccessException out of a function/lambda. If that is the case then we should constrain these to only be able to be constructed with a `StoreAccessException`. You could then override the `getCause()` method to return this type which might make some of the downstream exception handling simpler.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -994,59 +1002,54 @@ private long getSizeOfKeyValuePairs(K key, OnHeapValueHolder<V> holder) throws L
 
     @IgnoreSizeOf
     private final Supplier<ValueHolder<V>> source;
```
chrisdennis, line 1004: I think you should be able to remove the source reference here, and instead have the original initiating thread hold the source on-stack, call it, and then complete the Fault appropriately.

chrisdennis (changes requested):

core-spi-test/src/main/java/org/ehcache/internal/store/StoreSPITest.java
nnares: Deleted this ignore now, My code changes are ensuring these store contract.
Required to do some minor changes in - `StoreComputeIfAbsentTest#testStorePassThroughException()` to fix the break for `TieredStoreSPITest#testComputeIfAbsent, TieredStoreWith3TiersSPITest#testComputeIfAbsent`

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
nnares: After a thorough check, finds that ye, we can remove this -` Supplier<ValueHolder<V>> source` attribute from `Fault `class
Accommodated you review comment

zorglube:
Hello @nnares !!

Are you still working on this issue ?  
Since I'm currently setuping EHCache 3 with Hibernate, I have some Exception related to this issue. 

BR, 
Zorglube

PR #3044, opened 2022-07-19
Update dependency versions

cljohnso:
This commit updates dependency versions to adopt upstream
dependencies using version ranges for logging components.
In addition, this commit pins the following versions:

* slf4j 1.7.36
* terracotta-utilities 0.0.15

Except for demos, the logback version is inherited from
terracotta-core and terracotta-platform.

### DIFF
--- a/demos/build.gradle
+++ b/demos/build.gradle
@@ -6,10 +6,6 @@ subprojects {
   apply plugin: 'war'
   apply plugin: 'org.gretty'
 
-  repositories {
-    jcenter()
-  }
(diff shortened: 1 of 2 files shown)


mathieucarbou:
@cljohnso : can we merged that ? We need to bring a fix from platform to EE for cloud (so it will have to be in both branches)
thanks :-)

mathieucarbou (approved):

chrisdennis (approved):

PR #3056, opened 2022-08-09
feature-25363 : Deprecating size based cache at heap tier - updating user doc

nnares:


docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -162,6 +162,8 @@ You can create an XML file to configure a `CacheManager`.
 ----
 include::{sourcedir310}/ehcache-xml/src/test/resources/configs/docs/getting-started.xml[tags=gettingStarted]
 ----
+NOTE: The use of memory units (such as 'B', 'kB' or 'MB') for heap resources is deprecated and will be removed in a future version.
```
jhouserizer, line 165: I wonder if we should add "(This only applies to heap tiers)" ... I know the language says "for heap resources", but since there a multiple things in the examples, and readers may be new to the concepts, they might think too broadly the the feature in full is deprecated (for all tiers) - especially since the example does not show using for the heap tier!

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -73,13 +73,17 @@ include::{sourcedir310}/ehcache-impl/src/test/java/org/ehcache/docs/Tiering.java
 <2> A shortcut to specify 10 entries.
 <3> Only 10 MB allowed. Eviction will occur when full.
 
+NOTE: Byte based heap resources are deprecated and will be removed in a future version.
```
jhouserizer, line 76: We should probably add the note up at line 66 as well, or even just modify the statement at line 65, to be something like "A heap tier can be sized by entries or by size in bytes (though byte-based-sizing of heap tiers is now a deprecated feature)."

jhouserizer (changes requested):

docs/src/docs/asciidoc/user/tiering.adoc
nnares: Yes, line 66 should be updated as well, SO I have updated it to
`A heap tier can be sized by entries or by size(though byte-based-sizing of heap tiers is now a deprecated feature).`

docs/src/docs/asciidoc/user/getting-started.adoc
nnares: I got your point, Agreed as well, and Updated my note to 
`NOTE: The use of memory units (such as 'B', 'kB' or 'MB') for heap resources is deprecated and will be removed in a future version.This only applies to heap tiers.`

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -62,7 +62,7 @@ Here we use off-heap only.
 The starting point of every cache and also the faster since no serialization is necessary.
 You can optionally use copiers (see the section <<serializers-copiers.adoc#copiers,Serializers and Copiers>>) to pass keys and values by-value, the default being by-reference.
 
-A heap tier can be sized by entries or by size.
+A heap tier can be sized by entries or by size(though byte-based-sizing of heap tiers is now a deprecated feature).
```
jhouserizer, line 65: sorry, there should be a space between the word "size" and the parenthesis.

jhouserizer (approved):

docs/src/docs/asciidoc/user/tiering.adoc
nnares: incorporated this.

docs/src/docs/asciidoc/user/getting-started.adoc
```diff
@@ -162,6 +162,8 @@ You can create an XML file to configure a `CacheManager`.
 ----
 include::{sourcedir310}/ehcache-xml/src/test/resources/configs/docs/getting-started.xml[tags=gettingStarted]
 ----
+NOTE: The use of memory units (such as 'B', 'kB' or 'MB') for heap resources is deprecated and will be removed in a future version.This only applies to heap tiers.
```
chrisdennis, line 165: This sample doesn't even reference byte-based units. I think we can safely drop this note.

chrisdennis: Conversely there is a link to byte-based heap documentation at the end of this file that we should probably remove. Don't want to lead more people to using a deprecated feature (around line 260ish)

chrisdennis: There's also a reference to byte based sizing in the `performance.adoc` file that should probably get a warning/note.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -62,7 +62,7 @@ Here we use off-heap only.
 The starting point of every cache and also the faster since no serialization is necessary.
 You can optionally use copiers (see the section <<serializers-copiers.adoc#copiers,Serializers and Copiers>>) to pass keys and values by-value, the default being by-reference.
 
-A heap tier can be sized by entries or by size.
+A heap tier can be sized by entries or by size (though byte-based-sizing of heap tiers is now a deprecated feature).
```
chrisdennis, line 65: Let just drop this sentence. No need to note the deprecation if the sentence doesn't exist.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -73,13 +73,17 @@ include::{sourcedir310}/ehcache-impl/src/test/java/org/ehcache/docs/Tiering.java
 <2> A shortcut to specify 10 entries.
 <3> Only 10 MB allowed. Eviction will occur when full.
```
chrisdennis, line 74: Lets just delete this callout and the referenced code in the sample. Then you can delete the `NOTE:` below as well.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -73,13 +73,17 @@ include::{sourcedir310}/ehcache-impl/src/test/java/org/ehcache/docs/Tiering.java
...
+NOTE: Byte based heap resources are deprecated and will be removed in a future version.
+
 ==== Byte-sized heap
 
 For every tier except the heap tier, calculating the size of the cache is fairly easy.
 You more or less sum the size of all byte buffers containing the serialized entries.
 
 When heap is limited by size instead of entries, it is a bit more complicated.
 
+NOTE: Byte based heap resources are deprecated and will be removed in a future version.
```
chrisdennis, line 85: I would move this note to immediately beneath the section header so that the reader sees it before they read the section. I would also change it to either `WARNING:` or `IMPORTANT:` as they render much more visibly.

docs/src/docs/asciidoc/user/tiering.adoc
```diff
@@ -88,7 +92,7 @@ include::{sourcedir310}/ehcache-impl/src/test/java/org/ehcache/docs/Tiering.java
 ----
 
 <1> This will limit the amount of memory used by the heap tier for storing key-value pairs.
-There is a cost associated with sizing objects.
+There is a cost associated with sizing objects. Byte based heap resources are deprecated and will be removed in a future version.
```
chrisdennis, line 95: I think this sentence becomes excessive if we have a warning at the start of the section.

chrisdennis (commented):
I've added some comments, but some are at odds with what @jhouserizer suggested. Lets discuss tomorrow morning and see what we want to settle on.

docs/src/docs/asciidoc/user/performance.adoc
```diff
@@ -61,6 +61,8 @@ Size calculation is done using the https://github.com/ehcache/sizeof[SizeOf] lib
 This library uses multiple magic tricks to do so. It selects the fastest one for a given environment.
 Make sure of what is used to confirm you can't use a faster way on your platform.
 
+WARNING: Byte based heap resources are deprecated and will be removed in a future version.
```
chrisdennis, line 64: I would put this at the top of this section, immediately beneath the header, so that it's read first.

chrisdennis (approved):

docs/src/docs/asciidoc/user/getting-started.adoc
nnares: I have deleted this, I had put it for precautionary measure.

nnares: Done, removed the entire block

nnares: Added a warning note there as well (at the beginning) , I have changed all my prev NOTE to WARNING

docs/src/docs/asciidoc/user/tiering.adoc
nnares: just removed the comeplete sentence

docs/src/docs/asciidoc/user/tiering.adoc
nnares: Agreed, DONE

docs/src/docs/asciidoc/user/tiering.adoc
nnares: Agreed, Incorporated

docs/src/docs/asciidoc/user/tiering.adoc
nnares: Agreed, Incorporated

docs/src/docs/asciidoc/user/performance.adoc
nnares: Moved this to the beginning of paragraph.

chrisdennis (approved):

PR #3062, opened 2022-09-23
issue-3034 : correcting OccupiedBytes statistic value at heap tier

nnares:


ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteStatisticTest.java
```diff
@@ -0,0 +1,106 @@
...
+    cache.put("key2", "value2");
+    cache.put("key3", "value3");
+
+    long occupiedByteSize = statisticsService.getCacheStatistics("cacheName").getTierStatistics().get("OnHeap").getOccupiedByteSize();
+    assertThat(occupiedByteSize, is(984L));
+
+    cache.clear();
+    occupiedByteSize = statisticsService.getCacheStatistics("cacheName").getTierStatistics().get("OnHeap").getOccupiedByteSize();
+    assertThat(occupiedByteSize, is(984L));
+    assertThat(occupiedByteSize, not(0L));
```
GaryWKeim, line 55: I don't understand these redundant, secondary checks. If it's 984, it's not 0.

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteStatisticTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    cache.remove("key1");
+    cache.remove("key2");
+    cache.remove("key3");
+    long currentMemoryPostRemove = statisticsService.getCacheStatistics("cacheName").getTierStatistics().get("OnHeap").getOccupiedByteSize();
+    assertThat(occupiedByteSize, is(984L));
```
GaryWKeim, line 61: Don't you want to use `currentMemoryPostRemove` here?

nnares: Wanted to remove this `currentMemoryPostRemove `variable usage by `occupiedByteSize`, corrected this now
I am simply trying to get the `getOccupiedByteSize()` post `remove()`

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteStatisticTest.java
nnares: redundant assert I put to emphasize `getOccupiedByteSize()` is not 0 (which ideally should be)

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
```diff
@@ -105,6 +105,7 @@ public OnHeapValueHolder<V> compute(final K key, final BiFunction<K, OnHeapValue
   public void clear() {
     // This is faster than performing a clear on the underlying map
     realMap = realMapSupplier.get();
+    byteSize.set(0);
```
chrisdennis, line 108: This is going to be problematic when there are other threads modifying the cache at the same time. If wholesale replacement of the map is our strategy we likely need to scope the byteSize counter with the map itself so that they cannot drift apart.

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
nnares: I don't think there will be any issue in concurrent execution as well.
`byteSize.set(0)` it's a write operation, even the dirty writes will set byteSize to 0 only.

AbfrmBlr: The point here is to make clearing the map and resetting counter as an atomic operation, then other operations modifying cache cannot leave the map and counter in inconsistent state.

nnares: After a discussion With @AbfrmBlr, got the use case when this could be problematic in concurrent execution.
we need to execute below two statements as atomic :
```
    realMap = realMapSupplier.get();
    byteSize.set(0);
```
One Fix that I applied here, wrapping both the statement under a supplier

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
```diff
@@ -122,26 +117,45 @@ public Collection<Map.Entry<K, OnHeapValueHolder<V>>> removeAllWithHash(int hash
...
+    return mapHolder.getRealMap().putIfAbsent(key, valueHolder);
   }
 
   @Override
   public boolean replace(K key, OnHeapValueHolder<V> oldValue, OnHeapValueHolder<V> newValue) {
-    return realMap.replace(key, oldValue, newValue);
+    return mapHolder.getRealMap().replace(key, oldValue, newValue);
   }
+
+  static class MapWithByteSizeHolder<K, V> {
```
chrisdennis, line 143: This the right idea, but you should be able to make the `SimpleBackend` class the holder. That way the clear operation becomes a replacement of the SimpleBackend in the OnHeapStore layer (and the provider moves up a layer, and changes type appropriately).

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
```diff
@@ -28,43 +28,39 @@
...
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiFunction;
-import java.util.function.Supplier;
 
 /**
  * Simple passthrough backend, no key translation
  */
 class SimpleBackend<K, V> implements Backend<K, V> {
 
-  private volatile EvictingConcurrentMap<K, OnHeapValueHolder<V>> realMap;
```
chrisdennis, line 38: This field would stay but become final.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
```diff
@@ -28,43 +28,39 @@
...
-  private final Supplier<EvictingConcurrentMap<K, OnHeapValueHolder<V>>> realMapSupplier;
```
chrisdennis, line 39: This field will move up to the OnHeapStore and become a Supplier of Backend types.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
```diff
@@ -28,43 +28,39 @@
...
-  private final boolean byteSized;
-  private final AtomicLong byteSize = new AtomicLong(0L);
+  private volatile MapWithByteSizeHolder<K, OnHeapValueHolder<V>> mapHolder;
```
chrisdennis, line 37: This won't be needed.

nnares: Yes, Removed it.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
nnares: I have move the map Supplier to level up at - `OnHeapStore.java`

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteSizeTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.ehcache.core.internal.statistics.DefaultStatisticsService;
+import org.ehcache.core.internal.statistics.DefaultTierStatistics;
+import org.junit.Test;
+
+import java.time.Duration;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+public class OnHeapStoreByteSizeTest {
```
chrisdennis, line 20: This test needs an @BeforeClass assumption to prevent it failing under Java 17. You can test this by running:
```
./gradlew -PtestVM=${JAVA_HOME_17} test
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
```diff
@@ -231,11 +228,11 @@ public OnHeapStore(Configuration<K, V> config, TimeSource timeSource, Copier<K>
     this.storeEventDispatcher = eventDispatcher;
 
     if (keyCopier instanceof IdentityCopier) {
-      this.map = new SimpleBackend<>(byteSized, castBackend(backingMapSupplier));
+      this.backendSupplier = () -> new SimpleBackend<>(byteSized);
     } else {
-      this.map = new KeyCopyBackend<>(byteSized, keyCopier, castBackend(backingMapSupplier));
```
chrisdennis, line 236: I think the `castBackend(...)` method should be unused now.

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
nnares: After a discussion got the context, and produce the fix by moving MapSupplier up a level at - `OnHeapStore.java`

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/SimpleBackend.java
nnares: Yes, Initialise it at the attribute level and makes it final rather volatile

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/heap/OnHeapStore.java
nnares: Yes, I have removed it now.

chrisdennis (approved):

chrisdennis (changes requested):
Needs rebasing against the commit that introduced the bug... then we're good to go.

nnares:
opened another PR for daggy fix and get it merged, closing this PR

PR #3072, opened 2022-10-31
Upgrade to Terracotta Platform 5.9.13 (Java 17 Support)

chrisdennis:


clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -126,12 +129,12 @@ public LeasedConnection connect(Properties connectionProperties) throws Connecti
...
           @Override
           public void onNodeRemoval(Cluster cluster, UID stripeUID, Node removedNode) {
-            removedNode.getEndpoints().forEach(e -> servers.remove(e.getAddress()));
+            removedNode.getEndpoints().forEach(e -> servers.remove(e.getHostPort().createInetSocketAddress()));
           }
 
           @Override
           public void onNodeAddition(Cluster cluster, UID addedNodeUID) {
-            servers.add(cluster.determineEndpoint(addedNodeUID, servers).get().getAddress());
+            servers.add(cluster.determineEndpoint(addedNodeUID, servers.stream().map(HostPort::create).collect(toList())).get().getHostPort().createInetSocketAddress());
```
chrisdennis, line 137: I hate this code... I'm going to keep looking at it to see if I can rework the wider area in to something better. Suggestions are welcome.

tmesic99:
No need to add LinuxJava17 and WindowsJava17 checks like I did with Platform?

tmesic99 (approved):

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
mathieucarbou: wdym ? tou hate determineEndpoint or the fact that ehcache is not using HostPort ?

chrisdennis: Sorry... the "I hate this code" was about the code before this change.

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
```diff
@@ -167,24 +191,33 @@ public boolean isValid() {
...
     }
 
-    private List<InetSocketAddress> cloneServers(Iterable<InetSocketAddress> servers) {
-      List<InetSocketAddress> socketAddresses = new ArrayList<>();
-      servers.forEach(socketAddresses::add);
-      return socketAddresses;
+    public void refresh(Cluster cluster) {
+      Set<HostPort> newEndpoints = cluster.determineEndpoints(hostPorts).stream().map(Node.Endpoint::getHostPort).collect(toSet());
+      hostPorts.retainAll(newEndpoints);
+      hostPorts.addAll(newEndpoints);
```
mathieucarbou, line 220: I still prefer this variant, because HostPort is just an address and does not hold the group information that was initially decided by the user. 
So like you said, if in the node definition some nodes happen to have the same addresses both for internal and public, with your version you might lose the group information that was used initially by the user because the host port collection could contain "anything" new matching either public or internal addresses.

```
Collection<Endpoint> endpoints = dynamicTopologyEntity.getUpcomingCluster().determineEndpoints(servers.stream().map(HostPort::create).collect(toList()));

[...]

private static class HostPortSet extends AbstractSet<InetSocketAddress> {

    private final Endpoint initiator; // keep a ref to the group that was used
    private volatile Collection<Endpoint> endpoints;

    private HostPortSet(Collection<Endpoint> endpoints) {
      this.endpoints = endpoints;
      this.initiator = endpoints.iterator().next();
    }

    @Override
    public Iterator<InetSocketAddress> iterator() {
      return endpoints.stream()
        .map(Node.Endpoint::getHostPort)
        .map(HostPort::createInetSocketAddress)
        .iterator();
    }

    @Override
    public int size() {
      return endpoints.size();
    }

    public void refresh(Cluster cluster) {
      this.endpoints = cluster.determineEndpoints(initiator);
    }
  }
```

chrisdennis: I like the direction this is moving in, but aren't we forced (by the initial seeding) to do a blind match to the host names/IPs? It would be nice to be able to retrieve that in some more robust way? That might also allow us to solve the problem of the URI based connection process.

mathieucarbou:
@chrisdennis : maybe we could first merge these daggy in platform and to another release:

- Fixed Cluster#determineEndpoint(UID, Collection)
- Voter fixes regarding CLI parsing and topology used to start voter threads
- And the PR I am preparing to expose the group (daggy too)

(https://github.com/Terracotta-OSS/terracotta-platform/pulls)

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/ConnectionSource.java
mathieucarbou: no: using the initiator here is a trick: only the group string is used and the host/port is disregarded... This is because we didn't want to expose the concept of `EndpointType`... I am going to raise a PR for that in a few minutes... You'll see it will be far better :P

mathieucarbou: Here is the PR: https://github.com/Terracotta-OSS/terracotta-platform/pull/1135

With this the code would become:

```
Collection<HostPort> hp = servers.stream().map(HostPort::create).collect(toList()); // 🤮 
Cluster cluster = dynamicTopologyEntity.getUpcomingCluster();
EndpointType type = cluster.determineEndpointType();

[...]

private static class HostPortSet extends AbstractSet<InetSocketAddress> {

    private final EndpointType type; // keep a ref to the group that was used initially by the user
    private volatile Cluster cluster;

    private HostPortSet(Cluster cluster, EndpointType type) {
      this.type = type;
      this.cluster = cluster;
    }

    @Override
    public Iterator<InetSocketAddress> iterator() {
      return cluster. determineEndpoints(type)
        .stream()
        .map(Node.Endpoint::getHostPort)
        .map(HostPort::createInetSocketAddress)
        .iterator();
    }

    @Override
    public int size() {
      return cluster.getNodeCount();
    }

    public void refresh(Cluster cluster) {
      this.cluster = cluster;
    }
  }
```

mathieucarbou: I think this is best to keep the endpoint type because the public host/port can be set and unset at runtime. So the topology that is refreshed could have new supported endpoint types, and by default, public ones have priority over internal ones.

mathieucarbou (approved):

PR #3075, opened 2022-11-09
[master] Issue 2852

chrisdennis:


ehcache-107/src/main/java/org/ehcache/jsr107/EhcacheCachingProvider.java
```diff
@@ -87,7 +87,9 @@ public CacheManager getCacheManager(URI uri, ClassLoader classLoader, Properties
 
   /**
    * Enables to create a JSR-107 {@link CacheManager} based on the provided Ehcache {@link Configuration}.
-   *
+   * <p>
+   *   This method behaves exactly like {@link #getCacheManager(URI, ClassLoader, Properties)}.
```
jhouserizer, line 91: This is a poor clarification, try again.

jhouserizer (changes requested):

PR #3077, opened 2022-11-14
issue-3034 : fixing byteSize computation for clear() operation

nishchay-naresh:
this is daggy fix over - https://github.com/ehcache/ehcache3/commit/db0a495286e4c918e9044c1bb3cde032066739ac
the actual PR for this fix from the current master - https://github.com/ehcache/ehcache3/pull/3062

### DIFF
--- a/ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteSizeTest.java
+++ b/ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteSizeTest.java
@@ -1,0 +1,106 @@
+package org.ehcache.impl.internal.store.heap.bytesized;
+
+import org.ehcache.Cache;
+import org.ehcache.CacheManager;
+import org.ehcache.config.CacheConfiguration;
+import org.ehcache.config.builders.CacheConfigurationBuilder;
+import org.ehcache.config.builders.CacheManagerBuilder;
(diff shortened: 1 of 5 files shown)


ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/bytesized/OnHeapStoreByteSizeTest.java
```diff
@@ -0,0 +1,106 @@
...
+  @BeforeClass
+  public static void preconditions() {
+    assumeThat(parseInt(getProperty("java.specification.version").split("\\.")[0]), is(lessThan(16)));
+  }
+
+  @Test
+  public void testByteSizePostClearAndRemoveUsingDefaultStatisticsService() {
+
+    DefaultStatisticsService statisticsService = new DefaultStatisticsService();
+    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
```
chrisdennis, line 35: Would be better code-style wise to use try-with-resources blocks for both the CacheManager and the Cache here, e.g.:
```java
try (CacheManager cacheManager = newCacheManagerBuilder().using(statisticsService).build(true)) {
  ...
}
```

chrisdennis (changes requested):

chrisdennis (approved):

chrisdennis:
@nnares if you could follow this up with a PR for the same bug in the `release/3.9` branch that would be great.

PR #3089, opened 2022-11-22
Run check (but not depcheck) to catch checkstyle issues in PRs

akomakom:


akomakom:
@chrisdennis do other branches need this?  Also doesn't seem like daggy is possible.

PR #3091, opened 2022-11-29
Fixes  #1853 for missing threadpool configuration

### LINKED ISSUE #1853: No way to easily figure out where a thread pool configuration is missing
As soon as you have no default thread pool, you need to specify it everywhere. So far, so good.

However, when one is missing, the exception doesn't really obviously say where. Here is an example that will fail. Uncomment `event-dispatch` to make it work. You will see that the exception is far from being useful.

Also, the documentation on thread pools seems to be missing.

```xml
<config xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
           xmlns='http://www.ehcache.org/v3'
           xsi:schemaLocation="http://www.ehcache.org/v3
  http://www.ehcache.org/schema/ehcache-core-3.0.xsd">
    
    <thread-pools>
        <thread-pool alias="defaultDiskPool" min-size="1" max-size="3"/>
    </thread-pools>

    <!--<event-dispatch thread-pool="defaultDiskPool"/>-->

    <cache-template name="default">
        <key-type>java.lang.String</key-type>
        <value-type>java.lang.String</value-type>
        <heap unit="entries">1</heap>
    </cache-template>

    <cache alias="cache" uses-template="default"/>
</config>
```

satyen-chimulkar:


ehcache-impl/src/main/java/org/ehcache/impl/internal/events/CacheEventDispatcherFactoryImpl.java
```diff
@@ -67,11 +67,13 @@ public <K, V> CacheEventDispatcher<K, V> createCacheEventDispatcher(Store<K, V>
     if (config != null) {
       threadPoolAlias = config.getThreadPoolAlias();
     }
-
-    ExecutorService orderedExecutor = executionService.getOrderedExecutor(threadPoolAlias, new LinkedBlockingQueue<>());
-    ExecutorService unOrderedExecutor = executionService.getUnorderedExecutor(threadPoolAlias, new LinkedBlockingQueue<>());
-
-    return new CacheEventDispatcherImpl<>(unOrderedExecutor, orderedExecutor);
+    try {
```
chrisdennis, line 70: This code looks fine... but after discussion with @cljohnso we have some concerns with what this issue is asking for. We're not convinced that the benefits of this behavior change are worth the amount of code change that would result. We're also wondering if the idea of a configuration the lacks a default thread pool makes sense.

Put another way, in the following configuration is the "bug" in:
1. the lack of a defined thread pool for the event listener, or
2. the lack of a default thread pool?
```xml
<config xmlns='http://www.ehcache.org/v3'>
    <thread-pools>
        <thread-pool alias="thread-pool" min-size="1" max-size="3"/>
    </thread-pools>

    <cache alias="cache">
        <key-type>java.lang.String</key-type>
        <value-type>java.lang.String</value-type>
        <heap>1000</heap>
    </cache-template>
</config>
```

My feeling right now is that the lack of a default thread-pool is more likely to be the bug. If that's true, we should be working on highlighting that problem, so that the user fixes that bug.

integration-test/src/test/java/org/ehcache/integration/ThreadPoolsTest.java
```diff
@@ -35,13 +37,13 @@ public void testNoDefaultThreadPoolFails() throws Exception {
     PooledExecutionServiceConfiguration executionServiceConfiguration = new PooledExecutionServiceConfiguration();
     executionServiceConfiguration.addPool("foo", 2, 4);
     CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
-        .using(executionServiceConfiguration)
```
chrisdennis, line 38: Try to avoid disturbing existing indentation unless it's very obviously wrong.

integration-test/src/test/java/org/ehcache/integration/ThreadPoolsTest.java
```diff
@@ -102,4 +104,23 @@ public void testDefaultThreadPoolFailsWithExplicitConfig() throws Exception {
     cacheManager.close();
   }
 
+  @Test
+  public void testNoDefaultThreadPoolFailsWithNoCacheEventDispatcher() throws Exception {
+    PooledExecutionServiceConfiguration executionServiceConfiguration = new PooledExecutionServiceConfiguration();
+    executionServiceConfiguration.addPool("foo", 2, 4);
+    CacheManager cacheManager = CacheManagerBuilder.newCacheManagerBuilder()
```
chrisdennis, line 111: The pre-existing code here probably predates try-with-resources, but that shouldn't stop you using it in new code.

integration-test/src/test/java/org/ehcache/integration/ThreadPoolsTest.java
```diff
@@ -102,4 +104,23 @@ public void testDefaultThreadPoolFailsWithExplicitConfig() throws Exception {
...
+      .using(executionServiceConfiguration)
+      .build(true);
+    try {
+      cacheManager.createCache("testCache",
+        CacheConfigurationBuilder.newCacheConfigurationBuilder(Long.class, String.class, heap(10))
+          .build());
+      fail("expected IllegalStateException");
+    } catch (IllegalStateException ise) {
+      String message = "Exception occurred in creating cache event dispatcher with error message : Null pool alias provided and no default pool configured";
+      assertEquals(message, ise.getCause().getMessage());
```
chrisdennis, line 121: Try to use Hamcrest `assertThat(...)` and the appropriate matchers when possible, it gives much better failure messages.

chrisdennis (commented):
I think we need to have a meeting to discuss the wider issues here. If we get time to do that tomorrow that's great... if not we can try to schedule something on Thursday. I put comments on the test class just for your future reference, as depending on what we decide when we talk the test may not end up being needed.

integration-test/src/test/java/org/ehcache/integration/ThreadPoolsTest.java
satyen-chimulkar: applied Java indentation to file through inteliJ.
May need different code style used by project

ehcache-impl/src/main/java/org/ehcache/impl/internal/executor/PooledExecutionService.java
```diff
@@ -82,7 +81,7 @@ private ThreadPoolExecutor getThreadPoolExecutor(String poolAlias) {
 
     poolAlias = poolAlias == null ? defaultPoolAlias : poolAlias;
     if (poolAlias == null) {
-      throw new IllegalArgumentException("Null pool alias provided and no default pool configured");
+      throw new IllegalArgumentException("Did you miss to configure default pool? Null pool alias provided and no default pool configured.");
```
chrisdennis, line 84: I'm wondering if we can phrase so that we don't start with a question... but also lead with a hint to the default pool? Maybe something like:
```java
throw new IllegalArgumentException("No default pool configured, and no explicit alias provided. Consider configuring a default pool.");
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/executor/PooledExecutionService.java
```diff
@@ -106,7 +105,7 @@ public void start(ServiceProvider<Service> serviceProvider) {
         throw new IllegalStateException("Pool for default pool alias is null");
       }
     } else {
-      LOGGER.warn("No default pool configured, services requiring thread pools must be configured explicitly using named thread pools");
+      LOGGER.warn("No default pool configured, services requiring thread pools must be configured explicitly using named thread pools. This may result in subsequent failure ahead.");
```
chrisdennis, line 108: There's some redundancy in this new sentence ('subsequent' and 'ahead'). Perhaps we could consider folding this statement in with the second clause in the previous sentence:
```java
LOGGER.warn("No default pool configured. Services requiring thread pools will fail if not explicitly configured to use a named pool.");
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/events/CacheEventDispatcherFactoryImpl.java
```diff
@@ -67,11 +67,13 @@ public <K, V> CacheEventDispatcher<K, V> createCacheEventDispatcher(Store<K, V>
     if (config != null) {
       threadPoolAlias = config.getThreadPoolAlias();
     }
-
-    ExecutorService orderedExecutor = executionService.getOrderedExecutor(threadPoolAlias, new LinkedBlockingQueue<>());
```
chrisdennis, line 71: This code can go back as it was now.

integration-test/src/test/java/org/ehcache/integration/ThreadPoolsTest.java
```diff
@@ -102,4 +106,23 @@ public void testDefaultThreadPoolFailsWithExplicitConfig() throws Exception {
     cacheManager.close();
   }
 
+  @Test
+  public void testNoDefaultThreadPoolFailsWithNoCacheEventDispatcher() throws Exception {
```
chrisdennis, line 110: This test could stay and be adjusted to the new message, but it probably makes more sense to remove it. It could become a test of the PooledExecutorService though.

chrisdennis (changes requested):

integration-test/src/test/java/org/ehcache/integration/ThreadPoolsTest.java
satyen-chimulkar: Removed

chrisdennis (approved):

chrisdennis:
Tests need some attention... and we need to consider if this is a "bug fix" and should therefore be reworked to be daggy.

AbfrmBlr:
@satyen-chimulkar  Please raise a daggy pr against master branch.

PR #3101, opened 2022-12-15
fix for issue2742 null pointer exception

ajaymaddi31:
fix for issue2742 null pointer exception

### DIFF
--- a/ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
+++ b/ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
@@ -859,17 +859,17 @@ public abstract class AbstractOffHeapStore<K, V> extends BaseStore<K, V> impleme
     final StoreEventSink<K, V> eventSink = eventDispatcher.eventSink();
 
     try {
-        boolean result = backingMap().computeIfPinned(key, (k, valuePresent) -> {
-          if (valuePresent.getId() == valueFlushed.getId()) {
-            if (valueFlushed.isExpired(timeSource.getTimeMillis())) {
-              onExpiration(k, valuePresent, eventSink);
-              return null;
-            }
-            valuePresent.updateMetadata(valueFlushed);
-            valuePresent.writeBack();
+      boolean result = backingMap().computeIfPinned(key, (k, valuePresent) -> {
+        if (valuePresent.getId() == valueFlushed.getId()) {
+          if (valueFlushed.isExpired(timeSource.getTimeMillis())) {
+            onExpiration(k, valuePresent, eventSink);
+            return null;
           }
-          return valuePresent;
-        }, valuePresent -> valuePresent.getId() == valueFlushed.getId());
+          valuePresent.updateMetadata(valueFlushed);
+          valuePresent.writeBack();
+        }
+        return valuePresent;
+      }, valuePresent -> valuePresent.getId() == valueFlushed.getId());
       eventDispatcher.releaseEventSink(eventSink);
       if (result) {
         flushObserver.end(AuthoritativeTierOperationOutcomes.FlushOutcome.HIT);
--- a/ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
+++ b/ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
@@ -858,7 +858,6 @@ public abstract class AbstractOffHeapStore<K, V> extends BaseStore<K, V> impleme
     flushObserver.begin();
     final StoreEventSink<K, V> eventSink = eventDispatcher.eventSink();
 
-
     try {
         boolean result = backingMap().computeIfPinned(key, (k, valuePresent) -> {
           if (valuePresent.getId() == valueFlushed.getId()) {
--- a/ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
+++ b/ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
@@ -857,19 +857,22 @@ public abstract class AbstractOffHeapStore<K, V> extends BaseStore<K, V> impleme
 
     flushObserver.begin();
     final StoreEventSink<K, V> eventSink = eventDispatcher.eventSink();
+    boolean result = false;
 
     try {
-      boolean result = backingMap().computeIfPinned(key, (k, valuePresent) -> {
-        if (valuePresent.getId() == valueFlushed.getId()) {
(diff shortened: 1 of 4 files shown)


AbfrmBlr:
IMHO the way to look into the problem is to understand the tiering model, and contract of flush method in AuthoritativeTier. In addition to the java docs on interfaces you can read about the whole tiering design here https://github.com/ehcache/ehcache3/blob/master/docs/src/docs/asciidoc/developer/design.tiering.adoc.

ajaymaddi31:
we are getting valueHolder as null in getOrComputeIfAbsent method because it is unable to fetch the value for the key from cache and in the subsequent call to flush method throwing null pointer exception because of this issue.Now we are handling with this null check.After handling the null check we are not getting null pointer issue even after running the test case for 2 hours.

chrisdennis (approved):

chrisdennis (changes requested):
This looks like the correct fix. That said we need to remove the unrelated whitespace changes, and we need to add some test coverage. A unit test for the NoOpCachingTier, and ideally some kind of integration test (maybe using a mocked authoritative tier?) that can reproduce the original issue.

ajaymaddi31:
Unit test case for NoopCachingTier(private static class in TieredStore.java) the method(getOrComputeIfAbsent) in NoopCachingTier is covered in TieredStoreTest.java, CachingTierDoesNotSeeAnyOperationDuringClear covers this method in NoopCachingTier.

Unit test case to Reproduce the NullPointerexception:-
we are getting the null pointer exception in flush method in AbstractOffHeapStore.java class. In AbstractOffHeapStoreTest.java class has one test case named testFlushUpdateAccessStats covers flush method, by passing firstValuHolder and secondValueHolder. I wrote a similar test case(testFlushUpdateAccessStatsForNullPointerException) by passing the parameters(firstValueHolder,secondValueHolder) as null to reproduce the NullPointerException.

chrisdennis:
Yeah, the test here doesn't seem to make sense given the context of the bug and the way the tiered store works. Can you schedule a 30 minute slot some time next week so we can discuss this, and I can explain what's going on with this code, why the null-check is really necessary, and what test needs to be written to reproduce the original NPE, and therefore prove the fix is correct.

ajaymaddi31:
As per the yesterday's discussion added test case to cover the  scenario for not calling flush method if the value is null.

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStoreTest.java
```diff
@@ -241,6 +241,32 @@ public void testFlushUpdatesAccessStats() throws StoreAccessException {
     }
   }
 
+  @Test
+  public void testFlushUpdatesAccessStatsForNullPointerException() throws StoreAccessException {
```
chrisdennis, line 245: This test can be removed.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
```diff
@@ -572,7 +572,10 @@ public NoopCachingTier(final AuthoritativeTier<K, V> authoritativeTier) {
     @Override
     public ValueHolder<V> getOrComputeIfAbsent(final K key, final Function<K, ValueHolder<V>> source) {
       final ValueHolder<V> apply = source.apply(key);
-      authoritativeTier.flush(key, apply);
+      //this null check is to handle the null pointer exception that can be thrown from the flush method
```
chrisdennis, line 575: Can we phrase this more concretely, and maybe without using a negation. Something like:
```java
if (apply != null) {
  //immediately flush any entries faulted from the authority as this tier has no capacity
  authoritativeTier.flush(key, apply);
}
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/offheap/AbstractOffHeapStore.java
```diff
@@ -859,17 +859,17 @@ public boolean flush(K key, final ValueHolder<V> valueFlushed) {
     final StoreEventSink<K, V> eventSink = eventDispatcher.eventSink();
 
     try {
-      boolean result = backingMap().computeIfPinned(key, (k, valuePresent) -> {
```
chrisdennis, line 862: These indentation changes should be removed so that the file stays unmodified.

chrisdennis (changes requested):

ajaymaddi31:
A gentle remainder, Please let us know if there are any review comments?

chrisdennis:
> A gentle reminder, Please let us know if there are any review comments?

https://user-images.githubusercontent.com/954331/212095088-555f7832-e160-4e4f-ba97-15579ce5ab7e.mov

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
```diff
@@ -572,7 +573,11 @@ public NoopCachingTier(final AuthoritativeTier<K, V> authoritativeTier) {
     @Override
     public ValueHolder<V> getOrComputeIfAbsent(final K key, final Function<K, ValueHolder<V>> source) {
       final ValueHolder<V> apply = source.apply(key);
-      authoritativeTier.flush(key, apply);
+      Optional<ValueHolder<V>> ap = Optional.ofNullable(apply);
```
chrisdennis, line 576: Why the optional? It seems like all this does is create an object for no real benefit?

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/tiering/TieredStoreTest.java
```diff
@@ -560,6 +560,33 @@ public void CachingTierDoesNotSeeAnyOperationDuringClear() throws StoreAccessExc
...
+        throw new RuntimeException(e);
+      }
+    });
+
+    t.start();
+    barrier.await();
+    tieredStore.get("foo");
+    barrier.await();
+    t.join();
+    verify(stringAuthoritativeTier,never()).flush("foo",null);
```
chrisdennis, line 587: Can you fix the formatting here to confirm to some reasonable standard. A reasonable version could be:
```java
verify(stringAuthoritativeTier, never()).flush("foo", null);
```

chrisdennis (changes requested):

chrisdennis (changes requested):

chrisdennis:
Before this can be merged the fix will need to be rebased to be a daggy fix against the commit that introduced the bug. I believe that's going to be the commit that introduces the concept of the `NoopCachingTier`. It should probably also be squashed to be a more logical commit stream.

ajaymaddi31:
closing this PR.
#3114  is the replacement for this PR 
https://github.com/ehcache/ehcache3/pull/3114

Issue #2954, opened 2021-10-06
Recover from fatal invalid cache state


cdalexndr:
Ehcache 3.9.5

Stacktrace:
```
Caused by: java.lang.NullPointerException: null
	at org.ehcache.shadow.org.terracotta.offheapstore.disk.storage.FileBackedStorageEngine.findChunk(FileBackedStorageEngine.java:363) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
	at org.ehcache.shadow.org.terracotta.offheapstore.disk.storage.FileBackedStorageEngine.readKeyBuffer(FileBackedStorageEngine.java:264) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
	at org.ehcache.shadow.org.terracotta.offheapstore.storage.PortabilityBasedStorageEngine.readKey(PortabilityBasedStorageEngine.java:119) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
	at org.ehcache.shadow.org.terracotta.offheapstore.OffHeapHashMap$DirectEntry.<init>(OffHeapHashMap.java:1562) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
	at org.ehcache.shadow.org.terracotta.offheapstore.OffHeapHashMap$EntryIterator.create(OffHeapHashMap.java:1540) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
	at org.ehcache.impl.internal.store.disk.factories.EhcachePersistentSegmentFactory$EhcachePersistentSegment$EntrySet$1.create(EhcachePersistentSegmentFactory.java:133) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
	at org.ehcache.impl.internal.store.disk.factories.EhcachePersistentSegmentFactory$EhcachePersistentSegment$EntrySet$1.create(EhcachePersistentSegmentFactory.java:130) ~[ehcache-3.9.5.jar:3.9.5 ad079c948c87e261c36fd97fc51b65d927bbfde8]
[...]

PR #3103, opened 2022-12-23
Enhancement implemented #2954 : Recover from fatal invalid cache state

jitendra-nalwaya:


jhouserizer:
Hi @jitendra-nalwaya ,

Could you please explain the purpose of using two sentinel files ("cleanShutdown" and "unCleanShutdown")?  

What does using two files rather than one provide?  why can't the presence of "unCleanShutdown" indicate non-clean, and the absence of it represent clean?   or why can't the presence of "cleanShutdown" represent clean, and the absence represent non-clean?

Managing two files is non-atomic, and therefore more prone to leaving one behind when it need not be, or etc.?

There may be an advantage to using two files, and if so, can you please explain?

jitendra-nalwaya:
> 
Used two sentinel files to identify last shutdown status in below scenarios -

1) When customer upgrade their environment with this enhancement.
2) Remove directory structure explicitly by customer.
3) Remove directory structure using destroy/destroyall methods.

On startup,
- If both files do not exist then we can't identify last shutdown state, therefore we will create 'unCleanShutdown' file.
- If 'unCleanShutdown' file exist then we will remove directory structure and on shutdown we will delete 'unCleanShutdown' file and create 'cleanShutdown' file.
- If 'cleanShutdown' file exist then we will delete 'cleanShutdown' file and create 'unCleanShutdown' file.

We can use single sentinel file as well if we write some content in it to identify the last shutdown state.

Could you please elaborate below statement -
Managing two files is non-atomic, and therefore more prone to leaving one behind when it need not be, or etc.?

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -131,6 +136,18 @@ public synchronized void stop() {
...
+        if (cleanShutdown.createNewFile()) {
+          LOGGER.debug(".cleanShutdown file is created.");
+          if (unCleanShutdown.exists()) {
+            LOGGER.debug(".unCleanShutdown file is exists.");
+            Files.delete(unCleanShutdown.toPath());
+            LOGGER.debug(".unCleanShutdown file is deleted.");
+          }
+        }
+      } catch (IOException e) {
+        LOGGER.debug("Either '.cleanShutdown' file is not created or '.unCleanShutdown' is not deleted.");
```
chrisdennis, line 149: This would seem to deserve a little higher than debug logging.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -244,4 +261,33 @@ public File getRoot() {
       return safeSpace.directory;
     }
   }
+
+  private void verifyCleanShutdown(final File rootDirectory) {
```
chrisdennis, line 265: This logic appears to do more than just verify, it also deletes if the disk is believed to be unclean.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -65,6 +67,8 @@ public DefaultLocalPersistenceService(final DefaultPersistenceConfiguration pers
       throw new NullPointerException("DefaultPersistenceConfiguration cannot be null");
     }
     lockFile = new File(rootDirectory, ".lock");
+    cleanShutdown = new File(rootDirectory, ".cleanShutdown");
```
chrisdennis, line 70: Trying to understand the need for two files here:

| &darr; unclean.exists() / clean.exists() &rarr; | true | false |
|-|-|-|
|true| clean (no-op) | unclean (delete) |
|false| clean (no-op) | clean (no-op) |

Since there are only two output states... why not just have a single "clean" marker? Look for it on startup (and then delete it), and then create it on shutdown.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -244,4 +261,33 @@ public File getRoot() {
...
+        } else if (cleanShutdown.exists()) {
+          Files.delete(cleanShutdown.toPath());
+          LOGGER.debug("cleanShutdown file is deleted.");
+          if (unCleanShutdown.createNewFile()) {
+            LOGGER.debug("unCleanShutdown file is created.");
+          }
+        } else if (unCleanShutdown.exists()) {
+          LOGGER.debug("Probably unclean shutdown was done.");
+          if (rootDirectory.exists()) {
+            FileUtils.tryRecursiveDelete(rootDirectory);
```
chrisdennis, line 281: Isn't this recursive delete going to trash the lock file? This might break the existing protection against parallel use of the same directory by multiple instances? We should probably only do any of this marker file manipulation *after* we've validated it is safe for us to access (& modify) the directory.

chrisdennis: I'm also thinking that making the delete decision will break any users of the LocalPersistenceService that are fault tolerant? Maybe we need to think of a way of signaling the unclean shutdown to consumers of the service, or move this code to a different layer?

chrisdennis (changes requested):

jitendra-nalwaya:
@chrisdennis 
Creating a single "clean" marker on shutdown and verifying on startup (and then delete it). Please review.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -244,4 +275,38 @@ public File getRoot() {
       return safeSpace.directory;
     }
   }
+
+  private void onStartupIfLockFileExists() {
```
chrisdennis, line 279: This method appears to be duplicating a lot of the logic already in the `internalStart()` method. It seems this could be greatly simplified if you took a more strict "lock first, then check for clean" approach?

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -83,6 +85,27 @@ public synchronized void startForMaintenance(ServiceProvider<? super Maintainabl
   private void internalStart() {
     if (!started) {
       createLocationIfRequiredAndVerify(rootDirectory);
+      if (!clean.exists()) {
+        if (lockFile.exists()) {
+          onStartupIfLockFileExists();
+        } else {
+          LOGGER.debug("lock file is not exists.");
+        }
+        if (tryRecursiveDelete(rootDirectory)) {
```
chrisdennis, line 94: Sorry, I hinted to this in my previous review but probably wasn't explicit enough. The `LocalPersistenceService` implementation cannot be this dogmatic about what the unclean state of a directory means to the consumers of the service. It happens that all the existing OSS users of this service don't tolerate an unclean shutdown. That's not true for all the existing users of this service, and isn't necessarily true for all future OSS users. We'll need to find a way that the `LocalPersistenceService` can communicate the clean/unclean state of the service to consumers so they can make a decision on a case by case basis as to whether their files need to be deleted.

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -104,6 +106,18 @@ private void internalStart() {
       if (lock == null) {
         throw new RuntimeException("Persistence directory already locked by another process: " + rootDirectory.getAbsolutePath());
       }
+
+      if (!clean.exists()) {
+        LOGGER.warn("Probably unclean shutdown was done. Please take appropriate action, if needed.");
```
chrisdennis, line 111: Who do you expect to see this log message, and what do you expect them to do?

jitendra-nalwaya:
Introduce new API isClean() to identify state(normal/abnormal) of stopped service.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -198,6 +223,20 @@ public void destroyAll(String owner) {
     }
   }
 
+  /**
+   * {@inheritDoc}
+   * Abnormally stopped service may lead to data corruption.
+   * Can take appropriate action by identifying state of service stopped.
+   */
+  @Override
+  public final boolean isClean() {
```
chrisdennis, line 232: Services are started in dependency order. The return of this method is only correct before the service is started. Once started the clean file is deleted and the state will look dirty. This means any service which depends on the `LocalPersistenceService` will check the state of the service on start and always see it as dirty.

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
jitendra-nalwaya: @chrisdennis 
This scenario is addressed in isClean() API. Please review.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -61,10 +64,12 @@ public class DefaultLocalPersistenceService implements LocalPersistenceService {
   public DefaultLocalPersistenceService(final DefaultPersistenceConfiguration persistenceConfiguration) {
     if(persistenceConfiguration != null) {
       rootDirectory = persistenceConfiguration.getRootDirectory();
+      isRootDirectoryExists = rootDirectory.exists();
```
chrisdennis, line 67: You really can't be reading state in the constructor. You don't have exclusive access to the directory, and arbitrary amounts of time can elapse (and arbitrary state changes can occur) between this constructor call and the service being started. Consider for example if I create two cache managers, pointing at the same directory, but then initialize the first one, mutate the disk, and then shut it down uncleanly, and then start the second one?

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -48,10 +48,13 @@ public class DefaultLocalPersistenceService implements LocalPersistenceService {
 
   private final File rootDirectory;
   private final File lockFile;
+  private final File clean;
 
   private FileLock lock;
   private RandomAccessFile rw;
   private boolean started;
+  private final boolean isRootDirectoryExists;
+  private boolean isCleanFileDeleted;
```
chrisdennis, line 57: Wouldn't it make more sense to just have a `clean` variable here that you initialize during start based on what you learn about the state of the disk files. Then your `isClean()` method can be little more than a return of this value (maybe with some guarding to make sure it's been set correctly).

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -129,6 +148,14 @@ public synchronized void stop() {
       } catch (IOException e) {
         throw new RuntimeException("Couldn't unlock rootDir: " + rootDirectory.getAbsolutePath(), e);
       }
+
+      try {
```
chrisdennis, line 152: You probably want to be creating the clean file while you still have the file lock to avoid issues with another cache manager starting up while you are shutting down.

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -82,6 +85,9 @@ public synchronized void startForMaintenance(ServiceProvider<? super Maintainabl
 
   private void internalStart() {
     if (!started) {
+      if (!rootDirectory.exists() || (!cleanFile.exists() && !lockFile.exists())) {
```
chrisdennis, line 88: I'm think that second condition should really be that the root directory is empty.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -104,6 +110,18 @@ private void internalStart() {
...
       }
+
+      if (cleanFile.exists()) {
+        try {
+          LOGGER.debug("clean file exists, trying to delete the file.");
+          Files.delete(cleanFile.toPath());
+          clean = true;
+          LOGGER.debug("clean file is deleted.");
+        } catch (IOException e) {
+          LOGGER.warn("clean file was not deleted {}.", cleanFile.getPath());
```
chrisdennis, line 121: I'm thinking this should be a hard failure. If the clean file isn't deleted here then a subsequent dirty start would be seen as clean.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -198,6 +225,39 @@ public void destroyAll(String owner) {
...
+   * {@inheritDoc}
+   * Abnormally stopped service may lead to data corruption.
+   * Can take appropriate action by identifying state of service stopped.
+   */
+  @Override
+  public final synchronized boolean isClean() {
+    if (started) {
+      return clean;
+    }
+    if (rootDirectory.exists()) {
```
chrisdennis, line 238: Nobody should be calling anything on a stopped service (other than start). This should be able to be:
```java
if (started) {
  return clean;
} else {
  throw new IllegalStateException(...);
}
```

chrisdennis (changes requested):

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -115,6 +140,15 @@ private void internalStart() {
   @Override
   public synchronized void stop() {
     if (started) {
+      try {
+        if (cleanFile.createNewFile()) {
+          LOGGER.debug("clean file is created.");
+        } else {
+          LOGGER.warn("clean file already exists.");
```
chrisdennis, line 147: We should probably replace this with a more informative message. Since we're logging this at warn level what should a user do with the information? What's gone/going wrong, and what should they do to fix it?

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -115,6 +140,15 @@ private void internalStart() {
...
+        }
+      } catch (IOException e) {
+        LOGGER.warn("clean file is not created.");
```
chrisdennis, line 150: Same deal here... what does this mean for a user. They don't know what a clean file is... they care about the implications of this message for their usage of the product.

ehcache-core/src/main/java/org/ehcache/core/spi/service/LocalPersistenceService.java
```diff
@@ -75,4 +75,16 @@ interface SafeSpaceIdentifier {
      */
     File getRoot();
   }
+
+  /**
+   * Identify state(normal/abnormal) of stopped service.
+   * @return <tt>true</tt> if service stopped normally.
+   */
+  boolean isClean();
```
chrisdennis, line 83: Javadoc needs polish, and the `@throws` should be moved up to here:
```java
  /**
   * Return the cleanliness of the state stored in this service.
   * <p>
   * Stored state is assumed to be clean if the service detects
   * that the last started instantiation of this service was shutdown
   * successfully.
   *
   * @return {@code true} if the state is clean
   * @throws IllegalStateException if the service is not started
   */
   ```

File: ehcache-impl/src/test/java/org/ehcache/impl/persistence/DefaultLocalPersistenceServiceTest.java [hunk 18]
```diff
@@ -140,4 +142,38 @@ public void testExclusiveLock() throws IOException {
...
+    assertTrue(service.isClean());
+    service.stop();
+  }
+
+  @Test
+  public void testServiceStoppedStatusWhenStoppedUnexpectedly() throws IOException {
+    // Service stopped unexpectedly means directory exists with some data but without .clean file.
+    File f = folder.newFolder("testServiceStoppedStatusWhenStoppedUnexpectedly");
+    final DefaultLocalPersistenceService service = new DefaultLocalPersistenceService(new DefaultPersistenceConfiguration(f));
+    new File(f.getAbsolutePath()+"\\dummy.txt").createNewFile();
```
chrisdennis, line 162: This doesn't work correctly on any platform except Windows. Outside of Windows you'll just get a file named `<blahblahblah>\dummy.txt` because `\` is not the right separator. You should use the `File(File, String)` constructor instead.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -115,6 +140,17 @@ private void internalStart() {
   @Override
   public synchronized void stop() {
     if (started) {
+      try {
+        if (cleanFile.createNewFile()) {
+          LOGGER.debug("clean file is created.");
+        } else {
+          LOGGER.warn("clean file already exists. It's not deleted either user's permission or network issue." +
```
chrisdennis, line 147: This logging is verbose but doesn't really help the user much. What will happen as a result of this error/problem? What should the user do or expect?

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -198,6 +235,30 @@ public void destroyAll(String owner) {
...
+    } else {
+      throw new IllegalStateException("Service is not running");
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public final boolean isServiceStarted() {
```
chrisdennis, line 258: This method can go away.

chrisdennis (changes requested):

jitendra-nalwaya:
@chrisdennis 
I addressed review comments. Please review.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultLocalPersistenceService.java
```diff
@@ -237,10 +239,6 @@ public void destroyAll(String owner) {
 
   /**
    * {@inheritDoc}
```
chrisdennis, line 241: No need to keep an `{@inheritDoc}` around... javadoc will get inherited anyway.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
```diff
@@ -60,6 +60,10 @@ private boolean isStarted() {
   @Override
   public void start(final ServiceProvider<Service> serviceProvider) {
     innerStart(serviceProvider);
+    if (persistenceService != null && !persistenceService.isClean()) {
```
chrisdennis, line 63: We shouldn't need to null-check here. Persistence service should always be non-null because we depend on it.

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
```diff
@@ -73,6 +73,10 @@ public void startForMaintenance(ServiceProvider<? super MaintainableService> ser
   private void innerStart(ServiceProvider<? super MaintainableService> serviceProvider) {
     persistenceService = serviceProvider.getService(LocalPersistenceService.class);
     isStarted = true;
+    if (persistenceService!=null && persistenceService.isServiceStarted() && !persistenceService.isClean()) {
```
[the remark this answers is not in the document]
jitendra-nalwaya: @chrisdennis 
Seven existing test cases are getting failed in DefaultDiskResourceServiceTest.java ("WithoutPersistenceService") if not handling the null check. Please share your thoughts on this. Thanks.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/DefaultDiskResourceService.java
jitendra-nalwaya: @chrisdennis 
Seven existing test cases are getting failed in DefaultDiskResourceServiceTest.java ("WithoutPersistenceService") if not handling the null check. Please share your thoughts on this. Thanks.

chrisdennis: You need to look at the `@ServiceDependency` annotation that is applied to this class, and think about how that annotation is interpreted by the ServiceLocator. Then note that the tests you're looking at are not using the ServiceLocator to provide the services to the DefaultDiskResourceService.

chrisdennis (changes requested):
I would have expected this to consist of oney two commits (after being rebased against latest HEAD of main). First commit would be the removal of the null checks around the local persistence service, and the removal of the related testing. The second would then be the addition of the new code for the safe shutdown detection. This may just be a rebase away from that state, but it's very hard to see that given the current commit stream.

jitendra-nalwaya:
@chrisdennis 
New pull request is created with 2 commits - https://github.com/ehcache/ehcache3/pull/3161

PR #3104, opened 2022-12-28
Fixes #1554 : Fixed by updating response in UserManagedCacheBuilder

### LINKED ISSUE #1554: UserManagedCacheBuilder tries to create serializers that aren't needed
Can be reproduce by launching demo-01 with logging in demo. You will see warnings about serializers that can't be found (you need to open http://localhost:8080 to actually get the log).

This cache manager is using heap storage only and the keys and values of its only cache are stored by reference. So the serializer won't ever be used. There is no point in creating it nor looking for it.

Sandeep184299:


chrisdennis (changes requested):
From what I can see here you've moved the logging from "builder build time" to "builder configuration time", and made it unconditional. It logs regardless of whether there is or isn't a serializer. I'm not sure how this could be considered a "fix" for this issue. The issue references not creating serializers that aren't needed... the logging in just a symptom of that issue.

PR #3105, opened 2023-01-03
[master] Docker files for Ehcache

mathieucarbou:
This PR adds a docker folder in the Ehcache kit with instructions on how to build the 3 images, plus the README to place on Docker Hub.

- ~~`ehcache-server`~~ `ehcache-terracotta-server`
- ~~`terracotta-voter-oss`~~ `ehcache-terracotta-voter`
- ~~`terracotta-config-tool-oss`~~ `ehcache-terracotta-config-tool`

**Notes:**

- I have manually tested them, since it is not yet possible to add docker tests in Ehcache
- Naming still to be discussed
- Gradle / Docker build scripts and deployment to hub.docker.com still to be done in a subsequent PR
- A cherry-pick to the `release/3.9` branch can be made once merged

Some ref regarding these required perm / `uid:0` with OpenShift (also others): https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines

### TO TEST:

```bash
gw assemble
gw distDir
cd clustered/ehcache-clustered/build/kit/
./docker/buildAndTest.groovy
```

### DIFF
--- a/clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
+++ b/clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
@@ -1,0 +1,31 @@
+FROM azul/zulu-openjdk-alpine:17.0.5-17.38.21
+
+RUN apk update && apk upgrade && apk add bash
+RUN mkdir /terracotta && chgrp -R 0 /terracotta && chmod -R g+rX /terracotta
+RUN mkdir /terracotta/config && chgrp -R 0 /terracotta/config && chmod -R g+rX /terracotta/config
+RUN mkdir /terracotta/run && chgrp -R 0 /terracotta/run && chmod -R g+rwX /terracotta/run
+
+WORKDIR /terracotta
+
+# This volume can be mounted with "ro" flag.
+# This is where the configurations will be sourced and eventually copied elsewhere within the container.
+VOLUME "/terracotta/config"
+
+# This folder can be mounted with "rw" flag.
+# This is where the default configurations will output any logs, data, etc
+# If the volume is mounted, the container will support being restarted in the same state.
+# If a volume is mounted, be sure to set the correct chmod permissions in the host machine.
+# Also, you might need to use a special cleanup task to remove the created files in the host machine
+VOLUME "/terracotta/run"
+
+EXPOSE 9410
+
+ENV DEFAULT_OFFHEAP "offheap-1:512MB,offheap-2:512MB"
+ENV DEFAULT_ACTIVATE "false"
+ENV DEFAULT_CLUSTER_NAME "tc-cluster"
+ENV DEFAULT_FAILOVER "availability"
+
+COPY server /terracotta/server
+COPY docker/server/entrypoint.sh /terracotta/server/entrypoint.sh
+
+ENTRYPOINT ["./server/entrypoint.sh"]
--- a/clustered/ehcache-clustered/build.gradle
+++ b/clustered/ehcache-clustered/build.gradle
@@ -158,6 +158,7 @@ distributions {
       }
       into ('') {
         from 'src/assemble'
+        filter(org.apache.tools.ant.filters.ReplaceTokens, tokens: [version: project.version])
       }
     }
   }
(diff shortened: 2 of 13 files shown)


clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM azul/zulu-openjdk-alpine:17-jre-latest
```
chrisdennis, line 1: Should this be a reference to an immutable tag?

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+RUN apk update && apk upgrade && apk add bash
```
chrisdennis, line 3: Do we need bash because of the start server script, or the entrypoint.sh?

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN mkdir /terracotta && chgrp -R 0 /terracotta && chmod -R g+rX /terracotta
```
chrisdennis, line 4: Would it be simpler to mirror the structure of the existing internal Docker images here?

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# This folder can be mounted with "rw" flag.
+# This is where the default configurations will output any logs, data, etc
+# If the volume is mounted, the container will support being restarted in the same state.
+# If a volume is mounted, be sure to set the correct chmod permissions in the host machine.
+# Also, you might need to use a special cleanup task to remove the created files in the host machine
+VOLUME "/terracotta/run"
+
+EXPOSE 9410
+
+ENV DEFAULT_OFFHEAP "offheap-1:512MB,offheap-2:512MB"
```
chrisdennis, line 23: Any reason we need two offheap resources by default in an Ehcache setup?

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+ENV DEFAULT_ACTIVATE "false"
```
chrisdennis, line 24: Would it make sense for this to be true for OSS images?

chrisdennis (changes requested):

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
mathieucarbou: I've asked myself the question and I think that's better to use lastest. Otherwise we are forced to put a process in place to handle the update. Latest also does not harm for a sample file or something deployed OSS I think. Anybody not happy with that could grab the code and rebuild.

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
mathieucarbou: Yes to keep the content in sync as much as possible as what's done in enterprise... If a fix has to be made, easier like that.

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
mathieucarbou: I replicated the enterprise defaults. I prefer to keep them as-is because config tool and sever defaults also have 2 offheaps I think... Better to be consistent across all our maintained images.

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
mathieucarbou: Good question... It would help users I think but... The problem is that this won't be consistent with the use of the docker image with a custom CLI (not through using env vars by tweaking the default CLI). When using CLI, the auto activate flag has to be added to auto activate.

clustered/ehcache-clustered/src/assemble/docker/server/Dockerfile
mathieucarbou: Note: my answer would be different when we can add later a process to generate and deploy docker image to Docker hub ;-) for now for a sample in the kit i don't see any issue.

mathieucarbou:
Updated READMEs to add the necessary header and footer (licensing, etc) required to be deployed on hub.docker.com

mathieucarbou:
- Changed image names to:

    terracotta/ehcache-terracotta-server
    terracotta/ehcache-terracotta-config-tool
    terracotta/ehcache-terracotta-voter

- Updated `./docker/buildAndTest.groovy` to include an Ehcache put/get test
- Fixed base images to `azul/zulu-openjdk-alpine:17.0.5-17.38.21`
- Added resource filtering in Groovy build script to generate docker files with the correct project version

jhouserizer (approved):

PR #3106, opened 2023-01-04
issue-2283 : fix against of ConnectionClosedException

nnares:


jhouserizer:
The fix looks reasonable to me with surface review. Is there a test case that proves that it eliminates the problem?

PR #3108, opened 2023-01-05
issue - 1785 : Fix for the issue "FileBasedStateRepository uses the w…

vaibhavsrivastava07:
CustomObjectInputStream & CustomLoaderFileBasedObjectInputStream classes introduced to hanle the order of classloader for execution. The flow will first look for TCCL, if class is not loaded then it will look for Cachemanager classloader if found then it will load otherwise it will look for the latest user defined classloader. Loadmaps method is loaded lazily when requested for persistent context. This will allow us to use the classloaders provided.


ehcache-api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
```diff
@@ -39,7 +41,7 @@ public interface StateRepository {
    * @return a state holder
    */
   @Deprecated
-  default <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass) {
+  default <K extends Serializable, V extends Serializable> StateHolder<K, V> getPersistentStateHolder(String name, Class<K> keyClass, Class<V> valueClass) throws CachePersistenceException {
```
AbfrmBlr, line 44: Why we need a change in public api? Is this really required?

ehcache-impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
```diff
@@ -80,8 +85,12 @@ public static <T> Class<? extends Serializer<T>> asTypedSerializer() {
 
   @Override
   public void init(final StateRepository stateRepository) {
-    this.persistentState = stateRepository.getPersistentStateHolder("CompactJavaSerializer-ObjectStreamClassIndex", Integer.class, ObjectStreamClass.class, c -> true, null);
-    refreshMappingsFromStateRepository();
+    try {
+      this.persistentState = stateRepository.getPersistentStateHolder("CompactJavaSerializer-ObjectStreamClassIndex", Integer.class, ObjectStreamClass.class, c -> true, null);
+      refreshMappingsFromStateRepository();
+    }catch (CachePersistenceException e){
```
AbfrmBlr, line 91: Incorrect formatting.

AbfrmBlr, line 91: Not sure I understand this change here, the init throws exception and we eat it up without letting user know the init has failed. Care to elaborate?

ehcache-impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
```diff
@@ -100,8 +99,14 @@ public <K extends Serializable, V extends Serializable> StateHolder<K, V> getPer
                                                                                                      Class<K> keyClass,
                                                                                                      Class<V> valueClass,
                                                                                                      Predicate<Class<?>> isClassPermitted,
-                                                                                                     ClassLoader classLoader) {
-    // isClassPermitted and  classLoader are ignored because this state repository has already being read from file and cached in
+                                                                                                     ClassLoader classLoader) throws CachePersistenceException {
+    // Lazy loading of loadMaps using class loader in order TCCL, CacheManager CL, Custom Classloader
```
AbfrmBlr, line 103: What's the intent behind this change of moving the `loadMaps` call from constructor to this method? Why are we loading the map at every call of `getPersistentHolder`? What's the point of loading in the map, if we are calling it every time?

AbfrmBlr:
The logic in `CustomLoaderFileBasedObjectInputStream.java` looks up for the class in user provided classloader before trying to load it from CacheManager's classloader? Am I missing something @vaibhavsrivastava07 ?

ehcache-api/src/main/java/org/ehcache/spi/persistence/StateRepository.java
vaibhavsrivastava07: We can avoid changing public API by removing CachePersistenceException and re-throwing as RuntimeException in FileBasedStateRepository:getPersistentStateHolder() following the ehcahe current coding practice.

ehcache-impl/src/test/java/org/ehcache/impl/serialization/CompactJavaSerializerTest.java
```diff
@@ -42,8 +43,11 @@ public class CompactJavaSerializerTest {
   public void testStateHolderFailureRereadBehavior() throws ClassNotFoundException {
     StateHolder<Integer, ObjectStreamClass> stateMap = spy(new TransientStateHolder<>());
     StateRepository stateRepository = mock(StateRepository.class);
-    when(stateRepository.getPersistentStateHolder(eq("CompactJavaSerializer-ObjectStreamClassIndex"), eq(Integer.class), eq(ObjectStreamClass.class), any(), any())).thenReturn(stateMap);
-
+    try {
+      when(stateRepository.getPersistentStateHolder(eq("CompactJavaSerializer-ObjectStreamClassIndex"), eq(Integer.class), eq(ObjectStreamClass.class), any(), any())).thenReturn(stateMap);
+    }catch (CachePersistenceException e){
```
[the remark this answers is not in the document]
vaibhavsrivastava07: The exception re-throwing as a RuntimeException in FileBasedStateRepository:getPersistentStateHolder(), this change will be reverted and formatting will be corrected.

ehcache-impl/src/main/java/org/ehcache/impl/serialization/CompactJavaSerializer.java
vaibhavsrivastava07: The exception is re-thrown as a RuntimeException in FileBasedStateRepository:getPersistentStateHolder(), hence this change will be reverted and formatting will be corrected.

vaibhavsrivastava07: Explanation same as above comment. We will be reverting this change.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/CustomObjectInputStream.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.ehcache.impl.persistence;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InvalidClassException;
+import java.io.ObjectStreamClass;
+import java.util.function.Predicate;
+
+public class CustomObjectInputStream extends CustomLoaderFileBasedObjectInputStream {
```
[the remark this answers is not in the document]
vaibhavsrivastava07: Yes, we need this to decide the order of class loader used for class loading.
**Scenario 1 :**
    class loader passed as null --> 
        1. TCCL - It will look for TCCL
        2. Cache Manager - If not loaded from TCCL it will look into Cache Manager to load the classes
        3. latestUserDefinedLoader - If not loaded from Cache Manager it will use the Latest user defined(sun.misc.VM.latestUserDefinedLoader())

**Scenario 2 :** 
    Passing Custom class loader--> 
        1. Custom class loader - It will look for Custom class loader
        2. Cache Manager - If not loaded from Custom class loader it will look into Cache Manager to load the classes
        3. latestUserDefinedLoader - If not loaded from Cache Manager it will use the Latest user defined(sun.misc.VM.latestUserDefinedLoader())

ehcache-impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
vaibhavsrivastava07: To enable the Lazy loading and implement class loader ordering and passing custom class loader as parameter for de-serialization we have moved this method to getPersistentHolder(). 
getPersistentHolder() has the parameter class loader and to use this, we have moved loadMaps call from constructor to getPersistentHolder().

vaibhavsrivastava07:
> The logic in `CustomLoaderFileBasedObjectInputStream.java` looks up for the class in user provided classloader before trying to load it from CacheManager's classloader? Am I missing something @vaibhavsrivastava07 ?



> The logic in `CustomLoaderFileBasedObjectInputStream.java` looks up for the class in user provided classloader before trying to load it from CacheManager's classloader? Am I missing something @vaibhavsrivastava07 ?
Explained in above comments for the class loading order.

vaibhavsrivastava07:
Explained in above comments for class loading ordering.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/CustomObjectInputStream.java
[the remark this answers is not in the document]
AbfrmBlr: No I don't see any logic of classloading in this class, it is just doing the filtering as whether the class being loaded is permitted or not.
The class resolution is only being done in its super class, so the question remains - DO we need this kind if filtering for Filebased state repository.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
AbfrmBlr: There is nothing "lazy" about loading the user provided classes here. All these operations are called during the init phase of `Store.Provider` initialize the serializers. Please look up the call stack of this method and also you read more about Serializers in user doc https://www.ehcache.org/documentation/3.10/serializers-copiers.html. Effectively all the user provided classes which are Keys & Values types used in cache are loaded when the Cache is initialized.

AbfrmBlr:
Please add tests for respective change

AbfrmBlr:
To understand this issue better, I would recommend to implement a test case in which a disk persistent Cache uses both key & value which are user defined classes. This issue wants to address the issue of classloading of user defined classes.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/FileBasedStateRepository.java
AbfrmBlr: Ignore my previous comment, delaying loading the classes till the getPersistentHolder is the right thing to do.

ehcache-impl/src/main/java/org/ehcache/impl/persistence/CustomObjectInputStream.java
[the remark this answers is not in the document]
vaibhavsrivastava07: This kind predicate filtering is not required as class loading order logic is in parent class CustomLoaderFileBasedObjectInputStream. Hence We will remove CustomObjectInputStream class.

vaibhavsrivastava07:
We will add respective test case for disk persistent Cache uses both key & value which are user defined classes.

AbfrmBlr:
The added tests don't fail in absence of the fix, which means it does not reproduce the problem at the hand and is unrelated to the issue.

ehcache-impl/build.gradle
```diff
@@ -50,6 +50,7 @@ task slowTest(type: Test) {
 
 dependencies {
   api project(':ehcache-core')
+  implementation project(':clustered:ehcache-common')
```
AbfrmBlr, line 53: clustered modules should not be added as a dependency in ehcache impl

AbfrmBlr: Please look into possibility of extracting the files to a "shared" module which becomes a dependency for both ehcache-impl & clustered. That would also serve the intent of reusing the existing code.

AbfrmBlr:
As discussed in the yesterday's call , make sure to include only relevant changes in the PR.

clustered/ehcache-common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
```diff
@@ -17,17 +17,13 @@
 package org.ehcache.clustered.common.internal.store;
 
 import org.ehcache.clustered.common.internal.util.ByteBufferInputStream;
+import org.ehcache.impl.store.FilteredObjectInputStream;
```
AbfrmBlr, line 20: why this change?

clustered/ehcache-common/build.gradle
```diff
@@ -28,6 +28,7 @@ publishing.publications.withType(MavenPublication) {
 dependencies {
   api project(':ehcache-api')
   api project(':clustered:ehcache-common-api')
+  api project(':ehcache-impl')
```
AbfrmBlr, line 31: Similarly, why this one?

clustered/ehcache-common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
amitbharate: Thanks for the review. After moving FilteredObjectInputStream from clustered-common to ehcache-impl it requires a import and as currently there is no dependency of ehcache-impl in clustered common we added these 2 change. Please suggest.

amitbharate:
> As discussed in the yesterday's call , make sure to include only relevant changes in the PR.

Thanks Abhilash. We are working on cleaning up remaining 2 unwanted commits.

clustered/ehcache-common/src/main/java/org/ehcache/clustered/common/internal/store/Util.java
vaibhavsrivastava07: > 

Unwanted commits has been removed.

vaibhavsrivastava07:
> As discussed in the yesterday's call , make sure to include only relevant changes in the PR.

Unwanted commits has been removed.

clustered/ehcache-client/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
   implementation "org.terracotta:lease-api:$terracottaPlatformVersion"
   implementation "org.terracotta.dynamic-config.entities:dynamic-config-topology-entity-client:$terracottaPlatformVersion"
   implementation "org.terracotta:connection-api:$terracottaApisVersion"
+  implementation group: 'org.terracotta', name: 'terracotta-utilities-tools', version: parent.terracottaUtilitiesVersion
```
AbfrmBlr, line 39: It will be nice to maintain consistency while adding dependencies using format 
`implemenation "org.terracotta:terracotta-utilities-tools:$terracottaUtilitiesVersion`

clustered/ehcache-common/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
 
   implementation "org.terracotta:entity-common-api:$terracottaApisVersion"
   implementation "org.terracotta:runnel:$terracottaPlatformVersion"
+  implementation group: 'org.terracotta', name: 'terracotta-utilities-tools', version: parent.terracottaUtilitiesVersion
```
AbfrmBlr, line 34: Similarly here as well

PR #3109, opened 2023-01-10
issue-2494 : fix cache visibility for javaxCacheManager

nishchay-naresh:
Fixing cache visibility at JSR107 layer post unwrap()


ehcache-107/src/main/java/org/ehcache/jsr107/Eh107CacheManager.java
```diff
@@ -82,39 +79,41 @@ class Eh107CacheManager implements CacheManager {
     this.configurationMerger = configurationMerger;
     this.statisticsService = jsr107Service.getStatistics();
 
-    refreshAllCaches();
   }
 
   private void refreshAllCaches() {
```
AbfrmBlr, line 84: Why do we still need this method?

Issue #2742, opened 2020-01-30
NullPointerException in disk store under high load on clear()


rschuetz:
We're currently still using EhCache2 (2.10.6) that deadlocks under high load when removeAll() gets called (SSCCE can be provided if needed). I ran the same test against EhCache3 (3.8.1) in order to check whether we can upgrade. Fortunately, there seems to be no deadlock anymore but there are occasionally NPEs:

```
java.lang.NullPointerException
    at org.ehcache.impl.internal.store.offheap.AbstractOffHeapStore.lambda$flush$16(AbstractOffHeapStore.java:863)
    at org.ehcache.impl.internal.store.disk.EhcachePersistentConcurrentOffHeapClockCache.lambda$computeIfPinned$3(EhcachePersistentConcurrentOffHeapClockCache.java:213)
    at org.ehcache.shadow.org.terracotta.offheapstore.OffHeapHashMap.computeIfPresentWithMetadata(OffHeapHashMap.java:2108)
    at org.ehcache.shadow.org.terracotta.offheapstore.AbstractLockedOffHeapHashMap.computeIfPresentWithMetadata(AbstractLockedOffHeapHashMap.java:542)
    at org.ehcache.shadow.org.terracotta.offheapstore.concurrent.AbstractConcurrentOffHeapMap.computeIfPresentWithMetadata(AbstractConcurrentOffHeapMap.java:781)
    at org.ehcache.impl.internal.store.disk.EhcachePersistentConcurrentOffHeapClockCache.computeIfPinned(EhcachePersistentConcurrentOffHeapClockCache.java:210)
    at org.ehcache.impl.internal.store.offheap.AbstractOffHeapStore.flush(AbstractOffHeapStore.java:862)
    at org.ehcache.impl.internal.store.tiering.TieredStore$NoopCachingTier.getOrComputeIfAbsent(TieredStore.java:552)
    at org.ehcache.impl.internal.store.tiering.TieredStore.get(TieredStore.java:87)
    at org.ehcache.core.Ehcache.doGet(Ehcache.java:90)
    at org.ehcache.core.EhcacheBase.get(EhcacheBase.java:127)
    at TestDeadlockLowLevelEhCache3Minimal$ThreadRunner$GetThread.run(TestDeadlockLowLevelEhCache3Minimal.java:151)
```
[...]

PR #3114, opened 2023-01-19
Fix for issue#2742 : NullPointerException in disk store under high load on clear()

ajaymaddi31:
Fix for issue #2742  : NullPointerException in disk store under high load on clear()


chrisdennis (changes requested):
The changes look good... and I think the commit structure is legitimate. The merge commit needs to have a better commit message. It really needs to identify that it is a merge commit, and what it merges in to where, and why. So if you were merging `main` in to the fix branch then it would say that, if you merged the fix branch in to main (aka an integration branch) then it would mention that. Leaving the commit message as is would be confusing for anyone browsing the commit history in the future.

jhouserizer:
>  Leaving the commit message as is would be confusing for anyone browsing the commit history in the future.

I agree, because funny enough, it took me some effort to figure out what the difference was between the two commits.

chrisdennis (changes requested):
The changes are good, the commit structure is good, but the commit message on the merge commit is wrong. It doesn't merge f2c76ad85b41009800d736e755498e396e1559f3 into issue-2742 it appears it merges b7ffc81f97157e4294c8bca1e7298c2b6f865a23 in to the branch (or maybe the other way round, hard to tell from the GitHub UI).

chrisdennis (approved):

PR #3122, opened 2023-02-03
Fix for the issue#1554: UserManagedCacheBuilder tries to create serialize…

ajaymaddi31:
fix for ehcache#1554 : modified the catch block while creating serializers

### DIFF
--- a/ehcache-impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
+++ b/ehcache-impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
@@ -271,11 +271,12 @@ public class UserManagedCacheBuilder<K, V, T extends UserManagedCache<K, V>> imp
             valueSerializer = valueSer;
           }
         } catch (UnsupportedTypeException e) {
-          if (resources.contains(OFFHEAP) || resources.contains(DISK)) {
-            throw new RuntimeException(e);
-          } else {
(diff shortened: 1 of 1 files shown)


ehcache-impl/src/main/java/org/ehcache/config/builders/UserManagedCacheBuilder.java
```diff
@@ -271,11 +271,12 @@ public void close() throws Exception {
...
-          if (resources.contains(OFFHEAP) || resources.contains(DISK)) {
-            throw new RuntimeException(e);
-          } else {
-            LOGGER.debug("Serializers for cache '{}' failed creation ({}). However, depending on the configuration, they might not be needed", id, e.getMessage());
+          for (ResourceType<?> resource : resources) {
+            if (resource.requiresSerialization()) {
+              throw new RuntimeException(e);
+            }
           }
+          LOGGER.info("Serializers for cache '{}' failed creation ({}). However, depending on the configuration, they might not be needed", id, e.getMessage());
```
chrisdennis, line 279: We should probably leave this at debug to match `EhcacheManager`.

chrisdennis (changes requested):

chrisdennis (approved):

chrisdennis:
This is good but needs rebasing against latest main.

chrisdennis (approved):

PR #3127, opened 2023-02-08
issue-628 Handled failures which might occur during write-behind

hchavdapsl:
Handle the failure which may occur during following write-behind scenarios and removed failed entries from cache.
- Batching
- NonBatching

### DIFF
--- a/ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
+++ b/ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
@@ -22,6 +22,7 @@ import org.ehcache.impl.internal.loaderwriter.writebehind.operations.DeleteAllOp
 import org.ehcache.impl.internal.loaderwriter.writebehind.operations.SingleOperation;
 import org.ehcache.impl.internal.loaderwriter.writebehind.operations.WriteOperation;
 import org.ehcache.impl.internal.loaderwriter.writebehind.operations.WriteAllOperation;
+import org.ehcache.spi.loaderwriter.BulkCacheWritingException;
 import org.ehcache.spi.loaderwriter.CacheLoaderWriter;
 import org.ehcache.spi.loaderwriter.WriteBehindConfiguration;
 import org.ehcache.spi.loaderwriter.WriteBehindConfiguration.BatchingConfiguration;
@@ -45,6 +46,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.function.Consumer;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.ehcache.impl.internal.executor.ExecutorUtil.shutdown;
@@ -72,8 +74,8 @@ public class BatchingLocalHeapWriteBehindQueue<K, V> extends AbstractWriteBehind
   private final boolean coalescing;
 
   private volatile Batch openBatch;
-
-  public BatchingLocalHeapWriteBehindQueue(ExecutionService executionService, String defaultThreadPool, WriteBehindConfiguration<?> config, CacheLoaderWriter<K, V> cacheLoaderWriter) {
+  private final Consumer<K> keyCleanUpMethod;
+  public BatchingLocalHeapWriteBehindQueue(Consumer<K> keyCleanUpMethod, ExecutionService executionService, String defaultThreadPool, WriteBehindConfiguration<?> config, CacheLoaderWriter<K, V> cacheLoaderWriter) {
     super(cacheLoaderWriter);
     this.cacheLoaderWriter = cacheLoaderWriter;
     BatchingConfiguration batchingConfig = config.getBatchingConfiguration();
@@ -91,6 +93,7 @@ public class BatchingLocalHeapWriteBehindQueue<K, V> extends AbstractWriteBehind
     } else {
       this.scheduledExecutor = executionService.getScheduledExecutor(config.getThreadPoolAlias());
     }
+    this.keyCleanUpMethod = keyCleanUpMethod;
   }
 
   @Override
@@ -188,25 +191,30 @@ public class BatchingLocalHeapWriteBehindQueue<K, V> extends AbstractWriteBehind
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public void run() {
-      try {
-        List<BatchOperation<K, V>> batches = createMonomorphicBatches(operations());
-        // execute the batch operations
-        for (BatchOperation<K, V> batch : batches) {
+      List<BatchOperation<K, V>> batches = createMonomorphicBatches(operations());
+      // execute the batch operations
+      for (BatchOperation<K, V> batch : batches) {
+        try {
           try {
             batch.performOperation(cacheLoaderWriter);
-          } catch (Exception e) {
-            LOGGER.warn("Exception while bulk processing in write behind queue", e);
+          } finally {
+            try {
+              for (K key : batch.getKeys()) {
+                latest.remove(key);
+              }
+            } finally {
+              LOGGER.debug("Cancelling batch expiry task");
+              expireTask.cancel(false);
+            }
           }
         }
-      } finally {
-        try {
-          for (SingleOperation<K, V> op : operations()) {
-            latest.remove(op.getKey(), op);
+        catch (Exception ex) {
+          for (K key : batch.getKeys()) {
+            keyCleanUpMethod.accept(key);
           }
-        } finally {
-          LOGGER.debug("Cancelling batch expiry task");
-          expireTask.cancel(false);
+          LOGGER.warn("Exception while bulk processing in write behind queue", ex);
         }
       }
     }
--- a/ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/NonBatchingLocalHeapWriteBehindQueue.java
+++ b/ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/NonBatchingLocalHeapWriteBehindQueue.java
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.function.Consumer;
 
 import static org.ehcache.impl.internal.executor.ExecutorUtil.shutdown;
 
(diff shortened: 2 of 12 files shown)


clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/loaderwriter/writebehind/ClusteredWriteBehind.java
```diff
@@ -76,7 +80,11 @@ private void process(long hash) {
               try {
                 if (result != null) {
                   if (result != currentState.get(key) && !(operation instanceof PutOperation)) {
-                    cacheLoaderWriter.write(result.getKey(), result.getValue());
+                    try {
+                      cacheLoaderWriter.write(result.getKey(), result.getValue());
+                    } catch (Exception e) {
+                      keyCleanUpMethod.accept(key);
```
chrisdennis, line 86: I would be very uncomfortable making this change without lots of test coverage to back it up. You should be terrified. The clustered store code is pretty different to the unclustered, and I'm really not convinced this is safe. Worst case scenario here is probably some kind of deadlock...

clustered/ehcache-client/src/test/java/org/ehcache/clustered/client/internal/loaderwriter/writebehind/ClusteredWriteBehindTest.java
```diff
@@ -60,6 +61,9 @@
 public class ClusteredWriteBehindTest {
 
   private static final TimeSource TIME_SOURCE = SystemTimeSource.INSTANCE;
+  private static <T> Consumer<T> NOOP() {
```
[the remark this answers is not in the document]
chrisdennis: Also... even if this was a good approach... why a method that returns a static thing? Why not a static variable that holds a singleton?

ehcache-api/src/main/java/org/ehcache/spi/loaderwriter/CacheLoaderWriter.java
```diff
@@ -113,8 +115,18 @@ default Map<K, V> loadAll(Iterable<? extends K> keys) throws BulkCacheLoadingExc
    * @throws Exception in case no values could be written
    */
   default void writeAll(Iterable<? extends Map.Entry<? extends K, ? extends V>> entries) throws BulkCacheWritingException, Exception {
+    Set<K> successes = new HashSet<>();
```
chrisdennis, line 118: There's nothing wrong with this part of the changeset... but I don't think it belongs in this PR. Better to leave this to another time, all this does is confuse the purpose here (and its presence might be causing you to miss other issues with the real code).

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -188,13 +193,19 @@ public boolean add(SingleOperation<K, V> operation) {
...
+    @SuppressWarnings("unchecked")
     public void run() {
       try {
         List<BatchOperation<K, V>> batches = createMonomorphicBatches(operations());
         // execute the batch operations
         for (BatchOperation<K, V> batch : batches) {
           try {
             batch.performOperation(cacheLoaderWriter);
+          }  catch (BulkCacheWritingException e) {
+            Iterator<K> it = (Iterator<K>) e.getFailures().keySet().iterator();
```
chrisdennis, line 205: Always write the simplest code that you can... no need to challenge the runtime more than necessary:
```java
for (Object key : e.getFailures().keySet()) {
  keyCleanUpMethod.accept((K) key);
}
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -188,13 +193,19 @@ public boolean add(SingleOperation<K, V> operation) {
...
+            while(it.hasNext()) {
+              keyCleanUpMethod.accept(it.next());
+            }
           } catch (Exception e) {
```
chrisdennis, line 209: You need to cleanup on the regular exception path too. It's legal for a writer to throw any exception and expect cleanup to be done.

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
chrisdennis: You also need to think about this generic cast... will there be a concrete cast done somwhere down the line? Where will that ClassCastException go? Do we need to defend against? Good test coverage might be able to protect you here.

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/NonBatchingLocalHeapWriteBehindQueue.java
```diff
@@ -62,12 +67,18 @@ protected SingleOperation<K, V> getOperation(K key) {
...
 
   @Override
+  @SuppressWarnings("unchecked")
   protected void addOperation(final SingleOperation<K, V> operation) {
     latest.put(operation.getKey(), operation);
 
     submit(() -> {
       try {
         operation.performOperation(cacheLoaderWriter);
+      }  catch (BulkCacheWritingException e) {
```
chrisdennis, line 77: This operation only does one write, so why are we special casing BulkCacheWritingException?

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/NonBatchingLocalHeapWriteBehindQueue.java
```diff
@@ -62,12 +67,18 @@ protected SingleOperation<K, V> getOperation(K key) {
...
+        Iterator<K> it = (Iterator<K>) e.getFailures().keySet().iterator();
+        while (it.hasNext()) {
+          keyCleanUpMethod.accept(it.next());
+        }
       } catch (Exception e) {
         LOGGER.warn("Exception while processing key '{}' write behind queue : {}", operation.getKey(), e);
```
chrisdennis, line 83: There should be cleanup logic here.

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/operations/WriteOperation.java
```diff
@@ -51,7 +57,16 @@ public WriteOperation(K k, V v, long creationTime) {
 
   @Override
   public void performOperation(CacheLoaderWriter<K, V> cacheWriter) throws Exception {
-    cacheWriter.write(key, value);
+    Map<K, Exception> failures = new HashMap<>();
```
chrisdennis, line 60: Why are we doing this conversion? It makes no sense. The thing only writes on key... creating a BulkCacheWritingException is just churning heap for no reason.

ehcache-impl/src/test/java/org/ehcache/impl/internal/loaderwriter/writebehind/WriteBehindProviderFactoryTest.java
```diff
@@ -44,6 +45,10 @@
  */
 public class WriteBehindProviderFactoryTest {
 
+  public static <T> Consumer<T> NOOP() {
```
chrisdennis, line 48: Don't default to no-ops. You're weakening the test. If you don't expect this to be called then assert here... if you do then assert that this is called, or assert on it's effects somehow.

chrisdennis (changes requested):
This needs test coverage to prove it all works. Doing that should also expose some of the existing issues that need fixing.

ehcache-api/src/main/java/org/ehcache/spi/loaderwriter/CacheLoaderWriter.java
chrisdennis: I still think this change should be considered separately from this PR.

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/operations/DeleteAllOperation.java
```diff
@@ -45,4 +48,13 @@ public void performOperation(CacheLoaderWriter<K, V> cacheLoaderWriter) throws B
     cacheLoaderWriter.deleteAll(entries);
   }
 
+  @Override
+  public Set<K> getKeys() {
```
chrisdennis, line 52: I wonder if this should be typed as `Iterable<K>` so that we can implement it as a wrapper on the `entries` iterable?

chrisdennis (changes requested):
We still need test coverage here to show that the correct clean up is happening.

hchavdapsl:
Since dead lock was happening in earlier solution for batching write-behind, changed to place to run the cleanup for failures.

chrisdennis (changes requested):
The tests here are still failing, but they do it sporadically. You should go look at the `AbstractWriteBehind` class and the implications of the implementation of that class and the order in which the cache and the `latest` map held in both of the concrete write behind implementation are cleaned up.

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/NonBatchingLocalHeapWriteBehindQueue.java
```diff
@@ -53,22 +55,25 @@ public NonBatchingLocalHeapWriteBehindQueue(ExecutionService executionService, S
...
   @Override
+  @SuppressWarnings("unchecked")
   protected void addOperation(final SingleOperation<K, V> operation) {
     latest.put(operation.getKey(), operation);
 
     submit(() -> {
       try {
         operation.performOperation(cacheLoaderWriter);
       } catch (Exception e) {
+        latest.remove(operation.getKey(), operation);
```
chrisdennis, line 75: Wouldn't this be better represented by something like the following:
```java
      try {
        try {
          operation.performOperation(cacheLoaderWriter);
        } finally {
          latest.remove(operation.getKey(), operation);
        }
      } catch (Exception e) {
        keyCleanUpMethod.accept(operation.getKey());
        LOGGER.warn("Exception while processing key '{}' write behind queue : {}", operation.getKey(), e);
      }
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
```diff
@@ -188,14 +191,23 @@ public boolean add(SingleOperation<K, V> operation) {
...
     public void run() {
       try {
         List<BatchOperation<K, V>> batches = createMonomorphicBatches(operations());
         // execute the batch operations
         for (BatchOperation<K, V> batch : batches) {
           try {
             batch.performOperation(cacheLoaderWriter);
+          } catch (BulkCacheWritingException e) {
+            for (Object key : e.getFailures().keySet()) {
+              keyCleanUpMethod.accept((K) key);
```
chrisdennis, line 204: Doesn't the same problem with the ordering of the cleanup call and the `latest.remove(...)` call happen in the batching queue too?

chrisdennis (changes requested):

ehcache-impl/src/main/java/org/ehcache/impl/internal/loaderwriter/writebehind/BatchingLocalHeapWriteBehindQueue.java
hchavdapsl: Hi Chris,

I have incorporated the review comments.

chrisdennis (changes requested):

ehcache-api/src/main/java/org/ehcache/spi/loaderwriter/CacheLoaderWriter.java
```diff
@@ -113,8 +114,16 @@ default Map<K, V> loadAll(Iterable<? extends K> keys) throws BulkCacheLoadingExc
    * @throws Exception in case no values could be written
    */
   default void writeAll(Iterable<? extends Map.Entry<? extends K, ? extends V>> entries) throws BulkCacheWritingException, Exception {
+    Map<K, Exception> failures = new HashMap<>();
```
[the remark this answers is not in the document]
hchavdapsl: Restored to original.

chrisdennis (changes requested):
This needs rebasing against latest HEAD of main and the subsequent commits should probably be squashed to form a more compact and logical commit stream (potentially down to a single commit).

hchavdapsl:
Re-based with master

chrisdennis (approved):

PR #3130, opened 2023-02-23
issue-3097 : re-writting getAll() impl

nnares:


nnares:
Proposed changes are in early stage. Opening the PR to have effective feedback
PS - Not utilizing the mapping function while fetching the mappings. @Ignoring all the testing mappingFunction based fetch

chrisdennis (commented):
Test changes look like they're moving in the right direction. Lets chat tomorrow about how the tests look now.

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -554,6 +554,11 @@ public Map<K, ValueHolder<V>> bulkComputeIfAbsent(final Set<? extends K> keys, f
     }
   }
 
+  @Override
+  public Iterable<? extends Map.Entry<? extends K, ? extends ValueHolder<V>>> bulkComputeIfAbsentAndFault(Iterable<? extends K> keys, Function<Iterable<? extends K>, Iterable<? extends Map.Entry<? extends K, ? extends V>>> mappingFunction) throws StoreAccessException {
+    return bulkComputeIfAbsent((Set<? extends K>) keys,mappingFunction).entrySet();
```
chrisdennis, line 559: This doesn't fix the issue, it just masks it. When the broken runs the `bulkComputeIfAbsent(...)` method throws the `UnsupportedOperationException` at line 553 which happens to allow the test to pass, but it's not passing for the right reasons.

chrisdennis (changes requested):

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
nnares: This was the broken test - CacheLoaderWithoutReadThroughTest#shouldPropagateExceptionUsingLoadAll()

I have observed below behavior of broken test by running it in master branch(without my changes)
This test has the exact same behavior (which you have mentioned here), where it is getting failed at line 553 and throws `UnsupportedOperationException ` then validating the exception cause in test by -  `assertThat(e.getCause(), instanceOf(CacheLoaderException.class));`

After applying my fix for this test, the test is still behaving same and the same exception hierarchy is been thrown 
```
----issue-3097 branch----------
java.util.concurrent.ExecutionException:
javax.cache.integration.CacheLoaderException:
java.lang.UnsupportedOperationException:
This bulkComputeIfAbsent method is not yet capable of handling generic computation functions

----master branch-------
java.util.concurrent.ExecutionException:
javax.cache.integration.CacheLoaderException:
java.lang.UnsupportedOperationException:
This bulkComputeIfAbsent method is not yet capable of handling generic computation functions
```

core-spi-test/src/main/java/org/ehcache/internal/store/StoreSPITest.java
```diff
@@ -16,6 +16,7 @@
 
 package org.ehcache.internal.store;
 
+import org.junit.Ignore;
```
chrisdennis, line 19: Can be removed.

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
```diff
@@ -595,6 +600,7 @@ public void setInvalidationValve(InvalidationValve valve) {
     this.invalidationValve = valve;
   }
 
+
```
chrisdennis, line 603: Can be removed.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
```diff
@@ -75,7 +76,9 @@ public interface AuthoritativeTier<K, V> extends Store<K, V> {
    */
   void setInvalidationValve(InvalidationValve valve);
 
-  /**
+  Iterable<? extends Map.Entry<? extends K,? extends ValueHolder<V>>> bulkComputeIfAbsentAndFault(Iterable<? extends K> keyParam, Function<Iterable<? extends K>, Iterable<? extends Map.Entry<? extends K,? extends V>>> mappingFunction) throws StoreAccessException;
```
chrisdennis, line 79: I think we can javadoc this now that the signatures have stabilized.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -107,6 +108,8 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
    */
   void setInvalidationListener(InvalidationListener<K, V> invalidationListener);
 
+    Map<K, Store.ValueHolder<V>> bulkGetOrComputeIfAbsent(Set<? extends K> keys, Function<Set<? extends K>, Iterable<? extends Map.Entry<? extends K, ? extends Store.ValueHolder<V>>>> mappingFunction) throws StoreAccessException;
```
chrisdennis, line 111: I think we can relax the `keys` type here to `Iterable<? extends K>`. We should also add some javadoc.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/basic/NopStore.java
```diff
@@ -67,6 +67,11 @@ public void setInvalidationValve(InvalidationValve valve) {
 
   }
 
+  @Override
+  public Iterable<? extends Map.Entry<? extends K, ? extends ValueHolder<V>>> bulkComputeIfAbsentAndFault(Iterable<? extends K> keyParam, Function<Iterable<? extends K>, Iterable<? extends Map.Entry<? extends K, ? extends V>>> mappingFunction) throws StoreAccessException {
+    return null;
```
chrisdennis, line 72: We should implement this in the same manner as `bulkComputeIfAbsent`... it should return a map with null values.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
```diff
@@ -356,13 +357,19 @@ public Map<K, ValueHolder<V>> bulkCompute(Set<? extends K> keys, Function<Iterab
...
+      return cachingTier().bulkGetOrComputeIfAbsent(keys, missingKeys -> {
+        try {
+          return authoritativeTier.bulkComputeIfAbsentAndFault(missingKeys, mappingFunction);
+        } catch (StoreAccessException cae) {
+          throw new StorePassThroughException(cae);
+        }
+      });
+    } catch (StoreAccessException ce) {
+      handleStoreAccessException(ce);
+      return Collections.emptyMap();
```
chrisdennis, line 372: This return isn't right per the contract but will never happen because the method call above always throws. We should instead modify the method so that we can use it in the same manner in all places. I'm thinking either introduce a generic for the return type so that it can be inferred, or convert it to a `throw ...` and return an exception from the method.

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/OnHeapStoreBulkMethodsTest.java
```diff
@@ -66,6 +76,24 @@ protected <Number, CharSequence> OnHeapStore<Number, CharSequence> newStore() {
         new NoopSizeOfEngine(), NullStoreEventDispatcher.nullStoreEventDispatcher(), new DefaultStatisticsService());
   }
 
+  @SuppressWarnings("unchecked")
+  protected <K, V> Store.Configuration<K, V> mockStoreConfig(int heapCount) {
```
chrisdennis, line 80: These methods are duplicating too much of the existing code. You should be able to have the existing methods call these ones to reduce the duplication.

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/OnHeapStoreBulkMethodsTest.java
```diff
@@ -86,8 +114,8 @@ public void testBulkComputeFunctionGetsValuesOfEntries() throws Exception {
       Map<Number, Number> newValues = new HashMap<>();
       for (Map.Entry<? extends Number, ? extends Number> entry : entries) {
         final Number currentValue = entry.getValue();
-        if(currentValue == null) {
-          if(entry.getKey().equals(4)) {
+        if (currentValue == null) {
```
chrisdennis, line 117: I know the existing formatting is wrong, but I'd rather not see formatting changes on code that is otherwise untouched.

chrisdennis (changes requested):

core-spi-test/src/main/java/org/ehcache/internal/store/StoreSPITest.java
nnares: have removed it.

clustered/ehcache-client/src/main/java/org/ehcache/clustered/client/internal/store/ClusteredStore.java
nnares: have removed it.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
nnares: I have added method level javadoc

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
nnares: Relaxed the key type from `Set<? extends K>` to `Iterable<? extends K>`
Added method level javadoc as well

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/basic/NopStore.java
nnares: incorporated

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/OnHeapStoreBulkMethodsTest.java
nnares: Removed new overloaded methods, Now using existing methods for my junits

ehcache-impl/src/test/java/org/ehcache/impl/internal/store/heap/OnHeapStoreBulkMethodsTest.java
nnares: I have removed all formatting changes now

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
nnares: Can you elaborate this more?

chrisdennis: Returning `Collections.emptyMap()` here is wrong... but the line is unreachable because `handleStoreAccessException(...)` always throws. The other uses of `handleStoreAccessException(...)` get around this problem by having:
```
return handleStoreAccessException(e)
```
because its declared to return the right type (even though it doesn't return anything). You should be able to adapt the method to allow you to use it in that way here too.

nnares: modified the method signature to generic return type to accomodate both types of return type - ValueHolder<V> & Map<K, Store.ValueHolder<V>>

```
-  private ValueHolder<V> handleStoreAccessException(StoreAccessException ce) throws StoreAccessException
+  private <R> R handleStoreAccessException(StoreAccessException ce) throws StoreAccessException
```

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
```diff
@@ -604,6 +610,11 @@ public void setInvalidationListener(final InvalidationListener<K, V> invalidatio
       // noop
     }
 
+    @Override
+    public Map<K, Store.ValueHolder<V>> bulkGetOrComputeIfAbsent(Iterable<? extends K> keys, Function<Set<? extends K>, Iterable<? extends Map.Entry<? extends K, ? extends ValueHolder<V>>>> mappingFunction) throws StoreAccessException {
+      return null;
```
chrisdennis, line 615: Should this return a map with the full set of keys, each with a null value?

chrisdennis (approved):
Approved... with one minor question. Probably not a big deal... but would be good to know which it should be.

ehcache-impl/src/main/java/org/ehcache/impl/internal/store/tiering/TieredStore.java
nnares: this should return a full set of keys, each with a null value, I have put code change for this

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
```diff
@@ -75,7 +76,24 @@ public interface AuthoritativeTier<K, V> extends Store<K, V> {
...
   void setInvalidationValve(InvalidationValve valve);
 
+
   /**
+   * Bulk method to compute a value for every key passed in the {@link Iterable} <code>keys</code> argument using the <code>mappingFunction</code>
+   * to compute the value.
+   * <p>
+   * The function gets an {@link Iterable} of {@link java.util.Map.Entry} key/value pairs, where each entry's value is its currently stored value
+   * for each key that is not mapped in the store. It is expected that the function returns an {@link Iterable} of {@link java.util.Map.Entry}
+   * key/value pairs containing an entry for each key that was passed to it. This returned {@link Iterable} should also iterate in the same order as the input {@link Iterable}.
```
chrisdennis, line 86: I don't think the iteration order need be stable between the passed in keys and the function return. I think we can strike that sentence.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
```diff
@@ -75,7 +76,24 @@ public interface AuthoritativeTier<K, V> extends Store<K, V> {
    */
   void setInvalidationValve(InvalidationValve valve);
 
+
   /**
+   * Bulk method to compute a value for every key passed in the {@link Iterable} <code>keys</code> argument using the <code>mappingFunction</code>
+   * to compute the value.
+   * <p>
+   * The function gets an {@link Iterable} of {@link java.util.Map.Entry} key/value pairs, where each entry's value is its currently stored value
```
chrisdennis, line 84: `The function takes an...`

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
```diff
@@ -75,7 +76,24 @@ public interface AuthoritativeTier<K, V> extends Store<K, V> {
...
+   * for each key that is not mapped in the store. It is expected that the function returns an {@link Iterable} of {@link java.util.Map.Entry}
```
chrisdennis, line 85: `The function should return an...`

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
```diff
@@ -75,7 +76,24 @@ public interface AuthoritativeTier<K, V> extends Store<K, V> {
...
+   * key/value pairs containing an entry for each key that was passed to it. This returned {@link Iterable} should also iterate in the same order as the input {@link Iterable}.
+   * Note: This method guarantees atomicity of computations for each individual key in {@code keys}. Implementations may choose to provide coarser grained atomicity.
+   *
+   * @param keys the keys to compute a new value for, if they're not in the store.
+   * @param mappingFunction the function that generates new values.
+   * @return a {@link Map} of key/value pairs for each key in <code>keys</code> to the previously missing value.
```
chrisdennis, line 91: No need to `@link` a JDK type this common... plus given how the javadoc is currently generated a correct link will not be generated.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -107,6 +108,22 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
    */
   void setInvalidationListener(InvalidationListener<K, V> invalidationListener);
 
+  /**
+   * Bulk method to get its mapped value from CachingTier, then look at AuthoritativeTier for the missing entries Of CachingTier then Compute a value for every key passed in the {@link Set} <code>keys</code> argument using the <code>mappingFunction</code>
```
chrisdennis, line 112: If you can rewrite this javadoc without referencing the `AuthoritativeTier` that would be good. Just describe what the method does... the authority only comes in with the functions we normally pass to this method.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -107,6 +108,22 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
...
+   * to compute the value.
+   * <p>
+   * The function gets an {@link Iterable} of {@link java.util.Map.Entry} key/value pairs, where each entry's value is its currently stored value
```
chrisdennis, line 115: `The function takes an...`

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -107,6 +108,22 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
...
+   * for each key that is not mapped in the store. It is expected that the function returns an {@link Iterable} of {@link java.util.Map.Entry}
```
chrisdennis, line 116: `The function should return an...`

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -107,6 +108,22 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
...
+   * key/value pairs containing an entry for each key that was passed to it. This returned {@link Iterable} should also iterate in the same order as the input {@link Iterable}.
```
chrisdennis, line 117: Again, we can strike the sentence about iteration order.

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
```diff
@@ -107,6 +108,22 @@ public interface CachingTier<K, V> extends ConfigurationChangeSupport {
...
+   * Note: This method guarantees atomicity of computations for each individual key in {@code keys}. Implementations may choose to provide coarser grained atomicity.
+   *
+   * @param keys the keys to compute a new value for, if they're not in the store.
+   * @param mappingFunction the function that generates new values.
+   * @return a {@link Map} of key/value pairs for each key in <code>keys</code>.
```
chrisdennis, line 122: Can just use `@code` here instead of `@link`

chrisdennis (changes requested):

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
nnares: removed it

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
nnares: incorporated

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
nnares: incorporated

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
nnares: modified it, now not mentioning `AuthoritativeTier `

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/AuthoritativeTier.java
nnares: Its looks good using `@link`, rather than simply mentioning a Map, I am not  changing this

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
nnares: incorporated

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
nnares: incorporated

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
nnares: removed it

ehcache-core/src/main/java/org/ehcache/core/spi/store/tiering/CachingTier.java
nnares: incorporated

chrisdennis (approved):

Issue #2158, opened 2017-10-09
Decide on Java 9 module name(s)


ljacomet:
With Java 9 now released, we need to reserve module names by using the `MANIFEST.MF` entry approach - that is having a key `Automatic-Module-Name` in it.

As a reminder, there are two ways of consuming Ehcache:
* Individual jars, that do not contain any shaded dependencies. The end result is that you have many small jars on the uclasspath.
* Fat jars that contain all of the Ehcache code and shaded dependencies, but _not_ repackaged.

The moment we introduce modularity concepts, even as small as picking a module name, I believe we have to make a choice.
We cannot keep two options in the modular world. As it would mean having different modules containing the same package, which would break any app that would have dependencies which did not pick the same Ehcache distribution.

So I would suggest the following:

Only offer the small jars with module definitions
* that is effectively deprecating our fat jars for the modular world.
* this would solve the shading issue as well, since we would only need dependencies to also have a proper module name and we are good to go.

I would then suggest the following modules (from maven coordinates to module name):
* `org.ehcache.modules:ehcache-api` -> `org.ehcache.api`
* `org.ehcache.modules:ehcache-core` -> `org.ehcache.core`
* `org.ehcache.modules:ehcache-impl` -> `org.ehcache`
* `org.ehcache.modules:ehcache-xml` -> `org.ehcache.xml`
* `org.ehcache.modules:ehcache-107` -> `org.ehcache.jcache`
* `org.ehcache.modules:ehcache-client` -> `org.ehcache.clustered.client`
* `org.ehcache.modules:ehcache-common` ->`org.ehcache.clustered.common`
* `org.ehcache.modules:ehcache-management` -> `org.ehcache.management`
* `org.ehcache:ehcache-transactions` -> `org.ehcache.transactions`

The following jars would not get a module name:
* `org.ehcache:ehcahce` as it is a fat jar
* `org.ehcache:ehcache-clustered` as it is a fat jar
* `org.ehcache.modules:ehcache-server` as this is not required as long as the Terracotta Server does not run on the module path.

chrisdennis:
In the current world I think this makes sense... I would however like to then propose we push hard for progress on http://openjdk.java.net/projects/jigsaw/spec/issues/#MultiModuleJARs.  Without this implemented there is going to be constant tension when choosing module boundaries between producers and consumers of modules.

chrisdennis:
[...]

PR #3133, opened 2023-02-24
Fixes issue#2158 by adding Automatic-Module-Name

reinhapa:
Adds `Automatic-Module-Name` to the manifest logic.

The default calculation will take the module name replace all `-` characters with `.` and prefixes is with `org.` in case of the module `ehcache-107` a specific automatic module declaration `org.ehcache.jcache` was added to replace the calculated default name.

### DIFF
--- a/build-logic/src/main/java/org/ehcache/build/conventions/BndConvention.java
+++ b/build-logic/src/main/java/org/ehcache/build/conventions/BndConvention.java
@@ -14,6 +14,7 @@ import org.gradle.api.artifacts.ProjectDependency;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.provider.MapProperty;
 import org.gradle.api.publish.PublishingExtension;
+import org.gradle.api.publish.maven.MavenPom;
 import org.gradle.api.publish.maven.MavenPublication;
 import org.gradle.api.publish.plugins.PublishingPlugin;
 import org.gradle.api.tasks.bundling.Jar;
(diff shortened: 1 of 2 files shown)


chrisdennis (commented):
The module names are okay (I think)... but the `BndConvention` probably isn't the right place (congratulations on navigating the build-system though). I'm actually prototyping this atm after your comment on #2158 floated this up to my attention again.

The changes I have staged at the moment put the manifest changes in `InternalEhcacheModule` and `ClusteredEhcacheModule`. That's really only the start though - the second issue is getting some test coverage in place to make sure everything is working. I managed to hack that in (could probably do with some cleanup) - but that leaves issues with split packages (right now between `ehcache-xml-spi` and `ehcache-xml`). I'm still working on this, but I'll push a draft PR shortly with the changes I have.

reinhapa:
Would be cool to have those in the next release as I try to migrate the [TweetWallFx](https://github.com/TweetWallFX/TweetwallFX) project to the Java module system and got some other ones that would need this feature too...

reinhapa:
@chrisdennis Any progress on your prototype so far or would you give me a hint and I could look into it?

chrisdennis:
This is caught in an awkward place atm. Automatic modules cannot be made to work because the `org.ehcache.impl` module needs to depend on `jdk.unsupported` due to its Unsafe usage (modded JDK CHM). A fully modular setup probably the only way we can go. Right now I need to get some time internally to discuss which of the following three approaches we will be taking (in order of likelihood by my judgement):

1. Increment to Ehcache 4.0, drop support for JDK 1.8 (adopt JDK 11 baseline), and have a conventional Java 9 modular setup for the modules, with a non-modular uber-jar. This would also be accompanied by more JDK 11 enabled changes internally (and maybe in API).
2. Use Multi-Release Jars to support modules in the current 3.x line. This is more complex from a build-system approach but does have the least baggage in terms of alignment with other tasks.
3. Declare Java 9 modules "a bad idea" and decide to never implement them.

I'm hoping to get a decision made on this asap... but if I'm right and we go with a major version bump it's not going to happen massively quickly.

erilor:
@chrisdennis has any decision been made regarding which approach you're aiming for?

PR #3134, opened 2023-02-24
Initial draft of automatic modules support.

chrisdennis:


chrisdennis:
@reinhapa these are the draft changes I have for automatic module support. Still trying to work out kinks in the modules though. Note that this is assembled on top an as yet unmerged Gradle upgrade (#3057) because I needed test suite support to make it easier to implement the modularTest task - this makes the changes look bigger (and weirder) than they would otherwise be.

PR #3141, opened 2023-03-08
Issue 1785 3.9

vaibhavsrivastava07:
Fix for issue-1785 for release/3.9


AbfrmBlr (approved):

clustered/client/build.gradle
```diff
@@ -20,6 +20,8 @@ plugins {
 }
 
 dependencies {
+  implementation "org.terracotta:terracotta-utilities-tools:$terracottaUtilitiesVersion"
```
AbfrmBlr, line 23: Please use the existing format to add dependency. Also it will be nice if it ends up with other "implementation" dependencies

impl/src/test/java/org/ehcache/impl/persistence/FileBasedStateRepositoryTest.java
```diff
@@ -16,17 +16,14 @@
...
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
+import java.io.*;
```
AbfrmBlr, line 26: I think your IDE is collapsing the imports to .*, you can disable that. Please undo this change.

PR #3142, opened 2023-03-08
issue-628 Changes to handle failure during clustered write-behind

hchavdapsl:
Issue-628 Changes for handle failure during clustered write-behind
New EvenNumberLoaderWriter is added to tests to avoid touching any changes to existing clustered tests.


clustered/integration-test/src/test/java/org/ehcache/clustered/writebehind/BasicClusteredWriteBehindTest.java
```diff
@@ -127,4 +133,40 @@ public void testClusteredWriteBehindLoading() throws Exception {
...
+
+  @Test
+  public void testBasicClusteredWriteBehindWithFailure()  {
+    cache = null;
+    cacheManager.close();
+    cacheManager = null;
+    cacheManager = createCacheManagerWithLoaderWriterWithFailure(CLUSTER.getConnectionURI());
+    cache = cacheManager.getCache(testName.getMethodName(), Long.class, String.class);
+    long counter;
+    for (counter = 1; counter <= 4; counter++) {
```
chrisdennis, line 145: I think all threee of these loops will be easier to follow if you just unroll them.

clustered/integration-test/src/test/java/org/ehcache/clustered/writebehind/WriteBehindTestBase.java
```diff
@@ -102,4 +103,28 @@ PersistentCacheManager createCacheManager(URI clusterUri) {
       .withCache(testName.getMethodName(), cacheConfiguration)
       .build(true);
   }
+
+  PersistentCacheManager createCacheManagerWithLoaderWriterWithFailure(URI clusterUri) {
+    evenNumberLoaderWriter = new EvenNumberLoaderWriter();
+    CacheConfiguration<Long, String> cacheConfiguration =
+      newCacheConfigurationBuilder(Long.class, String.class, ResourcePoolsBuilder.newResourcePoolsBuilder()
```
chrisdennis, line 110: It might make sense to extract out the commonality between this method and the existing `createCacheManager(...)` method so that it's easier to see where the differences are.

chrisdennis (changes requested):

PR #3147, opened 2023-03-21
[master] Fix Ehcache Docker images: client folder is now required as a dependency for voter and tools

mathieucarbou:


chrisdennis (approved):
Assume this is change in the platform kit, right?

akomakom (approved):

mathieucarbou:
> Assume this is change in the platform kit, right?

there is nothing to change in platform: the kit works, but the tools now depend on the client libs which were not in docker images.

We will have to verify the EE kit too...

PR #3155, opened 2023-04-14
Upgrade Ehcache 3.9 to Terracotta Core/Platform 5.10.x

chrisdennis:


mathieucarbou (approved):

chrisdennis:
I've tested this downstream and everything seems to be good with cross-version compatibility, nothing obviously broken in the comms protocols, so I think we're safe to merge this.

PR #3170, opened 2023-06-06
Fixes #3169 : Track event listeners keyed by type to allow earlier event firing veto

### LINKED ISSUE #3169: Cache events listeners descend too far down the "firing sequence" for "unfireable events"
Cache events listeners will get all the way to dispatching a task, including potentially spawning a thread , for events of a type for which there are no listeners. We have all the knowledge available to trim the call tree much sooner, and should be doing so.

chrisdennis:


SamuelBussmann:
Thank you for the fix.
But doesn't InvocationScopedEventSink.fireOrdered still block on all events if any ordered listener was registered?
Maybe skip events without interested listeners and just consume the queue entry?

ehcache-core/src/main/java/org/ehcache/core/events/EventListenerWrapper.java
```diff
@@ -86,8 +87,8 @@ public void onEvent(CacheEvent<? extends K, ? extends V> event) {
     return listener;
   }
 
-  public boolean isForEventType(EventType type) {
```
muthumcapec31, line 89: Why this isForEventType check removed?

ehcache-impl/src/main/java/org/ehcache/impl/events/EventDispatchTask.java
```diff
@@ -40,12 +40,10 @@
   @Override
   public void run() {
     for(EventListenerWrapper<K, V> listenerWrapper : listenerWrappers) {
-      if (listenerWrapper.isForEventType(cacheEvent.getType())) {
```
muthumcapec31, line 43: Why this isForEventType check removed?

SamuelBussmann: listenerWrappers are already correctly preselected to only contain relevant entries from *syncListenersList.get(event.getType()) ?

chrisdennis:
> Thank you for the fix.
> But doesn't InvocationScopedEventSink.fireOrdered still block on all events if any ordered listener was registered?
> Maybe skip events without interested listeners and just consume the queue entry?

@SamuelBussmann you're right... we should be able to pull this higher and eliminate a lot more work. Let me move this back to draft and until I get time to rework it.

SamuelBussmann:
> @SamuelBussmann you're right... we should be able to pull this higher and eliminate a lot more work. Let me move this back to draft and until I get time to rework it.

Just add a filter, whether any listener is interested in this event at all?
And then a further optimization would be to group events by ordered listener to let the other events get past. But this will get tricky fast with Listeners implementing multiple events...

PR #3171, opened 2023-06-09
Introduce a SourceDirectorySet for Unsafe Java code and wire compilation appropriately (Avoid IDEA-321592)

chrisdennis:


mathieucarbou (approved):
tested and works

PR #3187, opened 2023-07-21
Add 'InstantiatorService' to allow customization of object construction

chrisdennis:
This is a proposal for an `InstantiatorService` which would allow users to customize how Ehcache creates instances of user types that support the cache. That means things like event listeners, cache loader/writers, resilience strategies etc. At this point I'm mainly interested in knowing if the interface presented here is suitable for wiring in to the Spring dependency injection system.


mathieucarbou (approved):

mathieucarbou:
Tested with the JCache changes in https://github.com/ehcache/ehcache3/pull/3190 in the sample at https://github.com/SoftwareAG/terracotta-samples/pull/32.

Happy with it!

PR #3197, opened 2023-09-08
ARC Java and Xml API

tmesic99:
Changes are for the Java and Xml Configuration APi for the Automatic Resource Control (ARC) feature.  Changes include:

- A new `SharedResourcePoolsBuilder` is used to add one or more 'shareable' `ResourcePool`s to the cache manager.  Phase 1 will only support shareable offheap.  But I added shareable `heap` configuration (targeted for Phase 2) to these changes in order to build out and test some of the expected configuration options.  However, I stopped short of adding the API hooks for the `heap` `SizeOfEngine` support which will be required when supporting ARC for heap in Phase 2.
- Like the existing `ResourcePoolsBuilder` used during cache configuration, the `SharedResourcePoolsBuilder` accepts instances of `SizedResourcePool`s.  But when added to the cache manager's `sharedResources`, the `SizedResourcePool`s are configured with `shared=true`.
- When a cache wishes to inform the cache manager that it intends to use a shared resource, defined on the the cache manager, the cache adds a `SizedResourcePool` to its `ResourcePools` with `shared=true` but with `size=0`.  This is the mechanism through which cache configurations can convey to the cache manager that they want to use the shared resource and not have one constructed for its dedicated use.
- When undergoing initialization (`CacheManagerBuilder.build(true)`), the `EhcacheManager` understands the above relationships and can perform the necessary consistency checking across the cache's declared wishes and what has actually been configured on the cache manager.
- As much as possible, supported configurations are handled at the point of configuration by only exposing valid options (e.g. creating a `SharedResourcePoolsBuilder` instead of re-using the existing `ResourcePoolsBuilder` when configuring the cache manager's shared resources).  When not feasible during configuration, runtime validation is performed.


ehcache-core/src/main/java/org/ehcache/core/config/DefaultConfiguration.java
```diff
@@ -56,6 +60,7 @@ public DefaultConfiguration(Configuration cfg) {
     this.caches = new ConcurrentHashMap<>(cfg.getCacheConfigurations());
     this.services = unmodifiableCollection(cfg.getServiceCreationConfigurations());
     this.classLoader = cfg.getClassLoader();
+    this.sharedResourcePools = new ConcurrentHashMap<>(cfg.getSharedResourcePools());
```
tmesic99, line 63: I elected to implement the config manager's shared resource pools outside of its service implementation.  If this should be refactored to be included as a new service then please let me know.

ehcache-impl/src/main/java/org/ehcache/impl/config/SizedResourcePoolImpl.java
```diff
@@ -46,17 +46,31 @@ public class SizedResourcePoolImpl<P extends SizedResourcePool> extends Abstract
...
+   * @param shared whether the pool is to be shared
+   */
+  public SizedResourcePoolImpl(ResourceType<P> type, long size, ResourceUnit unit, boolean persistent, boolean shared) {
+    super(type, persistent, shared);
     if (unit == null) {
       throw new NullPointerException("ResourceUnit can not be null");
     }
-    if (size <= 0) {
+    if (size <= 0 && !shared) {
       throw new IllegalArgumentException("Size must be greater than 0");
```
tmesic99, line 67: A SizedResourcePool that is marked for sharing and has zero bytes is used to inform the cache manager that the cache wants to use a shared heap or offheap sized pool.

PR #3200, opened 2023-10-17
Bump to offheap-store 2.5.5

mathieucarbou:


osgi-test/src/test/java/org/ehcache/osgi/SimpleOsgiTest.java
```diff
@@ -70,6 +70,7 @@ public Option[] individualModules() {
       gradleBundle("org.terracotta:statistics"),
       gradleBundle("org.ehcache:sizeof"),
       gradleBundle("org.terracotta:offheap-store"),
+      gradleBundle("org.terracotta:fast-restartable-store"),
```
chrisdennis, line 73: Hmmm... should the `offheap-store` manifest declare an optional dependency on `fast-restartable-store`?

mathieucarbou: yeah that was my question when I did that... I don't know. So we have other cases of that in Echache ?

chrisdennis: I kinda feel like this depends on when/if we move the features FRS supports in to OSS.

mathieucarbou: Since we don't know the answer now, what is best to do: mark is optional, and eventually later remove the optional, or mark it required from start, eventually by marking it optional later ?

PR #3218, opened 2024-03-22
terracottaPlatformVersion = 5.10.21

mathieucarbou:


chrisdennis (approved):
Are we confident that there are no tests that will fail sporadically due to only 'racily' reaching the newly required preconditions for a dynamic config change.

mathieucarbou:
> Are we confident that there are no tests that will fail sporadically due to only 'racily' reaching the newly required preconditions for a dynamic config change.

Yes: the only tests that I am marking as `-force` are testing special conditions when a server is down and do not bring back up a node while a DC change is triggered

PR #3236, opened 2024-07-18
Fixes #3169 : Avoid firing unneeded Events

### LINKED ISSUE #3169: Cache events listeners descend too far down the "firing sequence" for "unfireable events"

SamuelBussmann:
Based on #3170 Track event listeners keyed by type to allow earlier event firing veto 

- #3170 only triggers the listeners interested in the current event
- Don't even register an Event if no Listener is interested in this event
- Inline execution of Unordered Synchron listeners as it incurs a massive delay to spin up a separate Thread (20us vs 1us) an wait for it completion anyway


SamuelBussmann:
Can you retrigger the test pipelines? locally the test runs successful and it timed out on your infrastructure

SamuelBussmann:
Hello
It's great to see the renewed activity and releases.
Is there a chance and possibly a timeline that this MergeRequest or a different fix for the original issue can be merged?

PR #3251, opened 2024-11-20
TDB-19133 : If invalidations are not claimed by reconnect clients, then don't attempt to fire invalidation completion messages

chrisdennis:


GaryWKeim:
I've tested this change in terracotta-enterprise (clean) and done some manual fail-over/reconnect testing without any trouble. Code improvements look good. Don't know if the Java 21 `org.ehcache.clustered.DestroyLoopTest` failure is related to this change.

GaryWKeim (approved):

PR #3274, opened 2025-08-20
Issue 2494 Refix

chrisdennis:
We move back here to a non-lazy approach to building the map of 107 wrappers. This means we generate all the wrappers up front, and then we generate new wrappers by listening for added caches, and remove when we see a cache removed. This should mean that direct additions/removals to the Ehcache manager will be immediately reflected in the JSR-107 manager.


chrisdennis:
Local testing confirms this fixes the issue in the spring-boot test, and triggers no other failures there.

prasantakumar82 (approved):
Looks good to me

PR #3328, opened 2026-09-03
Update platform to 5.11.12 and core to 5.12.27

devendra-ibm:


Gen-SIQA-User:
No vulnerabilities found.

devendra-ibm:
@myronkscott Tests are failing because invoke context for passive has not valid passthrough client info instead it's still fake invalid ClientId. So OOMessageHandler skip it and doesn't track it.
Probably need to fix in core first and then core should be release and upgraded in ehcache.

Gen-SIQA-User:
No vulnerabilities found.

mathieucarbou-ibm (dismissed):

chrisdennis (approved):
Approved but the underlying bug in passthrough should be fixed and this test re-enabled in short order.

guptakshit (approved):

mobasherul-ibm (approved):

prasanta-ibm (approved):

